PR #12, opened 2023-02-02
Add linting!

nickgros:
- Fix all existing lint errors, either by changing errors to warnings or fixing code
- Fix TypeScript errors caused by fixing lint errors
- Move eslint config to project root
- Add lint to CI


nickgros:
Pulled into #11

PR #13, opened 2023-02-03
SWC-6384

nickgros:
Moving this over to the monorepo: https://github.com/Sage-Bionetworks/Synapse-React-Client/pull/1975


packages/synapse-react-client/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "synapse-react-client",
-  "version": "3.1.23",
+  "version": "3.1.22",
```
nickgros, line 3: 3.1.22 and 3.1.23 have not been released, and tags created for these versions have been deleted. I was just testing the new workflow

PR #15, opened 2023-02-06
Upgrade workflow, rename artifact

nickgros:


.github/workflows/build-test.yml
```diff
@@ -64,5 +64,5 @@
        if: success() || steps.test.conclusion == 'failure'
        uses: actions/upload-artifact@v3
        with:
-         name: my-artifact
+         name: test-results
```
nickgros, line 67: Trying to sift through tons of logs for an error is hard, so I added an HTML test report that GH Actions will upload. I forgot to rename it when I added this workflow.

apps/portals/run.sh
```diff
@@ -77,7 +77,7 @@ EOL
 
 elif [ "$1" = "push-staging" ]; then
   # sync current with staging
-  pnpm && pnpm build
+  pnpm i && pnpm build
```
nickgros, line 80: `pnpm` is not valid, so the portals build broke.

jay-hodgson (approved):

PR #17, opened 2023-02-07
Improve test stability

nickgros:
- Remove `failOnConsole` from EntityPageTitleBar test suite that is likely failing due to resources bleeding over from previous test suite.
- Include failure messages in the test report.

### DIFF
--- a/packages/synapse-react-client/jest.config.js
+++ b/packages/synapse-react-client/jest.config.js
@@ -25,6 +25,8 @@ module.exports = {
       {
         pageTitle: 'synapse-react-client Test Report',
         outputPath: './coverage/test-report.html',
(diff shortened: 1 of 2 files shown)


packages/synapse-react-client/jest.config.js
```diff
@@ -25,6 +25,8 @@ module.exports = {
       {
         pageTitle: 'synapse-react-client Test Report',
         outputPath: './coverage/test-report.html',
+        includeFailureMsg: true,
+        includeSuiteFailure: true,
```
jay-hodgson, line 29: Nice find!

jay-hodgson (approved):

PR #18, opened 2023-02-07
Fix builds

nickgros:
- synapse-react-client: Build tsconfig `incremental: true` broke output
- portals: in build machine script, use Nx to guarantee dependencies are built

### DIFF
--- a/apps/portals/run.sh
+++ b/apps/portals/run.sh
@@ -77,7 +77,7 @@ EOL
 
 elif [ "$1" = "push-staging" ]; then
   # sync current with staging
-  pnpm i && pnpm build
+  pnpm i && pnpm nx run portals:build
   node sitemap/generate-sitemap.js $2
   # generate robots.txt
 cat > ./build/robots.txt <<EOL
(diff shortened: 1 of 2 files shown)


packages/synapse-react-client/tsconfig.build.json
```diff
@@ -1,11 +1,12 @@
 {
   "extends": "./tsconfig.json",
   "compilerOptions": {
+    "rootDir": "src/lib",
     // Set the outDir to dist, where the output will be published to npm
     "outDir": "dist",
     "composite": false,
     "strict": true,
-    "rootDir": "src/lib"
+    "incremental": false
```
nickgros, line 9: This inherited `"incremental": true` from ./tsconfig.json, but that setting broke the output. This build config does not need to use incremental builds, which, in the context of our project, just support IDEs, for now at least.

apps/portals/run.sh
```diff
@@ -77,7 +77,7 @@ EOL
 
 elif [ "$1" = "push-staging" ]; then
   # sync current with staging
-  pnpm i && pnpm build
+  pnpm i && pnpm nx run portals:build
```
nickgros, line 80: Using Nx ensures that all of the local dependencies of portals will also be built.

jay-hodgson (approved):

PR #19, opened 2023-02-07
Replace `cpio` with platform-independent npm package copyfiles

nickgros:
The portals build machine doesn't have `cpio` installed. I think a developer on a Windows machine would hit the same issue. Let's use `copyfiles`, which will be platform-independent, and we can install it with the rest of our dependencies.


packages/synapse-react-client/package.json
```diff
@@ -223,7 +224,7 @@
     "test:coverage": "jest --coverage --runInBand",
     "build": "rimraf dist && pnpm build:js && pnpm build:copy-assets && pnpm build:esbuild",
     "build:js": "tsc --p tsconfig.build.json",
-    "build:copy-assets": "cd src/lib && find . -name '*.scss' -o -name '*.css' -o -name '*.svg' | cpio -pdm ../../dist/",
+    "build:copy-assets": "copyfiles -u 2 src/lib/**/*.{css,scss,svg} dist",
```
nickgros, line 227: `-u 2` will drop the `src/lib` part so we don't need to `cd`

jay-hodgson (approved):

PR #21, opened 2023-02-09
PORTALS-2518: Always show filtered facets

nickgros:
- Fix case where a filtered facet would become hidden after a query result update
- Rename facetsToFilter prop to availableFacets
- Add better loading UI for FacetFilterControls

## Loading UI

Before:

image

After:

image


apps/portals/src/configurations/adknowledgeportal/synapseConfigs/experimental_tools.ts
```diff
@@ -1,7 +1,10 @@
 import { GenericCardSchema } from 'synapse-react-client/dist/containers/GenericCard'
 import { SynapseConstants } from 'synapse-react-client'
 import { SynapseConfig } from 'types/portal-config'
-import { CardConfiguration, LabelLinkConfig } from 'synapse-react-client/dist/containers/CardContainerLogic'
+import {
```
nickgros, line 4: I think the code style hasn't been enforced across all portals files yet, so lots of whitespace changes here

apps/portals/src/configurations/adknowledgeportal/synapseConfigs/experimental_tools.ts
```diff
@@ -118,7 +122,7 @@ const experimentalTools: SynapseConfig = {
     shouldDeepLink: true,
     hideDownload: true,
     tableConfiguration: experimentalDetailsTableConfiguration,
-    facetsToFilter: [
+    availableFacets: [
```
nickgros, line 125: Accomodate breaking change (in the same PR as the breaking change!)

packages/synapse-react-client/mocks/mockQueryResponseData.ts
```diff
@@ -0,0 +1,377 @@
+import { QueryResultBundle } from '../src/lib/utils/synapseTypes'
+
+const mockQueryResponseData: QueryResultBundle = {
```
nickgros, line 3: Moved from JSON to TypeScript file and converted to constant for strong typing

packages/synapse-react-client/src/lib/containers/query_wrapper_plot_nav/QueryWrapperPlotNav.tsx
```diff
@@ -59,7 +61,7 @@ type OwnProps = {
     'queryContext' | 'queryVisualizationContext'
   >
   facetsToPlot?: string[]
-  facetsToFilter?: string[]
+  availableFacets?: FacetFilterControlsProps['availableFacets']
```
nickgros, line 64: This is the set of facets that will be available. The old name was always confusing to me, so I figured a rename was worth the breaking change.

packages/synapse-react-client/src/lib/containers/widgets/query-filter/FacetFilterControls.tsx
```diff
@@ -135,40 +137,105 @@ export const applyChangesToRangeColumn = (
...
-  const { data, isLoadingNewBundle, getLastQueryRequest, executeQueryRequest } =
-    useQueryContext()
+/**
+ * Determines which facet filters should be shown after loading a new bundle. The shown facets will be the first
+ * three available facets, plus any other facets that have a filter applied.
+ * @param facets
+ * @param selectedFacets
+ * @returns the columnNames of the facets that should be shown.
+ */
+export function getDefaultShownFacetFilters(
```
nickgros, line 147: This new function should fix the bug

packages/synapse-react-client/src/lib/containers/widgets/query-filter/FacetFilterControls.tsx
```diff
@@ -135,40 +137,105 @@ export const applyChangesToRangeColumn = (
...
 
-  let shownChips: string[]
-  if (facetFiltersToShow == null) {
-    shownChips = facets?.map(facet => facet.columnName) ?? []
-  } else {
-    // The facets to show have been explicitly specified
-    shownChips = facetFiltersToShow
-  }
-  const [facetFiltersShown, setFacetFiltersShown] = React.useState<string[]>([])
+function FacetFilterControlsSkeleton() {
```
nickgros, line 181: This component really needed a better loading state, so I added this simple skeleton.

packages/synapse-react-client/src/lib/containers/widgets/query-filter/FacetFilterControls.tsx
```diff
@@ -135,40 +137,105 @@ export const applyChangesToRangeColumn = (
...
+    >
+      <FacetFilterSkeleton />
+      <FacetFilterSkeleton />
+      <FacetFilterSkeleton />
+      <FacetFilterSkeleton />
+    </div>
+  )
+}
+
+function FacetFilterControls(props: FacetFilterControlsProps) {
```
nickgros, line 200: Note that this is no longer exported--there is a component below that renders the loader if loading, otherwise render this component. This will remove branching logic from this component that prevented us from using hooks without lots of type guards.

packages/synapse-react-client/src/lib/containers/widgets/query-filter/FacetFilterControls.tsx
```diff
@@ -135,40 +137,105 @@ export const applyChangesToRangeColumn = (
...
+    .facets!.filter(
+      facet =>
+        // If availableFacets is configured, remove those that don't match.
+        availableFacets == null || availableFacets.includes(facet.columnName),
+    )
+    .filter(
+      facet =>
+        // Don't show facets where there are no values
+        !isSingleNotSetValue(facet),
+    )
```
nickgros, line 219: We were previously applying these filters in different places. Apply them both here.

packages/synapse-react-client/src/lib/containers/widgets/query-filter/FacetFilterControls.tsx
```diff
@@ -135,40 +137,105 @@ export const applyChangesToRangeColumn = (
...
+
+  // Controls which facet filter sections are shown/hidden by clicking on chips
+  const [facetFiltersShown, setFacetFiltersShown] = React.useState<Set<string>>(
+    getDefaultShownFacetFilters(facets, lastRequest.query.selectedFacets),
+  )
```
nickgros, line 224: Use the new function to fix the bug

packages/synapse-react-client/src/lib/containers/widgets/query-filter/FacetFilterControls.tsx
```diff
@@ -177,12 +244,12 @@ export function FacetFilterControls(props: FacetFilterControlsProps) {
     executeQueryRequest(queryRequest)
   }
 
-  const onClick = (facet: FacetColumnResult) => {
-    let newFacetFilterShown = [...(facetFiltersShown ?? ['']), facet.columnName]
-    if (facetFiltersShown.includes(facet.columnName)) {
-      newFacetFilterShown = newFacetFilterShown.filter(
-        col => col != facet.columnName,
-      )
+  const toggleShowFacetFilter = (facet: FacetColumnResult) => {
```
nickgros, line 247: Rename and refactor to handle sets instead of arrays

packages/synapse-react-client/src/lib/containers/widgets/query-filter/FacetFilterControls.tsx
```diff
@@ -195,77 +262,78 @@ export function FacetFilterControls(props: FacetFilterControlsProps) {
...
+
+export default function FacetFilterControlsOrLoader(
+  props: FacetFilterControlsProps,
+) {
+  const { data, isLoadingNewBundle } = useQueryContext()
+  if (isLoadingNewBundle) {
+    return <FacetFilterControlsSkeleton />
+  } else if (data == null || data.facets == null) {
+    return <></>
+  }
```
nickgros, line 337: This matches the existing behavior, but I don't know if it can actually happen. I don't think this component is rendered if there are no facets.

apps/portals/src/configurations/adknowledgeportal/synapseConfigs/experimental_tools.ts
jay-hodgson: Well that's convenient!

packages/synapse-react-client/src/lib/containers/widgets/query-filter/FacetFilterControls.tsx
jay-hodgson: Great solution

packages/synapse-react-client/src/lib/containers/widgets/query-filter/FacetFilterControls.tsx
jay-hodgson: I was about to leave a comment about this other potential bug before I saw you returned a Set (not a Bag :))

jay-hodgson (approved):

PR #24, opened 2023-02-09
SWC-6375 - Add `overflow-wrap: break-word` to the expandable table data expanded state

nickgros:
No spaces in this WWWWWWW... cell

image


jay-hodgson:
Yes, that looks great!  Did you happen to do any cross-browser tests for this one (on Chrome, FF, and Safari)?

jay-hodgson (approved):

PR #25, opened 2023-02-10
SWC-6390: synapse-react-client - Refactor Login

nickgros:
- Refactor Login to be a function component. Drop incomplete functionality.
- Move authentication-related components to new folder
- Skip flaky test in SynapseTable


packages/synapse-react-client/src/lib/containers/auth/Login.tsx
```diff
@@ -0,0 +1,211 @@
...
+import { useState } from 'react'
+
+export const PROVIDERS = {
+  GOOGLE: 'GOOGLE_OAUTH_2_0',
+  ORCID: 'ORCID',
+}
+
+type Props = {
+  ssoRedirectUrl?: string
+  sessionCallback: () => void // Callback is invoked after login
```
nickgros, line 22: We had an additional prop `redirectUrl` but it was never used internally, so it was dropped.

packages/synapse-react-client/src/lib/containers/auth/Login.tsx
```diff
@@ -0,0 +1,211 @@
...
+
+/**
+ *  Demo of user session, show login screen and handling user login submission.
+ *
+ *  To support Google SSO in your portal, you must add your domain to the Authorized Redirect URIs
+ *  for Synapse authentication.
+ *  This can be done by visiting https://sagebionetworks.jira.com/servicedesk/customer/portal/9 to set up a collaboration.
+ *  Synapse engineers must add your redirect URL in the Google API console found at https://console.cloud.google.com/ for this functionality to work.
+ */
+function Login(props: Props) {
```
nickgros, line 40: My motivation for converting this to a function component is that I want to use a new custom hook to manage the login state as it grows increasingly complex. Using all function components will make it easier to integrate this custom hook into existing components.

jay-hodgson (approved):

PR #30, opened 2023-02-15
SWC-6390: BREAKING CHANGE: remove sessionCallback from SynapseClient

nickgros:
Remove sessionCallback from SynapseClient functions. If a callback is needed, it can be attached to the promise with `.then`/`await`.

### DIFF
--- a/apps/portals/src/AppInitializer.tsx
+++ b/apps/portals/src/AppInitializer.tsx
@@ -28,7 +28,7 @@ export type AppInitializerState = {
 export type SignInProps = {
   userProfile: UserProfile | undefined
   resetSession: Function
-  getSession: Function
+  getSession: () => Promise<void>
   showLoginDialog: boolean
   onSignIn: Function
   handleCloseLoginDialog: Function
@@ -62,13 +62,12 @@ function useSession(
     undefined,
   )
 
-  const initAnonymousUserState = useCallback(() => {
-    SynapseClient.signOut(() => {
-      // reset token and user profile
-      setToken(undefined)
-      setUserProfile(undefined)
-      setHasCalledGetSession(true)
-    })
+  const initAnonymousUserState = useCallback(async () => {
+    await SynapseClient.signOut()
+    // reset token and user profile
+    setToken(undefined)
+    setUserProfile(undefined)
+    setHasCalledGetSession(true)
   }, [])
 
   const getSession = useCallback(async () => {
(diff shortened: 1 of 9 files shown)


apps/portals/src/AppInitializer.tsx
```diff
@@ -97,16 +96,16 @@ function useSession(
...
+      // are loading as the anonymous user
+      window.location.reload()
     }
   }, [initAnonymousUserState])
 
-  const resetSession = useCallback(() => {
-    SynapseClient.signOut(getSession)
+  const resetSession = useCallback(async () => {
+    await SynapseClient.signOut()
+    await getSession()
```
nickgros, line 108: No reason to pass a callback argument to `SynapseClient.signOut` when we can easily do
`await SynapseClient.signOut()` or `signOut().then(...)`. IMO the function argument just makes it easier to accidentally write a race condition.

packages/synapse-react-client/src/lib/utils/SynapseClient.ts
```diff
@@ -301,11 +295,9 @@ import { InviteeVerificationSignedToken } from './synapseTypes/SignedToken/Invit
 const cookies = new UniversalCookies()
 
 // TODO: Create JSON response types for all return types
-export const IS_OUTSIDE_SYNAPSE_ORG = window.location.hostname
+export const IS_OUTSIDE_SYNAPSE_ORG = !window.location.hostname
```
nickgros, line 298: IDE recommended simplifying this logic

packages/synapse-react-client/src/lib/utils/SynapseClient.ts
```diff
@@ -504,7 +496,7 @@ export const doDelete = (
 
 export const doPut = <T>(
   url: string,
-  requestJsonObject: any,
+  requestJsonObject: unknown,
```
nickgros, line 499: Fixing various lint issues too

packages/synapse-react-client/src/lib/utils/SynapseClient.ts
```diff
@@ -1585,12 +1567,12 @@ If state is included, then we assume that this is being used for account creatio
 where we pass the username through the process.
 */
 export const detectSSOCode = (
-  registerAccountUrl?: string,
-  onError?: (err: any) => void,
+  registerAccountUrl = `${PRODUCTION_ENDPOINT_CONFIG.PORTAL}#!RegisterAccount:0`,
```
nickgros, line 1570: Move default value up here instead of using `??` operator later on

packages/synapse-react-client/stories/LoginPage.stories.ts
```diff
@@ -9,4 +11,16 @@ const meta = {
 export default meta
 type Story = StoryObj<typeof meta>
 
-export const Demo: Story = {}
+export const Demo: Story = {
+  args: {
+    sessionCallback: () => {
```
nickgros, line 16: Make login work in Storybook story for LoginPage

jay-hodgson (approved):

PR #35, opened 2023-02-17
SWC-6390: add useLogin hook

nickgros:


packages/synapse-react-client/src/lib/utils/hooks/useLogin.ts
```diff
@@ -0,0 +1,179 @@
...
+      if (!['VERIFICATION_CODE', 'RECOVERY_CODE', 'LOGGED_IN'].includes(step)) {
+        setStep('VERIFICATION_CODE')
+      }
+    }
+  }, [twoFaErrorResponse])
+
+  async function finishLogin(loginResponse: LoginResponse) {
+    // now get access token from cookie has to be called in the portals repo
+    await SynapseClient.setAccessTokenCookie(loginResponse.accessToken)
+    // Set the new receipt
```
nickgros, line 91:
```suggestion
    await SynapseClient.setAccessTokenCookie(loginResponse.accessToken)
```

packages/synapse-react-client/src/lib/utils/hooks/useLogin.ts
```diff
@@ -0,0 +1,179 @@
...
+    if (sessionCallback) {
+      sessionCallback()
+    }
+  }
+
+  const submitUsernameAndPassword: UseLoginReturn['submitUsernameAndPassword'] =
+    async (username, password) => {
+      setErrorMessage(undefined)
+      try {
+        // get last valid receipt
```
nickgros, line 107:
```suggestion
```

packages/synapse-react-client/src/lib/utils/hooks/useLogin.ts
```diff
@@ -0,0 +1,176 @@
...
+    } catch (e) {
+      setErrorMessage(e.reason)
+      if (
+        // The twoFaToken wasn't transmitted correctly
+        e.reason.includes('The provided twoFaToken is invalid') ||
+        // The user waited too long to enter the code.
+        e.reason.includes('Token has expired')
+      ) {
+        // Instruct the user refresh to start over.
+        setErrorMessage('Something went wrong. Refresh the page and try again.')
```
jay-hodgson, line 155: If we are overwriting the error message in the UI, could you send the actual reason to the console as well?
It's possible that the exact strings used ("The provided twoFaToken is invalid") could change over time, but the consequence is minimal (the user would see a confusing error), so I'm fine with this

jay-hodgson (approved):

PR #37, opened 2023-02-20
SWC-6390 - basic 2FA login support

nickgros:
- Add TOTP Form
- Use `useLogin` hook in Login component
- TOTP Form styling is currently broken, will be fixed upon removing `bootstrap-4-backport` class in future PR

https://user-images.githubusercontent.com/17580037/220185794-4b565786-66df-4044-b83e-bc0e5e6b4410.mov


apps/portals/src/AppInitializer.tsx
```diff
@@ -12,25 +12,17 @@ import { DOWNLOAD_FILES_MENU_TEXT } from 'synapse-react-client/dist/containers/t
 import { SynapseContextProvider } from 'synapse-react-client/dist/utils/SynapseContext'
 import { UserProfile } from 'synapse-react-client/dist/utils/synapseTypes'
 import useDetectSSOCode from 'synapse-react-client/dist/utils/hooks/useDetectSSOCode'
+import { TwoFactorAuthErrorResponse } from 'synapse-react-client/dist/utils/synapseTypes/ErrorResponse'
 import useAnalytics from 'useAnalytics'
 import docTitleConfig from './config/docTitleConfig.json'
 import palette from './config/paletteConfig'
 
-export type AppInitializerState = {
```
nickgros, line 19: Unused since this was converted to a function component months ago

apps/portals/src/Navbar.tsx
```diff
@@ -223,7 +224,24 @@ class Navbar extends React.Component<any, State> {
...
                     <Modal.Header closeButton />
                     <SynapseComponents.Login
-                      sessionCallback={() => getSession()}
+                      twoFactorAuthenticationRequired={twoFaErrorResponse}
+                      onBeginOAuthSignIn={() => {
+                        // save current route (so that we can go back here after SSO)
+                        localStorage.setItem(
+                          'after-sso-login-url',
+                          window.location.href,
+                        )
```
nickgros, line 233: This behavior was always called when the Google OAuth button was clicked, but that's not appropriate for SWC, which doesn't use this key. Moved the logic here. Other client-only apps will need to follow-suite on SRC upgrade.

nickgros: At some point I do want to collect all of our client-only React app code into one utility, but haven't wanted to attempt that until we get another app into the monorepo

packages/synapse-react-client/src/lib/containers/auth/TOTPForm.tsx
```diff
@@ -0,0 +1,61 @@
...
+type TOTPFormProps = {
+  onSubmit: (value: string) => void
+}
+
+export default function TOTPForm(props: TOTPFormProps) {
+  const { onSubmit } = props
+  const [verificationCode, setVerificationCode] = React.useState('')
+  return (
+    <Box>
+      <MuiOtpInput
```
nickgros, line 17: I tried to roll my own OTP component but using this library ended up being 100x easier

apps/portals/src/Navbar.tsx
```diff
@@ -223,7 +224,24 @@ class Navbar extends React.Component<any, State> {
                   >
                     <Modal.Header closeButton />
                     <SynapseComponents.Login
-                      sessionCallback={() => getSession()}
+                      twoFactorAuthenticationRequired={twoFaErrorResponse}
```
nickgros, line 227: Pass in 2FA error if we hit it after OAuth sign in

packages/synapse-react-client/src/lib/utils/hooks/useDetectSSOCode.ts
```diff
@@ -15,16 +15,29 @@ import {
...
 /*
  * During SSO login, the authorization provider (Google) will send the user back to the portal with an authorization code,
  * which can be exchanged for a Synapse user session. This function should be called whenever the root App is initialized
  * (to look for this code parameter and complete the round-trip). If state is included, then we assume that this is being
  * used for account creation, where we pass the username through the process.
  */
 export default function useDetectSSOCode(
-  registerAccountUrl = `${PRODUCTION_ENDPOINT_CONFIG.PORTAL}#!RegisterAccount:0`,
-  onError?: (err: unknown) => void,
+  opts: UseDetectSSOCodeOptions = {},
```
nickgros, line 34: too many optional arguments, so just use an object

jay-hodgson:
It's my understanding that the 2FA form should auto-submit when numbers are filled in (looked like you had to click the Submit button on the video)

jay-hodgson (approved):

PR #39, opened 2023-02-20
SWC-6399: Inability to get or clear the session cookie should not cause the web…

jay-hodgson:
…site to render a blank page

### DIFF
--- a/apps/portals/src/AppInitializer.tsx
+++ b/apps/portals/src/AppInitializer.tsx
@@ -56,11 +56,15 @@ function useSession(
   )
 
   const initAnonymousUserState = useCallback(async () => {
-    await SynapseClient.signOut()
     // reset token and user profile
     setToken(undefined)
     setUserProfile(undefined)
     setHasCalledGetSession(true)
+    try {
+      await SynapseClient.signOut()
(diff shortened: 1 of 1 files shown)


apps/portals/src/AppInitializer.tsx
```diff
@@ -56,11 +56,15 @@ function useSession(
...
-    await SynapseClient.signOut()
     // reset token and user profile
     setToken(undefined)
     setUserProfile(undefined)
     setHasCalledGetSession(true)
+    try {
+      await SynapseClient.signOut()
+    } catch (e) {
+      console.error('Unable to sign out: ', e)
+    }
```
nickgros, line 67: We might want to add general fallback logic if we hit a CORS issue when we hit the servlet. Not necessary to merge this though

nickgros (approved):

apps/portals/src/AppInitializer.tsx
jay-hodgson: The current fallback logic is found in getSession(), where it calls this function (initAnonymousUserState()) if getAccessTokenFromCookie() fails.

PR #43, opened 2023-02-23
PORTALS-2568: In CardContainerLogic, use the default client-side limi…

jay-hodgson:
…t if undefined

### DIFF
--- a/packages/synapse-react-client/src/lib/containers/CardContainerLogic.tsx
+++ b/packages/synapse-react-client/src/lib/containers/CardContainerLogic.tsx
@@ -19,6 +19,7 @@ import { InfiniteQueryWrapper } from './InfiniteQueryWrapper'
 import QuerySortSelector from './QuerySortSelector'
 import { NoContentPlaceholderType } from './table/NoContentPlaceholderType'
 import { IconOptions } from './row_renderers/utils/Icon'
+import { DEFAULT_PAGE_SIZE } from '../utils/SynapseConstants'
 
 /**
  *  Used when a column value should link to an external URL defined by a value in another column.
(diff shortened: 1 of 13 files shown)


packages/synapse-react-client/src/lib/containers/CardContainerLogic.tsx
```diff
@@ -159,7 +160,7 @@ export const CardContainerLogic = (props: CardContainerLogicProps) => {
     entityId: entityId,
     query: {
       sql: props.sql,
-      limit: props.limit,
+      limit: props.limit ?? DEFAULT_PAGE_SIZE,
```
nickgros, line 163: This is a good place for the fix. 

I think we should also fix the hook, since this would happen again if a new component used it and didn't fall back to the default limit. I wonder if it's as simple as changing `getNextPageParam` to use the number of rows for page size.

```tsx
// currently:
const pageSize = request.query.limit ?? DEFAULT_PAGE_SIZE
// suggested change
const pageSize =
  request.query.limit ??
  page.responseBody!.queryResult?.queryResults.rows.length ??
  DEFAULT_PAGE_SIZE
// or maybe just this
const pageSize = page.responseBody!.queryResult?.queryResults.rows.length ?? DEFAULT_PAGE_SIZE
```

I would be fine with having this suggested change tracked in a new ticket so we could dedicate a bit of time to test it properly, since there might be a few edge cases to think about.

Thoughts?

nickgros (approved):
LGTM, I want us to fix the bug in the hook but that can happen separately

packages/synapse-react-client/src/lib/containers/CardContainerLogic.tsx
jay-hodgson: I think this could result in an infinite loop.  If the row length is 0 (and the request query limit is undefined), then `getNextPageParam` will still return true (pageSize === row count)

PR #45, opened 2023-02-23
SWC-6391 - Add recovery code login

nickgros:
- Add and integrate RecoveryCodeForm.tsx
- Return isLoading state from useLogin, determined now by using `useMutation`
- Update UsernamePasswordForm to use HTML5 validation (required text fields)

https://user-images.githubusercontent.com/17580037/221054048-1b4da6cc-1cc7-4140-a3f2-bc1625dfacd6.mov

### DIFF
--- a/packages/synapse-react-client/src/lib/containers/auth/LoginPage.tsx
+++ b/packages/synapse-react-client/src/lib/containers/auth/LoginPage.tsx
@@ -35,6 +35,14 @@ const Tagline: StyledComponent<TypographyProps> = styled(Typography, {
   font: '300 24px/34px Lato, sans-serif',
 }))
 
+const BACKUP_CODE_INFO_JSX = (
+  <>
+    Your backup code is a 16-digit code. We generated 10 of these codes for you
+    when you set up 2FA. If you don’t have access to these codes, please{' '}
+    <Link href={LOST_ACCOUNT_ACCESS_CONTACT_URL}>contact us</Link>.
+  </>
+)
+
 export default function LoginPage(props: LoginPageProps) {
   const { ssoRedirectUrl, sessionCallback } = props
   const showDesktop = useShowDesktop(910)
@@ -87,9 +95,7 @@ export default function LoginPage(props: LoginPageProps) {
       )}
       {!showDesktop && step === 'RECOVERY_CODE' && (
         <Typography variant={'body1'} align={'center'} sx={{ my: 1 }}>
-          Your backup code is a 16-digit code. We generated 10 of these codes
-          for you when you set up 2FA. If you don’t have access to these codes,
-          please <Link href={LOST_ACCOUNT_ACCESS_CONTACT_URL}>contact us</Link>.
+          {BACKUP_CODE_INFO_JSX}
         </Typography>
       )}
       <LoginForm
@@ -156,13 +162,7 @@ export default function LoginPage(props: LoginPageProps) {
                     Enter your backup code
                   </Typography>
                   <Typography variant={'headline2'} sx={{ lineHeight: '30px' }}>
-                    Your backup code is a 16-digit code. We generated 10 of
-                    these codes for you when you set up 2FA. If you don’t have
-                    access to these codes, please{' '}
-                    <Link href={LOST_ACCOUNT_ACCESS_CONTACT_URL}>
-                      contact us
-                    </Link>
-                    .
+                    {BACKUP_CODE_INFO_JSX}
                   </Typography>
                 </>
               )}
--- a/packages/synapse-react-client/src/lib/containers/auth/LoginPage.tsx
+++ b/packages/synapse-react-client/src/lib/containers/auth/LoginPage.tsx
@@ -3,6 +3,7 @@ import {
   Box,
   Stack,
   styled,
+  Link,
   Typography,
   TypographyProps,
   useTheme,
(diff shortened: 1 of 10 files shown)


packages/synapse-react-client/src/lib/containers/auth/LoginPage.tsx
```diff
@@ -14,6 +15,14 @@ import useLogin from '../../utils/hooks/useLogin'
 import LoginForm from './LoginForm'
 import { StyledComponent } from '@emotion/styled'
 
+const LOST_ACCOUNT_SERVICE_DESK_ISSUE_TITLE = encodeURIComponent(
+  'Lost access to my Synapse account',
+)
+const LOST_ACCOUNT_SERVICE_DESK_DESCRIPTION = encodeURIComponent(
+  '<Please provide your username and/or email address associated with your account.>',
+)
```
nickgros, line 23: How could we make this better?

packages/synapse-react-client/src/lib/containers/auth/LoginPage.tsx
```diff
@@ -122,17 +139,34 @@ export default function LoginPage(props: LoginPageProps) {
...
+                    Enter your backup code
+                  </Typography>
+                  <Typography variant={'headline2'} sx={{ lineHeight: '30px' }}>
+                    Your backup code is a 16-digit code. We generated 10 of
+                    these codes for you when you set up 2FA. If you don’t have
+                    access to these codes, please{' '}
+                    <Link href={LOST_ACCOUNT_ACCESS_CONTACT_URL}>
+                      contact us
+                    </Link>
+                    .
```
nickgros, line 165: TODO: This text matches the above, so I should create a reusable constant or component fragment

packages/synapse-react-client/src/lib/utils/hooks/useLogin.ts
```diff
@@ -26,6 +28,7 @@ export type UseLoginReturn = {
     otpType?: TwoFactorAuthOtpType,
   ) => Promise<void>
   errorMessage: string | undefined
+  isLoading: boolean
```
nickgros, line 31: I noticed Adam changed the button text/disabled status in the designs. It makes the login components feel much more responsive

packages/synapse-react-client/src/lib/utils/hooks/useLogin.ts
```diff
@@ -98,28 +101,80 @@ export default function useLogin(
...
-        const data = await SynapseClient.login(
-          username,
-          password,
-          authenticationReceipt,
-        )
-        if ('errorCode' in data) {
+  const {
+    mutateAsync: mutateLoginWithUsernameAndPassword,
+    isLoading: isLoadingLoginWithUsernameAndPassword,
+  } = useMutation<
```
nickgros, line 107: The easiest way to track loading state in an asynchronous request is `useMutation`. I don't think we'd reuse these, and they don't use query keys, so it made more sense to me to keep them here, rather than putting them with the rest of the react-query hooks

packages/synapse-react-client/src/lib/containers/auth/UsernamePasswordForm.tsx
```diff
@@ -58,16 +61,14 @@ export default function UsernamePasswordForm(props: UsernamePasswordFormProps) {
...
         variant="contained"
+        disabled={isLoading}
         sx={{
           height: '50px',
           mt: 4,
           mb: 2,
         }}
-        onClick={e => {
-          handleLogin(e)
-        }}
```
nickgros, line 68: Removing this handler has us fall back to the `form`'s `onSubmit`, which will let HTML 5 validation happen. The handler still has `preventDefault`, so the page won't reload.

packages/synapse-react-client/test/lib/containers/Login.test.tsx
```diff
@@ -35,22 +49,34 @@ describe('Functionality of Login Component', () => {
...
-    screen.getByLabelText('Username or Email Address', { exact: false })
-    screen.getByLabelText('Password', { exact: false })
+    await userEvent.type(
+      screen.getByLabelText('Username or Email Address', { exact: false }),
+      username,
+    )
+    await userEvent.type(
+      screen.getByLabelText('Password', { exact: false }),
+      password,
+    )
```
nickgros, line 65: With HTML5 validation now we actually have to enter u/p in the test

packages/synapse-react-client/test/lib/containers/Login.test.tsx
```diff
@@ -81,22 +119,78 @@ describe('Functionality of Login Component', () => {
...
+        userId: twoFactorAuthErrorResponse.userId,
+        otpCode: '123456',
+        otpType: 'TOTP',
+      })
       expect(callback).toHaveBeenCalled()
     })
   })
 
-  it('Log in with 2FA code supplied via props', async () => {
+  it('Log in with email + 2FA recovery code', async () => {
```
nickgros, line 132: new test (bad diff)

packages/synapse-react-client/src/lib/containers/auth/LoginPage.tsx
jay-hodgson: Is there a way to disable 2FA for a user once their identity is verified?  I see https://rest-docs.synapse.org/rest/DELETE/2fa.html, but that's for the current user only.  Would this need to be a direct database update?
@marcomarasca

packages/synapse-react-client/src/lib/utils/hooks/useLogin.ts
```diff
@@ -98,28 +101,80 @@ export default function useLogin(
...
+  )
+
+  const {
+    mutateAsync: mutateLoginWith2FACode,
+    isLoading: isLoadingLoginWith2FACode,
+  } = useMutation<LoginResponse, SynapseClientError, TwoFactorAuthLoginRequest>(
+    SynapseClient.loginWith2fa,
+    {
+      onError: e => {
+        setErrorMessage(e.reason)
```
jay-hodgson, line 137: This error handling feels unstable (looking at the error reason text to determine how to handle instead of checking an associated error code, and the cleanup via regex).

jay-hodgson:
I'm confused by the build output.  It looks like the DetailsView.test.tsx suite passed...

jay-hodgson (approved):

packages/synapse-react-client/src/lib/utils/hooks/useLogin.ts
nickgros: https://sagebionetworks.jira.com/browse/PLFM-7709

nickgros:
> I'm confused by the build output. It looks like the DetailsView.test.tsx suite passed...

Because we have so many console warnings and errors, the console output isn't easy to read--and I think Nx truncates the output. The GitHub Actions workflow emits an artifact containing the HTML test report, which shows what actually failed.

PR #49, opened 2023-02-27
Ability to use generatePalette for future apps that want to integrate…

jay-hodgson:
… with OneSage


jay-hodgson:
To generate a palette in OneSage for apps outside of data dissemination portals and MTB.  See https://github.com/Sage-Bionetworks/SageAccountWeb#how-do-i-add-my-app-to-onesage

nickgros (approved):

PR #57, opened 2023-03-06
PORTALS-2527: support other links that require an intermediate screen

jay-hodgson:


jay-hodgson:
Screen Shot 2023-03-06 at 11 35 09 AM

jay-hodgson:
Screen Shot 2023-03-06 at 1 41 33 PM

Screen Shot 2023-03-06 at 1 41 49 PM

jay-hodgson:
Screen Shot 2023-03-06 at 2 03 47 PM
Screen Shot 2023-03-06 at 1 57 40 PM

apps/portals/src/portal-components/RedirectDialog.tsx
```diff
@@ -0,0 +1,125 @@
...
+    <strong>You are being redirected to Synapse to view this data.</strong>
+  </Typography>
+
+const isSynapseURL = (url:string) => {
+  if (!url) return false
+  return new URL(url).hostname.toLowerCase() === 'www.synapse.org'
+}
+
+const RedirectDialog = (props: RedirectDialogProps) => {
+  const [countdownSeconds, setCountdownSeconds] = React.useState<number | undefined>()
```
nickgros, line 41: Can we inline this so it's never undefined and also remove the effect below?

```suggestion
  const [countdownSeconds, setCountdownSeconds] = React.useState(isSynapseURL(redirectUrl) ? 10 : 30)
```

apps/portals/src/portal-components/RedirectDialog.tsx
```diff
@@ -0,0 +1,125 @@
...
+      setRedirectInstructions(isRedirectToSynapse ? synapseRedirectInstructions : 
+        redirectInstructionsMap[redirectUrl])
+    }
+  }, [redirectUrl])
+
+  React.useEffect(() => {
+    if (redirectUrl && !countdownSeconds) {
+      setCountdownSeconds(isSynapseURL(redirectUrl) ? 10 : 30)
+    }
+  }, [countdownSeconds, redirectUrl])
```
nickgros, line 73: Inlined initial state above
```suggestion
```

apps/portals/src/portal-components/RedirectDialog.tsx
```diff
@@ -0,0 +1,125 @@
...
+
+  const onClose = () => {
+    // cancel the redirect
+    onCancelRedirect()
+    // and reset countdown seconds
+    setCountdownSeconds(undefined)
```
nickgros, line 79:
```suggestion
    setCountdownSeconds(isSynapseURL(redirectUrl) ? 10 : 30)
```

Consider creating a variable to store `isSynapseURL(redirectUrl) ? 10 : 30` instead of duplicating it here and in the suggestion above

apps/portals/src/portal-components/RedirectDialog.tsx
```diff
@@ -0,0 +1,125 @@
...
+      <strong>You are currently being redirected to the Multi-Consortia Coordinating (MC2) Center Intranet.</strong>
+    </Typography>
+    <Typography variant="body1" sx={{paddingBottom: '20px'}}>
+      The MC2 Center Intranet is a protected website containing MC2 Center resources. If you are not a participating member of the MC2 Center community, you will not be able to access the intranet.
+    </Typography>
+    <Typography variant="body1" sx={{paddingBottom: '20px'}}>
+      Feel free to <Link target="_blank" rel="noreferrer" href="https://help.cancercomplexity.synapse.org/doc/contact-us">contact the MC2 Center</Link> to learn more, or if you have any questions.
+    </Typography>
+  </>
+}
```
nickgros, line 28: Seems like this should be in the routesConfig, but nbd for one case

nickgros (changes requested):

jay-hodgson:
Merged suggested changes, but they introduced a bug in the initialized value of countdownSeconds (when switching from a synapse.org link to this other link, or visa versa).  The value in the state variable was initialized once, where I was hoping it would be reinitialized when the redirect URL prop value changed.  Reverted back to the strategy of setting the countdownSeconds to undefined when it needs to be configured, and the useEffect (on mount) to figure out the correct value based on the current redirect URL

nickgros (approved):
I see. Seems like we need complexity in one way or another, so this is fine. Sorry to lead you astray.

PR #59, opened 2023-03-07
SWC-6394 - Add 2FA recovery code generation

nickgros:


packages/synapse-react-client/src/lib/assets/icons/FrontHand.tsx
```diff
@@ -0,0 +1,26 @@
+import React from 'react'
+import SvgIcon, { SvgIconProps } from '@mui/material/SvgIcon'
+
+export default function FrontHand(props: SvgIconProps) {
```
nickgros, line 4: image

packages/synapse-react-client/src/lib/utils/theme/DefaultTheme.ts
```diff
@@ -8,6 +8,18 @@ const defaultMuiTheme: ThemeOptions = {
   typography: typographyOptions,
   palette: palette,
   components: {
+    MuiAlert: {
```
nickgros, line 11: Small tweaks to make MUI alert look like our custom FullWidthAlert built on top of Bootstrap. Icons + buttons probably won't be right, but this will be usable for inline alerts.

jay-hodgson (approved):

PR #62, opened 2023-03-08
SWC-6393 - Add 2FA panel for settings page

nickgros:
image

image

### DIFF
--- a/packages/synapse-react-client/src/lib/containers/IconSvg.tsx
+++ b/packages/synapse-react-client/src/lib/containers/IconSvg.tsx
@@ -199,6 +199,7 @@ export const IconStrings = [
   'chevronLeft',
   'database',
   'close',
+  'cross',
   'verticalEllipsis',
   'sync',
   'clipboardCheck',
@@ -234,7 +235,7 @@ export const IconStrings = [
   'addConditions',
 ] as const
 
-export type Icon = typeof IconStrings[number]
+export type Icon = (typeof IconStrings)[number]
 
 export type IconSvgProps = {
   icon: Icon
@@ -359,6 +360,7 @@ function IconMapping(props: { icon: string; sx?: SxProps }) {
     case 'helpOutlined':
       return <HelpOutlined sx={sx}></HelpOutlined>
     case 'close':
+    case 'cross':
       return <CloseTwoTone sx={sx} />
     case 'expandLess':
       return <ExpandLess sx={sx}></ExpandLess>
--- a/packages/synapse-react-client/src/lib/containers/auth/TwoFactorAuthSettingsPanel.tsx
+++ b/packages/synapse-react-client/src/lib/containers/auth/TwoFactorAuthSettingsPanel.tsx
@@ -1,0 +1,100 @@
+import { Box, Button, Typography, Stack, Skeleton } from '@mui/material'
+import React from 'react'
+import IconSvg from '../IconSvg'
+import { TWO_FACTOR_DOCS_LINK } from './TwoFactorEnrollmentForm'
+import {
+  useDisableTwoFactorAuth,
+  useGetTwoFactorEnrollmentStatus,
+} from '../../utils/hooks/SynapseAPI/auth/useTwoFactorEnrollment'
+import ConditionalWrapper from '../utils/ConditionalWrapper'
+import { displayToast } from '../ToastMessage'
+
+export type TwoFactorAuthSettingsPanelProps = {
+  onRegenerateBackupCodes: () => void
+  onBeginTwoFactorEnrollment: () => void
+}
+
+export default function TwoFactorAuthSettingsPanel(
+  props: TwoFactorAuthSettingsPanelProps,
+) {
+  const { onRegenerateBackupCodes, onBeginTwoFactorEnrollment } = props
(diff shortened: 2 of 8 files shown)


packages/synapse-react-client/src/lib/containers/IconSvg.tsx
```diff
@@ -359,6 +360,7 @@ function IconMapping(props: { icon: string; sx?: SxProps }) {
     case 'helpOutlined':
       return <HelpOutlined sx={sx}></HelpOutlined>
     case 'close':
+    case 'cross':
```
nickgros, line 363: Same icon but different semantic meaning should make this a bit easier to refactor in the future, if needed.

packages/synapse-react-client/src/lib/containers/auth/TwoFactorAuthSettingsPanel.tsx
```diff
@@ -0,0 +1,100 @@
...
+) {
+  const { onRegenerateBackupCodes, onBeginTwoFactorEnrollment } = props
+  const { data: status, isLoading: isLoadingStatus } =
+    useGetTwoFactorEnrollmentStatus()
+  const isActivated = status?.status === 'ENABLED'
+
+  const { mutate: disable2FA, isLoading: isLoadingMutation } =
+    useDisableTwoFactorAuth({
+      onSuccess: () => {
+        displayToast('2FA removed from this account', 'info')
```
nickgros, line 28: Designs use different style toast in SageAccountWeb, not sure if that will use the same API or if this should be a prop.

packages/synapse-react-client/src/lib/containers/auth/TwoFactorAuthSettingsPanel.tsx
```diff
@@ -0,0 +1,100 @@
...
+        <Typography
+          variant={'body1'}
+          color={isActivated ? 'success.main' : 'error.main'}
+          sx={{ my: 2, display: 'flex', alignItems: 'center' }}
+        >
+          <IconSvg
+            icon={isActivated ? 'check' : 'cross'}
+            sx={{ mr: 1, height: '24px' }}
+          />
+          {isLoading ? 'Loading...' : isActivated ? 'Active' : 'Inactive'}
```
nickgros, line 49: `'Loading...'` is never seen but it makes it easier to right warning-free tests, since the 'Active'/'Inactive' text won't be queryable until the API call completes.

packages/synapse-react-client/stories/TwoFactorAuthSettingsPanel.stories.tsx
```diff
@@ -0,0 +1,36 @@
...
+import { displayToast } from '../src/lib/containers/ToastMessage'
+import TwoFactorAuthSettingsPanel from '../src/lib/containers/auth/TwoFactorAuthSettingsPanel'
+import { Paper } from '@mui/material'
+
+const meta = {
+  title: 'Authentication/TwoFactorSettingsPanel',
+  component: TwoFactorAuthSettingsPanel,
+  render: args => {
+    return (
+      <Paper sx={{ p: 7, mx: 'auto', width: '720px' }}>
```
nickgros, line 12: Wrapped in paper for Storybook, which we would probably do in SageAccountWeb, while we can wrap it in the lightbox in SWC

packages/synapse-react-client/src/lib/containers/auth/TwoFactorAuthSettingsPanel.tsx
jay-hodgson: I think this is fine for SageAccountWeb for now.  If design thinks this is a high priority issue, we can work it into a future sprint.

packages/synapse-react-client/src/lib/containers/auth/TwoFactorAuthSettingsPanel.tsx
```diff
@@ -0,0 +1,100 @@
...
+        </Typography>
+      </ConditionalWrapper>
+      <Typography variant={'body1Italic'} sx={{ my: 1 }}>
+        Required to satisfy certain access requirements, and recommended for
+        overall account security.
+      </Typography>
+      <Typography variant={'body1'} sx={{ my: 2 }}>
+        Synapse uses a time-based system, which you can set up using an app like
+        Google Authenticator, Duo Mobile, Microsoft Authenticator, or Authy.
```
jay-hodgson, line 58: I'm so happy we are suggesting Authy, I love that app.

packages/synapse-react-client/src/lib/containers/auth/TwoFactorAuthSettingsPanel.tsx
```diff
@@ -0,0 +1,100 @@
...
+              disabled={isLoading}
+              onClick={() => {
+                onRegenerateBackupCodes()
+              }}
+            >
+              Regenerate Backup Codes
+            </Button>
+          </ConditionalWrapper>
+        )}
+        <Button
```
jay-hodgson, line 91: I think we should hide this until the docs link is valid (I don't want this sneaking out into production).

jay-hodgson (approved):

PR #63, opened 2023-03-08
SWC-6365

jay-hodgson:


packages/synapse-react-client/src/lib/containers/oauth/CreateOAuthClient.tsx
```diff
@@ -64,17 +93,21 @@ export const CreateOAuthModal: React.FunctionComponent<
...
-    const propArr = client?.redirect_uris
-
-    client &&
-    (stateArr?.length !== propArr?.length ||
-      !stateArr?.every(el => propArr?.includes(el)) ||
-      sectorUri != client?.sector_identifier_uri)
-      ? setWarnTrigger(true)
-      : setWarnTrigger(false)
-  }, [redirectUris, sectorUri, client])
+  useDebouncedEffect(
```
jay-hodgson, line 96: This is called when the client-side oAuthClient is updated from the UI, so I debounce the effect.

packages/synapse-react-client/src/lib/containers/oauth/CreateOAuthClient.tsx
```diff
@@ -64,17 +93,21 @@ export const CreateOAuthModal: React.FunctionComponent<
...
+    () => {
+      if (accessToken) {
+        // SWC-6365: use the pre-check service to determine if we need to show a warning on edit
+        SynapseClient.isOAuthClientReverificationRequired(
+          getOAuthClient(),
+          accessToken,
+        ).then(precheckResult => {
+          setWarnTrigger(precheckResult.reverificationRequired)
+        })
```
jay-hodgson, line 105: If there's an error in the precheck, I think the update will also fail (at which point the user will see the error).

packages/synapse-react-client/src/lib/utils/SynapseClient.ts
```diff
@@ -2445,6 +2446,23 @@ export const updateOAuthClient = (
   )
 }
 
+/**
+ 
+ Note: Only the creator of a client can update it.
+ https://rest-docs.synapse.org/rest/PUT/oauth2/client/id/verificationPrecheck.html
+ */
+export const isOAuthClientReverificationRequired = (
```
jay-hodgson, line 2454: New API call

packages/synapse-react-client/src/lib/containers/oauth/CreateOAuthClient.tsx
```diff
@@ -53,6 +57,31 @@ export const CreateOAuthModal: React.FunctionComponent<
...
+  }, [
+    client?.client_id,
+    client?.etag,
+    clientName,
+    clientUri,
+    policyUri,
+    redirectUris,
+    sectorUri,
+    tosUri,
+  ])
```
nickgros, line 83: This is being invoked every render, so can simplify to `useMemo`
```suggestion
  const oAuthClient: OAuthClient = useMemo(() => {
    return {
      client_id: client?.client_id,
      client_name: clientName,
      redirect_uris: redirectUris?.map(str => str.uri) ?? [''],
      policy_uri: policyUri,
      client_uri: clientUri,
      sector_identifier_uri: sectorUri ?? '',
      tos_uri: tosUri,
      etag: client?.etag,
    }
  }, [
    client?.client_id,
    client?.etag,
    clientName,
    clientUri,
    policyUri,
    redirectUris,
    sectorUri,
    tosUri,
  ])
```

packages/synapse-react-client/src/lib/containers/oauth/CreateOAuthClient.tsx
```diff
@@ -64,17 +93,21 @@ export const CreateOAuthModal: React.FunctionComponent<
...
-      sectorUri != client?.sector_identifier_uri)
-      ? setWarnTrigger(true)
-      : setWarnTrigger(false)
-  }, [redirectUris, sectorUri, client])
+  useDebouncedEffect(
+    () => {
+      if (accessToken) {
+        // SWC-6365: use the pre-check service to determine if we need to show a warning on edit
+        SynapseClient.isOAuthClientReverificationRequired(
+          getOAuthClient(),
```
nickgros, line 101: Following above suggestion
```suggestion
          oAuthClient,
```

packages/synapse-react-client/src/lib/containers/oauth/CreateOAuthClient.tsx
nickgros: Makes sense, we probably don't want to show an outright error until they click submit

packages/synapse-react-client/src/lib/containers/oauth/CreateOAuthClient.tsx
```diff
@@ -128,16 +161,7 @@ export const CreateOAuthModal: React.FunctionComponent<
...
-          client_id: client?.client_id,
-          client_name: clientName,
-          redirect_uris: redirectUris?.map(str => str.uri) ?? [''],
-          policy_uri: policyUri,
-          client_uri: clientUri,
-          sector_identifier_uri: sectorUri ?? '',
-          tos_uri: tosUri,
-          etag: client?.etag,
-        }
+        const oAuthClient = getOAuthClient()
```
nickgros, line 164: Following above suggestion
```suggestion
```

packages/synapse-react-client/src/lib/containers/oauth/CreateOAuthClient.tsx
```diff
@@ -64,17 +93,21 @@ export const CreateOAuthModal: React.FunctionComponent<
...
+        // SWC-6365: use the pre-check service to determine if we need to show a warning on edit
+        SynapseClient.isOAuthClientReverificationRequired(
+          getOAuthClient(),
+          accessToken,
+        ).then(precheckResult => {
+          setWarnTrigger(precheckResult.reverificationRequired)
+        })
+      }
+    },
+    [accessToken, getOAuthClient],
```
nickgros, line 108: Following above suggestion
```suggestion
    [accessToken, oAuthClient],
```

nickgros (approved):
LGTM, one suggested simplification

jay-hodgson:
Memorized value (via useMemo()) rather than the function (useCallback()), as Nick recommended.

PR #70, opened 2023-03-15
PORTALS-2580 - refactor for synapse-oauth-signin + misc fixes

nickgros:
Changes I've made in synapse-oauth-signin are dependent upon the changes to useDetectSSOCode, which is how this work is related to PORTALS-2580

-----

- Refactor useDetectSSOCode so SSO redirect logic can be captured in-app
- Upgrade mui-one-time-password-input to 1.1.0 and use new `autoFocus`
- Fix dialog UI in portals

-----

Current portals login dialog needs more padding: 

image

after changes:
image

### DIFF
--- a/apps/portals/craco.config.js
+++ b/apps/portals/craco.config.js
@@ -2,19 +2,26 @@ const { ProvidePlugin } = require('webpack')
 
 module.exports = {
   webpack: {
-    configure: {
-      resolve: {
-        fallback: {
-          path: require.resolve('path-browserify'),
-          timers: require.resolve('timers-browserify'),
-          fs: require.resolve('memfs'),
-          https: require.resolve('https-browserify'),
-          stream: require.resolve('stream-browserify'),
-          http: require.resolve('stream-http'),
-          events: require.resolve('events'),
-        },
-      },
-      ignoreWarnings: [/Failed to parse source map/],
+    configure: (webpackConfig) => {
+      const scopePluginIndex = webpackConfig.resolve.plugins.findIndex(
+        ({ constructor }) =>
+          constructor && constructor.name === 'ModuleScopePlugin',
+      )
+
+      webpackConfig.resolve.plugins.splice(scopePluginIndex, 1)
+
+      webpackConfig.resolve.fallback = {
+        ...webpackConfig.resolve.fallback,
+        path: require.resolve('path-browserify'),
+        timers: require.resolve('timers-browserify'),
+        fs: require.resolve('memfs'),
+        https: require.resolve('https-browserify'),
+        stream: require.resolve('stream-browserify'),
+        http: require.resolve('stream-http'),
+        events: require.resolve('events'),
+      }
+      webpackConfig.ignoreWarnings = [/Failed to parse source map/]
+      return webpackConfig
     },
     plugins: {
       add: [
@@ -28,6 +35,7 @@ module.exports = {
   eslint: {
     enable: false,
   },
+
   // babel: {
   //   presets: ['@babel/preset-react'],
   //   plugins: [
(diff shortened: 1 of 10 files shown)


packages/synapse-react-client/src/lib/utils/hooks/useDetectSSOCode.ts
```diff
@@ -53,26 +52,17 @@ export default function useDetectSSOCode(
...
     if (code && provider) {
       const redirectUrl = `${redirectURL}?provider=${provider}`
-      const redirectAfterSuccess = () => {
-        // go back to original route after successful SSO login
-        const originalUrl = localStorage.getItem('after-sso-login-url')
-        localStorage.removeItem('after-sso-login-url')
-        if (originalUrl) {
-          window.location.replace(originalUrl)
-        }
-      }
```
nickgros, line 63: This redirect logic seems to me like it should be app-specific, so it's now an argument `onSignInComplete`

packages/synapse-react-client/src/lib/utils/hooks/useDetectSSOCode.ts
```diff
@@ -16,12 +16,10 @@ import { TwoFactorAuthErrorResponse } from '../synapseTypes/ErrorResponse'
 import { PROVIDERS } from '../../containers/auth/AuthenticationMethodSelection'
 
 type UseDetectSSOCodeOptions = {
+  onSignInComplete?: () => void
   registerAccountUrl?: string
   onError?: (err: unknown) => void
-  onTwoFactorAuthRequired?: (
-    resp: TwoFactorAuthErrorResponse,
-    callback: () => void,
```
nickgros, line 23: This `callback` parameter is unused, I think it was mistakenly thrown in when I originally refactored this.

apps/portals/src/utils.ts
```diff
@@ -12,3 +12,24 @@ export const DESKTOP_VIEWPORT_MIN_WIDTH_MEDIA_QUERY = `(min-width: ${DESKTOP_VIE
 
 export const useShowDesktop = () =>
   useMedia(DESKTOP_VIEWPORT_MIN_WIDTH_MEDIA_QUERY)
+
```
nickgros, line 15: May move these functions back into SRC or a different app utils project when we have multiple apps in the monorepo and we want to refactor duplicate code

apps/portals/craco.config.js
```diff
@@ -2,19 +2,26 @@ const { ProvidePlugin } = require('webpack')
...
-        },
-      },
-      ignoreWarnings: [/Failed to parse source map/],
+    configure: (webpackConfig) => {
+      const scopePluginIndex = webpackConfig.resolve.plugins.findIndex(
+        ({ constructor }) =>
+          constructor && constructor.name === 'ModuleScopePlugin',
+      )
+
+      webpackConfig.resolve.plugins.splice(scopePluginIndex, 1)
```
nickgros, line 11: Ran into a strange issue where a module within `@babel/runtime` was being imported from the root `node_modules` and it didn't recognize it as an external dependency, so the `ModuleScopePlugin` failed saying that the imported module wasn't in `./src`. 

It's not clear to me what change actually caused this issue to surface.

I couldn't find a real fix for this other than removing the plugin. This seemed to not cause problems for now, and I want to replace webpack with vite soon, so I think this is an acceptable stopgap.

apps/portals/src/Navbar.tsx
```diff
@@ -212,38 +214,38 @@ class Navbar extends React.Component<any, State> {
                   >
                     SIGN&nbsp;IN
                   </Button>
-                  <Modal
-                    animation={false}
-                    backdrop="static"
-                    keyboard={false}
-                    onHide={() => {
+                  <Dialog
```
nickgros, line 217: Use @mui instead of react-bootstrap

apps/portals/craco.config.js
jay-hodgson, line 11: Why does the ModuleScopePlugin need to be removed?

jay-hodgson (approved):

PR #81, opened 2023-03-20
Upgrade TypeScript to 5.0

nickgros:


shared/tsconfig.base.json
```diff
@@ -22,6 +22,7 @@
     "skipLibCheck": true,
     "strictNullChecks": true,
     "suppressImplicitAnyIndexErrors": true,
+    "ignoreDeprecations": "5.0",
```
nickgros, line 25: suppressImplicitAnyIndexErrors is deprecated and will be removed in TS5.5. We will need to refine our types in some places to remove this flag.

nickgros: https://sagebionetworks.jira.com/browse/PORTALS-2598

apps/synapse-oauth-signin/package.json
```diff
@@ -66,7 +66,7 @@
     "markdown-it-synapse-math": "^3.0.4",
     "markdown-it-synapse-table": "^1.0.6",
     "memfs": "^3.4.7",
-    "msw": "1.0.1",
```
nickgros, line 69: Upgrading MSW past v1 in SRC seems to break a few things. Let's use the same version (0.49.3) across the monorepo

jay-hodgson (approved):

bleu4422 (approved):
[]()

PR #85, opened 2023-03-21
PORTALS-2162 - Add IN, HAS_LIKE to sqlOperator options

nickgros:
- Refactor so SQLOperator uses ColumnSingleValueFilterOperator | ColumnMultiValueFunction
- Use native filter for 'EQUAL'
- Add 'IN', 'HAS_LIKE' options

### DIFF
--- a/packages/synapse-react-client/test/lib/utils/functions/sqlFunctions.test.ts
+++ b/packages/synapse-react-client/test/lib/utils/functions/sqlFunctions.test.ts
@@ -1,8 +1,8 @@
 import {
+  generateQueryFilterFromSearchParams,
+  parseEntityIdAndVersionFromSqlStatement,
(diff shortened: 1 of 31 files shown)


packages/synapse-react-client/test/lib/utils/functions/sqlFunctions.test.ts
```diff
@@ -217,4 +255,17 @@ describe('generateQueryFilterFromSearchParams', () => {
     const result = generateQueryFilterFromSearchParams(searchParams, operator)
     expect(result).toStrictEqual(expectedResults)
   })
+
+  it('handles every filter type', () => {
```
nickgros, line 259: If we add a new filter method, this test will fail unless we add a new case to `generateQueryFilterFromSearchParams`

PR #91, opened 2023-03-24
SWC-6395 - Allow line breaks in dropdown menu button tooltip

nickgros:


jay-hodgson (approved):

jay-hodgson:
Note, test failure is in synapse-oauth-signin (src/App.test.tsx) - App integration tests › Shows login when no token is provi...

PR #98, opened 2023-03-27
PORTALS-2581 - Google SSO + Synapse 2FA in SageAccountWeb

nickgros:
- Add SSO + 2FA handling to SageAccountWeb
- Refactor SSO redirect utils from portals to synapse-react-client
- Use SSO redirect utils in all 3 apps

### DIFF
--- a/apps/SageAccountWeb/src/AppInitializer.tsx
+++ b/apps/SageAccountWeb/src/AppInitializer.tsx
@@ -14,6 +14,9 @@ import { ThemeOptions } from '@mui/material'
 import { useSourceApp } from 'components/SourceApp'
 import { deepmerge } from '@mui/utils'
 import theme from './style/theme'
+import { TwoFactorAuthSSOContextProvider } from './TwoFactorAuthSSOContext'
+import { TwoFactorAuthErrorResponse } from 'synapse-react-client/dist/utils/synapseTypes/ErrorResponse'
+import { redirectAfterSSO } from 'synapse-react-client/dist/utils/AppUtils'
 
 /**
  * State and helpers for managing a user session in the portal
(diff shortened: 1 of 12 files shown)


apps/SageAccountWeb/src/AppInitializer.tsx
```diff
@@ -192,20 +202,25 @@ function AppInitializer(props: { children?: React.ReactNode }) {
...
-        synapseContext={{
-          accessToken: token,
-          isInExperimentalMode: SynapseClient.isInSynapseExperimentalMode(),
-          utcTime: SynapseClient.getUseUtcTimeFromCookie(),
-          downloadCartPageUrl: '',
-        }}
-        theme={themeOptions}
+      <TwoFactorAuthSSOContextProvider
+        context={{ twoFactorAuthErrorResponse: twoFactorAuthError }}
       >
```
nickgros, line 207: Needed to add `TwoFactorAuthSSOContext` because

- `useDetectSSOCode` gets the data we need to complete SSO sign in with 2FA. This is called from `AppInitializer`.

- The `LoginPage` components that need to receive the code are created by `App`, which renders both `LoginPage` as a child of `AppInitializer`

Since `AppInitializer` doesn't know anything about `LoginPage`, the only way to pass the data from `AppInitializer` down into `LoginPage` is through context. There are probably ways we could refactor this to avoid this indirection, but I'm not yet very familiar with this code base, so I'm not equipped to make that kind of change.

apps/SageAccountWeb/src/LoginPage.tsx
```diff
@@ -32,16 +32,16 @@ const LoginPage: React.FunctionComponent<LoginPageProps> = ({
               <SourceAppLogo />
             </div>
             <StandaloneLoginForm
-              sessionCallback={() => setIsSessionEstablished(true)}
+              sessionCallback={() => {
+                redirectAfterSSO(returnToUrl)
+              }}
```
nickgros, line 37: Not sure if the original state variable was used because this is unclear. If so, we could instead clarify the prop name by renaming it to something like `onSessionEstablished`

jay-hodgson (approved):

PR #110, opened 2023-03-30
PORTALS-2566: Polish Items for Verification Step 3 (Pledge)

hallieswan:
. | main | feature
:---:|:---:|:---:
validate flow | main_authenticated_validate | feature_authenticated_validate
registration flow | main_authenticated_signTermsOfUse | feature_authenticated_signTermsOfUse

Other changes: 
- Refactor right panel text into its own component, so can be used in both T&C pages
- Refactor T&C link into its own component, so can be used in both T&C pages
- Refactor T&C URL into SRC constant, so can be used in SRC, SageAccountWeb, and portals
- Add signature box below validate flow T&C, as in [Figma](https://www.figma.com/file/TOcmJ2QHr51v8Huv3wIec9/Sage-Registration-App?node-id=1746-44457&t=44Jb5cnH6IP9Bogj-4)


apps/SageAccountWeb/src/components/ProfileValidation/ProfileValidation.tsx
```diff
@@ -431,7 +425,17 @@ export const ProfileValidation = (props: ProfileValidationProps) => {
...
+          sx={
+            step === ValidationWizardStep.TERMS_AGREE
+              ? {
+                  width: '1200px',
+                  '& > div:nth-of-type(1)': { paddingTop: 10, width: '750px' },
+                  '& > div:nth-of-type(2)': { paddingTop: 10 },
+                }
+              : null
+          }
+        >
```
hallieswan, line 438: Extra width for T&C

apps/SageAccountWeb/src/components/ProfileValidation/TermsAndConditionsSignature.tsx
```diff
@@ -0,0 +1,49 @@
...
+}
+
+export const TermsAndConditionsSignature: React.FC<
+  TermsAndConditionsSignatureProps
+> = ({ canSign, onSigned, expectedSignature }) => {
+  return (
+    <>
+      <Typography variant="body1" sx={tcSignatureTextSx}>
+        After agreeing to the above terms, electronically sign this agreement by
+        typing your full name, first and last, to acknowledge your agreement
```
nickgros, line 29: We hardly ever do i18n or localization, and maybe check with Adam on this (or just ignore it)--asking for 'full name' seems to be sufficient where clarifying for 'first and last name' adds some cultural assumptions that might not make sense for every user
```suggestion
        typing your full name to acknowledge your agreement
```

apps/SageAccountWeb/src/components/ProfileValidation/ProfileValidation.tsx
```diff
@@ -431,7 +425,17 @@ export const ProfileValidation = (props: ProfileValidationProps) => {
...
     <StyledOuterContainer>
       {step !== ValidationWizardStep.THANK_YOU ? (
-        <StyledInnerContainer>
+        <StyledInnerContainer
+          sx={
+            step === ValidationWizardStep.TERMS_AGREE
+              ? {
+                  width: '1200px',
+                  '& > div:nth-of-type(1)': { paddingTop: 10, width: '750px' },
+                  '& > div:nth-of-type(2)': { paddingTop: 10 },
```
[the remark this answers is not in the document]
nickgros: Also would prefer to use the `useTheme` instead of importing the theme directly, but that might be an issue across the app so don't make that change now if it happens to break other things.

apps/SageAccountWeb/src/components/ProfileValidation/TermsAndConditionsSignature.tsx
```diff
@@ -0,0 +1,49 @@
+import { TextField } from '@mui/material'
+import { StyledFormControl } from 'components/StyledComponents'
+import React from 'react'
+import { Typography } from 'synapse-react-client'
```
nickgros, line 4: Would prefer to import `Typography` from MUI instead of SRC but if that ends up breaking things, don't worry about it for now. I suspect there are a few problems throughout the app related to how we're applying the theme

packages/synapse-react-client/src/lib/utils/SynapseConstants.ts
```diff
@@ -135,3 +135,7 @@ export const NETWORK_UNAVAILABLE_MESSAGE =
 export const REJECT_SUBMISSION_CANNED_RESPONSES_TABLE = 'syn50683097'
 export const AUTHENTICATION_RECEIPT_LOCALSTORAGE_KEY =
   'last_user_authentication_receipt'
+
+/** URL constants */
+export const URL_TERMS_CONDITIONS_AGREEMENT =
+  'https://s3.amazonaws.com/static.synapse.org/governance/SageBionetworksSynapseTermsandConditionsofUse.pdf?v=5'
```
nickgros, line 141: This is so much better here

nickgros (changes requested):
Looks great! Couple of minor suggestions, most can be disregarded if they aren't a simple change

apps/SageAccountWeb/src/components/ProfileValidation/TermsAndConditionsSignature.tsx
hallieswan: I'll update the import -- it doesn't cause any noticeable differences here. It does look like Typography is imported from synapse-react-client in several other components, so may be something to clean up in the future

apps/SageAccountWeb/src/components/ProfileValidation/ProfileValidation.tsx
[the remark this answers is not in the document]
hallieswan: Agreed, I was confused too! I'll change to theme.spacing and use the `useTheme` hook in these components, it doesn't seem to break anything here.

apps/SageAccountWeb/src/components/ProfileValidation/TermsAndConditionsSignature.tsx
hallieswan: Good point! Adam agreed and suggested using “…typing your full name, as it appears below, to acknowledge your agreement…” so the expected signature is clear. 

He also mentioned that since there's still [some discussion](https://sagebionetworks.slack.com/archives/C034RNAG800/p1680118990871579) around how to capture the signature, we should probably hold off on pushing out this feature. I'll comment out the component for now, so we can add it in the future if needed.

nickgros (approved):

PR #113, opened 2023-04-03
PORTALS-2567: Polish Items for Verification Step 4 (Upload)

hallieswan:
page | main | feature
:----:|:----:|:----:
sage account web attestation | main_authenticated_validate | feature_authenticated_validate
access requirements list | main_storybook_accessrequirementlist | feature_storybook_accessrequirementlist

Other Changes
- Refactor FileUpload to use MUI Button rather than bootstrap button
- Don't show error banner when user "cancels" FileUpload without selecting a file
- Remove outdated documentation and scripts for building portals with local build of SRC
- Restyle FileUpload button to be more visible

### DIFF
--- a/packages/synapse-react-client/src/lib/containers/FileUpload.tsx
+++ b/packages/synapse-react-client/src/lib/containers/FileUpload.tsx
@@ -1,19 +1,48 @@
 import * as React from 'react'
-import { Button } from 'react-bootstrap'
+import { Button, ButtonProps, styled } from '@mui/material'
 import { uploadFile } from '../utils/SynapseClient'
 import { useSynapseContext } from '../utils/SynapseContext'
 import { FileUploadComplete, UploadCallbackResp } from '../utils/synapseTypes'
+import { StyledComponent } from '@emotion/styled'
 
 export type FileUploadProps = {
   id?: string
   label?: string
-  variant?: string // allow you to change the appearance of the button to link (see react bootstrap doc)
+  buttonProps?: ButtonProps
   uploadCallback?: (response: UploadCallbackResp) => void
   context?: any // to distinguish which file was uploaded if we have several FileUpload components on the same page
 }
 
+// replicate bootstrap styling
+const StyledUploadButton: StyledComponent<ButtonProps> = styled(Button, {
+  label: 'StyledUploadButton',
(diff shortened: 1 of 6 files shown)


packages/synapse-react-client/src/lib/containers/FileUpload.tsx
```diff
@@ -24,7 +53,7 @@ const FileUpload: React.FC<FileUploadProps> = props => {
   }
 
   const changeHandler = async (e: React.ChangeEvent<HTMLInputElement>) => {
-    if (e.target.files) {
+    if (e.target.files && e.target.files.length > 0) {
       const file = e.target.files[0]
```
hallieswan, line 57: Don't show error banner when user "cancels" FileUpload without selecting a file

packages/synapse-react-client/src/lib/containers/FileUpload.tsx
```diff
@@ -1,19 +1,27 @@
...
+import { Button, ButtonProps } from '@mui/material'
 import { uploadFile } from '../utils/SynapseClient'
 import { useSynapseContext } from '../utils/SynapseContext'
 import { FileUploadComplete, UploadCallbackResp } from '../utils/synapseTypes'
 
 export type FileUploadProps = {
   id?: string
   label?: string
-  variant?: string // allow you to change the appearance of the button to link (see react bootstrap doc)
+  variant?: ButtonProps['variant']
```
nickgros, line 10: You could drop this prop and just include `variant` in the `buttonProps` object. You might need to add a default `buttonProps` value where `variant` is defined (I think it's a required prop). Up to you.

nickgros (approved):
LGTM!

packages/synapse-react-client/src/lib/containers/FileUpload.tsx
hallieswan: I originally included `variant` in `buttonProps`, but ended up moving it into `FileUploadProps` so it would be easier to switch between / see the possible variants in Storybook. I defer to your experience though -- if you think it fits better to group with the other `buttonProps`, then I'm happy to move it into `buttonProps`!

nickgros (approved):

PR #114, opened 2023-04-03
PORTALS-2582 - refactor to decouple SynapseContext from QueryClientContext and MUI ThemeContext

nickgros:
- decouple SynapseContext from QueryClientContext and ThemeContext
- Add FullContext to provide backwards-compatibility

SageAccountWeb handles its own MUI theme and QueryClient. It makes more sense to allow SageAccountWeb to just pull in the SynapseContext that it needs instead of having multiple (potentially-conflicting) MUI themes and QueryClient contexts. This commit prepares the repo so that change can be made.


packages/synapse-react-client/src/lib/umd.index.ts
```diff
@@ -54,19 +54,19 @@ import { SynapseClient, SynapseConstants } from './utils'
...
 import TwoFactorEnrollmentForm from './containers/auth/TwoFactorEnrollmentForm'
 import TwoFactorAuthSettingsPanel from './containers/auth/TwoFactorAuthSettingsPanel'
+import { FullContextProvider } from './utils/FullContextProvider'
 
 // Also include scss in the bundle
 import './style/main.scss'
 
 const SynapseContext = {
-  SynapseContextProvider,
+  SynapseContextProvider: FullContextProvider,
```
nickgros, line 69: This file is what we use to create the bundle of SRC components that SynapseWebClient uses. SynapseWebClient needs the FullContextProvider, and maintaining the name prevents this from being a breaking API change.

Having a breaking API change between SRC and SWC is not really a big deal, but this is one less change I have to worry about in a refactor with a lot of moving parts (beyond this PR).

packages/synapse-react-client/src/lib/utils/theme/useTheme.tsx
```diff
@@ -3,8 +3,11 @@ import { ThemeProvider as MuiThemeProvider } from '@mui/material/styles'
 import { merge } from 'lodash-es'
 import React, { useMemo } from 'react'
 import defaultMuiTheme from './DefaultTheme'
+import type { PartialDeep } from 'type-fest'
 
-export function mergeTheme(themeOverrides: ThemeOptions): ThemeOptions {
+export function mergeTheme(
+  themeOverrides: PartialDeep<ThemeOptions>,
+): ThemeOptions {
```
nickgros, line 10: No need to supply an entire theme. Use this `PartialDeep` utility to mark all fields and those fields' subfields as optional.

We have a `DeepPartial` utility, but I ended up running into recursive type limit issues using that, but not with this utility type.

packages/synapse-react-client/src/lib/umd.index.ts
```diff
@@ -66,7 +67,8 @@ import { FullContextProvider } from './utils/FullContextProvider'
 import './style/main.scss'
 
 const SynapseContext = {
-  SynapseContextProvider: FullContextProvider,
+  FullContextProvider,
+  SynapseContextProvider,
```
nickgros: I changed my mind on using an alias here too. Will have to make the change in SWC.

hallieswan (approved):
Looks good!

PR #124, opened 2023-04-11
Add missing ThemeProvider to synapse-oauth-signin

nickgros:
The theme was missing, so many components were unstyled.


hallieswan (approved):
Looks good!

PR #127, opened 2023-04-12
PORTALS-2604: button for 'view all charts' should live underneath the 'search field dropdown'

hallieswan:
Before: 

https://user-images.githubusercontent.com/26949006/231316716-582fc1a0-6bfc-48ec-90df-561c0d32134a.mov

After: 

https://user-images.githubusercontent.com/26949006/231316722-27556467-d54f-4428-8c77-4210d84fd157.mov


nickgros (approved):
Wonder why that had a z-index in the first place...

PR #129, opened 2023-04-12
PORTALS-2475: fix Storybook demo of toast without button or link text

hallieswan:
main | feature
:---:|:---:
main_storybook_toast | feature_storybook_toast

### DIFF
--- a/packages/synapse-react-client/demo/containers/playground/ToastDemo.tsx
+++ b/packages/synapse-react-client/demo/containers/playground/ToastDemo.tsx
@@ -15,19 +15,28 @@ export const ToastDemo = () => {
   const [linkText, setLinkText] = React.useState('')
 
   function fireToast() {
+    const primaryButtonConfig =
+      buttonText === ''
+        ? undefined
+        : {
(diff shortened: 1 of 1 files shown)


packages/synapse-react-client/demo/containers/playground/ToastDemo.tsx
```diff
@@ -15,19 +15,28 @@ export const ToastDemo = () => {
...
+              console.log('Primary button clicked!')
+            },
+          }
+    const secondaryButtonConfig =
+      linkText === ''
+        ? undefined
+        : {
+            text: linkText,
+            href: '#',
+          }
```
hallieswan, line 33: In ToastDemo, `buttonText` and `linkText` are empty strings by default. `FullWidthAlert`, which is called by `displayToast`, expects the button config to be `undefined` when there shouldn't be a button or link added to the alert. Update the demo to pass `undefined` when the user hasn't entered a value for `buttonText` or `linkText`, so buttons aren't added to the toast

nickgros (approved):

PR #130, opened 2023-04-13
PORTALS-2626

jay-hodgson:
Use a paragraph in a number of renderers to enable the expand behavior in the context of a Table Cell (but does not impact rendering in a Card context):
Screen Shot 2023-04-13 at 11 27 15 AM
Screen Shot 2023-04-13 at 12 21 36 PM


nickgros (approved):
LGTM

PR #132, opened 2023-04-13
SWC-6420: update rss-parser (and update plotly in a failed attempt to…

jay-hodgson:
… fix a d3-color vulnerable to ReDoS issue)


jay-hodgson:
Tested locally - the query plots and rss news feed component seem to function properly after the version bump

hallieswan (approved):
Looks good to me!

PR #133, opened 2023-04-14
PORTALS-2621 - Fix NF stale data issue by resetting table query on prop change

nickgros:
See PORTALS-2621

Fix a case where CardContainerLogic did not update when the `searchParams` prop updated. The `searchParams` prop is used to create the `initQueryRequest` object, which is passed into `useImmutableTableQuery`.

Before this change, `useImmutableTableQuery` was not configured to update the query state when the `initQueryRequest` argument changed.

### DIFF
--- a/apps/portals/src/SynapseComponent.tsx
+++ b/apps/portals/src/SynapseComponent.tsx
@@ -42,9 +42,8 @@ export const SynapseComponent: React.FC<SynapseComponentWithContextProps> = ({
 }) => {
   // return the synapse object but with token/search params injected into its props from the context created in AppInitializer
   const { props, ...rest } = synapseConfig
-  const key = JSON.stringify(props)
   return (
-    <SynapseContextConsumer key={key}>
+    <SynapseContextConsumer>
       {(ctx?: SynapseContextType) => {
         const propsWithSearchAndToken = {
           ...props,
--- a/packages/synapse-react-client/src/lib/containers/CardContainerLogic.tsx
+++ b/packages/synapse-react-client/src/lib/containers/CardContainerLogic.tsx
@@ -1,4 +1,4 @@
-import React, { useMemo } from 'react'
+import React from 'react'
 import { SynapseConstants } from '../utils'
 import {
   generateQueryFilterFromSearchParams,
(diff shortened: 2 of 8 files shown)


apps/portals/src/SynapseComponent.tsx
```diff
@@ -42,9 +42,8 @@ export const SynapseComponent: React.FC<SynapseComponentWithContextProps> = ({
 }) => {
   // return the synapse object but with token/search params injected into its props from the context created in AppInitializer
   const { props, ...rest } = synapseConfig
-  const key = JSON.stringify(props)
   return (
-    <SynapseContextConsumer key={key}>
```
nickgros, line 47: This key should not be necessary. I didn't notice any new issues after removing it and clicking around. It was added in 2019 with no linked issue in the associated PR - https://github.com/Sage-Bionetworks/portals/pull/145.

packages/synapse-react-client/src/lib/containers/CardContainerLogic.tsx
```diff
@@ -142,38 +142,49 @@ export type CardContainerLogicProps = {
  */
 export const CardContainerLogic = (props: CardContainerLogicProps) => {
   const entityId = parseEntityIdFromSqlStatement(props.sql)
-  const queryFilters = generateQueryFilterFromSearchParams(
+  const { sortConfig, columnAliases } = props
+  const initQueryRequest: QueryBundleRequest = useMemo(() => {
```
nickgros, line 146: Memoize initQueryRequest, since it would otherwise be recreated on each render.

packages/synapse-react-client/src/lib/containers/useImmutableTableQuery.ts
```diff
@@ -194,6 +195,11 @@ export default function useImmutableTableQuery(
     setQuery(initQueryRequest)
   }, [initQueryRequest, setQuery])
 
+  /* If the initial query changes, then reset the query to match the new prop */
+  useDeepCompareEffect(() => {
+    resetQuery()
+  }, [initQueryRequest])
+
```
nickgros, line 202: The issue occurred because `initQueryRequest` was only used to initialize state. Any updates to this prop would not trigger a re-render. In this hook, let's track the object for changes and re-render the component if it does change.

We will use `useDeepCompareEffect` because otherwise we would need to memoize this object in each component where this hook is used.

jay-hodgson (approved):

PR #134, opened 2023-04-14
update elportal logo

jay-hodgson:


jay-hodgson:
Note that the majority of this configuration is still just a copy of ADKP

jay-hodgson:
Here's the updated logo:
Screen Shot 2023-04-14 at 9 34 38 AM

hallieswan (approved):
Looks good!

PR #139, opened 2023-04-18
SWC-6412 - Input, Dialog component theme changes

nickgros:
In SWC-6412 I am reworking the AccessRequirementList component and submission flow. I'm using MUI where possible, so some theme updates were necessary.

- Move StyledFormControl to SRC
- Apply most StyledFormControl > InputBase styles to all InputBase components in the theme
- Use StyledFormControl in custom TextField component
- Apply 2px padding to modal content to ensure input borders and box shadow are shown; use `border-image` to hide first and last 2px of modal border

For input fields, SageAccountWeb uses StyledFormControl with an inner MUI TextField. This is flawed because TextField actually contains its own FormControl component. The input fields look fine for now (no visual change), but we should aim to fix this in the future.

#### Screenshots

Modal changes

|Before|After|
|---|---|
| image | image |

### DIFF
--- a/apps/SageAccountWeb/src/components/AccountSettings.tsx
+++ b/apps/SageAccountWeb/src/components/AccountSettings.tsx
@@ -42,7 +42,7 @@ import { InputLabel } from '@mui/material'
 import { TextField } from '@mui/material'
 import { useSourceAppConfigs } from './SourceApp'
 import TwoFactorAuthSettingsPanel from 'synapse-react-client/dist/containers/auth/TwoFactorAuthSettingsPanel'
-
+import SRCTextField from 'synapse-react-client/dist/containers/TextField'
 const CompletionStatus: React.FC<{ isComplete: boolean | undefined }> = ({
   isComplete,
 }) => {
@@ -399,24 +399,17 @@ export const AccountSettings = () => {
                       })}
                     </Grid>
                   </StyledFormControl>
-                  <StyledFormControl
+                  <SRCTextField
                     fullWidth
-                    variant="standard"
                     margin="normal"
-                    sx={{ marginBottom: '10px' }}
-                  >
-                    <InputLabel shrink htmlFor="bio">
-                      Bio
-                    </InputLabel>
-                    <TextField
-                      id="bio"
-                      name="bio"
-                      fullWidth
-                      multiline
-                      onChange={e => setBio(e.target.value)}
-                      value={bio}
-                    />
-                  </StyledFormControl>
+                    label="Bio"
+                    id="bio"
+                    name="bio"
+                    multiline
+                    rows={5}
+                    onChange={e => setBio(e.target.value)}
+                    value={bio}
+                  />
                   <div className="primary-button-container">
                     <Button
                       onClick={() => {
--- a/apps/SageAccountWeb/src/components/StyledComponents.ts
+++ b/apps/SageAccountWeb/src/components/StyledComponents.ts
@@ -1,19 +1,7 @@
-import {
-  alpha,
-  Box,
-  BoxProps,
-  Paper,
-  PaperProps,
-  FormControl,
-  FormControlProps,
-  formHelperTextClasses,
-  inputBaseClasses,
-  styled,
-  textFieldClasses,
-} from '@mui/material'
-import { latoFont } from 'style/theme'
+import { BoxProps, Paper, PaperProps, styled } from '@mui/material'
 import { StyledComponent } from '@mui/styles'
 import { StyledOuterContainer as _StyledOuterContainer } from 'synapse-react-client/dist/components/styled/LeftRightPanel'
+import { StyledFormControl as _StyledFormControl } from 'synapse-react-client/dist/components/styled/StyledFormControl'
 
 export const StyledOuterContainer: StyledComponent<BoxProps> =
   _StyledOuterContainer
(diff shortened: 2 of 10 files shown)


apps/SageAccountWeb/src/components/AccountSettings.tsx
```diff
@@ -399,24 +399,17 @@ export const AccountSettings = () => {
...
-                    sx={{ marginBottom: '10px' }}
-                  >
-                    <InputLabel shrink htmlFor="bio">
-                      Bio
-                    </InputLabel>
-                    <TextField
-                      id="bio"
-                      name="bio"
-                      fullWidth
-                      multiline
```
nickgros, line 415: The multiline styling broke here, but this is the only place where it's used, so just use the SRCTextField, which handles it properly.

apps/SageAccountWeb/src/components/StyledComponents.ts
```diff
@@ -38,56 +26,4 @@ export const StyledInnerContainer: StyledComponent<PaperProps> = styled(Paper, {
...
- usage: 
-        <StyledFormControl fullWidth variant="standard" margin="normal">
-          <InputLabel shrink htmlFor="someinput">
-            labelText
-          </InputLabel>
-          <TextField
-            id="someinput"/>
-        </StyledFormControl>
-*/
-export const StyledFormControl: StyledComponent<FormControlProps> = styled(
```
nickgros, line 51: Moved to SRC

apps/SageAccountWeb/src/components/StyledComponents.ts
```diff
@@ -38,56 +26,4 @@ export const StyledInnerContainer: StyledComponent<PaperProps> = styled(Paper, {
...
-      position: 'relative',
-      backgroundColor: theme.palette.mode === 'light' ? '#F1F3F5' : '#F1F3F5',
-      border: 'none',
-      padding: '14px 12px',
-      fontFamily: latoFont,
-      '&:focus': {
-        boxShadow: `${alpha(theme.palette.primary.main, 0.25)} 0 0 0 0.1rem`,
-        borderColor: theme.palette.primary.main,
-      },
-    },
```
nickgros, line 88: Mostly moved to DefaultTheme in SRC, to apply to all InputBase components.

packages/synapse-react-client/src/lib/components/styled/StyledFormControl.tsx
```diff
@@ -0,0 +1,48 @@
...
+import { StyledComponent } from '@emotion/styled'
+
+/* bootstrap-like label/text inputs
+ usage:
+        <StyledFormControl fullWidth variant="standard" margin="normal">
+          <InputLabel shrink htmlFor="someinput">
+            labelText
+          </InputLabel>
+          <TextField
+            id="someinput"/>
```
nickgros:
```suggestion
          <InputBase id="someinput"/>
```

packages/synapse-react-client/src/lib/containers/TextField.tsx
```diff
@@ -35,28 +35,26 @@ type TextFieldProps = Pick<
  */
 export default function TextField(props: TextFieldProps) {
```
nickgros, line 36: Apparently Alina and I had separately tried to make an input field using MUI that matches designs.

Alina's solution used `styled` and was closer to the design, but wasn't quite using the right MUI components (e.g. the CSS and existing usage required nesting a `TextField` (which contains a `FormControl`) in the `StyledFormControl`.

I created this custom `TextField` component, which is closer to a re-creation of the MUI TextField but tailored for our design requirements. I don't think the actual implementation matched designs as closely, but it's an "all-in-one" component that itself handles the FormControl wrapper and the label.

Here I combined the two--use the styled FormControl and fix some of the components in this custom component.

I think ideally, our theme would be configured so that we could just use the MUI TextField component instead of a custom component or styled component, but we'll get there eventually.

packages/synapse-react-client/src/lib/containers/TextField.tsx
```diff
@@ -35,28 +35,26 @@ type TextFieldProps = Pick<
...
+  const { noWrapInFormControl, label, ...rest } = props
   const Wrapper = useMemo(
     () =>
       noWrapInFormControl
         ? (props: React.PropsWithChildren<object>) => (
             <React.Fragment>{props.children}</React.Fragment>
           )
         : (props: React.PropsWithChildren<object>) => (
-            <FormControl fullWidth sx={{ my: 1 }}>
+            <StyledFormControl fullWidth sx={{ my: 1 }}>
```
nickgros, line 46: Use `StyledFormControl`

packages/synapse-react-client/src/lib/containers/TextField.tsx
```diff
@@ -35,28 +35,26 @@ type TextFieldProps = Pick<
...
-            </FormControl>
+            </StyledFormControl>
           ),
     [noWrapInFormControl],
   )
   return (
     <Wrapper>
       {props.label && (
-        <Typography
+        <InputLabel
```
nickgros, line 55: Should use InputLabel instead of Typography

packages/synapse-react-client/src/lib/containers/TextField.tsx
```diff
@@ -67,22 +65,9 @@ export default function TextField(props: TextFieldProps) {
...
-        sx={{
-          backgroundColor: 'grey.200',
-          px: 2,
-          py: 1,
-          ...props.sx,
-          '&.Mui-focused': {
-            backgroundColor: palette.background.default,
-            outline: `1px solid ${palette.primary.main}`,
-          },
-        }}
```
nickgros, line 84: `sx` on InputBase is now applied via theme or StyledFormControl

packages/synapse-react-client/src/lib/containers/auth/TOTPForm.tsx
```diff
@@ -27,6 +27,10 @@ export default function TOTPForm(props: TOTPFormProps) {
         }}
         gap={0}
         sx={{
+          '.MuiInputBase-root': {
+            paddingLeft: '5px',
+            paddingRight: '5px',
+          },
```
nickgros, line 33: The TOTP form (6 single-digit inputs) is a special case where the default theme InputBase styles should be overridden

packages/synapse-react-client/src/lib/utils/theme/DefaultTheme.ts
```diff
@@ -79,12 +81,19 @@ const defaultMuiThemeOptions: ThemeOptions = {
...
+          paddingBottom: ownerState.dividers ? theme.spacing(3) : 0,
+          // Hack - set add a small padding and offset with a negative margin so box-shadow effects on full width elements (like input fields) are shown
+          paddingLeft: DIALOG_INNER_PADDING,
+          paddingRight: DIALOG_INNER_PADDING,
+          marginLeft: `-${DIALOG_INNER_PADDING}`,
+          marginRight: `-${DIALOG_INNER_PADDING}`,
+          // Add a filter on the `dividers` borders so that the borders are transparent on the edges based on the DIALOG_INNER_PADDING length
+          borderImage: ownerState.dividers
+            ? `linear-gradient(to right, transparent 0px, transparent ${DIALOG_INNER_PADDING}, ${theme.palette.grey[400]} ${DIALOG_INNER_PADDING}, ${theme.palette.grey[400]} calc(100% - ${DIALOG_INNER_PADDING}), transparent calc(100% - ${DIALOG_INNER_PADDING})) 1`
+            : undefined,
```
nickgros, line 96: Before this change, our input border/box-shadow styles were cut off on the sides if they were full-width in a Dialog, because the DialogContent container had no side padding. These changes

- Add enough side padding to ensure border styles are visible
- Add a counteracting negative margin so the content aligns with the action buttons at the bottom and title at the top.
- Add a border-image filter that basically just hides the first 2px and last 2px of the border, so the border remains visually aligned with the content.

packages/synapse-react-client/src/lib/utils/theme/DefaultTheme.ts
```diff
@@ -101,6 +110,27 @@ const defaultMuiThemeOptions: ThemeOptions = {
...
+              0.25,
+            )} 0 0 0 0.1rem`,
+            borderColor: theme.palette.primary.main,
+          },
+        }),
+        input: ({ theme }) => ({
+          padding: '14px 12px',
+        }),
+      },
+    },
```
nickgros, line 133: Pulled from `StyledFormControl` to apply to all InputBase components.

hallieswan (approved):
Looks good to me!

PR #142, opened 2023-04-19
SWC-6421

jay-hodgson:


packages/synapse-react-client/src/lib/containers/download_list/DownloadConfirmation.tsx
```diff
@@ -65,7 +65,7 @@ async function addToDownload(
     closeConfirmationFn()
     return [StatusEnum.LOADING_INFO, '']
   } catch (error) {
-    displayToast('danger', undefined, error.reason)
+    displayToast(error.reason, 'danger')
```
jay-hodgson, line 68: Parameter order was wrong, so 'danger' was the message and undefined was the message

nickgros (approved):

PR #146, opened 2023-04-19
Upgrade Storybook to v7 stable

nickgros:
- Upgrade Storybook packages to latest (v7.0.6)
- Add Storybook manager for some basic branding of our Storybook
- Bump MSW script to match current version


packages/synapse-react-client/.storybook/preview.jsx
```diff
@@ -108,8 +108,9 @@ initialize({
   onUnhandledRequest: 'bypass',
 })
 
+export const loaders = [mswLoader]
+
```
nickgros, line 112: Using the (new-ish) loader instead of the decorator should ensure the service worker is initialized before the story runs, eliminating race conditions.

packages/synapse-react-client/public/mockServiceWorker.js
```diff
@@ -2,13 +2,13 @@
 /* tslint:disable */
```
nickgros, line 2: Auto-generated file created by `pnpm msw init public`

jay-hodgson (approved):

PR #151, opened 2023-04-20
Update Storybook MSW script URL to work in GitHub Pages

nickgros:


packages/synapse-react-client/.storybook/preview.jsx
```diff
@@ -106,6 +106,11 @@ export const globalTypes = {
 // Initialize MSW
 initialize({
   onUnhandledRequest: 'bypass',
+  serviceWorker: {
+    url: `${
+      location.hostname.endsWith('.github.io') ? '/synapse-web-monorepo' : ''
```
jay-hodgson, line 111: Thanks Nick!

PR #155, opened 2023-04-24
PORTALS-2645

jay-hodgson:


packages/synapse-react-client/src/lib/utils/hooks/SynapseAPI/entity/useExportToCavatica.ts
```diff
@@ -39,18 +40,17 @@ export function useExportToCavatica(
         downloadFromTableRequest,
         accessToken,
       )
-      const presignedURL = await SynapseClient.getFileHandleByIdURL(
+      const fileURL = SYNAPSE_ORG_RAW_FILE_HANDLE_URL_SERVLET(
```
nickgros, line 43: Makes sense, assuming this service works with OAuth2 access tokens. I can't think of a reason why it shouldn't

PR #158, opened 2023-04-25
PORTALS-2624: remove bootstrap-4-backport class from CreateProjectModal

hallieswan:
Before: 
main_CreateProjectModal

After:
feature_CreateProjectModal

### DIFF
--- a/packages/synapse-react-client/src/lib/containers/ConfirmationDialog.tsx
+++ b/packages/synapse-react-client/src/lib/containers/ConfirmationDialog.tsx
@@ -38,7 +38,7 @@ export type ConfirmationDialogProps = {
   confirmButtonText?: string
   confirmButtonColor?: ButtonProps['color']
   className?: string
-  onConfirm: () => void
+  onConfirm: () => unknown | Promise<void>
   onCancel: () => void
   hasCloseButton?: boolean
   helpMarkdown?: string
--- a/packages/synapse-react-client/src/lib/containers/ConfirmationDialog.tsx
+++ b/packages/synapse-react-client/src/lib/containers/ConfirmationDialog.tsx
@@ -38,7 +38,7 @@ export type ConfirmationDialogProps = {
   confirmButtonText?: string
   confirmButtonColor?: ButtonProps['color']
   className?: string
-  onConfirm: () => unknown | Promise<void>
+  onConfirm: () => void
   onCancel: () => void
   hasCloseButton?: boolean
   helpMarkdown?: string
--- a/packages/synapse-react-client/src/lib/containers/CreateProjectModal.tsx
+++ b/packages/synapse-react-client/src/lib/containers/CreateProjectModal.tsx
@@ -1,9 +1,10 @@
+import { Alert } from '@mui/material'
 import React, { useState } from 'react'
-import { Alert, Button, Modal } from 'react-bootstrap'
 import { SynapseClient } from '../utils'
 import { useSynapseContext } from '../utils/SynapseContext'
+import { ConfirmationDialog } from './ConfirmationDialog'
 import FullWidthAlert from './FullWidthAlert'
-import { Form } from 'react-bootstrap'
+import TextField from './TextField'
 
 export type CreateProjectModalProps = {
   isShowingModal?: boolean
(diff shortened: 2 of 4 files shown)


packages/synapse-react-client/src/lib/containers/CreateProjectModal.tsx
```diff
@@ -40,55 +41,42 @@ export const CreateProjectModal: React.FunctionComponent<
...
+      <TextField
+        id="projectInput"
+        label="Project Name"
+        value={projectName}
+        fullWidth
+        onChange={event => {
+          setProjectName(event.target.value)
+        }}
+        inputProps={{
+          onKeyDown: (event: React.KeyboardEvent<HTMLInputElement>) => {
```
hallieswan, line 56: TextField doesn't expose `onKeyDown` as a prop, so pass to Input via `inputProps`, see [here](https://github.com/mui/material-ui/issues/12946#issuecomment-423850042)

packages/synapse-react-client/src/lib/containers/CreateProjectModal.tsx
```diff
@@ -40,55 +41,42 @@ export const CreateProjectModal: React.FunctionComponent<
...
+
+  return (
+    <>
+      <ConfirmationDialog
+        open={isShowingModal}
+        title="Create a new Project"
+        content={dialogContent}
+        confirmButtonText="Save"
+        onConfirm={onCreateProject}
+        onCancel={hide}
```
hallieswan, line 77: Previously, clicking the modal close button called `hide`, which cleared the alert and text entered into the input before calling `onClose` whereas clicking "Cancel" only called `onClose`. 

Now, ConfirmationDialog will call `hide` both when clicking "Cancel" or when clicking the modal close button.

packages/synapse-react-client/test/lib/containers/CreateProjectModal.test.tsx
```diff
@@ -34,28 +43,67 @@ describe('CreateProjectModal tests', () => {
...
+
+    const alert = await screen.findByRole('alert')
+    expect(alert).toHaveTextContent('Invalid project name')
+
+    const closeButton = screen.getByRole('button', { name: /close/i })
+    await user.click(closeButton)
+
+    expect(alert).not.toBeInTheDocument()
+    expect(input.value).toBe('')
+  })
```
hallieswan, line 108: Test that clicking "Cancel" button or clicking modal close button result in the same side effects

packages/synapse-react-client/src/lib/containers/ConfirmationDialog.tsx
```diff
@@ -38,7 +38,7 @@ export type ConfirmationDialogProps = {
   confirmButtonText?: string
   confirmButtonColor?: ButtonProps['color']
   className?: string
-  onConfirm: () => void
+  onConfirm: () => unknown | Promise<void>
```
hallieswan, line 41: Update type to handle `CreateProjectModal` passing an async function (`onCreateProject`) to `onConfirm`. Using `unknown` rather than `void` because Typescript doesn't handle the union type of `void` and `Promise<void>`, see [here](https://github.com/microsoft/TypeScript/issues/43921). I wasn't sure if there was a better way to handle this, so happy to adjust as needed!

nickgros: Nothing is listening to `onConfirm`, so it doesn't need to return a promise. You could just do `onConfirm: () => void` and then in code wrap an asynchronous call in a non-async function.

```tsx
async function callback(): Promise<void> {
 //...
}
//...
return <ConfirmationDialog
        onConfirm={() => {
          callback()
        }}
  />
```

In IntelliJ (not sure about VSCode), this shows a warning that the returned value of the promise is ignored. Apparently, we can use the `void` operator to return `undefined` and essentially assert that we don't care about the result. See [StackOverflow](https://stackoverflow.com/a/64234381).

So it seems like the best thing to do would be to use `() => void` and write the prop as:
```tsx
<ConfirmationDialog
    onConfirm={() => {
      void callback()
    }}
  />
```

nickgros (approved):
Suggest reverting the prop to `() => void` as mentioned, but otherwise changes look great!

PR #162, opened 2023-04-26
CI improvements - Reuse project setup action and run build/test/lint in parallel

nickgros:
- Separate repo setup/dep install into local reusable action
- Split build/test/lint into separate parallel steps, making the entire process slightly faster and easier to see which step failed.

synapse-react-client:
- Remove react-scripts specific sections in README.md

### DIFF
--- a/.github/actions/pnpm-setup-action/action.yml
+++ b/.github/actions/pnpm-setup-action/action.yml
@@ -1,0 +1,37 @@
+name: "Set up pnpm and configure environment"
+description: "Installs Node, pnpm, fetches cache (if exists), and runs pnpm install"
+runs:
+  using: "composite"
+  steps:
+    - name: Use Node.js ${{ matrix.node-version }}
+      uses: actions/setup-node@v3
+      with:
+        node-version: ${{ matrix.node-version }}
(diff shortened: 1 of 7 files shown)


.github/actions/pnpm-setup-action/action.yml
nickgros, line 1: This is the same set of steps used across multiple workflows to install Node, pnpm, and set up caches. This is now a [Composite Action](https://docs.github.com/en/actions/creating-actions/creating-a-composite-action) that we can reuse in multiple workflows.

.github/workflows/build-test.yml
```diff
@@ -1,72 +1,54 @@
 # This workflow will do a clean installation of node dependencies, cache/restore them, build the source code and run tests across different versions of node
 # For more information see: https://help.github.com/actions/language-and-framework-guides/using-nodejs-with-github-actions
 
- name: Build/Test All Projects
+name: Build/Test All Projects
```
nickgros, line 4: Whitespace was not what prettier expected so the diff is bad 😠

jay-hodgson (approved):

PR #163, opened 2023-04-26
CI Improvements - Run only `msw` "integration" serially, only show logs for failed tests

nickgros:
Motivation: I have PRs for SWC-6412 where the `test` job fails, but the HTML reporter states that all tests pass. Nx is suppressing the full log so I can't see why the `test` job fails despite all individual tests passing, so I want to suppress warnings and errors when appropriate. Additionally, our test suite takes a long time, so running our unit tests in parallel could speed it up considerably.

synapse-react-client:
- Rename all tests using `msw` from `*.test.tsx` to `*.integration.test.tsx`
- Modify Jest config to run just `*.integration.test.tsx` files with `jest-serial-runner`. With this change, our unit tests (which don't depend on `msw` or another shared resource) will no longer have to run serially
- Add and use LogForFailedTestsOnlyReporter which will only log warnings and errors if a test suite fails.
- Few test suite changes
  - Move a handful of test files to more appropriate locations, and update imports accordingly
  - Replace some `jest.mock` usage with a module import and `jest.spyOn` to better handle cases where files are moved.
  - EntityBadgeIcons.integration.test.tsx: Attempt to remove `not wrapped in act` warnings by wrapping `render` in `act`
  - MarkdownSynapse.integration.test.tsx: Update mock implementations to prevent logged errors


packages/synapse-react-client/test/lib/containers/EntityBadgeIcons.integration.test.tsx
```diff
@@ -41,9 +41,11 @@ const defaultProps: EntityBadgeIconsProps = {
   renderTooltipComponent: true,
 }
 
-function renderComponent(wrapperProps?: SynapseContextType) {
-  render(<EntityBadgeIcons {...defaultProps} />, {
-    wrapper: createWrapper(wrapperProps),
+async function renderComponent(wrapperProps?: SynapseContextType) {
```
nickgros, line 44: Attempting to remove some `not wrapped in act` warnings

jay-hodgson (approved):

PR #164, opened 2023-04-26
clean up ELITE portal explore sections

jay-hodgson:


apps/portals/src/configurations/elportal/routesConfig.ts
```diff
@@ -117,7 +114,7 @@ const routes: GenericRoute[] = [
         outsideContainerClassName: 'home-spacer home-bg-dark',
         centerTitle: true,
         props: {
-          sql: `${peopleSql}`,
+          sql: `${peopleSql} where isFeatured=true`,
```
jay-hodgson, line 117: and an isFeatured column has been added to this table

PR #166, opened 2023-04-26
SWC-5637

nickgros:
Fix SWC-5637

- Fix alignment by switching to MUI
- Use MUI popover. Fixes cases where popovers would not appear over MUI dialogs

The alignment with the form and the dialog isn't perfect, but fixing that properly will require major changes to the SchemaDrivenAnnotationEditor

|Before|After|
|---|---|
|image|image|
|image|image|

### DIFF
--- a/packages/synapse-react-client/src/lib/containers/entity/metadata/EntityModal.tsx
+++ b/packages/synapse-react-client/src/lib/containers/entity/metadata/EntityModal.tsx
@@ -27,7 +27,6 @@ import {
   Tooltip,
 } from '@mui/material'
 import { CloseButton } from '../../ConfirmationDialog'
-import ConditionalWrapper from '../../utils/ConditionalWrapper'
 
 export enum EntityModalTabs {
   METADATA = 'METADATA', // non-annotation metadata about the entity
@@ -224,9 +223,6 @@ export function EntityModal(props: EntityModalProps) {
         </div>
       </DialogContent>
       <DialogActions>
-        <ConditionalWrapper condition={!entityBundle} wrapper={Skeleton}>
-          <Button></Button>
-        </ConditionalWrapper>
         {showCancelAnnotationEditsButton && cancelAnnotationEditsButton}
         {showSaveAnnotationsButton && saveAnnotationsButton}
         {showEditAnnotationsButton && editAnnotationsButton}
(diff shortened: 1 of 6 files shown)


packages/synapse-react-client/src/lib/containers/entity/metadata/EntityModal.tsx
```diff
@@ -60,147 +71,167 @@ export const EntityModal: React.FC<EntityModalProps> = ({
...
-        copy: `Open ${entityTypeToFriendlyName(entityBundle.entityType)}`,
-        onClick: () =>
-          window.open(
-            `${getEndpoint(
-              BackendDestinationEnum.PORTAL_ENDPOINT,
-            )}#!Synapse:${entityId}`,
-            '_blank',
-            'noopener',
-          ),
+  const showOpenEntityPageButton =
```
nickgros, line 74: Way easier to just create the buttons than to try to replicate this `action` -> `Button` mapping

packages/synapse-react-client/src/lib/style/components/_annotations.scss
```diff
@@ -1,156 +1,154 @@
 @use '../abstracts/variables' as SRC;
 
-.bootstrap-4-backport {
```
nickgros, line 3: Just remove this outer .bootstrap-4-backport class which was unnecessary (and broke styling when I removed it from the modal)

packages/synapse-react-client/src/lib/utils/hooks/useOverlay.tsx
```diff
@@ -70,35 +77,33 @@ export function useOverlay(
 
   const OverlayComponent = useCallback(
     () => (
-      <Overlay
-        target={targetRef.current}
-        show={isShowing}
-        placement={placement}
+      <Popover
```
nickgros, line 80: Use the MUI version. The bootstrap version didn't work in MUI modals

hallieswan: Nice! Removes another bootstrap-4-backport instance too

hallieswan (approved):
Looks good! I saw the same issue with the Popover in another MUI modal, so it's great that you've fixed it!

PR #167, opened 2023-04-26
add EL portal palette, and fix bug in study details page (data study …

jay-hodgson:
…column is not multi-value in this portal)


apps/portals/src/configurations/elportal/synapseConfigs/studies.ts
```diff
@@ -130,22 +130,6 @@ export const studiesDetailsPageProps: DetailsPageProps = {
       toolTip: 'All of the Data generated within this study',
       cssClass: 'tab-database',
       synapseConfigArray: [
-        {
-          name: 'RssFeedCards',
```
jay-hodgson, line 134: Removing rss feed for the initial release

PR #168, opened 2023-04-27
SWC-6412 - Add queries and mutations for ARs and submissions

nickgros:
- Improve types for ARs and submissions
- Add react-query queries and mutations for fetching ARs and fetching and updating submissions
- Rename SynapseClient#postAccessApproval to .createAccessApproval
- use Infinity cacheTime for react-query in test


packages/synapse-react-client/test/testutils/TestingLibraryUtils.tsx
```diff
@@ -10,6 +10,12 @@ type RtlWrapperProps = {
   children?: ReactNode
 }
 
+const reactQueryTestConfig = {
+  ...defaultQueryClientConfig,
+  // Recommended to set cacheTime to Infinity: https://tanstack.com/query/v3/docs/react/guides/testing#set-cachetime-to-infinity-with-jest
+  cacheTime: Infinity,
+}
```
nickgros, line 17: Just an optimization to prevent the garbage collector from inadvertently failing tests. I have yet to actually see it happen, but the react-query docs recommend it.

We use a new QueryClient for each test, and the default cacheTime is much longer than any particular test would run (5min), so this should not impact on our tests otherwise.

packages/synapse-react-client/test/lib/containers/discussion_forum/DiscussionThread.integration.test.tsx
```diff
@@ -105,7 +105,8 @@ function renderComponent() {
   })
 }
 
-describe('Discussion Thread test', () => {
+// TODO: Fix flakiness
+describe.skip('Discussion Thread test', () => {
```
nickgros, line 109: This suite fails often. Perhaps can come back to it if/when we use these discussion forum components.

packages/synapse-react-client/src/lib/utils/SynapseClient.ts
```diff
@@ -3303,19 +3304,19 @@ export const getDataAccessRequestForUpdate = (
...
 // http://rest-docs.synapse.org/rest/GET/accessRequirement/requirementId/dataAccessRequestForUpdate.html
-export const updateDataAccessRequest = (
-  requestObj: RequestInterface,
+export function updateDataAccessRequest(
+  requestObj: Request | Renewal,
   accessToken: string,
-) => {
-  return doPost<RequestInterface>(
+) {
+  return doPost<typeof requestObj>(
```
nickgros, line 3319: We know the return type will match the request object type, so update type signature accordingly.

packages/synapse-react-client/src/lib/utils/SynapseClient.ts
```diff
@@ -2859,11 +2859,12 @@ export const searchAccessRequirements = (
  * @param {string} requirementId id of entity to lookup
  * @returns {AccessRequirementStatus}
  */
-export const getAccessRequirementStatus = (
-  accessToken: string | undefined,
-  requirementId: string | number,
-): Promise<AccessRequirementStatus | ManagedACTAccessRequirementStatus> => {
-  return doGet(
+export function getAccessRequirementStatus<
```
nickgros, line 2862: Add optional generic parameter so we can (safely) assert the return type if we know what to expect . For example, if we're fetching the status of a `ManagedACTAccessRequirement`, then we know we're going to get a `ManagedACTAccessRequirementStatus`

packages/synapse-react-client/src/lib/containers/StorybookComponentWrapper.tsx
```diff
@@ -134,20 +135,21 @@ export function StorybookComponentWrapper(props: {
     [accessToken],
   )
 
+  const theme = createTheme(defaultMuiThemeOptions, {
+    palette: paletteMap[storybookContext.globals.palette],
+  })
```
nickgros, line 140: Small change, create the theme in Storybook the same way we create it in the apps.

packages/synapse-react-client/src/lib/containers/access_requirement_list/AccessRequirementList.tsx
```diff
@@ -57,7 +58,7 @@ export type RequestDataStepCallbackArgs = {
   managedACTAccessRequirement?: ManagedACTAccessRequirement
   step: number
   researchProjectId?: string
-  formSubmitRequestObject?: RequestInterface
+  formSubmitRequestObject?: Request | Renewal
```
nickgros, line 61: In TypeScript it's easier to use this union type than the interface

packages/synapse-react-client/src/lib/utils/SynapseClient.ts
```diff
@@ -2940,7 +2941,7 @@ export const getAccessApproval = async (
  * @param {AccessApproval} accessApproval access approval request object
  * @returns {AccessApproval}
  */
-export const postAccessApproval = async (
+export const createAccessApproval = async (
```
nickgros, line 2944: Renamed for clarity

packages/synapse-react-client/src/lib/utils/hooks/SynapseAPI/dataaccess/useAccessRequirements.ts
```diff
@@ -129,3 +148,181 @@ export function useCreateLockAccessRequirement(
...
+  const queryClient = useQueryClient()
+  const { keyFactory } = useSynapseContext()
+
+  return useMutation<
+    ACTSubmissionStatus,
+    SynapseClientError,
+    { submissionId: string; accessRequirementId: string }
+  >(
+    request =>
+      SynapseClient.cancelDataAccessRequest(request.submissionId, accessToken!),
```
nickgros, line 312: We don't need the the `accessRequirementId` to cancel the request, but we do need it to invalidate the `AccessRequirementStatus` query key, so we require it as an argument.

packages/synapse-react-client/src/lib/utils/SynapseClient.ts
```diff
@@ -4444,3 +4445,31 @@ export function getDOIAssociation(
...
+ * This effectively provides a stable URL where a user can request the contents of a file and get the same result, without
+ * concern for the 30s expiration of presigned file handles returned by the Synapse backend.
+ *
+ * Note that this URL will only work if the user is logged in to Synapse on the domain of the SWC instance, and the client
+ * (typically a browser) can provide the current access token cookie.
+ * @param fileHandleId The ID of the file handle to retrieve
+ * @param associatedObjectId ID of the object the file handle is associated with. Required if the user is not the one who uploaded the file handle
+ * @param associatedObjectType The type of the associated object. Required if the user is not the one who uploaded the file handle
+ */
+export function getPortalFileHandleServletUrl(
```
nickgros, line 4461: Add this functionality back--it's handy to have, even if it doesn't work for the Cavatica scenario.

packages/synapse-react-client/src/lib/utils/hooks/SynapseAPI/file/useFiles.ts
```diff
@@ -70,3 +74,34 @@ export function useGetProfileImage(
...
+  const { accessToken, keyFactory } = useSynapseContext()
+  const queryFn = async () => SynapseClient.getFiles(request, accessToken)
+
+  if (request.includePreSignedURLs || request.includePreviewPreSignedURLs) {
+    // Don't use this hook if you need pre-signed URLs. They expire every 30 seconds, so you will either end up giving the
+    // user an expired URL, or making requests to the backend more frequently than necessary.
+    console.error('useGetFileBatch does not support pre-signed URLs.')
+    request.includePreSignedURLs = false
+    request.includePreviewPreSignedURLs = false
+  }
```
nickgros, line 98: It would be totally inappropriate to use this hook to get a pre-signed URL, since we'd have to get a new one every 30s. Considered throwing an error but decided to just alter the request and log an error.

packages/synapse-react-client/src/lib/utils/synapseTypes/AccessRequirement/RenewalInterface.ts
nickgros, line 1: This is not an interface defined by the Synapse backend

packages/synapse-react-client/src/lib/utils/synapseTypes/AccessRequirement/RequestInterface.ts
```diff
@@ -19,3 +19,13 @@ export interface RequestInterface {
...
+
+export interface Request extends RequestInterface {
+  concreteType: 'org.sagebionetworks.repo.model.dataaccess.Request'
+}
+
+export interface Renewal extends RequestInterface {
+  concreteType: 'org.sagebionetworks.repo.model.dataaccess.Renewal'
+  publication?: string
+  summaryOfUse?: string
+}
```
nickgros, line 31: Define these concrete classes.

packages/synapse-react-client/mocks/user/mock_user_profile.ts
```diff
@@ -77,6 +77,43 @@ export const mockUserGroupHeader2: UserGroupHeader = {
   isIndividual: true,
 }
 
+export const MOCK_USER_ID_3 = 1001
+export const MOCK_USER_NAME_3 = 'heisenberg'
```
jay-hodgson, line 81: We should keep an eye on mock user 3

packages/synapse-react-client/src/lib/utils/hooks/SynapseAPI/file/useFiles.ts
jay-hodgson: My preference would be to throw an error to fail in a highly visible way rather than update the caller's request to a valid value, but if the caller is trying to get a batch of presigned URLs using this hook then they should eventually find this code

jay-hodgson (approved):

PR #176, opened 2023-04-28
PORTALS-2624: remove bootstrap-4-backport from ExperimentalMode

hallieswan:
Before: 

https://user-images.githubusercontent.com/26949006/235269863-cc1947be-ed55-4d85-aca4-f4f040d67eae.mov

After: 

https://user-images.githubusercontent.com/26949006/235743088-66cad998-735c-4d41-8f48-235383c77d9b.mov

### DIFF
--- a/packages/synapse-react-client/src/lib/containers/ExperimentalMode.tsx
+++ b/packages/synapse-react-client/src/lib/containers/ExperimentalMode.tsx
@@ -1,12 +1,21 @@
 import * as React from 'react'
-import { Button, OverlayTrigger, Popover } from 'react-bootstrap'
+import { Button, IconButton } from '@mui/material'
 import { useEffect, useState } from 'react'
 import UniversalCookies from 'universal-cookie'
 import { isInSynapseExperimentalMode } from '../utils/SynapseClient'
 import { EXPERIMENTAL_MODE_COOKIE } from '../utils/SynapseConstants'
 import { InfoOutlined } from '@mui/icons-material'
+import LightTooltip from '../components/styled/LightTooltip'
+
+const experimentalModeText =
+  'This mode gives you early access to features that are still in development. Please note that we do not guarantee an absence of errors, and that the data created using these features may be lost during product upgrade.'
 
 const ExperimentalMode: React.FC = () => {
+  const [open, setOpen] = useState<boolean>(false)
+  const toggleOpen = () => {
+    setOpen(!open)
+  }
+
   const [isExperimentalModeOn, setIsExperimentalModeOn] =
     useState<boolean>(false)
   const cookies = new UniversalCookies()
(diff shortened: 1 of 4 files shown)


packages/synapse-react-client/src/lib/containers/ExperimentalMode.tsx
```diff
@@ -60,9 +56,21 @@ const ExperimentalMode: React.FC = () => {
       >
         Experimental mode is {isExperimentalModeOn ? 'on' : 'off'}
       </Button>
-      <OverlayTrigger trigger="click" placement="top" overlay={popover}>
-        <InfoOutlined style={{ verticalAlign: 'middle' }} />
-      </OverlayTrigger>
+      <LightTooltip
```
hallieswan, line 59: Used a light tooltip that appears on clicking the info icon to replicate the current design. We could consider using a dark tooltip that appears on hover to match the tooltip on Version in the Portals footer (i.e. where this component is used)

packages/synapse-react-client/src/lib/containers/ExperimentalMode.tsx
```diff
@@ -60,9 +56,21 @@ const ExperimentalMode: React.FC = () => {
...
+        title={experimentalModeText}
+        arrow
+        placement="top"
+        open={open}
+      >
+        <IconButton
+          aria-label="info"
+          color="inherit"
+          onClick={toggleOpen}
+          sx={{ '&:hover': { backgroundColor: 'transparent' } }}
```
hallieswan, line 69: Match current styling

packages/synapse-react-client/src/lib/style/components/_experimental-mode.scss
```diff
@@ -7,6 +7,7 @@
     text-transform: uppercase;
     padding-right: 0.2rem;
     color: inherit;
+    text-decoration-color: inherit;
```
hallieswan, line 10: Ensures that the correct underline color is used in the Portals footer

packages/synapse-react-client/stories/ExperimentalMode.stories.tsx
```diff
@@ -1,10 +1,17 @@
...
+import React, { Meta, StoryObj } from '@storybook/react'
 
 import ExperimentalMode from '../src/lib/containers/ExperimentalMode'
 
 const meta = {
   title: 'Authentication/ExperimentalMode',
   component: ExperimentalMode,
+  render: () => {
+    return (
+      <div style={{ marginTop: '150px' }}>
```
hallieswan, line 10: Add margin so the entire tooltip can be seen in the Story

packages/synapse-react-client/src/lib/containers/ExperimentalMode.tsx
nickgros: Design has typically used the dark tooltip for 1-2 sentences, where you'd typically hover over a button to see it. For longer content, they would typically use a light popover (they call it a "super tooltip") opened by clicking an info button. The "super tooltip" may also contain buttons where we may link out to documentation.

See the `MarkdownPopover` component for the "super tooltip" case.

I like the idea of just using the regular (dark) tooltip here. The content isn't super long or interactive, so it would probably be appropriate.

You could consider keeping your LightTooltip to see if it's a better foundation for `MarkdownPopover`

nickgros (approved):

packages/synapse-react-client/src/lib/containers/ExperimentalMode.tsx
hallieswan: Thanks Nick! Good to know about the "super tooltip" -- I think the LightTooltip will be helpful since MUI Popvers don't come with [an arrow](https://github.com/mui/material-ui/issues/18928) by default. Happy to switch to the dark tooltip here!

PR #80, opened 2023-03-20
Use vite for apps

nickgros:

PR #178, opened 2023-05-01
PORTALS-2628 - Use Vite for all apps

nickgros:
See #80, #131

- Use Vite in each app
- Use vitest in each app, replacing Jest. This reduces the configuration overhead and allows our test configuration to match our dev/prod configurations more closely.
- Create a `packages/vite-config` project which stores shared vite/vitest configurations.
- Ensure dependencies are built before `lint` jobs are run, because our linter is type-aware.
- Ensure our unmatrixed jobs (test, lint) fall back to Node 18 (the default was 16)
- Break circular dependencies in synapse-react-client, mostly cases where `fileA.ts` would introduce a `concreteType` value and a base type, and import from `fileB.ts` an additional `concreteType` value. `fileB.ts` would import from `fileA.ts` the declared interface. In these cases, I moved `concreteType` values to a separate file just to break the cycle.


packages/synapse-react-client/src/lib/utils/SynapseClient.ts
```diff
@@ -331,13 +332,6 @@ export function delay(t: number) {
   })
 }
 
-/**
- * Error message returned by the Synapse Backend
- */
-export type SynapseError = {
-  reason: string
-}
-
```
nickgros, line 340: Move to separate file to break circular dependency.

.eslintrc.cjs
nickgros, line 1: Just renamed to `cjs` to make it clear (to our tooling) that this is a CommonJS file

.github/actions/pnpm-setup-action/action.yml
```diff
@@ -3,10 +3,10 @@ description: "Installs Node, pnpm, fetches cache (if exists), and runs pnpm inst
 runs:
   using: "composite"
   steps:
-    - name: Use Node.js ${{ matrix.node-version }}
+    - name: Use Node.js ${{ matrix.node-version || 18 }}
       uses: actions/setup-node@v3
       with:
-        node-version: ${{ matrix.node-version }}
```
nickgros, line 9: Add a fallback jobs where we don't declare a matrix (test, lint). Before it would fall back to 16!

.gitignore
```diff
@@ -20,7 +20,6 @@ build/
 
 # misc
 .DS_Store
-.env
```
nickgros, line 23: Using `.env` files to store portal-specific data

apps/SageAccountWeb/.env
nickgros, line 1: Unused since craco is gone

apps/SageAccountWeb/src/AppInitializer.tsx
```diff
@@ -1,16 +1,16 @@
 import React, { useEffect, useState } from 'react'
-import { AppContextProvider } from 'AppContext'
+import { AppContextProvider } from './AppContext'
 import { Redirect } from 'react-router-dom'
-import { SynapseComponents } from 'synapse-react-client'
```
nickgros, line 4: You will see that I avoided using `SynapseComponents` in many files. Importing that object busts tree-shaking, which caused us to import a file that broke the deployed version of the site. I have fixed the deployment issue, so I didn't remove every single import like this. If we want to optimize for smaller builds, then that should be a separate, focused effort.

apps/portals/src/react-app-env.d.ts
```diff
@@ -1,7 +1,17 @@
...
+/// <reference types="vite/client" />
+interface ImportMetaEnv {
+  readonly VITE_PORTAL_NAME: string
+  readonly VITE_PORTAL_DESCRIPTION: string
+  readonly VITE_PORTAL_KEY: string
+}
+
+interface ImportMeta {
+  readonly env: ImportMetaEnv
+}
```
nickgros, line 17: These declarations allow us to access the environment variables in a type-safe way.

packages/synapse-react-client/src/lib/utils/synapseTypes/SignedToken/SignedTokenConcreteType.ts
nickgros, line 1: Moving all the concreteType declarations helped fix some crazy dependency cycle issues.

packages/synapse-react-client/src/lib/utils/synapseTypes/Table/ConcreteType.ts
nickgros, line 1: Moving all the concreteType declarations fixed dependency cycle issues.

packages/synapse-react-client/tsconfig.json
```diff
@@ -3,7 +3,9 @@
   "compilerOptions": {
     "rootDir": "src/lib",
     "outDir": "dist",
-    "module": "CommonJS",
+    "module": "esnext",
+    "target": "esnext",
+    "moduleResolution": "Node",
```
nickgros, line 8: Emit ESM instead of CommonJS so it plays better with Vite. Only the apps use this output, so it should not be breaking.

packages/vite-config/src/vite-config.ts
```diff
@@ -0,0 +1,70 @@
...
+import svgr from "vite-plugin-svgr";
+import { NodeGlobalsPolyfillPlugin } from "@esbuild-plugins/node-globals-polyfill";
+import { NodeModulesPolyfillPlugin } from "@esbuild-plugins/node-modules-polyfill";
+import rollupNodePolyFill from "rollup-plugin-polyfill-node";
+import { fileURLToPath } from "url";
+
+const __filename = fileURLToPath(import.meta.url);
+const __dirname = dirname(__filename);
+
+export default defineConfig({
```
nickgros, line 13: Define a Vite config file for all the apps to use.

packages/vite-config/src/vitest-config.ts
```diff
@@ -0,0 +1,22 @@
+import { defineConfig, mergeConfig } from "vitest/config";
+import viteConfig from "./vite-config.js";
+
+export default mergeConfig(
```
nickgros, line 4: Extend the shared Vite config to create a Vitest config that all of the apps can use

patches/@jsdevtools__ono@7.1.3.patch
nickgros, line 1: This package, which is a dependent of `@apidevtools/json-schema-ref-parser`

apps/SageAccountWeb/src/components/AccountSettings.tsx
```diff
@@ -16,18 +19,15 @@ import {
 import {
   SynapseClient,
   SynapseConstants,
-  Typography,
```
nickgros, line 19: Also in many places import `Typography` from MUI because the SRC version is just aliased to MUI.

apps/SageAccountWeb/src/AppInitializer.tsx
jay-hodgson: Could you open a ticket to track this idea?  Something like "Investigate if dropping the SynapseComponents object reduces the build size significantly"

nickgros: https://sagebionetworks.jira.com/browse/PORTALS-2655

jay-hodgson (approved):

PR #181, opened 2023-05-02
PORTALS-2600: rebranding as NF Research Tools Central

jay-hodgson:
A-New-Day-For-Ivy-Coach-300x203

### DIFF
--- a/apps/portals/src/portal-components/nf/BrowseToolsPage.tsx
+++ b/apps/portals/src/portal-components/nf/BrowseToolsPage.tsx
@@ -59,12 +59,12 @@ const BrowseToolsPage = () => {
       <div className="header">
         <div className="home-container-description">
           <Typography variant="headline1" className="sectionTitle">
-            NF Research Tools Database
+            NF Research Tools Central
           </Typography>
           <div className="center-content">
             <div className="description">
               <Typography variant="body1">
-                The NF Research Tools Database aims to support the development
+                NF Research Tools Central aims to support the development
                 of a robust research toolkit and lower the barrier of entry to
                 neurofibromatosis (NF) research. The database includes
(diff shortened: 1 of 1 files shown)


cconrad8 (commented):
@jay-hodgson @allaway after reading this I think I would like  it better if it said "The NF Research Tools Central" a few places.

If "NF Research Tools Central" is a proper noun, it seems like we should add The before. 

Thoughts?

jay-hodgson:
@cconrad8 
> @jay-hodgson @allaway after reading this I think I would like it better if it said "The NF Research Tools Central" a few places.
> 
> If "NF Research Tools Central" is a proper noun, it seems like we should add The before.
> 
> Thoughts?
I'll edit it to be whatever you and Robert would like, just let me know (line by line) if you'd like to prefix with a "The".

allaway:
Hmm, I can't quite put my finger on it but adding "The" to the beginning doesn't sound quite right. For example, the train station is called Grand Central, not The Grand Central. But again...I can't quite put my finger on why this doesn't sound right to me. 

(or, it sounds kind of like the difference between calling I-5 "5" or "the 5" - or for the DC area, "495" vs "the 495". Both are correct, but I think people have strong preferences for one or the other...I know I-5 vs the 5 is quite contentious here 😆

cconrad8:
😂 it’s funny cuz I was using “Dog” as the noun replacement. So I was like
“The dog” sounds right!

It can definitely go either way. Maybe we can do a larger poll lol. I can
make one in slack?

allaway:
Yes, that sounds great @cconrad8 !

cconrad8 (approved):
dropping the the!

cconrad8 (approved):
All is good!

PR #182, opened 2023-05-02
PORTALS-2610: pass through generic card configuration into custom Dat…

jay-hodgson:
…aset row renderer, and respect the secondary labels definition

### DIFF
--- a/packages/synapse-react-client/src/lib/containers/row_renderers/Dataset.tsx
+++ b/packages/synapse-react-client/src/lib/containers/row_renderers/Dataset.tsx
@@ -2,10 +2,11 @@ import * as React from 'react'
 import { DATASET } from '../../utils/SynapseConstants'
 import { calculateFriendlyFileSize } from '../../utils/functions/calculateFriendlyFileSize'
 import * as Utils from './utils'
-import { getValueOrMultiValue } from '../GenericCard'
+import { GenericCardSchema, getValueOrMultiValue } from '../GenericCard'
 import { SelectColumn, ColumnModel } from '../../utils/synapseTypes'
 import { Button } from 'react-bootstrap'
 import { PRODUCTION_ENDPOINT_CONFIG } from '../../utils/functions/getEndpoint'
+import { QueryVisualizationContextType } from '../QueryVisualizationWrapper'
 
 export type DatasetProps = {
   data?: any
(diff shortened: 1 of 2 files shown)


packages/synapse-react-client/src/lib/containers/row_renderers/Dataset.tsx
```diff
@@ -2,17 +2,20 @@ import * as React from 'react'
...
+import { QueryVisualizationContextType } from '../QueryVisualizationWrapper'
 
 export type DatasetProps = {
   data?: any
   schema?: any
   secondaryLabelLimit?: number
   selectColumns?: SelectColumn[]
   columnModels?: ColumnModel[]
+  genericCardSchema?: GenericCardSchema
+  queryVisualizationContext: QueryVisualizationContextType
```
jay-hodgson, line 18: The Card already attempts to pass these into the special row renderers, we just need to define them in the props to gain access to them.

packages/synapse-react-client/src/lib/containers/row_renderers/Dataset.tsx
```diff
@@ -31,7 +34,14 @@ class Dataset extends React.Component<DatasetProps, never> {
...
 
   public render() {
-    const { data, schema, selectColumns, columnModels } = this.props
+    const {
+      data,
+      schema,
+      selectColumns,
+      columnModels,
+      genericCardSchema,
+      queryVisualizationContext: { getColumnDisplayName },
```
jay-hodgson, line 43: Need the queryVisualizationContext to have access to the getColumnDisplayName (which possibly transforms the camel case column name into a more friendly version)

packages/synapse-react-client/src/lib/containers/row_renderers/Dataset.tsx
```diff
@@ -63,6 +73,20 @@ class Dataset extends React.Component<DatasetProps, never> {
       ['SIZE', fileSize],
       ['FILES', fileCount],
     ]
+    if (genericCardSchema && genericCardSchema.secondaryLabels) {
```
jay-hodgson, line 76: Originally, FUNDER, SIZE, and FILES were the hard-coded secondary values for Datasets.  Now we add the ability to define your own in the configuration (no special rendering of the value today)

hallieswan (approved):
Looks good!

PR #187, opened 2023-05-04
PORTALS-2624: refactor UserMentionModal to use MUI components

hallieswan:
Before: 

https://user-images.githubusercontent.com/26949006/236078939-c8c15577-605e-45a9-b13e-401f31229af4.mov

After: 

https://user-images.githubusercontent.com/26949006/236078932-5b7c928c-7cec-421b-bbcf-363db73f4819.mov

### DIFF
--- a/packages/synapse-react-client/src/lib/containers/markdown/MarkdownEditor.tsx
+++ b/packages/synapse-react-client/src/lib/containers/markdown/MarkdownEditor.tsx
@@ -67,8 +67,8 @@ export const MarkdownEditor: React.FunctionComponent<MarkdownEditorProps> = ({
     setTagModalWithKeyboard(false)
   }
 
-  const handleTagModal = (e: React.KeyboardEvent) => {
-    if (e.key == '@') {
+  const handleTagModal = (text: string) => {
+    if (text.slice(-1) === '@') {
       setTagModalWithKeyboard(true)
       setIsShowingTagModal(true)
     }
@@ -159,13 +159,15 @@ export const MarkdownEditor: React.FunctionComponent<MarkdownEditorProps> = ({
       <div>
         {currentTab === MarkdownEditorTabs.WRITE ? (
           <textarea
-            onChange={e => setText(e.target.value)}
+            onChange={e => {
+              setText(e.target.value)
+              handleTagModal(e.target.value)
+            }}
             style={{ width: '100%' }}
             rows={6}
             value={text}
             ref={textAreaRef}
             placeholder={placeholder}
-            onKeyDown={handleTagModal}
           />
         ) : text ? (
           <MarkdownSynapse markdown={text} />
--- a/packages/synapse-react-client/src/lib/containers/markdown/MarkdownEditor.tsx
+++ b/packages/synapse-react-client/src/lib/containers/markdown/MarkdownEditor.tsx
@@ -56,7 +56,7 @@ export const MarkdownEditor: React.FunctionComponent<MarkdownEditorProps> = ({
       const start = textVal?.selectionStart
       const textBeforeTag = text.substring(0, start)
       const textAfterTag = text.substring(start, text.length)
-      setSelectionStart(start + user.length + 1)
+      setSelectionStart(start + user.length)
       newText.push(
         textBeforeTag,
         `${tagModalWithKeyboard ? '' : '@'}${user.replace(/\s/g, '')}`,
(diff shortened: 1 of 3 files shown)


packages/synapse-react-client/src/lib/containers/markdown/MarkdownEditor.tsx
```diff
@@ -159,13 +159,15 @@ export const MarkdownEditor: React.FunctionComponent<MarkdownEditorProps> = ({
       <div>
         {currentTab === MarkdownEditorTabs.WRITE ? (
           <textarea
-            onChange={e => setText(e.target.value)}
+            onChange={e => {
+              setText(e.target.value)
+              handleTagModal(e.target.value)
+            }}
```
hallieswan, line 165: When `onKeyDown` was used, the keydown event triggered the MUI dialog before the `onChange` handler updated the text (this didn't seem to be an issue with bootstrap modal). Due to how `handleUserTag` processes the text, this resulted in the `@` symbol being removed from the text after a user name or group tag was selected in the dialog.

I considered changing the event handler to `onKeyUp` instead, but typing the `@` symbol quickly resulted in a separate keyup event for each key (Shift, 2) and didn't trigger the dialog.

Instead, I moved both events into `onChange`, so we ensure that text is updated before the dialog is triggered and that dialog is triggered regardless of how quickly the user types the @ symbol.

packages/synapse-react-client/src/lib/containers/markdown/UserMentionModal.tsx
```diff
@@ -26,19 +26,25 @@ export const UserMentionModal: React.FC<UserMentionModalProps> = ({
...
           />
-        </Modal.Body>
-      </Modal>
+        }
+        sx={{
+          '.MuiDialog-paperFullWidth': {
+            overflowY: 'visible',
+            '.MuiDialogContent-root': { overflowY: 'visible' },
+          },
+        }}
```
hallieswan, line 46: Ensures the react-select dropdown menu is visible above the MUI dialog as described [here](https://stackoverflow.com/a/57946878). This does result in some styling changes to the TeamBadges, but the styling appears to more closely match the TeamBadge styling in other UserSearchBoxV2 instances, such as in the AccessSubmissionDashboard (i.e. no link styling applied to the TeamBadge) -- 
main_ReviewerDashboard_SubmissionsReviewer


I also considered updating UserSearchBoxV2 directly as described [here](https://react-select.com/advanced#portaling) (i.e. adjust menuPortal z-index in styles and add `menuPortalTarget` argument), but the menu was moved outside the parent div of the component and this resulted in larger styling changes.

packages/synapse-react-client/src/lib/containers/markdown/MarkdownEditor.tsx
```diff
@@ -67,8 +67,9 @@ export const MarkdownEditor: React.FunctionComponent<MarkdownEditorProps> = ({
     setTagModalWithKeyboard(false)
   }
 
-  const handleTagModal = (e: React.KeyboardEvent) => {
-    if (e.key == '@') {
+  const handleTagModal = (text: string) => {
+    const start = textAreaRef.current && textAreaRef.current.selectionStart
+    if (start && start > 0 && text.charAt(start - 1) === '@') {
```
hallieswan, line 72: Handle case where `@` is pressed in the middle of text -- 

https://user-images.githubusercontent.com/26949006/236282442-43739f66-6658-4dde-aa5d-922cabb3d1f6.mov

packages/synapse-react-client/src/lib/containers/markdown/MarkdownEditor.tsx
```diff
@@ -56,7 +56,7 @@ export const MarkdownEditor: React.FunctionComponent<MarkdownEditorProps> = ({
       const start = textVal?.selectionStart
       const textBeforeTag = text.substring(0, start)
       const textAfterTag = text.substring(start, text.length)
-      setSelectionStart(start + user.length + 1)
+      setSelectionStart(start + user.length)
```
hallieswan, line 59: Adjust so that after a user tag is inserted in the middle of text, the cursor is positioned just after the user tag, rather than one character further. 

See example of what happened previously -- after the user tag is inserted, the cursor is positioned after 5, but we would expect it to be positioned after `@test`: 

https://user-images.githubusercontent.com/26949006/236282101-be94abe0-0139-4528-a5ae-4055e089e6aa.mov

The updated behavior can be seen in this [comment](https://github.com/Sage-Bionetworks/synapse-web-monorepo/pull/187/files#r1185307858).

PR #195, opened 2023-05-05
PORTALS-2628 - Fix button styling in MarkdownSynapse

nickgros:
- Remove styling that was incorrectly setting text color for buttons that were anchors
- Refine MarkdownButton to only apply `secondary` background color on hover in the neutral case. Previously, the background color did not change if highlight was `'true'` when it should have.
- Add `neutral` to theme. Same as grey, except we can reference it as a button color.
- Let MUI handle `a` links.


apps/portals/src/_Markdown.scss
```diff
@@ -34,7 +34,6 @@
     padding: 0;
   }
 
-  a,
```
nickgros, line 37: This style was setting the text color of `a` buttons to the primary color when it is now handled by the MUI Button component. I suspect, but have not confirmed, that this style was not applied to buttons before because of more specific CSS that has since been removed.

packages/synapse-react-client/src/lib/containers/markdown/MarkdownSynapse.tsx
```diff
@@ -442,6 +442,12 @@ export default class MarkdownSynapse extends React.Component<
               {children}
             </Typography>
           )
+        case 'a':
+          return (
+            <Link {...props} component={Tag}>
+              {children}
+            </Link>
+          )
```
nickgros, line 450: Let MUI handle `a` tags. This doesn't include the `a` button case, because that's handled separately by `MarkdownWidget`.

packages/synapse-react-client/src/lib/containers/markdown/widget/MarkdownButton.tsx
```diff
@@ -21,15 +22,18 @@ export default function MarkdownButton(
     buttonClasses += 'floatright '
   }
   const buttonIsCenterAligned = alignLowerCase === 'center'
-  const buttonVariant = highlight === 'true' ? 'secondary' : 'inherit'
+  const buttonColor = highlight ? 'secondary' : 'neutral'
```
nickgros, line 25: `'inherit'` isn't a real button color, this is a holdover from bootstrap that doesn't make sense, even if it still worked. Replace with a new color palette 'neutral'

packages/synapse-react-client/src/lib/containers/markdown/widget/MarkdownButton.tsx
```diff
@@ -21,15 +22,18 @@ export default function MarkdownButton(
...
     <WideButton
       href={widgetParamsMapped.url}
       className={buttonClasses}
       variant="contained"
-      color={buttonVariant}
+      color={buttonColor}
       sx={{
-        '&:hover': { backgroundColor: 'secondary.main', color: 'white' },
+        '&:hover': {
+          backgroundColor: highlight ? undefined : 'secondary.main',
```
nickgros, line 34: Don't set the background color in the `highlight === true` case. In that case, MUI created a color that is slightly different, which we don't want to override.

PR #201, opened 2023-05-08
SWC-6410 - Support ARs with 2FA in Download Cart

nickgros:
SWC-6410

Blocked by PLFM-7839

- Add ActionRequiredCard, side-effect free component for rendering an action item
- Add EnableTwoFaRequirementCard which opens the AccessRequirementList with those ARs that require 2FA. The AccessRequirementList component will guide the user through 2FA enrollment.
- Refactor existing action required cards to use ActionRequiredCard component
- Include `span` case in MarkdownSynapse
- Upgrade Storybook from 7.0.6 to 7.0.9

image

image


packages/synapse-react-client/package.json
```diff
@@ -122,15 +122,15 @@
...
     "@juggle/resize-observer": "^3.3.1",
     "@rollup/plugin-inject": "^4.0.4",
-    "@storybook/addon-actions": "^7.0.6",
-    "@storybook/addon-essentials": "^7.0.6",
-    "@storybook/addon-interactions": "^7.0.6",
-    "@storybook/addon-links": "^7.0.6",
-    "@storybook/manager-api": "^7.0.6",
-    "@storybook/react": "^7.0.6",
-    "@storybook/react-vite": "^7.0.6",
+    "@storybook/addon-actions": "^7.0.9",
```
nickgros, line 125: Was having some local issues with Storybook so I upgraded to the latest version to try to fix them.

packages/synapse-react-client/src/lib/containers/download_list_v2/ActionRequiredCard.tsx
```diff
@@ -0,0 +1,75 @@
...
+  gap: theme.spacing(4),
+}))
+
+/**
+ * The ActionRequiredCard component renders a generic card that represents some action that a user must take to gain
+ * download access to a file.
+ * @param props
+ * @returns
+ */
+export function ActionRequiredCard(props: ActionRequiredCardProps) {
```
nickgros, line 40: No side-effects or fetching in this component.

packages/synapse-react-client/src/lib/containers/download_list_v2/ActionRequiredCard.tsx
```diff
@@ -0,0 +1,75 @@
...
+      count: number
+      isLoading?: false
+    }
+  | { isLoading: true }
+
+const ActionRequiredCardContainer: StyledComponent<PaperProps> = styled(Paper, {
+  label: 'ActionRequiredCardContainer',
+})(({ theme }) => ({
+  padding: `${theme.spacing(2.5)} ${theme.spacing(4)}`,
+  display: 'flex',
```
nickgros, line 28: Previous implementation used `grid`, but we can do it just fine with flex with a bit less complexity.

packages/synapse-react-client/src/lib/containers/download_list_v2/ActionRequiredCard.tsx
```diff
@@ -0,0 +1,75 @@
...
+export type ActionRequiredCardProps =
+  | {
+      title: React.ReactNode
+      description: React.ReactNode
+      actionNode: React.ReactNode
+      iconType: string
+      count: number
+      isLoading?: false
+    }
+  | { isLoading: true }
```
nickgros, line 22: If `isLoading` is true, no need to supply the other props

packages/synapse-react-client/src/lib/containers/download_list_v2/DownloadListActionsRequired.tsx
```diff
@@ -63,28 +58,34 @@ export const DownloadListActionsRequired: React.FunctionComponent<
   const renderActionRequired = (actionRequiredCount: ActionRequiredCount) => {
     switch (actionRequiredCount.action.concreteType) {
       case 'org.sagebionetworks.repo.model.download.MeetAccessRequirement': {
-        const meetARAction: MeetAccessRequirement =
-          actionRequiredCount.action as MeetAccessRequirement
```
nickgros, line 67: No need for this case because I changed the type definitions so the type can be inferred when the concreteType matches that string.

packages/synapse-react-client/src/lib/containers/download_list_v2/EnableTwoFaRequirementCard.tsx
```diff
@@ -0,0 +1,56 @@
...
+        }
+      />
+      {isShowingAccessRequirement && ar && (
+        <AccessRequirementList
+          entityId={ar.subjectIds[0].id}
+          accessRequirementFromProps={[ar]}
+          renderAsModal={true}
+          numberOfFilesAffected={count}
+          onHide={() => setIsShowingAccessRequirement(false)}
+        />
```
nickgros, line 52: The `AccessRequirementList` component will walk the user through 2FA enrollment

packages/synapse-react-client/src/lib/containers/download_list_v2/MeetAccessRequirementCard.tsx
```diff
@@ -27,25 +25,17 @@ export const LOCK_TITLE = 'Access Restricted'
...
-    error: newError,
-  } = useGetAccessRequirements(accessRequirementId)
-  useEffect(() => {
-    if (isError && newError) {
-      handleError(newError)
-    }
-  }, [isError, newError, handleError])
+  const { data: ar, isLoading } = useGetAccessRequirements(
+    accessRequirementId,
+    { useErrorBoundary: true },
```
nickgros, line 30: The `useErrorBoundary: true` option is functionally identical to calling `handleError` with the error object. Since this will throw right away, we don't need to worry about handling cases where `isError` is true.

jay-hodgson (approved):

PR #202, opened 2023-05-08
Fix requests being unauthenticated in dev

nickgros:
- Fix synapse-react-client package.json to include `"type": "module"`
- Align plotly versions
- Fix type issues caused using multiple plotly modules
- Remove synapse-react-client from optimizeDeps.include because it is ESM
- Add plotly.js-basic-dist to optimizeDeps.include to fix import issue


apps/SageAccountWeb/package.json
```diff
@@ -12,7 +12,7 @@
     "@mui/system": "^5.11.16",
     "@mui/utils": "^5.11.13",
     "lodash-es": "^4.17.21",
-    "plotly.js-basic-dist": "^1.47.3",
+    "plotly.js-basic-dist": "^2.22.0",
```
nickgros, line 15: We're already using v2 of plotly.js-basic-dist in synapse-react-client, so this should not cause new problems.

apps/SageAccountWeb/package.json
```diff
@@ -31,7 +31,7 @@
     "@testing-library/user-event": "^14.4.3",
     "@types/katex": "^0.5.0",
     "@types/node": "11.10.4",
-    "@types/plotly.js": "^1.54.10",
+    "@types/plotly.js": "^2.12.18",
```
nickgros, line 34: And since we're using Plotly v2, we should upgrade the types to v2

apps/portals/package.json
```diff
@@ -33,7 +33,7 @@
     "lint": "eslint src",
     "copy-assets": "cp src/config/style/favicon* public/; cp src/config/style/logo.svg public/; cp src/config/style/socialmedia.png public/",
     "save-build-date": "date > public/build-date.txt",
-    "start": "vite",
+    "start": "vite --force",
```
nickgros, line 36:
```suggestion
    "start": "vite",
```

apps/portals/src/SynapseComponent.tsx
```diff
@@ -5,7 +5,7 @@ import {
 } from 'synapse-react-client/dist/utils/SynapseContext'
 import { SynapseConfig } from './types/portal-config'
 import PortalComponents from './portal-components'
-import { SynapseComponents } from 'synapse-react-client/dist/index'
+import { SynapseComponents } from 'synapse-react-client'
```
nickgros, line 8: Revert the change in #200 just to prove this is fixed systemically

packages/synapse-react-client/package.json
```diff
@@ -147,8 +147,8 @@
     "@types/lodash-es": "4.17.3",
     "@types/markdown-it": "^12.0.1",
     "@types/node": "^14.14.31",
-    "@types/plotly.js": "^1.54.20",
-    "@types/plotly.js-basic-dist": "^1.54.0",
+    "@types/plotly.js": "^2.12.18",
+    "@types/plotly.js-basic-dist": "^1.54.1",
```
nickgros, line 151: @types/plotly.js-basic-dist just mirrors the types in @types/plotly.js, so it's fine that it's at v1

packages/synapse-react-client/src/lib/containers/widgets/facet-nav/FacetNavPanel.tsx
```diff
@@ -531,7 +530,7 @@ const FacetNavPanel: React.FunctionComponent<FacetNavPanelProps> = (
             className={`FacetNavPanel__body${isModalView ? '--expanded' : ''}`}
             role="graphics-object"
           >
-            <SizeMe monitorHeight>
+            <SizeMe monitorHeight noPlaceholder>
```
nickgros, line 533: I was spot-checking our Plotly components and noticed that some don't appear on the first render in dev, but this isn't a new problem. `noPlaceholder` fixes it.

package.json
```diff
@@ -10,7 +10,7 @@
     "prepare": "husky install",
     "build": "nx run-many --target=build",
     "lint": "nx run-many --target=lint",
-    "test": "nx run-many --target=test -- --watchAll=false --silent=true",
+    "test": "nx run-many --target=test:ci",
```
nickgros, line 13: Create `test:ci` jobs explicitly for running in parallel (vitest doesn't use `watchAll` so the interactive test runner would be launched if running this script locally)

PR #204, opened 2023-05-08
EC-79: add a featured projects card list to the home page

jay-hodgson:


jay-hodgson:
This does not match the design, so this may be changed before release.

PR #206, opened 2023-05-08
override ELITE portal contrast text

jay-hodgson:


packages/synapse-react-client/src/lib/utils/theme/palette/Palettes.ts
```diff
@@ -116,10 +120,14 @@ export const adKnowledgePortalPalette: PaletteOptions = {
   secondary: generatePalette('#2f8e94'),
 }
 
+const elPortalPrimaryPalette = {
+  ...generatePalette('#39AC97'),
+  contrastText: '#ffffff',
```
nickgros, line 125: Aha, this makes sense! Is it possible to make this change (and remove the modified `contrastThreshold`) in 
- The ARK Portal palette also defined in this file
- SageAccountWeb (SourceApp.tsx) (this might not be possible today with palettes coming from the Synapse table)

nickgros (approved):

PR #209, opened 2023-05-09
EC-113: add main nav home page section (Goals component)

jay-hodgson:


jay-hodgson:
Screen Shot 2023-05-08 at 5 14 44 PM

jay-hodgson:
Note that the count for files and individuals (cohort) is defined by the data curator (and these are not yet set up)

PR #211, opened 2023-05-09
PORTALS-2656: fix secondary color contrast color

jay-hodgson:
Before:
Screen Shot 2023-05-09 at 9 51 00 AM

After:
Screen Shot 2023-05-09 at 9 53 10 AM


apps/portals/src/configurations/elportal/routesConfig.ts
```diff
@@ -15,7 +15,6 @@ import {
 } from './synapseConfigs/studies'
 import {
   projectCardConfiguration,
-  projectHomePageCardConfiguration,
```
jay-hodgson, line 18: Not used

PR #212, opened 2023-05-09
PORTALS-2624: refactor CreateOAuthClient to use MUI components

hallieswan:
type | main | feature
:---:|:---:|:---:
Create | main_CreateOAuthClient_Create | feature_CreateOAuthClient_Create
Edit |main_CreateOAuthClient_Edit | feature_CreateOAuthClient_Edit


packages/synapse-react-client/src/lib/containers/oauth/CreateOAuthClient.tsx
```diff
@@ -199,246 +214,224 @@ export const CreateOAuthModal: React.FunctionComponent<
...
-  return (
-    <div className="bootstrap-4-backport">
-      <Modal
-        show={isShowingModal && !isShowingConfirmModal}
-        animation={false}
-        backdrop="static"
-        onHide={() => {
-          hide()
-          setError(undefined)
+  const sectorIdentifierURIInput = (
```
hallieswan, line 217: Consolidate Sector Identifier URI input definitions into one location for creating and editing a client

packages/synapse-react-client/src/lib/containers/oauth/CreateOAuthClient.tsx
```diff
@@ -199,246 +214,224 @@ export const CreateOAuthModal: React.FunctionComponent<
...
-                              >
-                                <IconSvg
-                                  icon="delete"
-                                  sx={{ color: 'error.main' }}
-                                />
-                              </button>
-                            )}
+        Sector Identifier URI
+      </InputLabel>
+      <MuiTextField
```
hallieswan, line 229: SRC TextField changes the label to the primary color on focus. Use MUI TextField instead, so that the error label color persists on focus when editing an existing client

packages/synapse-react-client/src/lib/containers/oauth/CreateOAuthClient.tsx
```diff
@@ -199,246 +214,224 @@ export const CreateOAuthModal: React.FunctionComponent<
...
-                  </Typography>
-                  <Row>
-                    <Col lg={6} md={6} sm={12} xs={12}>
-                      <Form.Label htmlFor="redirect-uri-0">
-                        Redirect URI(s)
-                      </Form.Label>
-                      <Tooltip title={uriHelpMessage} placement="top">
-                        <HelpOutlineTwoTone className={`HelpButton`} />
-                      </Tooltip>
+  const redirectURIInputs = (
```
hallieswan, line 239: Consolidate definitions for Redirect URI inputs to one location for creating vs editing a client

packages/synapse-react-client/src/lib/containers/oauth/CreateOAuthClient.tsx
```diff
@@ -199,246 +214,224 @@ export const CreateOAuthModal: React.FunctionComponent<
...
+              fullWidth
+              id={`redirect-uri-${idx}`}
+              onChange={e => handleUriChange(e, idx)}
+              value={singleUri.uri}
+              placeholder="https://"
+              type="text"
+            />
+            {redirectUris.length > 1 && (
+              <IconButton
+                sx={{ mt: '24px', width: '50px' }}
```
hallieswan, line 274: Add margin to the IconButton so that it is positioned inline with the input, rather than the entire TextField. Also, specify a width so that the delete button has a circle background on hover (vs an ellipse).

packages/synapse-react-client/src/lib/containers/oauth/CreateOAuthClient.tsx
```diff
@@ -199,246 +214,224 @@ export const CreateOAuthModal: React.FunctionComponent<
...
+                value={tosUri}
+              />
+            </Grid>
+          </Grid>
+          {isEdit && (
+            <Box
+              sx={{ backgroundColor: 'rgb(178, 36, 42, 0.03)' }}
+              mt="10px"
+              padding="8px"
+            >
```
hallieswan, line 379: Style components directly rather than using _oauth-editor.scss

packages/synapse-react-client/src/lib/style/components/_all.scss
```diff
@@ -20,5 +20,5 @@
   './row_renderers/observation-card', './subsection-row-renderer',
   './dataset-editor', './fluid-modal', './markdown-collapse',
   './discussion-search', './submission-page', './governance-dashboard',
-  './oauth-editor', './provenance', './certification-quiz', './discussion-forum',
+  './provenance', './certification-quiz', './discussion-forum',
```
hallieswan, line 23: No longer used -- styling components directly instead

packages/synapse-react-client/src/lib/utils/theme/DefaultTheme.ts
```diff
@@ -39,14 +39,13 @@ const defaultMuiThemeOptions: ThemeOptions = {
             transition: '0.2s',
           },
         },
-        text: ({ theme }) => ({
+        text: {
           '&:hover': {
             textDecoration: 'underline',
             textUnderlineOffset: '4px',
-            textDecorationColor: theme.palette.primary.main,
```
hallieswan, line 46: Removing so that the textDecorationColor will align with the text button color, e.g. when hovering on an "error" text button, the textDecorationColor should also be the "error" color

packages/synapse-react-client/src/lib/containers/oauth/CreateOAuthClient.tsx
```diff
@@ -199,246 +214,224 @@ export const CreateOAuthModal: React.FunctionComponent<
...
-            Save
-          </Button>
-        </Modal.Footer>
-      </Modal>
+            <Grid item md={GRID_WIDE} xs={GRID_NARROW}>
+              <TextField
+                label="Client Name"
+                required
+                onChange={e => setClientName(e.target.value)}
+                placeholder="Client Name"
```
hallieswan, line 329: SRC TextField doesn't display placeholder text until the input field is focused, whereas MUI TextField displays placeholder text in all states until new text is entered into the field. Is one of these styles preferred? If so, I can work to align the placeholder text handling between the TextField components

packages/synapse-react-client/src/lib/utils/theme/DefaultTheme.ts
nickgros: If we need this style, we could use something like `theme.palette[ownerState.color || 'primary']['main']`, but removing as you have done is ideal if it's really unnecessary.

packages/synapse-react-client/src/lib/containers/oauth/CreateOAuthClient.tsx
nickgros: Can you just do `alignItems: 'center'` on the flex container instead of the custom margin?

packages/synapse-react-client/src/lib/containers/oauth/CreateOAuthClient.tsx
nickgros: Sounds like a bug in SRC TextField. Placeholder text should always be shown if there's no value

packages/synapse-react-client/src/lib/containers/oauth/CreateOAuthClient.tsx
nickgros: I'd prefer we use the MUI versions when possible, so this is fine. There's just something strange with our styling/implementation. For example MUI's [TextField](https://mui.com/material-ui/react-text-field/) internally handles adding an InputLabel, so we shouldn't need to provide one.

nickgros (approved):

packages/synapse-react-client/src/lib/containers/oauth/CreateOAuthClient.tsx
hallieswan: Oh good call, I missed that! I updated the inputs to use MUI's TextField wrapped in a StyledFormControl instead

packages/synapse-react-client/src/lib/containers/oauth/CreateOAuthClient.tsx
hallieswan: Potentially! For ease of integration with MUI's TextField, I moved the delete icon to an endAdornment, which eliminates the custom styling. Happy to move the delete icon outside of the field, if that's preferred though!

image

packages/synapse-react-client/src/lib/utils/theme/DefaultTheme.ts
hallieswan: Oh great, thanks! I was looking for something like that. I'll add it back for now, in case it is needed.

packages/synapse-react-client/src/lib/containers/oauth/CreateOAuthClient.tsx
```diff
@@ -199,246 +213,237 @@ export const CreateOAuthModal: React.FunctionComponent<
...
-                                sx={{ color: 'error.main' }}
-                              />
-                            </button>
-                          )}
+  const redirectURIInputs = (
+    <StyledFormControl fullWidth variant="standard" margin="normal">
+      {redirectUris?.map((singleUri, idx) => (
+        <div key={idx}>
+          <TextField
+            required={idx === 0}
```
hallieswan, line 248: Only show `*` for the first redirect URI input field

packages/synapse-react-client/src/lib/containers/oauth/CreateOAuthClient.tsx
```diff
@@ -199,246 +213,237 @@ export const CreateOAuthModal: React.FunctionComponent<
...
+            InputLabelProps={{
+              sx: { ...dangerLabelSx, pointerEvents: 'auto', zIndex: 1 },
+            }}
+            label={
+              idx === 0 && (
```
hallieswan, line 253: Only display label for first redirect URI input field

packages/synapse-react-client/src/lib/containers/oauth/CreateOAuthClient.tsx
```diff
@@ -199,246 +213,237 @@ export const CreateOAuthModal: React.FunctionComponent<
...
+            type="text"
+            InputProps={{
+              endAdornment: redirectUris.length > 1 && (
+                <InputAdornment position="end">
+                  <IconButton onClick={() => handleRedirectUriRemove(idx)}>
+                    <DeleteTwoToneIcon sx={{ color: 'error.main' }} />
+                  </IconButton>
+                </InputAdornment>
+              ),
+            }}
```
hallieswan, line 280: Use endAdornment for delete icon, rather than positioning next to the input field

packages/synapse-react-client/src/lib/containers/oauth/CreateOAuthClient.tsx
```diff
@@ -199,246 +213,237 @@ export const CreateOAuthModal: React.FunctionComponent<
...
-            CANCEL
-          </Button>
-          <Button variant="primary" onClick={onCreateClient}>
-            Save
-          </Button>
-        </Modal.Footer>
-      </Modal>
+            <Grid item md={GRID_WIDE} xs={GRID_NARROW}>
+              <StyledFormControl fullWidth variant="standard" margin="normal">
+                <TextField
```
hallieswan, line 326: Use MUI TextField instead of SRC TextField

packages/synapse-react-client/src/lib/utils/theme/DefaultTheme.ts
```diff
@@ -131,6 +132,11 @@ const defaultMuiThemeOptions: ThemeOptions = {
         }),
       },
     },
+    MuiInputLabel: {
+      defaultProps: {
+        shrink: true,
+      },
+    },
```
hallieswan, line 139: Show placeholder text in TextField, even when input is not currently focused

hallieswan:
Styling after updates -- 

create | edit
:---:|:---:
feature_CreateOAuthClient_Create | feature_CreateOAuthClient_Edit

packages/synapse-react-client/test/lib/containers/oauth/CreateOAuthClient.integration.test.tsx
```diff
@@ -52,8 +52,8 @@ function setUp(props: CreateOAuthModalProps = defaultProps) {
   const user = userEvent.setup()
   const component = renderComponent(props)
   const inputs = {
-    name: screen.getByLabelText('Client Name'),
-    homePage: screen.getByLabelText('Client Homepage'),
+    name: screen.getByRole('textbox', { name: 'Client Name' }),
+    homePage: screen.getByRole('textbox', { name: 'Client Homepage' }),
```
nickgros, line 56: Strange that our MUI TextField isn't giving us accessible labels...I wonder if that's a bug on our end.

packages/synapse-react-client/src/lib/containers/oauth/CreateOAuthClient.tsx
```diff
@@ -199,246 +213,237 @@ export const CreateOAuthModal: React.FunctionComponent<
...
+            }}
+            label={
+              idx === 0 && (
+                <>
+                  Redirect URI(s)
+                  <Tooltip title={uriHelpMessage} placement="top">
+                    <HelpOutlineTwoTone
+                      className={`HelpButton`}
+                      sx={{ float: 'right' }}
+                    />
```
nickgros, line 260: Looking at the screenshots more closely, it looks like this shrank. Any idea why?

packages/synapse-react-client/src/lib/containers/oauth/CreateOAuthClient.tsx
nickgros: Tracking that set of work here: https://sagebionetworks.jira.com/browse/PORTALS-2658

packages/synapse-react-client/src/lib/containers/oauth/CreateOAuthClient.tsx
hallieswan: Great! Yes, the button still works and the input handles a long URL --

https://github.com/Sage-Bionetworks/synapse-web-monorepo/assets/26949006/60fe0abc-01e2-451d-b107-b936512372e4

packages/synapse-react-client/test/lib/containers/oauth/CreateOAuthClient.integration.test.tsx
hallieswan: I think when a field is required, there is an asterisk span included in the label, although it's hidden from the accessibility API (`aria-hidden="true"`) -- 

```
<label
  class="MuiFormLabel-root MuiInputLabel-root MuiInputLabel-formControl MuiInputLabel-animated MuiInputLabel-shrink MuiInputLabel-outlined MuiFormLabel-colorPrimary Mui-required MuiInputLabel-root MuiInputLabel-formControl MuiInputLabel-animated MuiInputLabel-shrink MuiInputLabel-outlined css-199xmkx-MuiFormLabel-root-MuiInputLabel-root"
  data-shrink="true"
  for="clientName"
  id="clientName-label"
>
    Client Name
    <span
      aria-hidden="true"
      class="MuiFormLabel-asterisk MuiInputLabel-asterisk css-1ygj4mw-MuiFormLabel-asterisk"
    >                     
      *
    </span>
</label>
```

In cases where the label text is broken up, [react-testing-library](https://testing-library.com/docs/queries/bylabeltext#name) recommends using getByRole with a name --

> The example above does NOT find the input node for label text broken up by elements. You can use getByRole('textbox', { name: 'Username' }) instead which is robust against switching to aria-label or aria-labelledby.

We could still find "Client Homepage" by label text though, since it's not a required field. I'll make that change in the next commit!

packages/synapse-react-client/src/lib/containers/oauth/CreateOAuthClient.tsx
```diff
@@ -213,55 +209,35 @@ export const CreateOAuthModal: React.FunctionComponent<
     }
   }
 
-  const labelSx = {
-    fontWeight: 700,
-    mb: '4px',
-  }
```
hallieswan: Use styling applied by MUI default theme to MUI TextField instead

packages/synapse-react-client/src/lib/containers/oauth/CreateOAuthClient.tsx
```diff
@@ -213,55 +209,35 @@ export const CreateOAuthModal: React.FunctionComponent<
...
-            label={
-              idx === 0 && (
-                <>
-                  Redirect URI(s)
-                  <Tooltip title={uriHelpMessage} placement="top">
-                    <HelpOutlineTwoTone
-                      className={`HelpButton`}
-                      sx={{ float: 'right' }}
-                    />
-                  </Tooltip>
```
hallieswan: Remove tooltip as it doesn't provide substantive information and is difficult to style when included as part of the label prop

packages/synapse-react-client/src/lib/containers/oauth/CreateOAuthClient.tsx
```diff
@@ -278,6 +254,7 @@ export const CreateOAuthModal: React.FunctionComponent<
                 </InputAdornment>
               ),
             }}
+            margin={idx === 0 ? 'normal' : 'dense'}
```
hallieswan, line 257: Reduce margin for subsequent redirection URI fields, so the fields appear closer together

packages/synapse-react-client/src/lib/containers/oauth/CreateOAuthClient.tsx
```diff
@@ -213,55 +209,35 @@ export const CreateOAuthModal: React.FunctionComponent<
...
-        onChange={e => setSectorUri(e.target.value)}
-        placeholder="https://"
-        type="text"
-        value={sectorUri}
-        id="sectorURI"
-        label="Sector Identifier URI"
-        InputLabelProps={{ sx: dangerLabelSx }}
-      />
-    </StyledFormControl>
+    <TextField
```
hallieswan, line 220: No longer need the StyledFormControl -- use the MUI theme styling instead

packages/synapse-react-client/src/lib/containers/oauth/CreateOAuthClient.tsx
```diff
@@ -359,37 +334,35 @@ export const CreateOAuthModal: React.FunctionComponent<
...
           </Grid>
           {isEdit && (
             <Box
               sx={{ backgroundColor: 'rgb(178, 36, 42, 0.03)' }}
               mt="10px"
-              padding="8px"
+              padding={1}
             >
-              <Typography sx={{ mt: '8px' }} color="error" variant="headline3">
+              <Typography sx={{ my: 1 }} color="error" variant="headline3">
```
hallieswan: Add extra margin below the Danger Zone text and use theme spacing

packages/synapse-react-client/src/lib/utils/theme/DefaultTheme.ts
```diff
@@ -111,6 +111,16 @@ const defaultMuiThemeOptions: ThemeOptions = {
         }),
       },
     },
+    MuiFormHelperText: {
```
hallieswan, line 114: Styling from StyledFormControl

packages/synapse-react-client/src/lib/utils/theme/DefaultTheme.ts
```diff
@@ -126,16 +136,39 @@ const defaultMuiThemeOptions: ThemeOptions = {
...
       },
     },
     MuiInputLabel: {
       defaultProps: {
         shrink: true,
       },
+      styleOverrides: {
+        root: ({ theme }) => ({
+          fontWeight: 700,
+          mb: '4px',
```
hallieswan, line 158: Styling from SRC TextField

packages/synapse-react-client/src/lib/utils/theme/DefaultTheme.ts
```diff
@@ -126,16 +136,39 @@ const defaultMuiThemeOptions: ThemeOptions = {
...
+        }),
+      },
+    },
+    MuiTextField: {
+      styleOverrides: {
+        root: ({ theme, ownerState }) => ({
+          [`& .${inputBaseClasses.root}`]: {
+            marginTop: ownerState.label && theme.spacing(3),
+          },
+        }),
```
hallieswan, line 170: Only add margin for the label when the TextField actually has a label, so that TextFields without labels don't have extra spacing added (e.g. CopyToClipboardInput)

packages/synapse-react-client/test/lib/containers/oauth/CreateOAuthClient.integration.test.tsx
```diff
@@ -42,6 +42,10 @@ const defaultProps: CreateOAuthModalProps = {
   setIsShowingConfirmModal: jest.fn(),
 }
 
+jest
+  .spyOn(SynapseClient, 'isOAuthClientReverificationRequired')
+  .mockReturnValue(Promise.resolve({ reverificationRequired: true }))
```
hallieswan, line 47: Move mock outside of specific test, so the mock applies in all cases, e.g. in "Shows a warning modal when deleteing a client" as well as "Shows a warning modal when changing redirect uri"

packages/synapse-react-client/test/lib/containers/oauth/CreateOAuthClient.integration.test.tsx
```diff
@@ -53,7 +57,7 @@ function setUp(props: CreateOAuthModalProps = defaultProps) {
   const component = renderComponent(props)
   const inputs = {
     name: screen.getByRole('textbox', { name: 'Client Name' }),
-    homePage: screen.getByRole('textbox', { name: 'Client Homepage' }),
+    homePage: screen.getByLabelText('Client Homepage'),
```
hallieswan, line 60: Use `getByLabelText` to find this field, as discussed [here](https://github.com/Sage-Bionetworks/synapse-web-monorepo/pull/212#discussion_r1190191428)

hallieswan:
Styling after latest updates -- 

create | edit
:---:|:---:
feature_CreateOAuthClient_Create | feature_CreateOAuthClient_Edit

packages/synapse-react-client/src/lib/utils/theme/DefaultTheme.ts
nickgros: Thanks for getting us closer to a unified solution!

nickgros (approved):

PR #222, opened 2023-05-11
PORTALS-2446: create challenge portal with placeholders

jay-hodgson:
Screen Shot 2023-05-11 at 3 38 33 PM


apps/portals/src/configurations/challengeportal/scripts/exportS3ProductionBucketName.sh
```diff
@@ -0,0 +1 @@
+export S3_PRODUCTION_BUCK_LOCATION=s3://prod.challengeportal.synapse.org
```
jay-hodgson, line 1: TBD by  @xschildw

apps/portals/src/configurations/challengeportal/scripts/exportS3StagingBucketName.sh
```diff
@@ -0,0 +1 @@
+export S3_STAGING_BUCKET_LOCATION=s3://staging.challengeportal.synapse.org
```
jay-hodgson, line 1: TBD by  @xschildw

apps/portals/src/configurations/challengeportal/headerConfig.tsx
```diff
@@ -0,0 +1,23 @@
...
+import { ReactComponent as HeaderImg } from './style/dna_header.svg'
+import { ReactComponent as SageLogo } from './style/SageLogo.svg'
+import { ReactComponent as DreamLogo } from './style/DreamLogo.svg'
+import { Box } from '@mui/material'
+
+const homePageHeader: HomePageHeaderConfig = {
+  title: 'Welcome to the Sage Challenge Portal!',
+  summary: (
+    <>
+      Crowd-sourced challenge-based benchmarking is an emerging paradigm for the development and unbiased assessment of tools and algorithms that aim to translate biological and biomedical big data into useful knowledge. By distributing problems to large communities of expert volunteers, complex biological, health care, and life sciences questions can be addressed efficiently and quickly, while incentivizing adoption of new standards. Challenges provide a successful resolution to the “self-assessment trap,” providing an objective framework for the evaluation of algorithms and methods
```
nickgros, line 12: you could do `{import.meta.env.VITE_PORTAL_DESCRIPTION}` if this should be exactly the same as the .env content

apps/portals/src/portal-components/challengeportal/ChallengeDetailsPage.tsx
nickgros, line 1: Wonder if this will be necessary or if the model works within the capabilities of our current DetailsPage. We'll see!

nickgros (approved):

apps/portals/src/portal-components/challengeportal/ChallengeDetailsPage.tsx
jay-hodgson: I think it starts to break down with the dynamic number of tasks associated with each challenge (and thus task tabs).  Discovery of task specific content is also complicated.

PR #231, opened 2023-05-15
Add test:ci, align plotly versions

nickgros:
- In SRC we're using Plotly.js v2, so bump that to the most recent v2 release in all packages
- Fix plotly types packages to avoid conflict
- Add test:ci so the Nx parallel runner doesn't enter interactive mode locally, and coverage will be collected in CI

### DIFF
--- a/package.json
+++ b/package.json
@@ -10,7 +10,7 @@
     "prepare": "husky install",
     "build": "nx run-many --target=build",
     "lint": "nx run-many --target=lint",
-    "test": "nx run-many --target=test -- --watchAll=false --silent=true",
+    "test": "nx run-many --target=test:ci",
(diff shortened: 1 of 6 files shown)


package.json
```diff
@@ -10,7 +10,7 @@
     "prepare": "husky install",
     "build": "nx run-many --target=build",
     "lint": "nx run-many --target=lint",
-    "test": "nx run-many --target=test -- --watchAll=false --silent=true",
```
nickgros, line 13: vitest doesn't have the `watchAll` option, so if you ran this locally, it would incorrectly enter interactive mode in the parallel runner. So let's use a `test:ci` job that is configured for each project.

jay-hodgson (approved):

PR #226, opened 2023-05-12
SWC-6410 - fix styling issues

nickgros:
- Fix styling within download cart by dropping some bootstrap
- Always use white for contrast text

PR #232, opened 2023-05-15
GenUI-ADO-2621: Extended MUI button to be aware of Synapse context login state

geoffroth:
### ADO Ticket:
[FEATURE 2621](https://dev.azure.com/GenUIADO/GenUI%20Sage%20Bionetworks%20-%20Challenge%20Portal/_workitems/edit/2621)

### Description:
Provides a button that extends both MUI Button and React Router's Link. 
- When logged in, the component will act as an anchor and accepts either the 'href' or 'to' properly. 
- When logged out, the component will add the `SRC_SIGN_IN_CLASS` class to it's button to trigger the login flow on click.
- It may be visually customized in the same manner as any MUI button. The Storybook example was passed sx params to match the "Register for this Challenge" button on the Challenge Details Page

#### Register
Storybook: 
image
Figma: 
image

#### Leave
Storybook:
image
Figma:
image

### Testing
New tests were provided to verify logged in vs logged out behavior.


packages/synapse-react-client/test/lib/containers/widgets/LoginAwareButton.test.tsx
```diff
@@ -0,0 +1,129 @@
...
+const defaultProps: LoginAwareButtonProps = {
+  variant: 'contained',
+  color: 'secondary',
+  sx: {
+    color: 'white',
+    fontWeight: 400,
+    ':hover': { color: 'white' },
+    ':active': { color: 'white' },
+    ':visited': { color: 'white' },
+  },
```
nickgros, line 23: Hopefully [this change in Palettes.ts](https://github.com/Sage-Bionetworks/synapse-web-monorepo/pull/226/files#r1192714592) makes hard-coding the text color unnecessary

nickgros (approved):

packages/synapse-react-client/test/lib/containers/widgets/LoginAwareButton.test.tsx
geoffroth: I removed the styling from this test because it was superfluous. I've confirmed white is now the default in effect for button variant="contained". For variant="outlined" specifying white appears to still be necessary however:
image

PR #234, opened 2023-05-16
PORTALS-2449: Add Challenge type generic card (with associated icon) …

jay-hodgson:
…using the old ways, a generic Tabs portals component (that takes SynapseConfigs as input), and the new cards list for Active and Closed Challenges in the Challenge portal


jay-hodgson:
Screen Shot 2023-05-15 at 5 52 25 PM
This is linked to the existing placeholder Challenge Detail page.

apps/portals/src/portal-components/TabbedSynapseObjects.tsx
```diff
@@ -0,0 +1,33 @@
...
+  return <Box className="TabbedSynapseObjects">
+    <Tabs
+      value={selectedTabConfig.label}
+      centered={centerTabs}
+    >
+      {tabConfigs.map((tabConfig, index) => {
+        return <Tab key={index} value={tabConfig.label} disableRipple={true}
+        label={tabConfig.label}
+        disableTouchRipple
+        onClick={() => setSelectedTabConfig(tabConfig)}/>
```
nickgros, line 28: `disableRipple` and `disableTouchRipple` could be added as `defaultProps` in the theme if we never want them

apps/portals/src/portal-components/TabbedSynapseObjects.tsx
nickgros, line 1: Hopefully these tabs don't need custom routing! Is it likely we'll have multiple tab groups per page?

nickgros (approved):
Looks great!

apps/portals/src/portal-components/TabbedSynapseObjects.tsx
jay-hodgson: That's a good point.  I don't think we need to restore the tab selection state in this case, but in the future we could add an optional variable that could be utilized (to store in a search parameter if sharing is important, or local storage).

PR #236, opened 2023-05-16
PORTALS-2447: Challenge Details page in progress

jay-hodgson:


apps/portals/src/portal-components/DetailsPage/DetailsPage.tsx
```diff
@@ -340,9 +340,7 @@ export const SplitStringToComponent: React.FC<{
   })
   if (overrideSqlSourceTable) {
     // use the search param value to override the sql param.
-    injectedProps['sql'] = `SELECT  *  FROM  ${value}${
-      rowVersionNumber ? `.${rowVersionNumber}` : ''
-    }`
+    injectedProps['sql'] = `SELECT  *  FROM  ${value}`
```
jay-hodgson, line 343: Using the row version number with the column value only makes sense if the column value is the ID

jay-hodgson:
Screen Shot 2023-05-16 at 10 50 43 AM
Screen Shot 2023-05-16 at 10 50 55 AM
Screen Shot 2023-05-16 at 10 51 06 AM

apps/portals/src/portal-components/DetailsPage/DetailsPage.tsx
```diff
@@ -341,7 +341,7 @@ export const SplitStringToComponent: React.FC<{
   if (overrideSqlSourceTable) {
     // use the search param value to override the sql param.
     injectedProps['sql'] = `SELECT  *  FROM  ${value}${
-      rowVersionNumber ? `.${rowVersionNumber}` : ''
+      rowVersionNumber && columnName == 'id' ? `.${rowVersionNumber}` : ''
     }`
   }
```
nickgros, line 346: Wait, why are we using the `rowVersionNumber` at all in this query? Shouldn't a version number, if any, come from the `value`?

apps/portals/src/portal-components/DetailsPage/DetailsPage.tsx
```diff
@@ -341,7 +341,7 @@ export const SplitStringToComponent: React.FC<{
   if (overrideSqlSourceTable) {
     // use the search param value to override the sql param.
```
nickgros, line 342:
```suggestion
    // use the search param value to override the sql param.
    // TODO: Refactor to consider the the type of the original table
    // For datasets and views, the rowVersionNumber corresponds to the actual version of the table. For TableEntities, the rowVersionNumber is meaningless. For now, just see if the columnName is `id`, which is always true for current view/dataset cases, and is never the case for current table cases.
    // 
```

PR #239, opened 2023-05-16
PORTALS-2659: Use tsup to bundle synapse-react-client

nickgros:
- Move Synapse type definitions from synapse-react-client to new package `@sage-bionetworks/synapse-types`
- Change synapse-react-client bundle strategy for CJS and ESM. The UMD bundle (used by SWC) is unchanged.
  - Before, `tsc` would emit ESM code and type definitions matching the directory structure. Assets were copied so that non-native features like SVG import could be processed by downstream bundlers. This was not really a "bundle" in any sense.
  - Now, use [tsup](https://github.com/egoist/tsup) to bundle and emit a singular files with CJS, ESM, and a singular type definition file. Configure tsup to use the same esbuild plugins we use for UMD, so those features enabled by bundling (like SVG import) are handled in these bundles.
- Reorganize synapse-react-client
  - Flatten to remove `src/lib` (contents of `lib` are moved up one level)
  - All components are now in `src/components` (removed `src/lib/containers`)
  - Organized many components to be contained in their own folders, and exported with an `index.ts`
  - All components are now exported in the main bundle
  - Most utils are now exported in the main bundle
  - Move `SynapseClient` from `src/lib/utils` to `src/synapse-client` (I think this may be valuable to move to a new package later, just starting that process)
  - Move Synapse-related react-query hooks in `src/lib/utils/hooks/SynapseAPI` to `src/synapse-queries`
 

Also includes a fix for PORTALS-2643


apps/SageAccountWeb/src/LoginPage.tsx
```diff
@@ -1,28 +1,28 @@
...
+import { backButtonSx } from './components/BackButton.js'
 
 export type LoginPageProps = {
   returnToUrl?: string
 }
 
 function LoginPage(props: LoginPageProps) {
   const { returnToUrl } = props
-  const { refreshSession, twoFactorAuthErrorResponse } =
+  const { refreshSession, twoFactorAuthSSOErrorResponse } =
```
nickgros, line 25: Was using the wrong name for the object. I have a feeling this will fix https://sagebionetworks.jira.com/browse/PORTALS-2643 but need to confirm

.eslintrc.cjs
```diff
@@ -24,6 +24,7 @@ module.exports = {
       "./packages/**/tsconfig.json",
       "./apps/**/tsconfig.json",
     ],
+    ignorePatterns: ["dist/*"],
```
nickgros, line 27: Don't lint `dist`

packages/synapse-react-client/introduction.md
nickgros, line 1: This file was redundant, we have a README

packages/synapse-react-client/package.json
```diff
@@ -213,28 +214,35 @@
...
     "test:coverage": "jest --coverage",
-    "build": "pnpm build:js && pnpm build:copy-assets && pnpm build:esbuild",
-    "build:js": "tsc --build",
-    "build:copy-assets": "copyfiles -u 2 \"src/lib/**/*.{css,scss,svg}\" dist",
-    "build:esbuild": "rimraf dist/umd && node esbuild.config.mjs",
-    "serve:docs": "pnpm storybook",
+    "build": "rimraf dist && pnpm build:copy-assets && pnpm build:js && pnpm build:esbuild",
+    "build:js": "tsup && pnpm build:fix-extension",
+    "build:fix-extension": "mv dist/synapse-react-client.js dist/synapse-react-client.mjs",
+    "build:copy-assets": "copyfiles -u 1 \"src/**/*.{css,scss,svg}\" dist",
```
nickgros, line 226: Change depth from 2 to 1. This may affect SWC, since the SCSS that it loads will no longer be in `dist/lib/assets`, now in `dist/assets`

.github/workflows/build-test.yml
```diff
@@ -52,3 +52,13 @@ jobs:
...
+  typecheck:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3.1.0
+        with:
+          fetch-depth: 0
+      - uses: ./.github/actions/pnpm-setup-action
+      # Portals test configuration must be loaded for a successful type check
+      - run: pnpm nx run portals:copy-test-configuration
+      - run: pnpm nx affected --target=type-check --parallel=2 --base=remotes/origin/main
```
nickgros, line 64: Type-checking is not included in our current build processes, so add it as a step in CI

packages/synapse-react-client/package.json
```diff
@@ -214,29 +215,36 @@
...
     "test:ci": "jest --watchAll=false --coverage",
     "test:coverage": "jest --coverage",
-    "build": "pnpm build:js && pnpm build:copy-assets && pnpm build:esbuild",
-    "build:js": "tsc --build",
-    "build:copy-assets": "copyfiles -u 2 \"src/lib/**/*.{css,scss,svg}\" dist",
-    "build:esbuild": "rimraf dist/umd && node esbuild.config.mjs",
-    "serve:docs": "pnpm storybook",
+    "build": "rimraf dist && pnpm build:copy-assets && pnpm build:js && pnpm build:esbuild",
+    "build:js": "tsup && pnpm build:fix-extension",
+    "build:fix-extension": "mv dist/synapse-react-client.js dist/synapse-react-client.mjs",
```
nickgros, line 226: For some reason, vitest won't read these when they are `.js`, even though we have `"type": "module"' in both projects. Just rename them after bundling.

packages/synapse-react-client/package.json
```diff
@@ -214,29 +215,36 @@
...
+    "build:copy-assets": "copyfiles -u 1 \"src/**/*.{css,scss,svg}\" dist",
```
nickgros, line 227: SWC may be affected by this change since the scss files will no longer be in `dist/lib/style`, now `dist/style`

packages/synapse-types/package.json
```diff
@@ -0,0 +1,17 @@
+{
+  "name": "@sage-bionetworks/synapse-types",
```
nickgros, line 2: New project! Just moved the types here. Much work to be done to clean it up

packages/vite-config/src/vite-config.ts
```diff
@@ -35,7 +35,9 @@ export default defineConfig({
     __DEV__: JSON.stringify(false),
   },
   optimizeDeps: {
-    include: ["synapse-react-client"],
+    // In the dev server, Vite doesn't automatically optimize plotly.js-basic-dist when it should. This causes a broken import
+    // This is probably because plotly.js-basic-dist is a UMD module, and we load the entire object returned by the UMD module into react-plotly.js.
+    include: ["plotly.js-basic-dist"],
     esbuildOptions: {
```
nickgros, line 41: We could also remove SRC from this list because it's bundled as ESM now.

.eslintrc.cjs
```diff
@@ -9,6 +9,7 @@ module.exports = {
     "jest-dom",
     "@typescript-eslint",
   ],
+    ignorePatterns: ["apps/portals/src/tests/*"],
```
nickgros, line 12: We don't want to lint these right now

packages/synapse-react-client/src/components/index.ts
```diff
@@ -0,0 +1,58 @@
+export * from './AccessTokenPage'
```
nickgros, line 1: If you want to export a new component, make a new folder and export the contents of that folder here

jay-hodgson (approved):

PR #240, opened 2023-05-16
additional enhancement to the challenge detail page

jay-hodgson:
- Added a wrapper around the SRC Google Map for the Challenge Portal - it needs to find the participant team ID based on the Challenge Project ID found in the search params of the details page.
- Added a useGetEntityChallenge() hook to get the Challenge associated to an Entity
- Add ability to hide a tab if a column value is undefined, then added Task tabs (that are hidden if annotation is not populated for that task)


jay-hodgson:
Correct participant team is found, and special marker for Synapse user is added, but I don't see it on the map (yet).

Screen Shot 2023-05-16 at 3 09 02 PM

jay-hodgson:
Screen Shot 2023-05-16 at 4 23 22 PM

apps/portals/src/types/portal-util-types.ts
```diff
@@ -59,4 +59,6 @@ export type DetailsPageTabProps = DetailsPageContent & {
   iconName?: string
   cssClass?: string
   toolTip?: string
+  /** Given this column name, if the assocated value in this column is not set, then this tab will be hidden */
+  hideIfColumnValueUndefined?: string
```
nickgros, line 63: The actual JSON values Synapse returns for a row are string or null, so l think this name would be a bit more accurate

```suggestion
  /** Given this column name, if the assocated value in this column is not set, then this tab will be hidden */
  hideIfColumnValueNull?: string
```

apps/portals/src/portal-components/DetailsPage/DetailsPageTabs.tsx
```diff
@@ -40,6 +42,16 @@ const DetailsPageTabs: React.FunctionComponent<DetailsPageTabsProps> = (
       </Switch>
       <div className="tab-groups">
         {tabConfigs.map((tab, index) => {
+          if (tab.hideIfColumnValueUndefined) {
+            if (rowValues && headers) {
+              const colIndex = headers.findIndex(h => h.name == tab.hideIfColumnValueUndefined)
+              if (!rowValues[colIndex]) {
```
nickgros, line 48: I'd expect us to filter nullish values rather than falsy values here. Since these are always strings or null, I think the only case where this will make a difference is the empty string. Does Synapse ever give us empty strings, and should we treat those like `null`?
```suggestion
              if (rowValues[colIndex] == null) {
```

apps/portals/src/portal-components/challengeportal/ChallengeParticipantGoogleMap.tsx
```diff
@@ -0,0 +1,18 @@
+import { Box } from '@mui/material'
+import * as React from 'react'
+import Map from 'synapse-react-client/dist/containers/GoogleMap/GoogleMap'
+import { useGetEntityChallenge } from 'synapse-react-client/dist/utils/hooks/SynapseAPI/entity/useGetEntityChallenge'
+
+const ChallengeParticipantGoogleMap = () => {
+  const urlSearchParams = new URLSearchParams(window.location.search)
+  const entityId = urlSearchParams.get('id') ?? ''
+  const {data: challenge } = useGetEntityChallenge(entityId)
```
nickgros, line 9: Looks like the precommit code formatter isn't running here

nickgros (approved):
Looks good in general, just some considerations about naming/edge cases for null vs undefined vs falsy

PR #241, opened 2023-05-17
PORTALS-2652 - Integrate with Chromatic

nickgros:
- Integrate with Chromatic

For an example, see nickgros/synapse-web-monorepo#3 . There will be two new checks:

image

The **Storybook Publish** check links to a deployed Storybook. A PR submitter could link to a specific story in the Storybook if desired, [like this](https://644bff4128662f11cf28efb6-oeoubvfloi.chromatic.com/?path=/story/governance-data-access-request-flow-accessrequirementlist--has-unmet-requirements&globals=stack:mock).

The **UI Tests** check links to a list of all of the components where the base snapshot changed since the last deployment. [Here's the list of changes from the example PR above](https://www.chromatic.com/build?appId=644bff4128662f11cf28efb6&number=10). Snapshot diffs can be reviewed and approved. [Here's a specific example of a snapshot diff for a component that changed in the example PR](https://www.chromatic.com/test?appId=644bff4128662f11cf28efb6&id=6464f99c102217be13223874). I don't think we'll want to require UI approval to merge for now, but this could be a useful resource to validate a subset of changes. 

The number of snapshots we can create per month is limited by the plan that we're on, so we shouldn't rely on this as a part of our SDLC until we understand the value, get a sense of our usage, and can determine if the cost is worth it (or if we should consider alternatives).


packages/synapse-react-client/package.json
```diff
@@ -228,7 +229,8 @@
     "build:esbuild": "rimraf dist/umd && node esbuild.config.mjs",
     "serve:docs": "pnpm storybook",
     "build:docs": "pnpm build:storybook",
-    "build:storybook": "storybook build && touch storybook-static/.nojekyll",
+    "build:build:gh-pages": "pnpm build-storybook && touch storybook-static/.nojekyll",
+    "build-storybook": "storybook build",
```
nickgros, line 233: Chromatic expects a `build-storybook` command. They let us override with a custom command, but that was causing issues for me where the job would fail without a meaningful error message.

jay-hodgson (approved):

PR #244, opened 2023-05-18
add Forum object, and a Community tab to the Challenge Details Page

jay-hodgson:
https://github.com/Sage-Bionetworks/synapse-web-monorepo/assets/1864447/a98d3572-5364-46b8-a7eb-a921c9a6c123

### DIFF
--- a/apps/portals/src/portal-components/ProjectDiscussionForum.tsx
+++ b/apps/portals/src/portal-components/ProjectDiscussionForum.tsx
@@ -5,20 +5,22 @@ import { ForumPage } from 'synapse-react-client/dist/containers/discussion_forum
 import ArrowBackIcon from '@mui/icons-material/ArrowBack'
 import { useHistory, useLocation } from 'react-router-dom'
 import { Box, IconButton } from '@mui/material'
+import { generateExcludedSearchParamKey } from 'synapse-react-client/dist/utils/functions/sqlFunctions'
 
 const ProjectDiscussionForum = () => {
   const urlSearchParams = new URLSearchParams(window.location.search)
   const entityId = urlSearchParams.get('id') ?? ''
-  const threadId = urlSearchParams.get('threadIdNotAQueryFilterKey') ?? ''
+  const threadIdSearchParamKey:string = generateExcludedSearchParamKey('threadId', 'forum')
+  const threadId = urlSearchParams.get(threadIdSearchParamKey) ?? ''
   const location = useLocation()
   const history = useHistory()
   const { data: forum } = useGetEntityForum(entityId)
   const updateThreadId = (threadId?:string) => {
     const searchParams = new URLSearchParams(location.search)
     if (threadId) {
-      searchParams.append('threadIdNotAQueryFilterKey', threadId)
+      searchParams.append(threadIdSearchParamKey, threadId)
     } else {
-      searchParams.delete('threadIdNotAQueryFilterKey')
+      searchParams.delete(threadIdSearchParamKey)
     }
     history.replace({ pathname: location.pathname, search: searchParams.toString() })
   }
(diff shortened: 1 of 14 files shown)


apps/portals/src/portal-components/ProjectDiscussionForum.tsx
```diff
@@ -0,0 +1,43 @@
+import * as React from 'react'
+import { useGetEntityForum } from 'synapse-react-client/dist/utils/hooks/SynapseAPI/entity/useGetEntityForum'
+import { DiscussionThread } from 'synapse-react-client/dist/containers/discussion_forum/DiscussionThread'
+import { ForumPage } from 'synapse-react-client/dist/containers/discussion_forum/ForumPage'
+import ArrowBackIcon from '@mui/icons-material/ArrowBack'
+import { useHistory, useLocation } from 'react-router-dom'
+import { Box, IconButton } from '@mui/material'
+
+const ProjectDiscussionForum = () => {
```
jay-hodgson, line 9: Wrap the ForumPage and DiscussionThread components from SRC, using a search parameter to determine which one should be shown.  Introduced the idea of a search parameter that does not have impact on the query filters of a DetailsPage (suffix "NotAQueryFilterKey"!).  Let me know what you think about this approach

packages/synapse-react-client/src/lib/containers/discussion_forum/ForumPage.tsx
```diff
@@ -15,7 +15,7 @@ import { ConfirmationDialog } from '../ConfirmationDialog'
 export type ForumPageProps = {
   forumId: string
   limit: number
-  onClickLink: () => void
+  onClickLink: (threadId: string) => void
```
jay-hodgson, line 18: Need to know what thread was clicked on

packages/synapse-react-client/src/lib/containers/discussion_forum/ForumTable.tsx
```diff
@@ -142,10 +143,10 @@ export const ForumTable: React.FC<ForumTableProps> = ({
             return (
               <tr key={item.id}>
                 <td>
-                  <a onClick={() => onClickLink()}>
+                  <Link onClick={() => onClickLink(item.id)}>
```
jay-hodgson, line 146: Fix link styling

apps/portals/src/types/portal-util-types.ts
```diff
@@ -44,6 +44,7 @@ export type DetailsPageContent =
 
 export type DetailsPageProps = DetailsPageContent & {
   showMenu?: boolean // default to true
+  /** Note, to exclude from processing, use the suffix "NotAQueryFilterKey" */
```
nickgros, line 47: I wonder if we could use [Template Literal Types](https://www.typescriptlang.org/docs/handbook/2/template-literal-types.html) to enforce a prefix or suffix

packages/synapse-react-client/src/lib/utils/functions/sqlFunctions.ts
```diff
@@ -35,7 +35,8 @@ export const generateQueryFilterFromSearchParams = (
   if (!searchParams) {
     return
   }
-  const isQueryWrapperKey = (key: string) => key.startsWith('QueryWrapper')
+  const isQueryWrapperKey = (key: string) =>
+    key.startsWith('QueryWrapper') || key.endsWith('NotAQueryFilterKey')
```
nickgros, line 39: Should the suffix case be negated? Or is QueryWrapper key != QueryFilter key (in which case maybe this needs a new name?)
```suggestion
  const isQueryWrapperKey = (key: string) =>
    key.startsWith('QueryWrapper') || !key.endsWith('NotAQueryFilterKey')
```

jay-hodgson: QueryWrapper key != QueryFilter key.  I think this is when the Query in a DetailsPage is changed (filtering takes place), which automatically updates the URL with the QueryWrapperX parameter.  But these should be ignored here, when the DetailsPage interprets every search param key/value pair as a query condition

jay-hodgson:
Latest changes result in the Challenge discussion forum URL in this form:
http://localhost:3000/Challenges/DetailsPage/Community?id=syn51208606&__forum_threadId=10182

nickgros:
> Latest changes result in the Challenge discussion forum URL in this form:
> http://localhost:3000/Challenges/DetailsPage/Community?id=syn51208606&__forum_threadId=10182

awesome!

nickgros (approved):
Looks great! Please fix the failing test before merging.

jay-hodgson:
Unrelated EntityFinderModal test failure, merging

PR #246, opened 2023-05-19
Stabilize a few UI checks in Chromatic

nickgros:
- Update action to reference the pull request ref so the actual changes get deployed and tested
- Load Lato font in Storybook preview-head, since it looks like UI tests commonly get flagged as changed when the font doesn't load in time for the screenshot. See https://www.chromatic.com/docs/resource-loading
- Update SubsectionRowRenderer demo to use existing table. The previous table used in the demo no longer exists, and when that happens we show a skeleton that uses a random number to determine its appearance, which is obviously not stable for snapshots.

### DIFF
--- a/.github/workflows/deploy-to-chromatic.yml
+++ b/.github/workflows/deploy-to-chromatic.yml
@@ -17,6 +17,7 @@ jobs:
     steps:
       - uses: actions/checkout@v3.1.0
         with:
+          ref: ${{ github.event.pull_request.head.ref }}
           # Chromatic requires full git history to track changes
           fetch-depth: 0
       - uses: ./.github/actions/pnpm-setup-action
--- a/packages/synapse-react-client/.storybook/preview-head.html
+++ b/packages/synapse-react-client/.storybook/preview-head.html
@@ -1,8 +1,8 @@
 <script>
   window.global = window
-  window.sql = {}
 </script>
 
+<link href="https://fonts.googleapis.com/css?family=Lato" rel="stylesheet" />
 <link
   rel="stylesheet"
   href="https://maxcdn.bootstrapcdn.com/bootstrap/3.3.7/css/bootstrap.min.css"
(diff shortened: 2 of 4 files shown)


packages/synapse-react-client/.storybook/preview-head.html
```diff
@@ -1,8 +1,8 @@
 <script>
   window.global = window
-  window.sql = {}
```
nickgros, line 3: no longer needed

packages/synapse-react-client/.storybook/preview-head.html
```diff
@@ -51,18 +51,3 @@
...
-<script>
-  const currentStack = JSON.parse(
-    window.localStorage.getItem('SynapseReactClientStyleGuideStack') ?? 'null',
-  )
-  if (currentStack) {
-    window.SRC = {
-      OVERRIDE_ENDPOINT_CONFIG: currentStack,
-    }
-  }
-</script>
```
nickgros, line 68: This was replaced by Storybook controls long ago

.github/workflows/deploy-to-chromatic.yml
```diff
@@ -17,6 +17,7 @@ jobs:
     steps:
       - uses: actions/checkout@v3.1.0
         with:
+          ref: ${{ github.event.pull_request.head.ref }}
```
nickgros, line 20: Deploy the pull request, not the main branch

packages/synapse-react-client/.storybook/preview.jsx
```diff
@@ -113,8 +114,16 @@ initialize({
...
 
+const fontLoader = async () => ({
+  fonts: await Promise.all([document.fonts.load('700 1em Lato')]),
+})
+
 export const loaders = [mswLoader]
 
+if (isChromatic && document.fonts) {
+  loaders.push(fontLoader)
+}
```
nickgros, line 125: When deployed to Chromatic, wait for the font to load before rendering the story. This should make our screenshots more stable

packages/synapse-react-client/stories/SubsectionRowRenderer.stories.ts
```diff
@@ -9,9 +9,12 @@ export default meta
 type Story = StoryObj<typeof meta>
 
 export const Demo: Story = {
+  name: 'SubsectionRowRenderer',
   args: {
-    sql: 'SELECT funderName as "Funding Agency" FROM syn26344829',
```
nickgros, line 13: This table has been deleted and the fallback is a skeleton with a random length. Update the story with a different table.

jay-hodgson (approved):

PR #250, opened 2023-05-19
PORTALS-2660, EC-120: only show the Actions Required if the user is already logged in

jay-hodgson:


jay-hodgson:
And add the virtually free implementation of a beta banner that we get on the portals.  We can spend more time if needed (perhaps we need a special "aboveHomePageHeader" area for this kind of alert?)

nickgros (approved):

PR #251, opened 2023-05-20
EC-122

jay-hodgson:
Moving a few items around in the ELITE portal config, and a few design-led changes to color and background.

Query plot nav header (gray instead of portal primary color based color, with slight change in size of panel):
Screen Shot 2023-05-19 at 5 57 12 PM

Home page alternating background colors, now based on the portal primary color instead of hard-coded to a light grey.
Screen Shot 2023-05-19 at 5 56 35 PM

Screen Shot 2023-05-19 at 5 56 05 PM

### DIFF
--- a/apps/portals/src/_Core.scss
+++ b/apps/portals/src/_Core.scss
@@ -23,7 +23,7 @@
 }
 
 %home-bg-util {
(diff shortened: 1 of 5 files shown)


hallieswan (approved):
Looks good to me!

PR #270, opened 2023-05-30
EC-119: Upset plot on the ELITE portal home page

jay-hodgson:
is not pretty (with the current data) and it seems to take ~12 seconds to load?


jay-hodgson:
Screen Shot 2023-05-30 at 1 40 02 PM

PR #273, opened 2023-06-01
PORTALS-2674: support cross-page multiple row selection with custom c…

jay-hodgson:
…ommands, and add a demo in the Storybook

### DIFF
--- a/apps/portals/src/configurations/crc-researcher/handleParticipantWorkflowChange.tsx
+++ b/apps/portals/src/configurations/crc-researcher/handleParticipantWorkflowChange.tsx
@@ -7,7 +7,7 @@ import { SynapseClient } from 'synapse-react-client'
 
 type CustomControlCallbackData = {
   data: QueryResultBundle | undefined
-  selectedRowIndices: number[] | undefined
+  selectedRows: Row[] | undefined
   refresh: () => void
 }
 
@@ -24,10 +24,9 @@ const handleParticipantWorkflowChange = async (
   )
   // collect all selected rows (create PartialRow objects)
   const rowUpdates: PartialRow[] = []
-  const rows: Row[] = event.data?.queryResult!.queryResults!.rows!
-  for (let index = 0; index < event.selectedRowIndices!.length; index++) {
+  for (let index = 0; index < event.selectedRows!.length; index++) {
     rowUpdates.push({
-      rowId: rows[event.selectedRowIndices![index]].rowId!,
+      rowId: event.selectedRows![index].rowId!,
       values: [
         {
           key: targetColumn?.id!,
--- a/packages/synapse-react-client/src/components/QueryVisualizationWrapper.tsx
+++ b/packages/synapse-react-client/src/components/QueryVisualizationWrapper.tsx
@@ -12,7 +12,7 @@ import { NoContentPlaceholderType } from './SynapseTable/NoContentPlaceholderTyp
 import SearchResultsNotFound from './SynapseTable/SearchResultsNotFound'
 import ThisTableIsEmpty from './SynapseTable/TableIsEmpty'
 import { unCamelCase } from '../utils/functions/unCamelCase'
-import { ColumnType } from '@sage-bionetworks/synapse-types'
+import { ColumnType, Row } from '@sage-bionetworks/synapse-types'
 import { getDisplayValue } from '../utils/functions/getDataFromFromStorage'
 
 export type QueryVisualizationContextType = {
@@ -22,8 +22,8 @@ export type QueryVisualizationContextType = {
   >
   columnsToShowInTable: string[]
   setColumnsToShowInTable: (newState: string[]) => void
-  selectedRowIndices: number[]
-  setSelectedRowIndices: (newState: number[]) => void
+  selectedRows: Row[]
+  setSelectedRows: (newState: Row[]) => void
   rgbIndex?: number
   unitDescription?: string
   /** Whether to show when the table or view was last updated. */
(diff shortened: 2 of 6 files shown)


jay-hodgson:
Screen Shot 2023-06-01 at 3 50 47 PM

apps/portals/src/configurations/crc-researcher/handleParticipantWorkflowChange.tsx
```diff
@@ -7,7 +7,7 @@ import { SynapseClient } from 'synapse-react-client'
 
 type CustomControlCallbackData = {
   data: QueryResultBundle | undefined
-  selectedRowIndices: number[] | undefined
+  selectedRows: Row[] | undefined
```
jay-hodgson, line 10: Keep track of Rows rather than the row indices of the current page.

packages/synapse-react-client/src/components/SynapseTable/RowSelectionControls.tsx
```diff
@@ -0,0 +1,71 @@
...
+      <Box sx={{ display: 'grid', gridTemplateColumns: '30px 140px' }}>
+        <Badge
+          badgeContent={selectedRows.length}
+          color="primary"
+          sx={{ right: '18px', top: '12px' }}
+        />{' '}
+        <Typography variant="body1">Rows Selected</Typography>
+      </Box>
+      <div>
+        {customControls.map(customControl => {
```
jay-hodgson, line 53: Block moved from TopLevelControls, and updated to be MUI buttons

packages/synapse-react-client/src/components/SynapseTable/RowSelectionControls.tsx
```diff
@@ -0,0 +1,71 @@
...
+
+export type RowSelectionControlsProps = {
+  customControls: CustomControl[]
+  data?: QueryResultBundle
+  selectedRows: Row[]
+  setSelectedRows: (newState: Row[]) => void
+  refresh: () => void
+}
+
+export const RowSelectionControls: React.FunctionComponent<
```
jay-hodgson, line 14: New UI that is shown when there is non-empty a row selection (and custom commands)

packages/synapse-react-client/src/components/SynapseTable/TopLevelControls.tsx
```diff
@@ -126,6 +128,9 @@ const TopLevelControls = (props: TopLevelControlsProps) => {
   }
 
   const refresh = () => {
+    // clear selection
+    setSelectedRows([])
+    // refresh the data
     executeQueryRequest(getLastQueryRequest())
```
jay-hodgson, line 134: This does not seem execute the query (on CRC researcher).  This is an existing bug, so I'm ignoring for now, but may want to track down what is happening here (do we need to invalidate the queries for react-query before this?)

jay-hodgson: I don't think we'll need to fix this for the Cohort Builder either, since it will switch views rather that refresh the current page data.

packages/synapse-react-client/src/components/SynapseTable/TopLevelControls.tsx
```diff
@@ -210,25 +215,6 @@ const TopLevelControls = (props: TopLevelControlsProps) => {
           )}
         </div>
         <div className="TopLevelControls__actions">
-          {customControls &&
```
jay-hodgson, line 213: Moved

packages/synapse-react-client/src/components/QueryVisualizationWrapper.tsx
```diff
@@ -139,15 +134,8 @@ export function QueryVisualizationWrapper(
       }))
     }
   }, [isFacetsAvailable])
-
-  useEffect(() => {
-    if (isLoadingNewBundle) {
-      setSelectedRowIndices([])
-    }
-  }, [isLoadingNewBundle])
```
nickgros, line 147: Location of this state is going to be interesting for cohort builder. Might need to move it elsewhere (maybe in a new context provider that wraps QueryVisualizationWrapper?), but that can be figured out later.

packages/synapse-react-client/src/components/SynapseTable/RowSelectionControls.tsx
```diff
@@ -0,0 +1,71 @@
...
+          return (
+            <Button
+              key={customControl.buttonText}
+              variant="contained"
+              style={{ marginRight: '10px' }}
+              onClick={() =>
+                customControl.onClick({ data, selectedRows, refresh })
+              }
+            >
+              {customControl.icon}&nbsp;
```
nickgros, line 63: Use `startIcon` prop: https://mui.com/material-ui/api/button/#props
```suggestion
              startIcon={customControl.icon}
            >
```

packages/synapse-react-client/src/components/SynapseTable/TopLevelControls.tsx
nickgros: I'm not that familiar with this, but I would guess that you're right about it being a query invalidation issue. Is this invoked after mutating the table? The re-fetch would probably just result in a cache hit if the update call doesn't invalidate the cache (which it probably doesn't, since this hasn't been closely looked at since before we started using react-query).

Can you open a Jira ticket with some repro steps?

nickgros (approved):
Looks good! Just one suggestion that doesn't block approval

packages/synapse-react-client/src/components/SynapseTable/TopLevelControls.tsx
jay-hodgson: Opened https://sagebionetworks.jira.com/browse/PORTALS-2682

PR #277, opened 2023-06-02
PORTALS-2435: RJSF v5 + MUI restyling

nickgros:
- Upgrade RJSF to v5
- Drop custom RJSF fork
- Include RJSF in UMD bundle

SchemaDrivenAnnotationEditor:
- Add basic support for derived annotations
- Use @rjsf/mui
- Replace Bootstrap components with MUI equivalents
- Refactor to match changes in v5
- Replace SelectWidget react-select implementation with MUI Autocomplete


packages/synapse-react-client/src/components/Evaluation/CalendarWithIconFormGroup.tsx
```diff
@@ -46,7 +46,13 @@ export const CalendarWithIconFormGroup: React.FunctionComponent<
...
             )}
           />
-          <span style={{ margin: 'auto 8px' }}>
+          <span
+            style={{
+              display: 'table-cell',
+              verticalAlign: 'middle',
+              paddingLeft: '8px',
+            }}
+          >
```
nickgros, line 55: Fix alignment

packages/synapse-react-client/src/components/QuerySortSelector.tsx
```diff
@@ -3,18 +3,30 @@ import { SortConfiguration } from './CardContainerLogic'
...
+export const Control = ({
+  children,
+  ...rest
+}: ControlProps<any, boolean, GroupBase<any>>) => {
+  return (
+    <components.Control {...rest} className="form-control">
+      {children}
+    </components.Control>
+  )
+}
```
nickgros, line 28: Moved from `CustomSelectWidget`, since these are now only used in this file.

packages/synapse-react-client/src/components/SchemaDrivenAnnotationEditor/SchemaDrivenAnnotationEditor.tsx
```diff
@@ -215,16 +227,33 @@ export const SchemaDrivenAnnotationEditor = (
...
             className="AnnotationEditorForm"
             liveValidate={liveValidate}
             noHtml5Validate={true}
-            ArrayFieldTemplate={CustomArrayFieldTemplate}
-            ObjectFieldTemplate={CustomObjectFieldTemplate}
-            FieldTemplate={CustomDefaultTemplate}
-            ref={formRefFromParent ?? formRef}
+            experimental_defaultFormStateBehavior={{
+              emptyObjectFields: 'skipDefaults',
+            }}
```
nickgros, line 236: This snippet ensures that `default`/`const` fields (which yield "derived annotations") are not pre-filled in the form, so they are not automatically hard-coded and override the derived value.

packages/synapse-react-client/src/components/SchemaDrivenAnnotationEditor/SchemaDrivenAnnotationEditor.tsx
```diff
@@ -237,64 +266,72 @@ export const SchemaDrivenAnnotationEditor = (
               } as JSONSchema7
             }
             uiSchema={{
-              'ui:DuplicateKeySuffixSeparator': '_',
+              'ui:options': {
+                copyable: true,
```
nickgros, line 270: New feature in RJSF v5.x - array items are copyable. Might as well surface this to our users!

packages/synapse-react-client/src/components/SchemaDrivenAnnotationEditor/SchemaDrivenAnnotationEditor.tsx
```diff
@@ -305,8 +342,8 @@ export const SchemaDrivenAnnotationEditor = (
                     onCancel={() => {
                       onCancel && onCancel()
                     }}
-                    onConfirm={() => {
-                      formRef.current!.submit()
+                    onConfirm={e => {
+                      ref.current!.formElement.current!.requestSubmit()
```
nickgros, line 346: fixes bug in Firefox where `preventDefault` would not work.

packages/synapse-react-client/src/theme/DefaultTheme.tsx
```diff
@@ -176,18 +176,19 @@ export const defaultMuiThemeOptions: ThemeOptions = {
...
-        root: ({ theme }) => ({
+        root: ({ theme, ownerState }) => ({
           borderRadius: '3px',
           fontSize: '14px',
           position: 'relative',
           backgroundColor: theme.palette.grey[200],
-          border: 'none',
+          border: ownerState.error
+            ? `1px solid ${theme.palette.error.main}`
+            : 'none',
```
nickgros, line 186: Apply border on error

packages/synapse-react-client/src/theme/DefaultTheme.tsx
```diff
@@ -176,18 +176,19 @@ export const defaultMuiThemeOptions: ThemeOptions = {
...
           '&.Mui-focused': {
             boxShadow: `${alpha(
               theme.palette.primary.main,
               0.25,
             )} 0 0 0 0.1rem`,
-            borderColor: theme.palette.primary.main,
```
nickgros, line 190: Box shadow is probably enough, applying a border is causing a layout shift.

packages/synapse-react-client/test/containers/entity/annotations/AdditionalPropertiesSchemaField.test.tsx
```diff
@@ -91,29 +104,15 @@ describe('AdditionalPropertiesSchemaField unit tests', () => {
...
-
-    it('gets CheckboxWidget for boolean', () => {
-      expect(getWidgetFromPropertyType(PropertyType.BOOLEAN)).toBe(
-        'CheckboxWidget',
-      )
-    })
-
-    it('gets TextWidget for string', () => {
-      expect(getWidgetFromPropertyType(PropertyType.STRING)).toBe('TextWidget')
+  describe('getSchemaForPropertyType tests', () => {
```
nickgros, line 107: Instead of mapping custom annotation type to Widget, now we custom annotation type to schema, and let RJSF pick the widget based on the mapping we have already set up for regular schema properties.

packages/synapse-react-client/test/containers/entity/annotations/SchemaDrivenAnnotationsEditor.integration.test.tsx
```diff
@@ -105,19 +129,42 @@ describe.skip('SchemaDrivenAnnotationEditor tests', () => {
     },
   )
 
-  const annotationsHandler = rest.get(
+  // Returns an entity with annotations that match the schema
+  const annotationsWithSchemaHandler = rest.get(
```
nickgros, line 133: Following [PLFM-6811](https://sagebionetworks.jira.com/browse/PLFM-6811), the annotation returned by Synapse may change depending on the presence of a schema.

For example, all "custom annotations" are arrays, but if a single annotation is described by a bound schema as a string, Synapse will return a string.

For this reason, this handler needed to be split into two different handlers.

packages/synapse-react-client/src/components/SchemaDrivenAnnotationEditor/field/AdditionalPropertiesSchemaField.tsx
nickgros, line 1: Changed enough from old version that the diff thinks it's new, but I'll call out changes with comments

packages/synapse-react-client/src/components/SchemaDrivenAnnotationEditor/field/AdditionalPropertiesSchemaField.tsx
```diff
@@ -0,0 +1,282 @@
...
+    list.every(item => typeof item === 'string') &&
+    (list as string[]).every((item: string) => !!ISO_TIMESTAMP_REGEX.exec(item))
+  ) {
+    return PropertyType.DATETIME
+  }
+  // otherwise, default type is 'string'
+  return PropertyType.STRING
+}
+
+export function transformDataFromPropertyType(
```
nickgros, line 52: Changed fallback/null cases from `undefined` to 0 (for integer), current date (for datetime) to reduce the likelihood of an empty property being dropped while edited. This wouldn't result in lost data but could be frustrating.

packages/synapse-react-client/src/components/SchemaDrivenAnnotationEditor/field/AdditionalPropertiesSchemaField.tsx
```diff
@@ -0,0 +1,282 @@
...
+    default:
+      return list.map(item => String(item))
+  }
+}
+
+/**
+ * Maps a Synapse Annotation PropertyType to a JSON Schema that captures the type and format.
+ * @param propertyType
+ */
+export function getSchemaForPropertyType(propertyType: PropertyType) {
```
nickgros, line 93: Previously mapped `PropertyType` to a `Widget`. Now, map `PropertyType` to a schema and RJSF will render the correct widget, since it's already doing that for the rest of the form.

packages/synapse-react-client/src/components/SchemaDrivenAnnotationEditor/field/AdditionalPropertiesSchemaField.tsx
```diff
@@ -0,0 +1,282 @@
...
+
+  const itemsSchema = getSchemaForPropertyType(propertyType)
+
+  // If the additional property has not yet been coerced to an array, trigger an update to coerce and render nothing.
+  // We must do this because RJSF does not currently batch these kinds of updates
+  // See https://github.com/rjsf-team/react-jsonschema-form/issues/3367
+  if (!Array.isArray(formData)) {
+    onChange(convertToArray(formData) as unknown as T)
+    return <></>
+  }
```
nickgros, line 213: Instead of setting a timeout, it seems to suffice to `onChange` each render, and display nothing until we get the right format.

I also added a link to the relevant issue that would need to be fixed so we can get this to work without this kind of workaround.

packages/synapse-react-client/src/components/SchemaDrivenAnnotationEditor/field/AdditionalPropertiesSchemaField.tsx
```diff
@@ -0,0 +1,282 @@
...
+
+  return (
+    <AdditionalPropertyContextProvider
+      value={{
+        dropProperty: e => {
+          onDropPropertyClick(name)(e)
+        },
+      }}
+    >
```
nickgros, line 222: Added a custom context to pass this callback around, since changes in RJSF have made it so this callback is not accessible from where we need to invoke it.

packages/synapse-react-client/src/components/SchemaDrivenAnnotationEditor/field/AdditionalPropertiesSchemaField.tsx
```diff
@@ -0,0 +1,282 @@
...
+      <Grid item xs={2}>
+        <InputLabel htmlFor={`${id}-type`}>Type</InputLabel>
+        <SelectWidget
```
nickgros, line 225: Use the `SelectWidget` for the "Type" field, so it aesthetically matches the rest of the form

packages/synapse-react-client/esbuild.config.mjs
```diff
@@ -15,7 +15,6 @@ const globals = {
   'rss-parser': 'Parser',
   'react-mailchimp-subscribe': 'ReactMailchimpSubscribe',
   'plotly.js-basic-dist': 'Plotly',
-  '@rjsf/core': 'JSONSchemaForm',
```
nickgros, line 18: Let's include RJSF in the UMD bundle. Our fork was included before removing it in this PR, so bundle size should not change much.

packages/synapse-react-client/src/components/SchemaDrivenAnnotationEditor/SchemaDrivenAnnotationEditor.tsx
```diff
@@ -215,16 +229,33 @@ export const SchemaDrivenAnnotationEditor = (
...
+              BaseInputTemplate: BaseInputTemplate,
+              FieldErrorTemplate: FieldErrorTemplate,
+              FieldTemplate: FieldTemplate,
+              ObjectFieldTemplate: ObjectFieldTemplate,
+              WrapIfAdditionalTemplate: WrapIfAdditionalTemplate,
+              ButtonTemplates: ButtonTemplate,
+              DescriptionFieldTemplate: DescriptionFieldTemplate,
+              /* Errors are displayed by an Alert component below, so we don't show the builtin ErrorList */
+              ErrorListTemplate: () => null,
+            }}
```
nickgros, line 256: RJSF v5 added many new templates to simplify customization.

packages/synapse-react-client/src/components/SchemaDrivenAnnotationEditor/template/ArrayFieldItemTemplate.tsx
nickgros, line 1: This is a new template in v5

packages/synapse-react-client/src/components/SchemaDrivenAnnotationEditor/template/ArrayFieldItemTemplate.tsx
```diff
@@ -0,0 +1,104 @@
...
+            />
+          )}
+          {(hasMoveUp || hasMoveDown) && (
+            <MoveDownButton
+              disabled={disabled || readonly || !hasMoveDown}
+              onClick={onReorderClick(index, index + 1)}
+              uiSchema={uiSchema}
+              registry={registry}
+            />
+          )}
```
nickgros, line 73: We weren't rendering the up/down buttons for array items before, but I figure we should add them anyways, because they're very useful.

packages/synapse-react-client/src/components/SchemaDrivenAnnotationEditor/template/ArrayFieldTemplate.tsx
```diff
@@ -0,0 +1,126 @@
...
+                  isAdditionalProperty &&
+                  items.length == 1 &&
+                  additionalPropertyContext
+                    ? i => {
+                        return e => {
+                          additionalPropertyContext.dropProperty(e)
+                        }
+                      }
+                    : itemProps.onDropIndexClick
+                }
```
nickgros, line 107: Some complexity here where we have to drop the entire field if this is an AdditionalProperty.

packages/synapse-react-client/src/components/SchemaDrivenAnnotationEditor/template/BaseInputTemplate.tsx
```diff
@@ -0,0 +1,123 @@
...
+  const _onBlur = ({ target: { value } }: FocusEvent<HTMLInputElement>) =>
+    onBlur(id, value)
+  const _onFocus = ({ target: { value } }: FocusEvent<HTMLInputElement>) =>
+    onFocus(id, value)
+  const DisplayInputLabelProps = TYPES_THAT_SHRINK_LABEL.includes(type)
+    ? {
+        ...InputLabelProps,
+        shrink: true,
+      }
+    : InputLabelProps
```
nickgros, line 77: This portion is all copied from `@rjsf/mui`

packages/synapse-react-client/src/components/SchemaDrivenAnnotationEditor/template/BaseInputTemplate.tsx
```diff
@@ -0,0 +1,123 @@
...
+
+  let derivedValuePlaceholder: string | undefined = undefined
+  if (schema.default || schema.const) {
+    derivedValuePlaceholder =
+      JSON.stringify(schema.default || schema.const) + ' (derived)'
+  }
```
nickgros, line 83: For derived annotations, render the `default`/`const` value as a placeholder

packages/synapse-react-client/src/components/SchemaDrivenAnnotationEditor/template/BaseInputTemplate.tsx
```diff
@@ -0,0 +1,123 @@
...
+    options,
+    schema,
+    uiSchema,
+    rawErrors = [],
+    formContext,
+    registry,
+    InputLabelProps,
+    ...textFieldProps
+  } = props
+  const hideLabel = true
```
nickgros, line 53: Always hide the label, since it's rendered at the field level.

packages/synapse-react-client/src/components/SchemaDrivenAnnotationEditor/template/FieldDescriptionTable.tsx
```diff
@@ -0,0 +1,60 @@
...
+  }
+
+  if (Array.isArray(type)) {
+    return type.reduce(
+      (typesCommaSeparated, nextType) => `${typesCommaSeparated}, ${nextType}`,
+      '',
+    )
+  }
+  return type || 'unknown'
+}
```
nickgros, line 33: Slightly improved this to handle more cases

packages/synapse-react-client/src/components/SchemaDrivenAnnotationEditor/template/FieldDescriptionTable.tsx
```diff
@@ -0,0 +1,60 @@
...
+            <tr>
+              <th>Description</th>
+              <td>{description}</td>
+            </tr>
+          )}
+          <tr>
+            <th>Type</th>
+            <td>{typeString}</td>
+          </tr>
+        </tbody>
```
nickgros, line 56: No longer render a row for `required` because
1. I don't think we were accurately determining this value, since `required` is captured in the parent object in JSONSchema
2. The form already tells you if it's required with the `*`

packages/synapse-react-client/src/components/SchemaDrivenAnnotationEditor/template/FieldErrorTemplate.tsx
```diff
@@ -0,0 +1,41 @@
...
+    return null
+  }
+  const id = errorId<T>(idSchema)
+
+  return (
+    <List dense={true} disablePadding={true}>
+      {errors.map((error, i: number) => {
+        return (
+          <ListItem key={i} disableGutters={true}>
+            <FormHelperText id={id} sx={{ color: 'error.main' }}>
```
nickgros, line 33: For some reason we were not getting the right color, so I added it here.

packages/synapse-react-client/src/components/SchemaDrivenAnnotationEditor/template/FieldTemplate.tsx
```diff
@@ -0,0 +1,83 @@
...
+  getTemplate,
+  getUiOptions,
+  RJSFSchema,
+  StrictRJSFSchema,
+} from '@rjsf/utils'
+import { HelpOutline } from '@mui/icons-material'
+import { Collapse, Grid, IconButton, InputLabel, Tooltip } from '@mui/material'
+
+// TODO: Wrap and pull out class names to remove bootstrap classes
+//  https://rjsf-team.github.io/react-jsonschema-form/docs/advanced-customization/custom-templates/#baseinputtemplate
```
nickgros, line 14: Leaving this as a future task, for now.

packages/synapse-react-client/src/components/SchemaDrivenAnnotationEditor/template/ObjectFieldTemplate.tsx
nickgros, line 1: Only minor changes called out

packages/synapse-react-client/src/components/SchemaDrivenAnnotationEditor/template/ObjectFieldTemplate.tsx
```diff
@@ -0,0 +1,162 @@
...
+    'DescriptionFieldTemplate',
+    T,
+    S,
+    F
+  >('DescriptionFieldTemplate', registry, options)
+
+  /*
+     Tracks the property keys that were last known to exist in the object.
+     */
+  const previousSchemaDefinedProperties = useRef<Set<string>>(new Set())
```
nickgros, line 64: Track this in a ref instead of useState to prevent multiple checks from occurring in strict mode, causing the toast to be displayed multiple times.

packages/synapse-react-client/src/components/SchemaDrivenAnnotationEditor/template/ObjectFieldTemplate.tsx
```diff
@@ -0,0 +1,162 @@
...
+            // Form data exists
+            !!formData &&
+            // Property still exists in form data and is not null
+            // if the data is null, then we don't need to worry about it; user data wasn't lost
+            formData[schemaDefinedProperty] != null &&
+            // Property matches `additionalProperties` requirements: is an array and contains non-null values
+            Array.isArray(formData[schemaDefinedProperty]) &&
+            (formData[schemaDefinedProperty] as Array<any>).filter(
+              item => item != null,
+            ).length > 0
```
nickgros, line 103: Just added these conditions to prevent notifications on cases where the data is empty and will be dropped.

packages/synapse-react-client/src/components/SchemaDrivenAnnotationEditor/widget/BooleanWidget.tsx
```diff
@@ -0,0 +1,17 @@
...
+export const BooleanWidget: Widget = (props: WidgetProps) => {
+  const { SelectWidget } = props.registry.widgets
+
+  const options = {
+    enumOptions: [
+      { value: true, label: 'true' },
+      { value: false, label: 'false' },
+    ],
+  }
+  return <SelectWidget {...props} options={options} />
```
nickgros, line 16: Use the `SelectWidget` for uniformity

packages/synapse-react-client/src/components/SchemaDrivenAnnotationEditor/widget/SelectWidget.tsx
nickgros, line 1: Completely re-implemented using MUI `Autocomplete` instead of `react-select`

nickgros, line 1: Completely re-implemented using MUI Autocomplete instead of react-select

packages/synapse-react-client/src/components/QuerySortSelector.tsx
jay-hodgson: We still want to use the bootstrap form-control css class here?

packages/synapse-react-client/src/components/SchemaDrivenAnnotationEditor/SchemaDrivenAnnotationEditor.tsx
jay-hodgson: That makes sense, we don't want to automatically override the derived value on edit.  Does the user see the derived value in the form somehow (does it set the input placeholder value for example)?

packages/synapse-react-client/src/components/ModalDownload/ModalDownload.tsx
```diff
@@ -24,7 +25,7 @@ type ModalDownloadState = {
   isLoading: boolean
   step: number
   data?: DownloadFromTableResult
-  formData: any
+  formData: Record<string, unknown>
```
jay-hodgson, line 28: Love all of the type improvements in this PR

packages/synapse-react-client/src/components/SchemaDrivenAnnotationEditor/template/ArrayFieldDescriptionTemplate.tsx
```diff
@@ -0,0 +1,46 @@
...
+  ArrayFieldDescriptionProps,
+  FormContextType,
+  RJSFSchema,
+  StrictRJSFSchema,
+} from '@rjsf/utils'
+
+/** The `ArrayFieldDescriptionTemplate` component renders a `DescriptionFieldTemplate` with an `id` derived from
+ * the `idSchema`.
+ *
+ * Identical to @rjsf/core ArrayFieldDescriptionTemplate except this doesn't render nothing if there is no description,
```
jay-hodgson, line 14: "except this doesn't render nothing if there is no description" -> "except this renders something if there is no description"?

packages/synapse-react-client/src/components/SchemaDrivenAnnotationEditor/template/BaseInputTemplate.tsx
jay-hodgson: Oh, that answers my question from earlier!  Thanks!

packages/synapse-react-client/src/components/SchemaDrivenAnnotationEditor/template/DescriptionFieldTemplate.tsx
```diff
@@ -0,0 +1,18 @@
...
+import React from 'react'
+import { Typography } from '@mui/material'
+import FieldDescriptionTable from './FieldDescriptionTable'
+
+export default function DescriptionFieldTemplate(props: DescriptionFieldProps) {
+  const { description, schema = {} } = props
+
+  if (schema.type === 'object') {
+    return (
+      <Typography variant={'body1'} sx={{ my: 2 }}>
```
jay-hodgson, line 11: What does this my:2 (margin-top/margin-bottom) result in today?  (theme spacing x2?)

packages/synapse-react-client/src/components/SchemaDrivenAnnotationEditor/template/ObjectFieldTemplate.tsx
jay-hodgson: You should call out major changes, too  ;)

packages/synapse-react-client/src/components/SchemaDrivenAnnotationEditor/widget/SelectWidget.tsx
```diff
@@ -0,0 +1,112 @@
...
+      onChange={(event, newValue) => {
+        if (isObject(newValue) && 'inputValue' in newValue) {
+          // Create a new value from the user input
+          onChange(newValue.inputValue)
+        } else if (isObject(newValue) && 'value' in newValue) {
+          onChange(newValue.value)
+        } else if (typeof newValue === 'string') {
+          onChange(newValue)
+        } else {
+          onChange(newValue)
```
jay-hodgson, line 81: Does `onChange` handle other cases gracefully?  Are there cases when `newValue` is an object but doesn't have an `inputValue` or `value` property?

packages/synapse-react-client/stories/stubs/TextField/MuiTextField.stories.ts
```diff
@@ -0,0 +1,19 @@
...
+  component: TextField,
+} satisfies Meta
+export default meta
+
+type Story = StoryObj<typeof meta>
+
+export const Demo: Story = {
+  name: 'Text Field',
+  args: {
+    label: 'My Button Text Here',
```
jay-hodgson, line 15: Button Text?

jay-hodgson (approved):

packages/synapse-react-client/src/components/QuerySortSelector.tsx
nickgros: Not sure. Left out of scope for this PR, since this isn't part of the annotations editor, it just borrowed a component from the previous implementation.

packages/synapse-react-client/src/components/SchemaDrivenAnnotationEditor/template/DescriptionFieldTemplate.tsx
nickgros: Yep, since our spacing is 8px this is equivalent to `margin-top: 16px; margin-bottom: 16px;`

packages/synapse-react-client/src/components/SchemaDrivenAnnotationEditor/widget/SelectWidget.tsx
nickgros: Not totally sure if this is possible, MUI seems to handle fallback like this in their [Creatable example](https://mui.com/material-ui/react-autocomplete/#creatable)

packages/synapse-react-client/stories/stubs/TextField/MuiTextField.stories.ts
nickgros: Copied n pasted, will fix

PR #278, opened 2023-06-02
PORTALS-2675: Ability to pass QueryFilters via localStorage that will…

jay-hodgson:
… be picked up by all of our Synapse Table/View components (in the same way that Search parameters are supported)

### DIFF
--- a/packages/synapse-react-client/src/components/CardContainerLogic/CardContainerLogic.tsx
+++ b/packages/synapse-react-client/src/components/CardContainerLogic/CardContainerLogic.tsx
@@ -146,6 +146,7 @@ export type CardContainerLogicProps = {
 export const CardContainerLogic = (props: CardContainerLogicProps) => {
   const entityId = parseEntityIdFromSqlStatement(props.sql)
   const queryFilters = getAdditionalFilters(
+    entityId,
     props.searchParams,
     props.sqlOperator,
   )
--- a/packages/synapse-react-client/src/components/CardContainerLogic/CardContainerLogic.tsx
+++ b/packages/synapse-react-client/src/components/CardContainerLogic/CardContainerLogic.tsx
@@ -1,7 +1,7 @@
 import React from 'react'
 import { SynapseConstants } from '../../utils'
 import {
-  generateQueryFilterFromSearchParams,
+  getAdditionalFilters,
   parseEntityIdFromSqlStatement,
   SQLOperator,
 } from '../../utils/functions/SqlFunctions'
@@ -145,7 +145,7 @@ export type CardContainerLogicProps = {
  */
 export const CardContainerLogic = (props: CardContainerLogicProps) => {
   const entityId = parseEntityIdFromSqlStatement(props.sql)
-  const queryFilters = generateQueryFilterFromSearchParams(
+  const queryFilters = getAdditionalFilters(
     props.searchParams,
     props.sqlOperator,
   )
--- a/packages/synapse-react-client/src/components/StandaloneQueryWrapper/StandaloneQueryWrapper.tsx
+++ b/packages/synapse-react-client/src/components/StandaloneQueryWrapper/StandaloneQueryWrapper.tsx
@@ -1,6 +1,6 @@
 import React from 'react'
 import {
-  generateQueryFilterFromSearchParams,
+  getAdditionalFilters,
   parseEntityIdFromSqlStatement,
 } from '../../utils/functions/SqlFunctions'
 import SynapseTable, { SynapseTableProps } from '../SynapseTable/SynapseTable'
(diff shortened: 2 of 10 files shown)


jay-hodgson:
Also updated the example Storybook for table row selection.  The custom command gathers the row ID values, creates a QueryFilter[], and stores it in local storage.  If you refresh the screen after executing this command, you'll see the QueryFilter[] has been applied (as additionalFilters in the Query).
Screen Shot 2023-06-02 at 4 10 51 PM
Screen Shot 2023-06-02 at 4 11 13 PM

packages/synapse-react-client/src/components/CardContainerLogic/CardContainerLogic.tsx
```diff
@@ -1,7 +1,7 @@
 import React from 'react'
 import { SynapseConstants } from '../../utils'
 import {
-  generateQueryFilterFromSearchParams,
+  getAdditionalFilters,
```
jay-hodgson, line 4: Renamed generateQueryFilterFromSearchParams() to getAdditionalFilters() since it derives the additionalFilters field value based on what it finds in local storage as well as the given search parameters.

packages/synapse-react-client/src/components/StandaloneQueryWrapper/StandaloneQueryWrapper.tsx
```diff
@@ -105,10 +105,8 @@ const StandaloneQueryWrapper: React.FunctionComponent<
 
   const derivedQueryRequestFromSearchParams = generateInitQueryRequest(sql)
 
-  if (searchParams) {
```
jay-hodgson, line 108: Removed check of searchParams, since this is done by getAdditionalFilters().  getAdditionalFilters() will return undefined if there are no search params and nothing found in local storage.

packages/synapse-react-client/src/utils/functions/SqlFunctions.ts
```diff
@@ -27,93 +28,112 @@ export const getIgnoredQueryFilterSearchParamKey = (
...
-          }
-          return filter
-        }
-      }
-    })
+        }),
+    )
+  }
+
+  return additionalFilters.length === 0 ? undefined : additionalFilters
```
jay-hodgson, line 136: Return undefined if it is an empty array

packages/synapse-react-client/src/utils/functions/SqlFunctions.ts
```diff
@@ -27,93 +28,112 @@ export const getIgnoredQueryFilterSearchParamKey = (
...
+  let additionalFilters: QueryFilter[] = []
+  if (localStorageQueryFiltersString) {
+    additionalFilters = JSON.parse(
+      localStorageQueryFiltersString,
+    ) as QueryFilter[]
+    // delay clearing out the value so re-rendering during React Strict mode does not destroy the filter
+    setTimeout(
+      () => localStorage.removeItem(QUERY_FILTERS_LOCAL_STORAGE_KEY),
+      1000,
+    )
```
nickgros, line 55: Why clear it out after loading? Can't we just keep it synced with selections and never clear it (unless the user clicks a 'clear button')?

packages/synapse-react-client/src/utils/functions/SqlFunctions.ts
```diff
@@ -27,93 +28,112 @@ export const getIgnoredQueryFilterSearchParamKey = (
...
   }
+  if (searchParams) {
+    const isQueryWrapperKey = (key: string) =>
+      key.startsWith('QueryWrapper') || key.startsWith('__')
+    const searchParamKeys = Object.keys(searchParams)
+    if (
+      searchParamKeys.length === 0 ||
+      searchParamKeys.every(isQueryWrapperKey)
+    ) {
+      return
```
nickgros, line 65: Bug here where filters from localStorage will not be returned if we have a searchParams object with no query filter keys.

You could return `additionalFilters.length === 0 ? undefined : additionalFilters` here, but see next comment for suggested fix

packages/synapse-react-client/src/utils/functions/SqlFunctions.ts
```diff
@@ -27,93 +28,112 @@ export const getIgnoredQueryFilterSearchParamKey = (
...
-          // parse this out. This will cause a bug if something matches the synId regex but is in free text.
-          const filter: ColumnSingleValueQueryFilter = {
-            concreteType:
-              'org.sagebionetworks.repo.model.table.ColumnSingleValueQueryFilter',
-            columnName: key,
-            operator: operator,
-            // Add wildcards around the value
-            values: [`%${removePrefixIfSynId(searchParams[key])}%`],
+    additionalFilters = additionalFilters.concat(
+      Object.keys(searchParams)
```
nickgros, line 69: I think you can just get rid of L61-L66 if you do this:

```suggestion
      Object.keys(searchParams || {})
```

nickgros (changes requested):

packages/synapse-react-client/src/utils/functions/SqlFunctions.ts
jay-hodgson: Rather not have to keep this in sync if we don't need to (since this is stored in the Query after load)

packages/synapse-react-client/src/utils/functions/SqlFunctions.ts
jay-hodgson: Changed the existing test (that has no search params) to provide an empty search params object instead, which failed before this change was made.

nickgros (approved):

PR #299, opened 2023-06-12
PORTALS-2665 - Visual update for FacetPlotsCard

nickgros:
- Use grid layout to ensure contents of each card are aligned across card
- Add new FacetPlotLegendTable to be used in this card
- Fix bug where FeaturedDataTabs sql would incorrectly override the inner plot config sql parameter (present in dhealth).
- Convert stories to use mock data

Recommend viewing in [Storybook](https://6464de5f2e01a2af614b79c7-aoewmodrpa.chromatic.com/?path=/story/home-page-featureddatatabs--single-query) and reviewing [UI Tests](https://www.chromatic.com/build?appId=6464de5f2e01a2af614b79c7&number=213) (note additional viewports were added for this component, but diffs are visible for the existing viewport)


apps/portals/src/configurations/adknowledgeportal/routesConfig.ts
```diff
@@ -74,6 +74,7 @@ const routes: GenericRoute[] = [
                     selectFacetColumnValue: 'WGS_Harmonization',
                     detailsPagePath:
                       '/Explore/Studies/DetailsPage?Study=syn22264775',
+                    unitDescription: 'Files',
```
nickgros, line 77: Determines the plot legend table column header: 
image

packages/synapse-react-client/mocks/query/mockFeaturedDataTabsQueryData.ts
nickgros, line 1: Mock data for Storybook

packages/synapse-react-client/src/components/FeaturedDataTabs/FacetPlotsCard.tsx
```diff
@@ -49,19 +59,59 @@ const layout: Partial<Plotly.Layout> = {
...
   description,
   facetsToPlot,
   detailsPagePath,
 }: FacetPlotsCardProps): JSX.Element => {
   const { accessToken } = useSynapseContext()
-  const { data, isLoadingNewBundle } = useQueryContext()
-  const { getColumnDisplayName, rgbIndex } = useQueryVisualizationContext()
+  const { data, isLoadingNewBundle } = useQueryContext<
+    'columnModels' | 'facets',
+    true
```
nickgros, line 109: I added some generics to make using `useQueryContext` a bit more friendly. Since the `partMask` is hard-coded, we can safely set this generic to basically say `columnModels` and `facets` will be defined. The boolean generic says to remove the other optional parts from the type of the returned object (e.g. the result rows).

packages/synapse-react-client/src/components/FeaturedDataTabs/FacetPlotsCard.tsx
```diff
@@ -70,7 +120,7 @@ const FacetPlotsCard: React.FunctionComponent<FacetPlotsCardProps> = ({
       const getColumnType = (
         facetToPlot: FacetColumnResult,
       ): ColumnTypeEnum | undefined =>
-        data?.columnModels?.find(
+        data?.columnModels.find(
```
nickgros, line 123: as a result of specifying the generic above, we know `columnModels` is defined if `data` is defined, so no `?` is needed.

packages/synapse-react-client/src/components/FeaturedDataTabs/FacetPlotsCard.tsx
```diff
@@ -113,73 +163,68 @@ const FacetPlotsCard: React.FunctionComponent<FacetPlotsCardProps> = ({
...
     !facetPlotDataArray ||
     !facetDataArray ||
     facetDataArray.length === 0
   ) {
-    return (
-      <div className="FacetPlotsCard FacetPlotsCard__loading SRC-centerContentColumn">
-        {loadingScreen}
-      </div>
-    )
+    return <LoadingCard numPlots={(facetsToPlot ?? []).length} />
```
nickgros, line 174: Add a generic loading card that fills the space properly using the grid layout. It's really nice to have a skeleton that closely match the actual result in this case especially, because if plots come from independent queries, they will finish at different times.

packages/synapse-react-client/src/components/FeaturedDataTabs/FacetPlotsCard.tsx
```diff
@@ -189,18 +234,22 @@ const FacetPlotsCard: React.FunctionComponent<FacetPlotsCardProps> = ({
                     )}
                   </SizeMe>
                   <div className="FacetPlotsCard__body__legend">
-                    <FacetPlotLegend
+                    <FacetPlotLegendTable
+                      facetName={
+                        getColumnDisplayName(facetDataArray[index].columnName)!
+                      }
```
nickgros, line 240: Determines the table column header (in this case, "Data Type")

image

packages/synapse-react-client/src/components/FeaturedDataTabs/FacetPlotsCardGrid.tsx
```diff
@@ -0,0 +1,109 @@
...
+    gap: theme.spacing(2),
+  },
+  [theme.breakpoints.up('md')]: {
+    /* 
+    The grid is broken up into cards which have subcomponents that should be aligned
+    Ideally we could use CSS Subgrid, but it is not widely supported as of the time of writing
+    See https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_grid_layout/Subgrid
+
+    Instead we will use grid with `display: contents`, and style each section to look like a card.
+     */
```
nickgros, line 28: This is super hacky, so I tried to call attention to this as fixable if we ever get `subgrid`

packages/synapse-react-client/src/components/FeaturedDataTabs/FacetPlotsCardGrid.tsx
```diff
@@ -0,0 +1,109 @@
...
+    display: 'grid',
+    gridTemplateRows: '[title] auto [plot0] auto',
+    gridTemplateColumns: '[col1] 32% [col2] 32% [col3] 32%',
+    columnGap: '2%',
+
+    '.FacetPlotsCard': {
+      display: 'contents',
```
nickgros, line 35: Display `contents` basically moves everything up one level, and any styles on this element are not rendered, even though it remains in the DOM.

packages/synapse-react-client/src/components/FeaturedDataTabs/FeaturedDataPlots.tsx
```diff
@@ -23,35 +26,63 @@ const FeaturedDataPlots: React.FunctionComponent<
...
+        {chunk(configs as QueryPerFacetPlotsCardProps[], CARDS_PER_ROW).map(
+          (configRow, rowIndex) => {
+            let cssGridTemplateRows = '[title] auto'
+
+            // There may be more than one plot per card, so add a row for each one
+            times(configRow[0].facetsToPlot?.length ?? 1).forEach(
+              (value, index) => {
+                cssGridTemplateRows += ` [plot${index}] auto`
+              },
+            )
```
nickgros, line 59: Dynamically update the grid to support `n` plots

packages/synapse-react-client/src/components/FeaturedDataTabs/QueryPerFacetPlotsCard.tsx
```diff
@@ -29,9 +33,7 @@ export function getQueryRequest(
     concreteType: 'org.sagebionetworks.repo.model.table.QueryBundleRequest',
     partMask:
       SynapseConstants.BUNDLE_MASK_QUERY_COLUMN_MODELS |
-      SynapseConstants.BUNDLE_MASK_QUERY_FACETS |
-      SynapseConstants.BUNDLE_MASK_QUERY_SELECT_COLUMNS |
-      SynapseConstants.BUNDLE_MASK_QUERY_RESULTS,
```
nickgros, line 34: We didn't need this much data to render the plots (and I was able to figure this out with type-safety with the new generic capabilities added to `QueryContext`)

packages/synapse-react-client/src/components/FeaturedDataTabs/SingleQueryFacetPlotsCards.tsx
```diff
@@ -1,59 +1,73 @@
...
-      </QueryWrapper>
-    </div>
+    <QueryWrapper initQueryRequest={initQueryRequest}>
+      <QueryVisualizationWrapper
+        rgbIndex={rgbIndex}
+        columnAliases={columnAliases}
+        unitDescription={unitDescription}
+      >
+        <QueryWrapperErrorBanner />
+        {chunk(facetsToPlot, CARDS_PER_ROW).map((facets, rowIndex) => {
```
nickgros, line 52: Since this is one-plot-per-facet, we chunk into groups of 3 at this level.

packages/synapse-react-client/src/components/QueryContext/QueryContext.tsx
```diff
@@ -23,11 +24,33 @@ export type LockedColumn = {
...
+  | 'maxRowsPerPage'
+  | 'columnModels'
+  | 'facets'
+  | 'sumFileSizes'
+  | 'combinedSql'
+>
+
+export type QueryContextType<
+  TIncludedFields extends OptionalQueryBundleRequestFields = never,
+  ExcludeOtherFields extends true | false = false,
```
nickgros, line 41: These default generics maintain the existing behavior--all of the optional properties remain optional.

packages/synapse-react-client/src/components/QueryContext/QueryContext.tsx
```diff
@@ -23,11 +24,33 @@ export type LockedColumn = {
...
   entity: Table | undefined
   /** The query results, which will be undefined while initially fetching a new bundle, but will not be unloaded when fetching new pages */
-  data: QueryResultBundle | undefined
+  data:
+    | Omit<
+        SetRequired<QueryResultBundle, TIncludedFields>,
+        ExcludeOtherFields extends true
+          ? Exclude<OptionalQueryBundleRequestFields, TIncludedFields>
+          : never
+      >
```
nickgros, line 52: `SetRequired<QueryResultBundle, TIncludedFields>` makes the specified fields non-optional (e.g. we can enforce that the columnModels were returned).

The `Omit` removes the other fields if `ExcludeOtherFields` is true. If it's false, the other fields will remain optional aka possibly undefined.

packages/synapse-react-client/src/components/widgets/facet-nav/FacetNavPanel.tsx
```diff
@@ -71,14 +72,6 @@ export type GraphData = {
   colors: string[]
 }
 
-export function truncate(str: string | undefined, n: number) {
```
nickgros, line 74: Moved

packages/synapse-react-client/src/components/widgets/facet-nav/FacetNavPanel.tsx
```diff
@@ -188,7 +181,7 @@ export async function extractPlotDataArray(
       plotType === 'BAR'
         ? facetToPlot.facetValues.map(facet => facet.count)
         : undefined,
-    // @ts-ignore
+    // @ts-expect-error
```
nickgros, line 184: Prefer `@ts-expect-error` because if some change causes this to become valid, `tsc` will fail and we will know to remove this comment.

packages/synapse-react-client/src/components/widgets/facet-nav/FacetNavPanel.tsx
```diff
@@ -261,62 +254,6 @@ export function getPlotStyle(
   }
 }
 
-export type FacetWithLabel = {
```
nickgros, line 264: Extracted to new file

packages/synapse-react-client/src/components/widgets/facet-nav/FacetPlotLegendList.tsx
nickgros, line 1: The "List" variant of the FacetPlotLegend was just moved from `FacetNavPanel`. It was also used by `FacetPlotsCard`, but that is no longer the case.

packages/synapse-react-client/src/components/widgets/facet-nav/FacetPlotLegendTable.tsx
nickgros, line 1: The "Table" variant of the `FacetPlotLegend` is used by the `FacetPlotsCard`

packages/synapse-react-client/src/style/components/_all.scss
```diff
@@ -7,15 +7,15 @@
   './query_filter/enum-facet-filter', './query_filter/facet-filter-header',
   './query_filter/facet-chip', './error', './goals', './feed-cards',
   './upset-plot', './resources', 'cards-rotate', './expandable_content',
-  './bar-loader', './facet-plots-card', './featured-data-plots',
```
nickgros, line 10: only removed `featured-data-plots.scss`, which was deleted

packages/synapse-react-client/stories/FeaturedDataTabs.stories.ts
```diff
@@ -41,6 +57,16 @@ export const SingleQuery: Story = {
 }
 
 export const QueryPerCard: Story = {
+  parameters: {
+    chromatic: { viewports: [600, 1200] },
+    msw: {
+      handlers: [
+        // In reality, we actually make multiple table queries that return different data
+        // It will suffice if all queries return the same data for this example
+        ...getHandlersForTableQuery(mockQueryPerCardResult, MOCK_REPO_ORIGIN),
```
nickgros, line 66: We'd need to somehow map query requests to generated result IDs to support multiple table query handlers at one time (so we would basically need a database of some sort)

packages/synapse-react-client/src/components/FeaturedDataTabs/FacetPlotsCardGrid.tsx
jay-hodgson: How easy would this be to do with `subgrid` support?  If straight-forward, could you provide an example implementation in the comments here?

packages/synapse-react-client/src/components/FeaturedDataTabs/FacetPlotsCard.tsx
jay-hodgson: So it's still possible that the type is not aligned with the partMask.  In general, I think these generics will be helpful for compile-time confidence in the response.  However, if these do not match at runtime, a developer unfamiliar with this requirement will face an interesting mystery!

jay-hodgson (approved):

packages/synapse-react-client/src/components/FeaturedDataTabs/FacetPlotsCardGrid.tsx
nickgros: While it would be much easier and less hacky than this implementation, I don't think it would be trivial, so I don't think it would be very valuable to flesh out an implementation now.

jay-hodgson (approved):

PR #303, opened 2023-06-13
SWC-6465 - VirtualTable support

nickgros:


jay-hodgson (approved):
Approved once error in EntityTypeUtils.ts is fixed

PR #305, opened 2023-06-14
PORTALS-2447: Integrate GenUI feature 1, 2, 3 into Challenge portal

jay-hodgson:
Needed to track down an infinite rendering loop in ChallengeTeamWizard


apps/portals/src/configurations/challengeportal/synapseConfigs/challenges.ts
```diff
@@ -278,6 +278,12 @@ export const challengeDetailsPageConfig: DetailsPageProps = {
 }
 
 export const challengeDetailsLandingPage: SynapseConfig[] = [
+  // Register For Challenge component is here, and the wrapper positions it properly in the header card
+  {
+    name: 'ChallengeDetailPageWrapper',
```
jay-hodgson, line 283: Wrapper pulls challenge project ID from the search param, and does some hacky things to make sure the registration button shows up in a relatively nice position on screen

apps/portals/src/portal-components/challengeportal/ChallengeParticipantGoogleMap.tsx
```diff
@@ -1,11 +1,10 @@
...
-import { SynapseQueries, Map } from 'synapse-react-client'
+import { AppUtils, SynapseQueries, Map } from 'synapse-react-client'
 
 const ChallengeParticipantGoogleMap = () => {
-  const urlSearchParams = new URLSearchParams(window.location.search)
-  const entityId = urlSearchParams.get('id') ?? ''
-  const { data: challenge } = SynapseQueries.useGetEntityChallenge(entityId)
+  const projectId = AppUtils.useQuerySearchParam('id')
+  const { data: challenge } = SynapseQueries.useGetEntityChallenge(projectId)
   if (challenge && challenge.participantTeamId) {
```
jay-hodgson, line 8: Use a new useQuerySearchParam hook (defined later in the PR) to grab a single parameter from the URL.

jay-hodgson: Also found that for the GenUI team, there was no geolocation file, which caused the app to crash (because it was attempting to parse json from the html response).  So I fixed this call as well (to only attempt to parse the json if the status code is 200)

apps/portals/src/types/portal-config.ts
```diff
@@ -204,7 +205,14 @@ type Image = {
   name: 'Image'
   props: ImageProps
 }
-
+type ChallengeDetailPageWrapper = {
+  name: 'ChallengeDetailPageWrapper'
+  props: undefined
+}
+type ChallengeDataDownload = {
```
jay-hodgson, line 212: Prep, not used in this PR

packages/synapse-react-client/src/components/AccessRequirementList/AccessRequirementList.tsx
```diff
@@ -50,6 +50,7 @@ export type AccessRequirementListProps = {
   renderAsModal?: boolean
   numberOfFilesAffected?: number // if provided, will show this instead of the entity information
   requestObjectName?: string // if provided, will show this instead of the entity information or numberOfFilesAffected
+  dialogTitle?: string // if provided, will show this instead of Data Access Request
```
jay-hodgson, line 53: I thought this was available at one point, but it's not in there.  Need to override the dialog title in our case!

packages/synapse-react-client/src/components/ChallengeRegisterButton/ChallengeRegisterButton.tsx
```diff
@@ -135,17 +100,19 @@ const ChallengeRegisterButton = ({
         <SpinnerButton
           disableElevation={true}
           variant="contained"
-          color="secondary"
-          onClick={() => (onJoinClick ? onJoinClick() : undefined)}
+          onClick={() =>
+            accessToken && onJoinClick ? onJoinClick() : undefined
```
jay-hodgson, line 104: The challenge register button already handles the case when the user is not logged in (it pops up a sign-in dialog), so don't call back until user is logged in.  In this case, after login, it will call back and cause the AccessRequirementList dialog to pop up (which will at least say that you need to be logged in!)

packages/synapse-react-client/src/components/ChallengeRegisterButton/ChallengeRegisterButton.tsx
```diff
@@ -135,17 +100,19 @@ const ChallengeRegisterButton = ({
...
+          }
           sx={{
             color: 'white',
+            backgroundColor: '#109488',
```
jay-hodgson, line 108: Quick and dirty background color override since we can't use the secondary version of the button here either.

packages/synapse-react-client/src/components/ChallengeTeamWizard/ChallengeTeamWizard.tsx
```diff
@@ -112,92 +111,57 @@ const ChallengeTeamWizard: React.FunctionComponent<
...
-      setLoading(false)
-      setErrorMessage(
-        `Error: Could not retrieve challenge for project "${projectId}".`,
-      )
-    },
-  })
-
+  const { data: challenge } = useGetEntityChallenge(projectId)
+  const participantTeamId = challenge ? challenge.participantTeamId : EMPTY_ID
+  const userId = userProfile ? userProfile.ownerId : EMPTY_ID
```
jay-hodgson, line 124: This was changed during debugging (for sanity).  No logical change

jay-hodgson:
Screen Shot 2023-06-13 at 8 53 33 PM

Screen Shot 2023-06-13 at 8 53 41 PM

Screen Shot 2023-06-13 at 8 53 49 PM

packages/synapse-react-client/src/components/ChallengeRegisterButton/ChallengeRegisterButton.tsx
nickgros: In Storybook it looks like we can just add the `color="secondary"` prop. Why can't we do that here?

nickgros (changes requested):

packages/synapse-react-client/src/components/ChallengeRegisterButton/ChallengeRegisterButton.tsx
jay-hodgson: If we set the secondary color to this in the challenge portal, it changes the colors of many other elements that we don't want to change

pnpm-lock.yaml
```diff
@@ -1356,7 +1356,7 @@ importers:
         version: 5.0.4
       vite:
         specifier: ^4.3.4
-        version: 4.3.4(@types/node@18.16.12)(sass@1.57.1)
+        version: 4.3.4(@types/node@14.18.36)(sass@1.57.1)
```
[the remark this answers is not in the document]
jay-hodgson: Oops, I usually nvm to 18.  Must have been on 14, I'll revert

packages/synapse-react-client/src/components/ChallengeTeamWizard/ChallengeTeamWizard.tsx
```diff
@@ -122,24 +121,23 @@ const ChallengeTeamWizard: React.FunctionComponent<
...
-          })
-        }
-      })
-      .catch(error => {
+  const { data: challengeTeamMembershipStatus } = useGetMembershipStatus(
+    participantTeamId,
+    userId,
+  )
+  useEffect(() => {
+    if (!challengeTeamMembershipStatus?.isMember && accessToken) {
```
nickgros, line 131: Will run if `challengeTeamMembershipStatus` is undefined, which will happen when it's not fetched

```suggestion
    if (challengeTeamMembershipStatus && !challengeTeamMembershipStatus?.isMember && accessToken) {
```

packages/synapse-react-client/src/components/ChallengeTeamWizard/ChallengeTeamWizard.tsx
```diff
@@ -122,24 +121,23 @@ const ChallengeTeamWizard: React.FunctionComponent<
...
+      addTeamMemberAsAuthenticatedUserOrAdmin(
+        participantTeamId,
+        userId,
+        accessToken,
+      ).catch(error => {
```
nickgros, line 136: Would also recommend invalidating the cache for `useGetMembershipStatus` when this completes, but it's not strictly necessary, since `challengeTeamMembershipStatus` won't change and cause this effect to run multiple times.

nickgros (approved):

PR #306, opened 2023-06-14
PORTALS-2598: Remove deprecated tsconfig property

nickgros:
- Remove `suppressImplicitAnyIndexErrors` and `ignoreDeprecations` from tsconfig
- Adjust logic or add casts to account for tsconfig change


packages/synapse-react-client/src/components/Markdown/MarkdownEditor.tsx
```diff
@@ -9,10 +9,7 @@ import MarkdownSynapse from './MarkdownSynapse'
 import { UserMentionModal } from './UserMentionModal'
 import { startCase } from 'lodash-es'
 
-export enum MarkdownEditorTabs {
-  WRITE = 'WRITE',
-  PREVIEW = 'PREVIEW',
-}
+const MarkdownEditorTabs = ['WRITE', 'PREVIEW'] as const
```
nickgros, line 12: Due to how TS handles string enums, you can't get accurate types when you key into them. For this reason, I changed most of our simple enums like this to static arrays.

packages/synapse-react-client/src/components/SchemaDrivenAnnotationEditor/field/AdditionalPropertiesSchemaField.tsx
```diff
@@ -16,50 +16,50 @@ const ISO_TIMESTAMP_REGEX = /\d{4}-\d{2}-\d{2}T\d{2}:\d{2}:\d{2}\.\d{3}Z/
 
 // Types that correspond to the different input fields that the annotation editor supports
 export enum PropertyType {
```
jay-hodgson, line 18: Just so I understand, why was this simple enum not converted into a static array?

packages/synapse-react-client/src/components/SynapseForm/SynapseFormWrapper.tsx
```diff
@@ -122,7 +122,7 @@ export class SynapseFormWrapper extends React.Component<
       }
 
       //for not new form we need to get the data
-      //and if it is submitted form we need to pull appropriave schema versions
+      //and if it is submitted form we need to pull appropriaves schema versions
```
jay-hodgson, line 125: Is this supposed to say "appropriate"?

jay-hodgson (approved):

packages/synapse-react-client/src/components/SchemaDrivenAnnotationEditor/field/AdditionalPropertiesSchemaField.tsx
nickgros: No real reason, it's basically equivalent if the keys and values are the same. But that invariant could easily be broken unintentionally, so I think I'll also change this to a string union defined by array

packages/synapse-react-client/src/components/SynapseForm/SynapseFormWrapper.tsx
nickgros: Don't recall seeing/changing this. Fixed now

PR #309, opened 2023-06-14
PORTALS-2447: and invalidate on membership changes

jay-hodgson:
this causes the registration button to properly update state


packages/synapse-react-client/src/components/ChallengeDetailPage/ChallengeDetailPage.tsx
```diff
@@ -47,7 +49,14 @@ export function ChallengeDetailPage({ projectId }: ChallengeDetailPageProps) {
...
         accessToken,
-      )
+      ).then(() => {
+        queryClient.invalidateQueries(
+          keyFactory.getIsUserMemberOfTeamQueryKey(
+            challenge.participantTeamId,
+            userProfile.ownerId,
+          ),
+        )
+      })
```
nickgros, line 59: If you have time, would prefer to refactor so that we have a `useDeleteTeamMembership` mutation colocated  with the other react-query hooks, which internally handles invalidation in the `onSuccess` callback.

Then the hooks are more reusable and we don't need to worry about invalidation at the component level. See `useDeleteThread` in useThread.tsx for an example

nickgros (approved):

PR #312, opened 2023-06-15
Run chromatic builds for PRs when "chromatic" label is applied

nickgros:
Our Chromatic trial has expired so we are now limited to 5000 snapshots per month. The Chromatic deployments are not necessarily useful for every PR, so one measure we can take to stay close to our limit is to explicitly enable the Chromatic deployment and UI Snapshot tests for those PRs where a developer or reviewer knows they would be particularly useful.

Once this PR is merged, to trigger a Chromatic deployment, a PR author or reviewer can apply the "chromatic" label in GitHub (applied to this issue for clarity). Once applied, a job will be triggered that deploys the latest commit to Chromatic.

Jobs will not be automatically triggered on new commits after the label is applied; to trigger additional runs on the same PR, remove and re-apply the "chromatic" label again.

For an example, see https://github.com/nickgros/synapse-web-monorepo/pull/10

### DIFF
--- a/.github/workflows/deploy-to-chromatic.yml
+++ b/.github/workflows/deploy-to-chromatic.yml
@@ -7,40 +7,24 @@ on:
   push:
     branches:
       - main
-  issue_comment:
-    types: [created]
+  pull_request:
+    types: [labeled]
 
 jobs:
   deploy-to-chromatic:
     name: Deploy to Chromatic
     runs-on: ubuntu-latest
     steps:
-      - name: Get PR if triggered by comment
-        if: github.event.issue.pull_request
-        uses: alessbell/pull-request-comment-branch@v1.1
-        id: comment-branch
-      - name: Configure to use main
-        if: github.event_name == 'push'
+      - name: Check PR label
+        if: ${{ github.event_name == 'pull_request' && github.event.label.name != 'chromatic' }}
         env:
-          FULL_REPO_NAME: ${{ github.repository }}
-          HEAD_REF: ${{ github.ref }}
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        # The following script will cancel the rest of the job
         run: |
-          echo "FULL_REPO_NAME=$FULL_REPO_NAME" >> "$GITHUB_ENV"
-          echo "HEAD_REF=$HEAD_REF" >> "$GITHUB_ENV"
-      - name: Configure to use PR
-        if: github.event.issue.pull_request
-        env:
-          FULL_REPO_NAME: ${{ steps.comment-branch.outputs.head_owner }}/${{ steps.comment-branch.outputs.head_repo }}"
-          HEAD_REF: ${{ steps.comment-branch.outputs.head_ref }}
-        run: |
-          echo "FULL_REPO_NAME=$FULL_REPO_NAME" >> "$GITHUB_ENV"
-          echo "HEAD_REF=$HEAD_REF" >> "$GITHUB_ENV"
-      - name: Output configuration
-        run: echo "Deploying $FULL_REPO_NAME to Chromatic using branch $HEAD_REF"
+          gh run cancel ${{ github.run_id }}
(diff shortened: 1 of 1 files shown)


.github/workflows/deploy-to-chromatic.yml
```diff
@@ -34,3 +31,51 @@ jobs:
           exitZeroOnChanges: true
           exitOnceUploaded: true
           autoAcceptChanges: 'main'
+  pr-deploy-to-chromatic:
+    if: github.event.issue.pull_request && contains(github.event.comment.body, '/chromatic')
```
jay-hodgson, line 35: Is this the only time a chromatic build would be deployed (if the comment contains this string)?

.github/workflows/deploy-to-chromatic.yml
```diff
@@ -34,3 +31,51 @@ jobs:
...
+        uses: chromaui/action@v1
+        with:
+          projectToken: ${{ secrets.CHROMATIC_PROJECT_TOKEN }}
+          workingDir: packages/synapse-react-client
+          exitZeroOnChanges: true
+          exitOnceUploaded: false # Wait for snapshot tests to complete so we can report the results
+          autoAcceptChanges: 'main'
+          repositorySlug: ${{ steps.comment-branch.outputs.head_owner }}/${{ steps.comment-branch.outputs.head_repo }}
+          branchName: ${{ steps.comment-branch.outputs.head_ref }}
+      - name: Add comment to PR
```
jay-hodgson, line 63: Like the idea of an automatically generated comment to surface these results in the PR

.github/workflows/deploy-to-chromatic.yml
nickgros: Yep, you would have to comment "/chromatic"

.github/workflows/deploy-to-chromatic.yml
nickgros: I'd rather use the built-in status check on the commit (as we have now), but I can't seem to get that to work--looking at the Chromatic GH Action source code, I think they override any environment variables I might be able to pass. So this will have to do for now.

.github/workflows/deploy-to-chromatic.yml
```diff
@@ -32,5 +41,5 @@ jobs:
           projectToken: ${{ secrets.CHROMATIC_PROJECT_TOKEN }}
           workingDir: packages/synapse-react-client
           exitZeroOnChanges: true
-          exitOnceUploaded: true
+          exitOnceUploaded: true # Wait for snapshot tests to complete so we can report the results
```
nickgros, line 44:
```suggestion
          exitOnceUploaded: true
```

.github/workflows/deploy-to-chromatic.yml
```diff
@@ -8,14 +8,23 @@ on:
...
     runs-on: ubuntu-latest
     steps:
+      - name: Check PR label
+        if: ${{ github.event_name == 'pull_request' && github.event.label.name != 'chromatic' }}
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        # The following script will cancel the rest of the job
+        run: |
+          gh run cancel ${{ github.run_id }}
+          gh run watch ${{ github.run_id }}
```
nickgros, line 27: If the event is a label on a PR, but that label isn't "chromatic", end the run now.

.github/workflows/deploy-to-chromatic.yml
nickgros: Opted for a different approach to continue to get the native GitHub status checks

.github/workflows/deploy-to-chromatic.yml
nickgros: If we use labels instead of comments, then we can continue to use the `pull_request_target` event, which Chromatic properly supports.

jay-hodgson (approved):

PR #318, opened 2023-06-16
PORTALS-2698: add optional subscription link to the portal header

jay-hodgson:


apps/portals/src/_Core.scss
```diff
@@ -421,10 +421,6 @@ button.dropdown-toggle.nav-button:hover {
 }
 
 .header-text {
-  h2 {
```
jay-hodgson, line 424: Moved to Typography sx

apps/portals/src/configurations/elportal/headerConfig.tsx
```diff
@@ -4,10 +4,17 @@ import { HomePageHeaderConfig } from 'types/portal-config'
...
-      Discover and explore exceptional longevity data, analyses, and tools generated by the translational longevity research projects supported by the National Institute on Aging (NIA).
+      Discover and explore exceptional longevity data, analyses, and tools
+      generated by the translational longevity research projects supported by
+      the National Institute on Aging (NIA).
     </>
   ),
-  title: 'Welcome to the Exceptional Longevity Translational Resources (ELITE) Portal',
+  title:
+    'Welcome to the Exceptional Longevity Translational Resources (ELITE) Portal',
+  subscriptionConfig: {
```
jay-hodgson, line 14: Add optional subscriptionConfig

jay-hodgson:
Screen Shot 2023-06-16 at 12 47 34 PM

apps/portals/src/configurations/elportal/headerConfig.tsx
```diff
@@ -1,13 +1,36 @@
...
+      generated by the translational longevity research projects supported by
+      the National Institute on Aging (NIA).
+      <>
+        <Link
+          href="https://news.eliteportal.org/elite-portal-newsletter/"
+          target="_blank"
+          sx={{
+            color: 'white',
+            '&:hover': { color: 'white' },
+            '&:focus': { color: 'white' },
```
nickgros, line 19: Is the text in the header already white? You may be able to add the prop `variant='inherit'`. Not totally sure if that will work, since bootstrap 3 styles might still override the hover/focus selectors.

nickgros (approved):

apps/portals/src/configurations/elportal/headerConfig.tsx
jay-hodgson: The variant `inherit` did not fix the base color (which was still set to the default mui-based link primary color), or hover/focus state.

PR #321, opened 2023-06-19
PORTALS-2701 - refactor QueryWrapper to deprecate `getLastQueryRequest`

nickgros:
- Deprecate `getLastQueryRequest` and just use a readonly `lastQueryRequest` object
- Expose dispatch signature for `executeQueryRequest`

Every call to `getLastQueryRequest` triggered an unnecessary `deepClone` call. Most of these instances are able to just use the object. Cases where the query is updated are able to use the dispatch signature for `executeQueryRequest`. This may have a slight impact on performance.

### DIFF
--- a/packages/synapse-react-client/src/synapse-client/SynapseClient.ts
+++ b/packages/synapse-react-client/src/synapse-client/SynapseClient.ts
@@ -626,7 +626,7 @@ export const getAsyncResultFromJobId = async <TRequest, TResponse>(
   )
   setCurrentAsyncStatus?.(response)
   while (response.jobState && response.jobState === 'PROCESSING') {
-    await delay(500)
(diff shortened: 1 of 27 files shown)


packages/synapse-react-client/src/synapse-client/SynapseClient.ts
```diff
@@ -626,7 +626,7 @@ export const getAsyncResultFromJobId = async <TRequest, TResponse>(
   )
   setCurrentAsyncStatus?.(response)
   while (response.jobState && response.jobState === 'PROCESSING') {
-    await delay(500)
+    await delay(1000)
```
nickgros, line 629: TODO - maybe remove? Did not mean to commit this, but I was getting 429s

PR #325, opened 2023-06-20
Feature/challenge submission (Feature 5)

geoffroth:
### Description

This branch adds the ability to submit a file or docker commit to an evaluation queue belonging to a challenge.

The following new components were added to the `ChallengeSubmission` directory:
• `AddExternalRepo` - (NB: due to a change in the desired flow, this component is no longer being used, please feel free to delete if desired)
• `ChallengeSubmission` - the top-level component for this feature. It displays a list of files or docker repos belonging to the challenge team's project, and will display a stepper for submitting a selected file or repo commit to the challenge evaluation queue.
• `ChallengeSubmissionStepper` - this manages the steps for selecting commits and submitting them to an eval queue
• `EvaluationQueueList` - displays a table of active evaluation queues belonging to the given challenge
• `SubmissionCommitList` - displays a table of available docker commits for the selected repo
• `SubmissionDirectoryList` - displays a table of available files or docker repos for submission


packages/synapse-react-client/src/components/ChallengeSubmission/ChallengeSubmission.tsx
```diff
@@ -0,0 +1,281 @@
...
+      }
+    },
+  })
+
+  useGetEntityPermissions(challengeProjectId!, {
+    enabled: !!challengeProjectId,
+    refetchInterval: Infinity,
+    onSettled: (data, error) => {
+      if (data && data.canView && data.canAddChild) {
+        setCanSubmit(true)
```
jay-hodgson, line 195: This is misnamed.  
It looks like you have determined that the user can upload files (or create new Docker repo entities) to the challenge project associated to this user submission team.  
(it's technically possible that they will not have the ability to submit to the evaluation queue)

packages/synapse-react-client/src/components/ChallengeSubmission/ChallengeSubmission.tsx
```diff
@@ -0,0 +1,281 @@
...
+          ],
+        }
+        updateACL({
+          ...data,
+          resourceAccess: [...data.resourceAccess, teamResourceAccess],
+        })
+        setIsProjectNewlyCreated(false)
+      }
+      if (error) {
+        setErrorMessage(`Error: Could not retrieve project ACL.`)
```
jay-hodgson, line 185: Is this needed, if you have an error boundary and you've told it to useErrorBoundary?

packages/synapse-react-client/src/components/ChallengeSubmission/ChallengeSubmission.tsx
jay-hodgson: Maybe keep the name though, if it makes sense

packages/synapse-react-client/src/components/ChallengeSubmission/ChallengeSubmission.tsx
geoffroth: Removed

packages/synapse-react-client/src/components/ChallengeSubmission/ChallengeSubmissionStepper.tsx
```diff
@@ -0,0 +1,206 @@
...
+  const submitRepoForEvaluation = async () => {
+    if (!entity.id || !selectedCommit) return
+
+    const submission: EvaluationSubmission = {
+      userId: userId,
+      evaluationId: selectedEval!,
+      entityId: entity.id,
+      dockerRepositoryName: entity.repositoryName,
+      dockerDigest: selectedCommit.digest,
+      versionNumber: 1,
```
jay-hodgson, line 110: As you mentioned in slack, this version number should be the version returned with the entity.
Also need to set contributors.
https://rest-docs.synapse.org/rest/org/sagebionetworks/evaluation/model/Submission.html

packages/synapse-react-client/src/components/ChallengeSubmission/ChallengeSubmissionStepper.tsx
```diff
@@ -0,0 +1,206 @@
...
+            selectedEvaluation={selectedEval}
+            onEvaluationChange={value => {
+              setStep({ ...step, confirmEnabled: true })
+              setSelectedEval(value)
+            }}
+          />
+        )
+      case StepsEnum.SUBMISSION_SUCCESS:
+        return (
+          <Typography variant="body1" sx={{ fontSize: '16px' }}>
```
jay-hodgson, line 181: Is this really 'body2' (if the fontSize needs to be adjusted)?

geoffroth: I agree it's not. There are several font-size mismatches between the available variants and the styles called for in Figma. How would it be best to handle these cases?

packages/synapse-react-client/src/components/ChallengeSubmission/SubmissionDirectoryList.tsx
```diff
@@ -0,0 +1,376 @@
...
+    if (response.success && response.resp) {
+      // Create Entity
+      const { fileHandleId, fileName } = response.resp
+      const newFileEntity: FileEntity = {
+        parentId: challengeProjectId,
+        name: fileName,
+        concreteType: FILE_ENTITY_CONCRETE_TYPE_VALUE,
+        dataFileHandleId: fileHandleId,
+      }
+      SynapseClient.createEntity(newFileEntity, accessToken)
```
jay-hodgson, line 247: This file may already exist and they may want to overwrite the value (create a new version).  Please try to find child  using https://rest-docs.synapse.org/rest/POST/entity/child.html
(either before you attempt to create, or after you get receive a 409 status code error when trying to create the entity and you run into the conflict).  If child exists with this name, then you should just update the FileEntity dataFileHandleId and update the entity in the backend (preferably after a confirmation dialog, but you can ignore for now if that saves time).

packages/synapse-react-client/src/components/ChallengeSubmission/ChallengeSubmissionStepper.tsx
geoffroth: Added the `entity.versionNumber` if defined, otherwise defaulting to 1 (versionNumber is required on `EvaluationSubmission`)

jay-hodgson: It's better to fail than to submit the wrong entity version to the evaluation queue.  You could default to -1 and probably let the service call fail

packages/synapse-react-client/src/components/ChallengeSubmission/ChallengeSubmissionStepper.tsx
jay-hodgson: Your override is fine for this PR.  This might be a new variant from design, so web engineering should follow up.

packages/synapse-react-client/src/components/ChallengeSubmission/ChallengeSubmissionStepper.tsx
jay-hodgson: I don't actually see entity.versionNumber being used here, am I missing something?

packages/synapse-react-client/src/components/ChallengeSubmission/ChallengeSubmissionStepper.tsx
```diff
@@ -0,0 +1,243 @@
...
+      )
+      .map(member => {
+        return { principalId: member.principalId.toString() }
+      })
+
+    const submission: EvaluationSubmission = {
+      userId: userId,
+      evaluationId: selectedEval!,
+      entityId: entity.id,
+      versionNumber: entity.versionNumber ?? 1,
```
geoffroth, line 155: @jay-hodgson Is this what you're looking for?

jay-hodgson: Yes!  I must have been looking at an old commit.  
Is entity.versionNumber generally defined here?  Still recommend to fallback to an invalid number (like -1) if this must be set, so it will show an error if it's attempting to use the fallback

PR #326, opened 2023-06-21
Challenge submission followup

jay-hodgson:


jay-hodgson:
fix chromatic deployment, and remove fallback entity version value

PR #327, opened 2023-06-21
move away from onSettled fixes lifecycle issue

jay-hodgson:
onSettled is deprecated and will be removed in the next major version of react query


jay-hodgson:
Merging quickly to test the staging challenge portal site while GenUI is still available

PR #331, opened 2023-06-23
PORTALS-2679: get Terms of Use table data as the anonymous user only

jay-hodgson:


packages/synapse-react-client/src/synapse-queries/entity/useGetQueryResultBundle.ts
```diff
@@ -304,12 +304,16 @@ export function useInfiniteQueryResultBundle(
 export function useGetFullTableQueryResults(
   queryBundleRequest: QueryBundleRequest,
   options?: UseQueryOptions<QueryResultBundle, SynapseClientError>,
+  forceAnonymous: boolean = false,
 ): UseQueryResult<QueryResultBundle, SynapseClientError> {
   const { accessToken, keyFactory } = useSynapseContext()
   return useQuery<QueryResultBundle, SynapseClientError>(
     keyFactory.getFullTableQueryResultQueryKey(queryBundleRequest),
```
nickgros, line 311: Need to add the `forceAnonymous` parameter to the query key

PR #342, opened 2023-06-29
fix markdown-it-synapse publish and deployment

nickgros:
- add back IIFE/UMD distribution
- If version did not increment, `pnpm publish` should include `--dry-run`

### DIFF
--- a/packages/synapse-react-client/.storybook/preview-head.html
+++ b/packages/synapse-react-client/.storybook/preview-head.html
@@ -36,7 +36,7 @@
 ></script>
 
 <script src="https://cdn.jsdelivr.net/npm/markdown-it@8.4.2/dist/markdown-it.min.js"></script>
-<script src="https://unpkg.com/markdown-it-synapse@latest/dist/index.js"></script>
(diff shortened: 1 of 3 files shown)


jay-hodgson (approved):

packages/synapse-react-client/.storybook/preview-head.html
```diff
@@ -36,7 +36,7 @@
 ></script>
 
 <script src="https://cdn.jsdelivr.net/npm/markdown-it@8.4.2/dist/markdown-it.min.js"></script>
-<script src="https://unpkg.com/markdown-it-synapse@latest/dist/markdown-it-synapse.min.js"></script>
+<script src="https://unpkg.com/markdown-it-synapse@latest/dist/index.js"></script>
```
nickgros, line 39:
```suggestion
<script src="https://unpkg.com/markdown-it-synapse@latest/dist/index.global.js"></script>
```

nickgros, line 39:
```suggestion
<script src="https://unpkg.com/markdown-it-synapse@latest/dist/index.global.js"></script>
```

PR #346, opened 2023-07-03
PORTALS-2701: refactor/remove topLevelControlsState for clarity

nickgros:
The motivation for this change is mostly "addition by subtraction". The present name and grouping of state in `topLevelControlsState` implies that it's the state of the TopLevelControls component. The object actually represents the state of various components that is usually (but not always) modified by the TopLevelControls. This is the wrong abstraction, which is worse than no abstraction.

- QueryVisualizationWrapper: break `topLevelControlsState` down into individual state variables.
- QueryWrapperPlotNav: Apply `.isShowingFacetFilters`/`.isHidingFacetFilters` classes only to the top-level container and remove its application on each child
- Fix copy where "Download List" should be "Download Cart" Rewrite Download Cart related components and refactor state management


packages/synapse-react-client/src/components/DownloadCart/DownloadCartPage.tsx
```diff
@@ -72,7 +72,7 @@ export const DownloadCartPage: React.FunctionComponent<
           <div className="grid">
             <h3 className="pageHeaderTitle">Your Download Cart</h3>
             <Tooltip
-              title="Immediately removes all items from your download list"
+              title="Immediately removes all items from your download cart"
```
nickgros, line 75: Fix couple places where "download list" was used inconsistent with the rest of the copy

packages/synapse-react-client/src/components/FullTextSearch.tsx
```diff
@@ -73,13 +67,7 @@ export const FullTextSearch: React.FunctionComponent<FullTextSearchProps> = ({
   }
 
   return (
-    <div
-      className={`QueryWrapperSearchInput ${
-        showFacetFilter
-          ? QUERY_FILTERS_EXPANDED_CSS
-          : QUERY_FILTERS_COLLAPSED_CSS
-      }`}
```
nickgros, line 81: remove this check from every component, except QueryWrapperPlotNav. In `QueryWrapperPlotNav`, we apply the class once and update selectors accordingly.

packages/synapse-react-client/src/components/QueryContext/QueryContext.tsx
```diff
@@ -11,9 +11,6 @@ import {
 import { ImmutableTableQueryResult } from '../useImmutableTableQuery'
 import { SetRequired } from 'type-fest'
 
-export const QUERY_FILTERS_EXPANDED_CSS: string = 'isShowingFacetFilters'
-export const QUERY_FILTERS_COLLAPSED_CSS: string = 'isHidingFacetFilters'
```
nickgros, line 15: Now only used in QueryWrapperPlotNav, so move these there.

packages/synapse-react-client/src/components/QueryVisualizationWrapper.tsx
```diff
@@ -142,31 +141,32 @@ export function QueryVisualizationWrapper(
...
+  useEffect(() => {
+    if (showSearchBar) {
+      setShowDownloadConfirmation(false)
+    }
+  }, [showSearchBar])
+  useEffect(() => {
+    if (showDownloadConfirmation) {
+      setShowSearchBar(false)
+    }
+  }, [showDownloadConfirmation])
```
nickgros, line 165: I bet there are other "invalid" states that we should probably guard against.

packages/synapse-react-client/src/components/SynapseTable/TopLevelControls/TopLevelControls.tsx
```diff
@@ -274,55 +212,54 @@ const TopLevelControls = (props: TopLevelControlsProps) => {
...
-          {controls.map(control => {
-            const { key, icon, tooltipText } = control
-            if (
-              (key === 'showDownloadConfirmation' && hideDownload) ||
-              (key === 'showFacetVisualization' && hideVisualizationsControl) ||
-              (key === 'showSqlEditor' && hideSqlEditorControl) ||
-              (key === 'showCopyToClipboard' && !showCopyToClipboard)
-            ) {
-              // needs to be a file view in order for download to make sense
-              return <React.Fragment key={key}></React.Fragment>
```
nickgros, line 286: This was also confusing IMO. Better to colocate the "hide" logic with the control instead of with "hide" conditions for other controls.

packages/synapse-react-client/src/components/download_list/DownloadConfirmation.tsx
nickgros, line 1: Changes in this file are temporary, there will be a substantial refactor for this component in #347

packages/synapse-react-client/src/utils/hooks/useMutuallyExclusiveState.ts
nickgros, line 1: New utility

packages/synapse-react-client/test/utils/hooks/useMutuallyExclusiveState.test.ts
```diff
@@ -0,0 +1,60 @@
...
+      .mockImplementation(() => {})
+
+    expect(() =>
+      renderHook(() => useMutuallyExclusiveState(true, true)),
+    ).toThrowError()
+
+    consoleErrorSpy.mockRestore()
+  })
+  it('allows both values to be false', () => {
+    const hookReturn = renderHook(() => useMutuallyExclusiveState(false, false))
```
jay-hodgson, line 17: Rather the return object has explicitly named values rather than accessing via index.

```
const [
    showValue1,
    setValue1,
    showValue2,
    setValue2,
  ] = useMutuallyExclusiveState(false, false)
```

jay-hodgson (approved):

packages/synapse-react-client/test/utils/hooks/useMutuallyExclusiveState.test.ts
nickgros: Decided to return an array to match the API for `useState`. I think the reason they do that is because 
1. you're always going to use every single return value
2. the variable name will always be overwritten with a better semantic name

PR #347, opened 2023-07-03
PORTALS-2701: Refactor download confirmation in synapse-react-client

nickgros:
Depends on #346 

- DownloadConfirmation component was split into three components:
   - TableQueryDownloadConfirmation - business logic for adding a TableQuery to the download list
   - FolderDownloadConfirmation - business logic for adding a folder to the download list
   - DownloadConfirmationUI - view logic for the component used by the above

This separation of concerns should make it easier to add table-specific row selection logic for adding to the DL list as needed in PORTALS-2701


packages/synapse-react-client/src/umd.index.ts
```diff
@@ -99,7 +99,7 @@ const SynapseComponents = {
   EntityBadgeIcons,
   DownloadCartPage,
   ShowDownloadV2,
-  DownloadConfirmation,
+  DownloadConfirmation: FolderDownloadConfirmation,
```
nickgros, line 102: Maintain the same API for SynapseWebClient

packages/synapse-react-client/src/components/download_list/DownloadConfirmation.tsx
nickgros, line 1: Replaced with TableQueryDownloadConfirmation, FolderDownloadConfirmation, DownloadConfirmationUI

See diagram (aside: ERD semantics does not work very well for UI components)

```mermaid
erDiagram
    TableQueryDownloadConfirmation ||--|| DownloadConfirmationUI : contains
    FolderDownloadConfirmation ||--|| DownloadConfirmationUI : contains
    DownloadConfirmationUI ||--|| DownloadOptions : contains
```

packages/synapse-react-client/src/components/download_list/DownloadConfirmationUI.tsx
```diff
@@ -0,0 +1,198 @@
...
+export type DownloadConfirmationUIProps = {
+  onAddToDownloadCart: () => void
+  fileCount: number
+  fileSize?: number
+  onCancel: () => void
+  isLoadingStats?: boolean
+  isAddingToDownloadCart?: boolean
+}
+
+type UiStateDictionary = {
```
nickgros, line 25: No changes to rendered UI states

nickgros: image

packages/synapse-react-client/src/components/download_list/FolderDownloadConfirmation.tsx
```diff
@@ -0,0 +1,62 @@
...
+} from '../../synapse-queries'
+import { displayToast } from '../ToastMessage'
+import { DownloadConfirmationUI } from './DownloadConfirmationUI'
+import { displayFilesWereAddedToDownloadListSuccess } from './DownloadConfirmationUtils'
+import { EntityType } from '@sage-bionetworks/synapse-types'
+
+export type FolderDownloadConfirmationProps = {
+  folderId: string
+  fnClose: () => void
+}
```
nickgros, line 15: Keep same prop signature as original DownloadConfirmation as this is used in SWC

packages/synapse-react-client/src/components/download_list/FolderDownloadConfirmation.tsx
```diff
@@ -0,0 +1,62 @@
...
+      includeSumFileSizes: true,
+      includeTotalChildCount: true,
+      includeTypes: [EntityType.FILE],
+    })
+
+  const fileCount = entityChildrenData?.totalChildCount ?? 0
+  const fileSizeTotal = entityChildrenData?.sumFileSizesBytes
+
+  const { mutate: addToDownloadList, isLoading: isAddingToDownloadCart } =
+    useAddQueryToDownloadList({
```
nickgros, line 35: Both implementations now use a mutation that will automatically invalidate the correct queries.

packages/synapse-react-client/test/components/download_list/TableQueryDownloadConfirmation.test.tsx
```diff
@@ -0,0 +1,204 @@
...
+      }),
+    )
+
+    mutationMockReturnValue = getUseMutationMock<
+      AddToDownloadListResponse,
+      SynapseClientError,
+      AddToDownloadListRequest
+    >(addFilesToDownloadListResponse)
+
+    mockUseAddQueryToDownloadList.mockReturnValue(mutationMockReturnValue)
```
nickgros, line 116: Starting to think instead of mocking `SynapseClient`, we create reusable utilities to mock our query/cache layer.

Ideally we will also have unit tests for the query/cache layer that validate the right APIs are called, and that the right cache keys are invalidated.

jay-hodgson:
nice mermaid diagram

packages/synapse-react-client/test/components/download_list/TableQueryDownloadConfirmation.test.tsx
jay-hodgson: That works.  At one point I had the idea that all APIs would be mocked, but that was so painful (and potentially contaminating).  This seems like the right layer to mock

jay-hodgson (approved):

PR #349, opened 2023-07-03
PORTALS-2714

nickgros:
- Fix issue where `selectFileColumn` would never be undefined, even if it should


packages/synapse-react-client/src/synapse-queries/entity/useActionsRequiredForTableQuery.ts
```diff
@@ -26,7 +26,9 @@ export function useGetActionsRequiredForTableQuery(
   const queryRequestCopy = useMemo(() => {
     const request = cloneDeep(queryBundleRequest)
     const fileColumnId = fileColumnModelId || getFileColumnModelId(columnModels)
-    request.query.selectFileColumn = Number(fileColumnId)
+    if (fileColumnId) {
+      request.query.selectFileColumn = Number(fileColumnId)
```
nickgros, line 30: `Number(undefined)` returns `NaN` which fails the `!== undefined` check below

PR #350, opened 2023-07-03
SWC-6490 - patch release

nickgros:
- Fix bug where subjectId could be null when creating a data access submission
- Fix issue where ResearchProjectForm would not re-populate when closing and re-opening the modal
- Fix z-index issue where form labels would appear above entries for UserSearchBoxV2 in the data access request flow.

### DIFF
--- a/packages/synapse-react-client/src/components/AccessRequirementList/ManagedACTAccessRequirementRequestFlow/ResearchProjectForm.tsx
+++ b/packages/synapse-react-client/src/components/AccessRequirementList/ManagedACTAccessRequirementRequestFlow/ResearchProjectForm.tsx
@@ -1,4 +1,4 @@
-import React, { useState } from 'react'
+import React, { useEffect, useState } from 'react'
 import {
   ManagedACTAccessRequirement,
   ResearchProject,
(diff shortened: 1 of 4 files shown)


packages/synapse-react-client/src/components/AccessRequirementList/ManagedACTAccessRequirementRequestFlow/ResearchProjectForm.tsx
```diff
@@ -35,28 +35,17 @@ export type ResearchProjectFormProps = {
...
-        if (data?.projectLead) {
-          setProjectLead(data?.projectLead)
-        }
-        if (data?.institution) {
-          setInstitution(data?.institution)
-        }
-        if (data?.intendedDataUseStatement) {
-          setIntendedDataUseStatement(data?.intendedDataUseStatement)
-        }
-      },
```
nickgros, line 59: Fix the following bug:
1. Open the access request modal
2. Step to the research project form, load or fill in data
3. Close the modal
4. Open again and navigate to the same form. 

The data does not re-populate, because onSuccess is not called (this type of bug is exactly why this API is going to be removed in react query v5)


Instead use an effect that will always populate the form when appropriate.

packages/synapse-react-client/src/components/UserSearchBoxV2.tsx
```diff
@@ -172,6 +172,7 @@ const UserSearchBoxV2: React.FC<UserSearchBoxProps> = props => {
             gridArea: '1 / 2 / 4 / 4 !important',
           },
         }),
+        menu: base => ({ ...base, zIndex: 2 }),
```
nickgros, line 175: Fix minor issue where form labels (with z-index of 1) would appear above the menu options.

packages/synapse-react-client/src/synapse-queries/dataaccess/useDataAccessSubmission.ts
```diff
@@ -114,8 +114,18 @@ export function useSubmitDataAccessRequest(
...
+    async ({ request, accessRequirementId }): Promise<ACTSubmissionStatus> => {
+      // The subjectId or subjectType may not be known, so pick an arbitrary one off of the AR (SWC-6490)
+      if (request.subjectId == null || request.subjectType == null) {
+        const { subjects } = await SynapseClient.getSubjects(
+          accessToken,
+          accessRequirementId,
+        )
+        request.subjectId = subjects[0].id
+        request.subjectType = subjects[0].type
+      }
```
nickgros, line 126: We were doing this in SWC but didn't port this behavior over to synapse-react-client:

https://github.com/Sage-Bionetworks/SynapseWebClient/blob/e7910e64d26e19a7ae68064822c26d7fee37453f/src/main/java/org/sagebionetworks/web/server/servlet/DataAccessClientImpl.java#L83

.github/workflows/build-and-deploy-app.yml
nickgros, line 1: I guess the code formatter ran on these files when I merged

packages/synapse-react-client/src/synapse-client/SynapseClient.ts
```diff
@@ -4886,3 +4887,23 @@ export const getDockerTag = (
...
   )
 }
+
+// https://rest-docs.synapse.org/rest/GET/accessRequirement/requirementId/subjects.html
+export function getSubjects(
+  accessToken: string | undefined,
+  requirementId: string,
+  nextPageToken?: string,
+) {
+  const params = new URLSearchParams()
```
jay-hodgson, line 4897: nice use of URLSearchParams.toString to do the right thing

jay-hodgson (approved):

PR #355, opened 2023-07-07
[bug fix]: set version number as optional in DownloadListItemResult

linglp:
Related to: https://sagebionetworks.jira.com/browse/SWC-6332

## Changelog

* Updated type `DownloadListItemResult` to allow versionNumber to be optional
* In `DownloadListTable.tsx`, call entityVersionNumber.toString() only when entityVersionNumber is not null. If entityVersionNumber is null, then exclude it from prop.

### DIFF
--- a/packages/synapse-react-client/src/components/DownloadCart/DownloadListTable.tsx
+++ b/packages/synapse-react-client/src/components/DownloadCart/DownloadListTable.tsx
@@ -322,12 +322,7 @@ export default function DownloadListTable(props: DownloadListTableProps) {
                               associatedObjectType={
                                 FileHandleAssociateType.FileEntity
                               }
-                              {...(item.versionNumber !== null
-                                ? {
-                                    entityVersionNumber:
-                                      item.versionNumber.toString(),
-                                  }
-                                : {})}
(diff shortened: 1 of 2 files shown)


packages/synapse-react-client/src/components/DownloadCart/DownloadListTable.tsx
```diff
@@ -322,7 +322,12 @@ export default function DownloadListTable(props: DownloadListTableProps) {
                               associatedObjectType={
                                 FileHandleAssociateType.FileEntity
                               }
-                              entityVersionNumber={item.versionNumber.toString()}
+                              {...(item.versionNumber !== null
+                                ? {
+                                    entityVersionNumber:
+                                      item.versionNumber.toString(),
+                                  }
+                                : {})}
```
nickgros, line 330: Can we just do something like this?
```suggestion
                              entityVersionNumber={item.versionNumber?.toString()}
```

linglp: Thanks for the suggestion. Yes, we could definitely do that.

nickgros (approved):
LGTM! Thanks @linglp !

PR #354, opened 2023-07-06
dependency maintenance

nickgros:
Ran `pnpm update --recursive` and followed up with a few manual version changes to achieve the following
- Fix all issues that appear in pnpm audit that have an upgrade path
- Fix all but one peer dependency issues
- Remove project-specific `resolutions` fields (these had no effect on dependency resolution)
- Small changes to account for package updates

Additionally, update our SBOM upload workflow to continue getting vulnerability alerts (still needed while we wait on https://github.com/dependabot/dependabot-core/issues/7434).

PR #361, opened 2023-07-11
Bump dependencies, take 2

nickgros:
Re-include the changes in #354 plus

- move react-switch to "dependencies"
- add "prop-types" as a dependency
- upgrade SVGR plugins (must now use @svgr/plugin-jsx)
- simplify vite.config.ts files


packages/synapse-react-client/package.json
```diff
@@ -78,6 +76,7 @@
     "mui-one-time-password-input": "1.1.0",
     "plotly.js-basic-dist": "^2.24.3",
     "pluralize": "^8.0.0",
+    "prop-types": "^15.8.1",
```
nickgros, line 79: For some reason, `tsup` isn't bundling/properly writing a dynamic import for the `prop-types` dependency, which is imported via `react-switch` (but also many other dependencies?)

This issue resolves if we include it in our dependencies.

packages/synapse-react-client/package.json
```diff
@@ -95,7 +94,7 @@
     "react-router-dom": "^5.3.4",
     "react-select": "^5.7.3",
     "react-sizeme": "^3.0.2",
-    "react-test-renderer": "^18.2.0",
+    "react-switch": "^7.0.0",
```
nickgros, line 97: move from devDependencies

packages/synapse-react-client/package.json
```diff
@@ -95,7 +94,7 @@
     "react-router-dom": "^5.3.4",
     "react-select": "^5.7.3",
     "react-sizeme": "^3.0.2",
-    "react-test-renderer": "^18.2.0",
```
nickgros, line 101: remove from dependencies (it's already in devDependencies)

packages/synapse-react-client/esbuild.config.mjs
```diff
@@ -45,6 +45,7 @@ const esBuildOptions = {
       loadPaths: ['../../node_modules', 'node_modules'],
     }),
     svgrPlugin({
+      plugins: ['@svgr/plugin-jsx'],
```
nickgros, line 48: Upgraded svgr plugins use new versions of svgr which require specifying the plugins we use. We just use the `jsx` plugin.

jay-hodgson (approved):
Additional changes also look ok.  Thank you for tracking this down!

PR #362, opened 2023-07-11
PORTALS-2697

jay-hodgson:
- Add custom commands that do not reference the row selection.  I think this will be necessary for the 'Make cohort from files' command in the Cohort Builder files perspective.
- Allow the two QueryWrapperPlotNav components that point to the same entity to load different additional filters (via local storage).  I think will be the case for the Virtual Table that drives the two perspectives of the cohort builder.


packages/synapse-react-client/src/components/SynapseTable/TopLevelControls/TopLevelControls.tsx
```diff
@@ -174,6 +199,29 @@ const TopLevelControls = (props: TopLevelControlsProps) => {
...
+                      })
+                    }
+                    startIcon={customControl.icon}
+                  >
+                    {customControl.buttonText}
+                  </Button>
+                  <Divider orientation="vertical" variant="middle" flexItem />
+                </>
+              )
+            })}
```
nickgros, line 224: Nit, need a `key` for mapping jsx
```suggestion
            topLevelCustomControls.map((customControl, index) => {
              return (
                <React.Fragment key={index}>
                  <Button
                    variant="text"
                    disabled={!numberOfResultsToInvokeAction}
                    onClick={() =>
                      customControl.onClick({
                        data,
                        selectedRows,
                        refresh,
                        request: getLastQueryRequest(),
                      })
                    }
                    startIcon={customControl.icon}
                  >
                    {customControl.buttonText}
                  </Button>
                  <Divider orientation="vertical" variant="middle" flexItem />
                </React.Fragment>
              )
            })}
```

packages/synapse-react-client/src/components/SynapseTable/TopLevelControls/TopLevelControls.tsx
```diff
@@ -64,13 +72,25 @@ const TopLevelControls = (props: TopLevelControlsProps) => {
...
   } = props
 
+  const rowSelectionCustomControls = getFilteredCustomControls(
+    true,
+    customControls,
+  )
+  const topLevelCustomControls = getFilteredCustomControls(
+    false,
+    customControls,
+  )
```
nickgros, line 82: I probably would have used lodash [partition](https://lodash.com/docs/4.17.15#partition), but having a custom utility like you have here is likely more readable.

packages/synapse-react-client/stories/QueryWrapperPlotNav.stories.tsx
```diff
@@ -236,12 +278,26 @@ export const TableRowSelectionWithCustomCommand: Story = {
...
+        },
+        isRowSelectionSupported: true,
+      },
+      {
+        buttonText: 'All Results Custom Command',
         onClick: event => {
-          handleCustomCommandClick(event)
+          handleAllDataCustomCommandClick(event)
         },
+        isRowSelectionSupported: false,
```
nickgros, line 300: Super cool!!

packages/synapse-react-client/stories/QueryWrapperPlotNav.stories.tsx
```diff
@@ -227,6 +231,44 @@ const handleCustomCommandClick = async (event: CustomControlCallbackData) => {
...
+    values: ids,
+  }
+  localStorage.setItem(
+    QUERY_FILTERS_LOCAL_STORAGE_KEY('syn51186974-selectedfiles'),
+    JSON.stringify([localStorageFilter]),
+  )
+  console.log(
+    'Local Storage value set, refresh table to see additionalFilter QueryFilter being utilized',
+  )
+  // TODO: PORTALS-2682: event.refresh() should refresh the data but it currently doesn't
```
nickgros, line 268: I see now, `refresh` just sends QueryWrapper the current query, so of course we get a cache hit and just see the same results. We really want this command to apply localStorage filters (likely by re-mounting the component).

nickgros (approved):
Looks great

packages/synapse-react-client/src/components/SynapseTable/TopLevelControls/TopLevelControls.tsx
jay-hodgson: Changed to use partition (less code, and less code to test)

packages/synapse-react-client/src/components/SynapseTable/TopLevelControls/TopLevelControls.tsx
jay-hodgson: Added key

packages/synapse-react-client/stories/QueryWrapperPlotNav.stories.tsx
jay-hodgson: Added a way for the toplevelcontrols to remount the query wrapper (increments the key).  This does the trick!

PR #368, opened 2023-07-14
PORTALS-2726, PORTALS-2729

jay-hodgson:


apps/portals/src/configurations/adknowledgeportal/synapseConfigs/data.ts
```diff
@@ -9,6 +9,7 @@ const data: SynapseConfig = {
     rgbIndex,
     name: 'Data',
     visibleColumnCount: 10,
+    isRowSelectionVisible: true,
```
jay-hodgson, line 12: ADKP also uses a File View, they can use the latest enhancement (made for the ELITE portal)

apps/portals/src/configurations/elportal/synapseConfigs/studies.ts
```diff
@@ -26,9 +26,9 @@ export const studyCardConfiguration: CardConfiguration = {
   labelLinkConfig: [
     {
       isMarkdown: false,
-      matchColumnName: 'grantNumber',
-      URLColumnName: 'Grant Number',
```
jay-hodgson, line 30: Fix bug (URL Column Name was incorrect)

packages/synapse-react-client/src/components/SynapseTable/SynapseTable.tsx
```diff
@@ -440,7 +440,9 @@ export class SynapseTable extends React.Component<
       this.allRowsHaveId()
 
     const isShowingAddToV2DownloadListColumn: boolean = !!(
-      rowsAreDownloadable && !this.props.hideDownload
+      rowsAreDownloadable &&
+      !this.props.hideDownload &&
+      !isRowSelectionVisible
```
jay-hodgson, line 445: It's redundant to show the Add To Download List table column if row selection is available (since it provides the same functionality in the RowSelectionControls if rows are downloadable)

nickgros (approved):

PR #371, opened 2023-07-15
[bug fix] Showed version number when it is available

linglp:
Related to: https://sagebionetworks.jira.com/browse/SWC-6332

## Changelog
Set conditional statement so that version number only gets shown in download cart when it is available

### DIFF
--- a/packages/synapse-react-client/src/components/DownloadCart/DownloadListTable.tsx
+++ b/packages/synapse-react-client/src/components/DownloadCart/DownloadListTable.tsx
@@ -257,7 +257,6 @@ export default function DownloadListTable(props: DownloadListTableProps) {
             <tbody>
               {allRows.map((item: DownloadListItemResult) => {
                 if (item) {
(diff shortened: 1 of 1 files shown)


packages/synapse-react-client/src/components/DownloadCart/DownloadListTable.tsx
```diff
@@ -257,6 +257,7 @@ export default function DownloadListTable(props: DownloadListTableProps) {
             <tbody>
               {allRows.map((item: DownloadListItemResult) => {
                 if (item) {
+                  item.versionNumber = null
```
jay-hodgson, line 260: Did you intend to check this in (always setting the item version number to null)?

linglp: Sorry! This is a line that I put to make sure the solution work. Let me remove it

jay-hodgson (approved):

PR #374, opened 2023-07-17
PORTALS-2715 - confirmation dialog on selection change

nickgros:
- show confirmation dialog if attempting to change filter criteria when making a selection. this does not affect changes the current page or sort
- refactor to remove InfiniteQueryWrapper, now can use `QueryWrapper` with `isInfinite` prop
- refactor to move row selection into QueryContext from QueryVisualizationContext
- add stories for EnumFacetFilter and RangeFacetFilter

known issue: adding a new filter and cancelling the change will not reset the form to the previous query state. To fix this I think we may need to track the "next"/"uncommitted" query in state (which is what the facet filter forms show), and revert this object when the change is cancelled. That itself is a fairly meaty refactor, so suggest following up in a separate ticket and PR.

https://github.com/Sage-Bionetworks/synapse-web-monorepo/assets/17580037/848d5111-a835-4abe-bf06-d0002b909d00


packages/synapse-react-client/src/components/InfiniteQueryWrapper.tsx
nickgros, line 1: Functionality folded into QueryWrapper, goodbye duplicate code

packages/synapse-react-client/src/components/QueryContext/QueryContext.tsx
```diff
@@ -53,7 +54,7 @@ export type QueryContextType<
   /** Returns a deep clone of the initial query bundle request */
   getInitQueryRequest: () => QueryBundleRequest
   /** Updates the current query with the passed request */
-  executeQueryRequest: (param: QueryBundleRequest) => void
+  executeQueryRequest: (param: React.SetStateAction<QueryBundleRequest>) => void
```
nickgros, line 57: Allow passing an updater function (this just uses a `useState` setter, so this was always supported)

packages/synapse-react-client/src/components/QueryContext/QueryContext.tsx
```diff
@@ -81,6 +82,18 @@ export type QueryContextType<
...
+  /** The collection of selected rows */
+  selectedRows: Row[]
+  /** State updater for `selectedRows`  */
+  setSelectedRows: React.Dispatch<React.SetStateAction<Row[]>>
+  /** The set of columns that defines a uniqueness constraint on the table for the purposes of filtering based on row selection.
+   * Note that Synapse tables have no internal concept of a primary key.
+   */
+  rowSelectionPrimaryKey?: string[]
+  /** Whether the user has selected any rows */
+  hasSelectedRows: boolean
```
nickgros, line 96: Move row selection to `QueryContext` since `QueryWrapper` needs to know about row selection in order to conditionally block changing the query.

packages/synapse-react-client/src/components/QueryContext/QueryContext.tsx
```diff
@@ -104,7 +117,7 @@ export type InfiniteQueryContextType<
   /** Returns true when loading a new page of query results */
   isLoadingNewPage: boolean
   /** Whether the query result bundle has a next page */
-  hasNextPage: boolean
+  hasNextPage?: boolean
```
nickgros, line 120: Match the type of react-query's `hasNextPage`, which is `boolean | undefined`.

packages/synapse-react-client/src/components/QueryWrapper/QueryWrapper.tsx
```diff
@@ -47,56 +49,52 @@ export function QueryWrapper(props: QueryWrapperProps) {
...
+    isInfinite = false,
   } = props
 
-  const [currentAsyncStatus, setCurrentAsyncStatus] = useState<
-    AsynchronousJobStatus<QueryBundleRequest, QueryResultBundle> | undefined
-  >(undefined)
+  // Must store requireConfirmationOnChange in state to break dependency cycle between useImmutableTableQuery and useTableRowsSelection
+  // TODO: consider combining these two hooks, or decoupling them in some other way
+  const [requireConfirmationOnChange, setRequireConfirmationOnChange] =
+    useState(false)
```
nickgros, line 60: This works, so I'm intentionally sticking with this for now. I don't have a clear idea for how this should be designed to fix this (ideas are welcome).

packages/synapse-react-client/src/components/QueryWrapper/QueryWrapper.tsx
```diff
@@ -47,56 +49,52 @@ export function QueryWrapper(props: QueryWrapperProps) {
...
-      // We use `keepPreviousData` because we don't want to clear out the current data when the query is modified via the UI
-      keepPreviousData: true,
-    },
-    setCurrentAsyncStatus,
-  )
-
-  const data = asyncJobStatus?.responseBody
+  const queryWrapperData = useQueryWrapperData(lastQueryRequest, isInfinite)
+  const { data, isLoadingNewBundle, error, currentAsyncStatus } =
+    queryWrapperData
```
nickgros, line 91: `useQueryWrapperData` abstracts whether we're fetching infinite or paginated data.

packages/synapse-react-client/src/components/QueryWrapper/QueryWrapper.tsx
```diff
@@ -47,56 +49,52 @@ export function QueryWrapper(props: QueryWrapperProps) {
...
 
-  // Indicate if we're fetching data for the first time (queryIsLoading) or if we're fetching data for a brand-new query (newQueryIsFetching)
-  const isLoadingNewBundle = queryIsLoading || newQueryIsFetching
+  const paginationControls = useQueryWrapperPaginationControls({
+    immutableTableQueryResult,
+    queryWrapperData,
+    isInfinite,
+  })
```
nickgros, line 97: When we have infinite data, we have a custom layer of pagination functions that sit on top of the useInfiniteQuery API.

If we're using infinite data, `useQueryWrapperPaginationControls` creates these functions.

packages/synapse-react-client/src/components/QueryWrapper/QueryWrapper.tsx
```diff
@@ -132,11 +130,25 @@ export function QueryWrapper(props: QueryWrapperProps) {
...
+    rowSelectionPrimaryKey,
+    hasSelectedRows,
+    selectedRows,
+    setSelectedRows,
+  } = useTableRowSelection({
+    entity: entity,
+    columnModels: data?.columnModels,
+    isRowSelectionVisible: isRowSelectionVisibleFromProps,
+    rowSelectionPrimaryKey: rowSelectionPrimaryKeyFromProps,
+  })
```
nickgros, line 144: Capture all of the selection logic into a custom hook

packages/synapse-react-client/src/components/QueryWrapper/QueryWrapper.tsx
```diff
@@ -132,11 +130,25 @@ export function QueryWrapper(props: QueryWrapperProps) {
...
+
+  useEffect(() => {
+    setRequireConfirmationOnChange(hasSelectedRows)
+  }, [hasSelectedRows])
```
nickgros, line 148: Effect used to break the dependency cycle mentioned earlier.

packages/synapse-react-client/src/components/QueryWrapper/QueryWrapper.tsx
```diff
@@ -155,13 +166,34 @@ export function QueryWrapper(props: QueryWrapperProps) {
...
+          'Changing the current query will cause your current selection to be lost. Are you sure you want to proceed?'
+        }
+        confirmButtonText={'Clear Selection and Update Query'}
+        onConfirm={() => {
+          onConfirmChange()
+        }}
+        onCancel={() => {
+          onCancelChange()
+        }}
+      />
```
nickgros, line 196: New feature dialog

packages/synapse-react-client/src/components/QueryWrapper/useQueryWrapperData.ts
nickgros, line 1: This hook basically captures logic for fetching table query data that was in both `QueryWrapper` and `InfiniteQueryWrapper`

packages/synapse-react-client/src/components/QueryWrapper/useQueryWrapperData.ts
```diff
@@ -0,0 +1,167 @@
...
+
+export type UseQueryWrapperDataReturn<TIsInfiniteData extends boolean> = {
+  data?: QueryResultBundle
+  currentAsyncStatus?: AsynchronousJobStatus<
+    QueryBundleRequest,
+    QueryResultBundle
+  >
+  error: SynapseClientError | null
+  isLoadingNewBundle: boolean
+} & (TIsInfiniteData extends true ? InfiniteDataQueryPageControls : object)
```
nickgros, line 40: Taking advantage of conditional types on generics; if we have infinite data, the return object will also include the `useInfiniteQuery` pagination controls.

packages/synapse-react-client/src/components/QueryWrapper/useQueryWrapperData.ts
```diff
@@ -0,0 +1,167 @@
...
+        } as QueryResultBundle
+      }
+      return infiniteData?.pages[infiniteDataCurrentPage].responseBody
+    }
+  }, [
+    infiniteData,
+    infiniteDataCurrentPage,
+    isInfinite,
+    paginatedData?.responseBody,
+  ])
```
nickgros, line 133: Here's where we decide whether to use the paginated data or the infinite data.

packages/synapse-react-client/src/components/QueryWrapper/useQueryWrapperData.ts
```diff
@@ -0,0 +1,167 @@
...
+      ? {
+          hasNextPage: infiniteDataHasNextPage,
+          fetchPreviousPage: infiniteDataFetchPreviousPage,
+          fetchNextPage: infiniteDataFetchNextPage,
+          isFetchingNextPage: infiniteDataIsFetchingNextPage,
+          currentPage: infiniteDataCurrentPage,
+          setCurrentPage: setInfiniteDataCurrentPage,
+        }
+      : {}
+  ) as T extends true ? InfiniteDataQueryPageControls : Record<string, never>
```
nickgros, line 158:
```suggestion
  ) as T extends true ? InfiniteDataQueryPageControls : object
```

packages/synapse-react-client/src/components/QueryWrapper/useQueryWrapperPaginationControls.ts
```diff
@@ -0,0 +1,111 @@
...
+  /** Whether the query result bundle has a next page */
+  hasNextPage?: boolean
+  /** Invoke this method to fetch and append the next page of rows to the data  */
+  appendNextPageToResults: () => Promise<void>
+  /** Invoke to fetch and update the data with the next page of query results */
+  goToNextPage: () => Promise<void>
+  /** Whether the query result bundle has a previous page */
+  hasPreviousPage: boolean
+  /** Invoke to fetch and update the data with the previous page of query results */
+  goToPreviousPage: () => Promise<void>
```
nickgros, line 27: InfiniteQueryWrapper provided these custom functions to manage pagination, so we need to continue to provide them to limit the scope of this change.

The hook in this file will provide these controls when we have an infinite data query.

packages/synapse-react-client/src/components/useImmutableTableQuery.ts
```diff
@@ -161,6 +172,42 @@ export default function useImmutableTableQuery(
...
   )
 
+  const setQueryOrPromptConfirmation: ImmutableTableQueryResult['setQuery'] =
+    useCallback(
+      (queryRequest: React.SetStateAction<QueryBundleRequest>) => {
+        const nextQueryRequest =
+          typeof queryRequest === 'function'
+            ? queryRequest(getLastQueryRequest())
+            : queryRequest
+        // Check if we need to confirm the change, and eventually call _setQuery
```
nickgros, line 182:
```suggestion
        // Check if we need to confirm the change, and eventually call setQuery
```

packages/synapse-react-client/src/components/QueryWrapper/QueryWrapper.tsx
nickgros, line 1: (still) missing tests for infinite query behavior (we never had any good tests to begin with)

packages/synapse-react-client/src/utils/functions/deepLinkingUtils.ts
```diff
@@ -93,24 +93,18 @@ export function updateUrlWithNewSearchParam(
...
     const query = JSON.parse(searchParamValue) as Query
     if (query.sql) {
       initQueryRequest = {
         concreteType: 'org.sagebionetworks.repo.model.table.QueryBundleRequest',
-        partMask:
-          SynapseConstants.BUNDLE_MASK_QUERY_COLUMN_MODELS |
-          SynapseConstants.BUNDLE_MASK_QUERY_FACETS |
-          SynapseConstants.BUNDLE_MASK_QUERY_SELECT_COLUMNS |
-          SynapseConstants.BUNDLE_MASK_QUERY_RESULTS |
-          SynapseConstants.BUNDLE_MASK_QUERY_COUNT,
```
nickgros, line 113: This was probably a bug waiting to happen - when we populate the query from a search parameter, we shouldn't override the `partMask`. Now, this partial request gets merged in with the value passed from props. The `partMask` used will be specified by the component, which asks for exactly what it needs.

packages/synapse-react-client/test/containers/QueryWrapper.test.tsx
```diff
@@ -64,191 +77,354 @@ const initialQueryRequest: QueryBundleRequest = {
   },
 }
 
-describe('basic functionality', () => {
-  it('renders without crashing', async () => {
-    renderComponent({ initQueryRequest: initialQueryRequest })
-    await screen.findByText(renderedTextConfirmation)
-  })
+describe('QueryWrapper', () => {
```
nickgros, line 80: Wrapped in a describe to easily run all at once in IDE

packages/synapse-react-client/test/containers/QueryWrapper.test.tsx
```diff
@@ -64,191 +77,354 @@ const initialQueryRequest: QueryBundleRequest = {
...
+        ),
+      ).toBeDefined()
+    })
   })
 
-  it('removeLockedColumnData should not remove any data if locked facet value is not set', async () => {
-    renderComponent({
-      initQueryRequest: initialQueryRequest,
-      lockedColumn: noLockedColumn,
+  describe('query change when rows are selected', () => {
```
nickgros, line 307: Tests for new functionality in this block

packages/synapse-react-client/test/containers/QueryWrapper.test.tsx
nickgros, line 1: Sorry, this is a horrible diff

packages/synapse-react-client/test/containers/QueryWrapper.test.tsx
nickgros: The existing tests are essentially unchanged, I would recommend just reviewing the new tests.

packages/synapse-react-client/src/components/QueryWrapper/QueryWrapper.tsx
```diff
@@ -155,13 +166,35 @@ export function QueryWrapper(props: QueryWrapperProps) {
...
+        }
+        confirmButtonText={'Clear Selection and Update Query'}
+        onConfirm={() => {
+          setSelectedRows([])
+          onConfirmChange()
+        }}
+        onCancel={() => {
+          onCancelChange()
+        }}
+      />
```
nickgros, line 197: New feature dialog

jay-hodgson:
> known issue: adding a new filter and cancelling the change will not reset the form to the previous query state. 

What happens if you cancel the confirmation dialog today?  Will it update the query and clear selection (and if so, is there any point in showing this dialog until this issue has been fixed)?

packages/synapse-react-client/src/components/QueryWrapper/QueryWrapper.tsx
jay-hodgson: I'm having trouble understanding these dependencies - Could we set up some time so you can help me draw a picture?

nickgros:
> > known issue: adding a new filter and cancelling the change will not reset the form to the previous query state.
> 
> What happens if you cancel the confirmation dialog today? Will it update the query and clear selection (and if so, is there any point in showing this dialog until this issue has been fixed)?

The query (and data) will not change, but the form will still indicate that the query was changed. This is only an issue with the facet filter controls, since they maintain their own state that synchronizes (and triggers the confirmation) after a delay (for the EnumFacetFilter) or after clicking apply (for the RangFacetFilter.

jay-hodgson (approved):

PR #379, opened 2023-07-20
SWC-6506 - Don't fetch duplicate profile pictures

nickgros:
Use react-query cache to dedupe requests file content using presigned URLs, such as profile pictures.

### DIFF
--- a/packages/synapse-react-client/src/utils/hooks/usePreFetchResource.ts
+++ b/packages/synapse-react-client/src/utils/hooks/usePreFetchResource.ts
@@ -1,4 +1,5 @@
 import { useEffect, useState } from 'react'
+import { useQuery } from 'react-query'
 
 export async function fetchBlob(url: string): Promise<Blob> {
   const response = await fetch(url)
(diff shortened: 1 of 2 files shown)


packages/synapse-react-client/src/utils/hooks/usePreFetchResource.ts
```diff
@@ -18,35 +19,36 @@ export function releaseResourceUrl(resourceUrl: string) {
...
-    }
-  }, [preSignedURL])
-
-  return url
+  // While the retrieved data is cached by react-query, the object created by URL.createObjectURL is not. More specifically,
+  // the stored object is different for each instance of this hook, so when one instance tears down and deletes its object,
+  // other instances of the hook are not affected.
+  // This uses more memory than necessary in cases where the same object is fetched multiple times, but it's a tradeoff
+  // for simple cache and resource management.
+  return useCreateUrlForData(blob)
```
nickgros, line 45: The old behavior was that with _n_ instances of this hook with the same URL, you would send _n_ HTTP requests and have _n_ objects stored in memory.

With react-query we now have 1 HTTP request, but still storing _n_ things in memory. I don't think the storage issue is causing any problems today. I also think that solution would add complexity in a way that's hard to get right, so I'm intentionally keeping unoptimal behavior to avoid introducing a new error.

We'd probably need to revisit this if we start downloading files larger than profile pictures with this hook.

jay-hodgson (approved):

jay-hodgson:
This is great.  I want this deployed to Synapse.org yesterday! :)

PR #380, opened 2023-07-20
SWC-6509

jay-hodgson:


jay-hodgson:
https://github.com/Sage-Bionetworks/synapse-web-monorepo/assets/1864447/470fb986-6b0b-4c59-95d4-a554e93406c7

packages/synapse-react-client/src/assets/icons/CommunicateFindings.tsx
```diff
@@ -0,0 +1,20 @@
+import React from 'react'
```
jay-hodgson, line 1: There were issues exporting 2 of the icons from figma, so Stockard did the export (which resulted in some styling attributes that React svg parsing did not approve).  I kept those 2 icons as svgs, and wrapped the rest

packages/synapse-react-client/src/components/SynapseHomepage.tsx
```diff
@@ -132,132 +151,182 @@ export const SynapseHomepage: React.FunctionComponent<SynapseHomepageProps> = ({
...
-            >
-              {accessToken ? 'View Your Dashboard' : 'Get Started Now'}
-            </Button>
-          </div>
-        </div>
+        <Typography
+          variant="h2"
+          sx={{ textAlign: 'center', paddingBottom: '30px' }}
+        >
+          Synapse Features
```
jay-hodgson, line 158: New section, "Synapse Features"

packages/synapse-react-client/src/components/SynapseHomepage.tsx
```diff
@@ -132,132 +151,182 @@ export const SynapseHomepage: React.FunctionComponent<SynapseHomepageProps> = ({
...
-            </div>
-            <Button
-              variant="contained"
-              color="primary"
-              sx={buttonSx}
-              href={accessToken ? DASHBOARD_LINK : REGISTRATION_LINK}
-            >
-              {accessToken ? 'View Your Dashboard' : 'Get Started Now'}
-            </Button>
+        <div className="SynapseHomepage__Section__Centered">
```
jay-hodgson, line 276: This is a terrible diff.  Best to look at the source file, specifically at the 2 new sections (no other changes were made)

packages/synapse-react-client/src/components/SynapseHomepage.tsx
```diff
@@ -363,67 +434,265 @@ export const SynapseHomepage: React.FunctionComponent<SynapseHomepageProps> = ({
...
-        <div className="SynapseHomepage__Section__Centered">
-          <h2>Powered By Synapse</h2>
-          <p>
-            Our knowledge portals are community-specific interfaces that enable
-            researchers to explore and share data, analyses, and tools.
+          <Typography
+            variant="h2"
+            sx={{ textAlign: 'center', paddingBottom: '10px' }}
+          >
+            Sage’s Data Platform Offerings
```
jay-hodgson, line 441: New section, "Sage’s Data Platform Offerings"

packages/synapse-react-client/src/components/SynapseHomepage.tsx
```diff
@@ -363,67 +434,265 @@ export const SynapseHomepage: React.FunctionComponent<SynapseHomepageProps> = ({
...
+            Contact us for more information
+          </Button>
+          <Link
+            target="_blank"
+            href="https://help.synapse.org/docs/FAQ.2047967233.html"
+            sx={{ display: 'block', marginTop: '25px', marginBottom: '50px' }}
+          >
+            View Frequently Asked Questions
+          </Link>
+          <Box sx={{ display: 'grid', gridTemplateColumns: '40% 20% 20% 20%' }}>
```
jay-hodgson, line 461: Header is a bit verbose, but the items in each Accordion reuse the same component (SynapsePriceFeatureRow)

packages/synapse-react-client/src/components/SynapseHomepage.tsx
```diff
@@ -363,67 +434,265 @@ export const SynapseHomepage: React.FunctionComponent<SynapseHomepageProps> = ({
...
+          <Typography
+            className="text-align-left"
+            sx={{
+              fontWeight: 300,
+              fontSize: '12px',
+              lineHeight: '16px',
+              paddingTop: '15px',
+            }}
+          >
+            Sage can create a customized quote for you if, for example, you need
```
jay-hodgson, line 661: Still need to figure out these footnote references.  There are asterisks all over the content, but they don't seem to be linked up properly to footnote markers

packages/synapse-react-client/src/components/SynapsePriceFeatureRow.tsx
```diff
@@ -0,0 +1,70 @@
+import { Box, Typography } from '@mui/material'
+import React from 'react'
+import { SynapsePriceTableCell } from './SynapsePriceTableCell'
+import { CheckTwoTone } from '@mui/icons-material'
+
+export type SynapsePriceFeatureRowProps = {
+  featureDescription: string
+  standardPlanValue?: boolean | string
```
jay-hodgson, line 8: Will render nothing if undefined, a checkmark if true (bug, it's currently if the type is boolean!), or the text if it's a string

packages/synapse-react-client/src/style/components/_synapse-homepage.scss
```diff
@@ -9,6 +9,9 @@
     margin-top: 0;
     margin-bottom: 1rem;
   }
+  p.no-margin {
```
jay-hodgson, line 12: I was unable to work around some of these scss styles with mui inline styles, so I needed to add a couple of classes

packages/synapse-react-client/src/components/SynapseHomepage.tsx
jay-hodgson: Footnote references linked, and reviewed/approved by Kevin

packages/synapse-react-client/src/components/SynapseHomepage.tsx
```diff
@@ -363,67 +434,271 @@ export const SynapseHomepage: React.FunctionComponent<SynapseHomepageProps> = ({
...
+          </Button>
+          <Link
+            target="_blank"
+            href="https://help.synapse.org/docs/FAQ.2047967233.html"
+            sx={{ display: 'block', marginTop: '25px', marginBottom: '50px' }}
+          >
+            View Frequently Asked Questions
+          </Link>
+          <Box sx={{ display: 'grid', gridTemplateColumns: '40% 20% 20% 20%' }}>
+            <p></p>
```
nickgros, line 462: Is this just a placeholder so everything else just slots nicely into the grid? If so, add simple comment (and maybe change it to a div?)

packages/synapse-react-client/src/components/SynapseHomepage.tsx
```diff
@@ -363,67 +434,271 @@ export const SynapseHomepage: React.FunctionComponent<SynapseHomepageProps> = ({
...
+            Contact us for more information
+          </Button>
+          <Link
+            target="_blank"
+            href="https://help.synapse.org/docs/FAQ.2047967233.html"
+            sx={{ display: 'block', marginTop: '25px', marginBottom: '50px' }}
+          >
+            View Frequently Asked Questions
+          </Link>
+          <Box sx={{ display: 'grid', gridTemplateColumns: '40% 20% 20% 20%' }}>
```
nickgros, line 461: Same `gridTemplateColumns` value as in the SynapsePriceFeatureRow, so I'd suggest either
* define a constant value for gridTemplateColumns in that file and reuse it here, or
* create a custom styled box component and reuse it in both places

packages/synapse-react-client/src/components/SynapseHomepage.tsx
```diff
@@ -363,67 +434,271 @@ export const SynapseHomepage: React.FunctionComponent<SynapseHomepageProps> = ({
...
+            <p></p>
+            <Box
+              sx={{
+                display: 'grid',
+                gridTemplateRows: '65px auto',
+                justifyContent: 'stretch',
+                alignItems: 'stretch',
+              }}
+            >
+              <SynapsePriceTableCell sx={{ backgroundColor: '#b6cedf' }}>
```
nickgros, line 471: Could consider adding aria roles for table headers here. Not to be too nitpicky but I feel like we should probably be particular about accessibility and content discovery on our homepage.

packages/synapse-react-client/src/components/SynapseHomepage.tsx
```diff
@@ -363,67 +434,271 @@ export const SynapseHomepage: React.FunctionComponent<SynapseHomepageProps> = ({
...
+                </Typography>
+              </SynapsePriceTableCell>
+              <SynapsePriceTableCell sx={{ backgroundColor: '#d2e6d5' }}>
+                <Typography variant="body2" sx={{ fontWeight: 700 }}>
+                  Contact Sage**
+                </Typography>
+              </SynapsePriceTableCell>
+            </Box>
+          </Box>
+          <Accordion expanded={true}>
```
nickgros, line 548: I appreciate that the `expanded` state of these Accordion sections are not mutually exclusive...I think that's the default behavior in an old version of Bootstrap, and it drives me nuts when I can't open more than one at one time on other websites. It might just be me who is bothered by that 🙂

packages/synapse-react-client/src/components/SynapseHomepage.tsx
```diff
@@ -363,67 +434,271 @@ export const SynapseHomepage: React.FunctionComponent<SynapseHomepageProps> = ({
...
-      <div className="SynapseHomepage__Section OffWhiteBackground">
+      {/* /**
+       * 
+Data access management
+
+Seamless data access integration with compute environments 
+
+Customized data portal 
+
+Contract length
```
nickgros, line 692: Is this future content that is coming later, or should be deleted?

packages/synapse-react-client/src/components/SynapsePriceFeatureRow.tsx
nickgros: That's a pretty fun props signature, I like it

packages/synapse-react-client/src/components/SynapsePriceFeatureRow.tsx
```diff
@@ -0,0 +1,93 @@
...
+  managedPlanValue?: boolean | string
+  dataCoordinationPlanValue?: boolean | string
+}
+
+export function getPriceFeatureCellContent(value?: boolean | string) {
```
nickgros, line 13: Prefer naming like a component using since this returns a ReactNode in all cases
```suggestion
export function PriceFeatureCellContent(props: {value?: boolean | string}) {
  const { value } = props
```

packages/synapse-react-client/src/components/SynapsePriceFeatureRow.tsx
```diff
@@ -0,0 +1,93 @@
...
+> = ({
+  featureDescription,
+  standardPlanValue,
+  managedPlanValue,
+  dataCoordinationPlanValue,
+}) => {
+  const standardPlanElement = useMemo(
+    () => getPriceFeatureCellContent(standardPlanValue),
+    [standardPlanValue],
+  )
```
nickgros, line 49: Use JSX in line with above recommendation

Since the value is always a primitive (string/boolean/undefined), I don't think you need to `useMemo` here

```suggestion
  const standardPlanElement = <PriceFeatureCellContent value={standardPlanValue} />
```

nickgros (approved):

packages/synapse-react-client/src/components/SynapseHomepage.tsx
jay-hodgson: Haha, should be deleted.  (it's a scratch pad while adding content!)

packages/synapse-react-client/src/components/SynapseHomepage.tsx
jay-hodgson: I think MUI differentiates "controlled" vs "uncontrolled" components (when managed by a parent component vs local state).  https://mui.com/material-ui/react-accordion/
I used the wrong prop though, needed to use defaultExpanded!

nickgros (approved):

PR #381, opened 2023-07-24
SWC-6495: fix unreadable success notification when creating a project

hallieswan:
before: 

https://github.com/Sage-Bionetworks/synapse-web-monorepo/assets/26949006/1055b76a-8c4c-43dc-b1d4-5b077b958a5d

after: 

https://github.com/Sage-Bionetworks/synapse-web-monorepo/assets/26949006/a3352188-a012-418f-bffe-7c00e2c527e6

### DIFF
--- a/packages/synapse-react-client/src/components/SynapseNavDrawer.tsx
+++ b/packages/synapse-react-client/src/components/SynapseNavDrawer.tsx
@@ -1,4 +1,9 @@
 import { Badge, Drawer, List, ListItemButton, Tooltip } from '@mui/material'
+import {
+  Direction,
+  SubmissionSortField,
+  SubmissionState,
+} from '@sage-bionetworks/synapse-types'
 import React, { useState } from 'react'
 import { Form } from 'react-bootstrap'
 import SynapseIconWhite from '../assets/icons/SynapseIconWhite'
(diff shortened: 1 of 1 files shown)


packages/synapse-react-client/src/components/SynapseNavDrawer.tsx
```diff
@@ -227,317 +230,319 @@ export const SynapseNavDrawer: React.FunctionComponent<
...
-              />
-              <NavDrawerListItem
-                tooltip="Trash Can"
-                iconName="delete"
-                onClickGoToUrl="/#!Trash:0"
-                handleDrawerClose={handleDrawerClose}
-                handleDrawerOpen={handleDrawerOpen}
-              />
-              {currentUserBundle?.isARReviewer && (
+    <>
```
hallieswan, line 233: The success alert is created within the `CreateProjectModal`, which was included within the `.SynapseNavDrawer` div. There is a [background color](https://github.com/Sage-Bionetworks/synapse-web-monorepo/blob/e1bfaaf8a450406096dbf48ac26e6407092b1ad4/packages/synapse-react-client/src/style/components/_synapse-nav-drawer.scss#L31C3-L31C3) applied to `.MuiPaper-root` within `.SynapseNavDrawer`. When the success alert was refactored to a MUI Alert, the background color applied to the alert as well, since the MUI Alert is based on the MUI Paper component.

The only change here is adding a wrapping empty JSX tag, so the `CreateProjectModal` can be moved out  of the SynapseNavDrawer div and the background color will no longer apply to the alert.

jay-hodgson (approved):
Deceptive diff, thanks for the explanation!

PR #382, opened 2023-07-25
PORTALS-2752 - add useCommittedState hook

nickgros:
useCommittedState will be used to track two versions of a query request in state. This could be useful for any form with an intermediate set of changes that should all be committed in one transaction.

### DIFF
--- a/packages/synapse-react-client/src/utils/hooks/useCommittedState.ts
+++ b/packages/synapse-react-client/src/utils/hooks/useCommittedState.ts
@@ -1,0 +1,41 @@
+import React from 'react'
+
+type UseCommittedStateReturn<T> = {
+  committedState: T
+  uncommittedState: T
+  commit: () => void
+  setUncommittedState: React.Dispatch<React.SetStateAction<T>>
+  isDirty: boolean
+}
+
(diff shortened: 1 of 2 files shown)


jay-hodgson (approved):

packages/synapse-react-client/src/utils/hooks/useCommittedState.ts
```diff
@@ -0,0 +1,51 @@
...
+      }
+    }
+  }, [shouldCommit, uncommittedState])
+
+  return {
+    committedState,
+    uncommittedState,
+    commit,
+    setUncommittedState,
+    isDirty: committedState !== uncommittedState,
```
jay-hodgson, line 49: Should this be a deep compare?

jay-hodgson (approved):

packages/synapse-react-client/src/utils/hooks/useCommittedState.ts
nickgros: The intent was to compare references, since `commit` updates `committedState` with the `uncommittedState` reference.

PR #383, opened 2023-07-25
PORTALS-2752 , PORTALS-2758 - refactor QueryWrapper deep linking/onChange, fix lockedColumn issue, drop extraneous query params

nickgros:
Refactoring / improvements related to PORTALS-2752 - query state management must be overhauled, so I made a few changes to streamline side effects and improve UX
- onChange will be invoked when setState is called internally. Before, it was wrapped in `useEffect` where we had logic to prevent it from running conditionally on the first render. `useEffect` is the wrong mechanism for this kind of behavior
- Deep linking (URL search params) code will always synchronize with the current query.
- If the current query matches the initial query, and is deep linking, then the URL search parameter will be removed
- Remove extraneous query parameters like null values and extra arrays to better match queries for deep linking and cache hits

Also fix PORTALS-2758 - lockedColumn matching should be case-insensitive

### DIFF
--- a/packages/synapse-react-client/src/components/useImmutableTableQuery.ts
+++ b/packages/synapse-react-client/src/components/useImmutableTableQuery.ts
@@ -14,7 +14,10 @@ import {
   isColumnSingleValueQueryFilter,
   isFacetColumnValuesRequest,
 } from '../utils/types/IsType'
-import { queryRequestsHaveSameTotalResults } from '../utils/functions/queryUtils'
+import {
+  queryRequestsHaveSameTotalResults,
+  removeEmptyQueryParams,
+} from '../utils/functions/queryUtils'
 
 export type ImmutableTableQueryResult = {
   /** The ID of the table parsed from the SQL query */
@@ -86,15 +89,33 @@ export default function useImmutableTableQuery(
   options: UseImmutableTableQueryOptions,
 ): ImmutableTableQueryResult {
   const {
-    initQueryRequest,
+    initQueryRequest: initQueryRequestFromProps,
     componentIndex = 0,
     shouldDeepLink = false,
     onQueryChange,
     requireConfirmationOnChange = false,
   } = options
 
-  const [lastQueryRequest, setLastQueryRequest] =
+  const initQueryRequest = useMemo(() => {
+    const request = cloneDeep(initQueryRequestFromProps)
+    request.query = removeEmptyQueryParams(request.query)
+    return request
+  }, [initQueryRequestFromProps])
+
+  const [lastQueryRequest, setLastQueryRequestInternal] =
     useState<QueryBundleRequest>(initQueryRequest)
+
+  function setLastQueryRequest(
+    newState: React.SetStateAction<QueryBundleRequest>,
+  ) {
+    setLastQueryRequestInternal(newState)
+    const request =
+      typeof newState === 'function' ? newState(lastQueryRequest) : newState
+
+    if (onQueryChange) {
+      onQueryChange(JSON.stringify(request.query))
+    }
+  }
 
   const [isConfirmingChange, setIsConfirmingChange] = useState(false)
 
@@ -104,26 +125,6 @@ export default function useImmutableTableQuery(
       setIsConfirmingChange(false)
     },
   )
-
-  /**
-   * Inspect the URL on mount to see if we have a particular query request that we must show.
-   */
-  useEffect(() => {
-    const queryRequestFromLink = DeepLinkingUtils.getQueryRequestFromLink(
-      'QueryWrapper',
-      componentIndex,
-    )
-    if (queryRequestFromLink) {
-      setLastQueryRequest(prevState => ({
-        ...prevState,
-        ...queryRequestFromLink,
-        query: {
-          ...prevState.query,
-          ...queryRequestFromLink.query,
-        },
-      }))
-    }
-  }, [componentIndex])
 
   /**
    * Pass down a deep clone (so no side effects on the child's part) of the
@@ -159,24 +160,13 @@ export default function useImmutableTableQuery(
           ? queryRequest(getLastQueryRequest())
           : queryRequest,
       )
-      setLastQueryRequest(newQueryRequest)
 
-      if (newQueryRequest.query) {
-        const clonedQueryJson = JSON.stringify(newQueryRequest.query)
-        if (shouldDeepLink) {
-          const encodedQuery = encodeURIComponent(clonedQueryJson)
-          DeepLinkingUtils.updateUrlWithNewSearchParam(
-            'QueryWrapper',
-            componentIndex,
-            encodedQuery,
-          )
-        }
-        if (onQueryChange) {
-          onQueryChange(clonedQueryJson)
-        }
-      }
+      // Remove null/empty array fields
+      lastQueryRequest.query = removeEmptyQueryParams(lastQueryRequest.query)
+
+      setLastQueryRequest(newQueryRequest)
     },
-    [componentIndex, getLastQueryRequest, onQueryChange, shouldDeepLink],
+    [getLastQueryRequest],
   )
 
   const setQueryOrPromptConfirmation: ImmutableTableQueryResult['setQuery'] =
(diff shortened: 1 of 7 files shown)


packages/synapse-react-client/src/components/useImmutableTableQuery.ts
```diff
@@ -86,16 +89,34 @@ export default function useImmutableTableQuery(
...
     componentIndex = 0,
     shouldDeepLink = false,
     onQueryChange,
     requireConfirmationOnChange = false,
   } = options
 
-  const [lastQueryRequest, setLastQueryRequest] =
+  const initQueryRequest = useMemo(() => {
+    const request = cloneDeep(initQueryRequestFromProps)
+    request.query = removeEmptyQueryParams(request.query)
```
nickgros, line 101: Remove empty parameter fields, e.g. empty `selectedFacets`

packages/synapse-react-client/src/components/useImmutableTableQuery.ts
```diff
@@ -86,16 +89,34 @@ export default function useImmutableTableQuery(
...
+  function setLastQueryRequest(
+    newState: React.SetStateAction<QueryBundleRequest>,
+  ) {
+    setLastQueryRequestInternal(newState)
+    const request =
+      typeof newState === 'function' ? newState(lastQueryRequest) : newState
+
+    if (onQueryChange) {
+      onQueryChange(JSON.stringify(request.query))
+    }
```
nickgros, line 117: `onQueryChange` is now called when `setLastQueryRequest` is called, which is the change event.

Before, `onQueryChange` was called in a `useEffect` block, but we didn't want it to run on every change. For example, it shouldn't run on mount.

This will be replaced by new, similar code once `useCommittedState` from #382 is integrated.

packages/synapse-react-client/src/components/useImmutableTableQuery.ts
```diff
@@ -105,26 +126,6 @@ export default function useImmutableTableQuery(
...
-        ...prevState,
-        ...queryRequestFromLink,
-        query: {
-          ...prevState.query,
-          ...queryRequestFromLink.query,
-        },
-      }))
-    }
-  }, [componentIndex])
-
```
nickgros, line 127: Moved to co-locate with other deep linking code, without changes.

packages/synapse-react-client/src/components/useImmutableTableQuery.ts
```diff
@@ -159,24 +160,13 @@ export default function useImmutableTableQuery(
...
-          )
-        }
-        if (onQueryChange) {
-          onQueryChange(clonedQueryJson)
-        }
-      }
+      // Remove null/empty array fields
+      lastQueryRequest.query = removeEmptyQueryParams(lastQueryRequest.query)
+
+      setLastQueryRequest(newQueryRequest)
```
nickgros, line 167: Remove the empty fields before saving in state. This better handles the scenario where a facet filter is added then removed. Before this change, the new query would contain `selectedFacets: []`, which semantically matches the original query, but the component would treat it as a different query.

For this scenario, this change would treat the latest query exactly like the initial query. This would result in a cache hit, and also allow us to clear the search parameters (see below).

packages/synapse-react-client/src/components/useImmutableTableQuery.ts
```diff
@@ -159,24 +160,13 @@ export default function useImmutableTableQuery(
...
-          const encodedQuery = encodeURIComponent(clonedQueryJson)
-          DeepLinkingUtils.updateUrlWithNewSearchParam(
-            'QueryWrapper',
-            componentIndex,
-            encodedQuery,
-          )
-        }
-        if (onQueryChange) {
-          onQueryChange(clonedQueryJson)
-        }
```
nickgros, line 176: onQueryChange code moved to setState handler above, deep linking code moved to its own effect below.

packages/synapse-react-client/src/components/useImmutableTableQuery.ts
```diff
@@ -211,6 +201,47 @@ export default function useImmutableTableQuery(
...
+        },
+      }))
+    }
+  }, [componentIndex])
+
+  // If `shouldDeepLink` is true, synchronize the URL
+  useEffect(() => {
+    if (shouldDeepLink) {
+      if (isEqual(initQueryRequest, lastQueryRequest)) {
+        // reset URL...somehow
```
nickgros, line 228:
```suggestion
```

packages/synapse-react-client/src/components/useImmutableTableQuery.ts
```diff
@@ -211,6 +201,46 @@ export default function useImmutableTableQuery(
...
+      } else {
+        const queryJsonString = JSON.stringify(lastQueryRequest.query)
+        DeepLinkingUtils.updateUrlWithNewSearchParam(
+          'QueryWrapper',
+          componentIndex,
+          queryJsonString,
+        )
+      }
+    }
+  }, [componentIndex, lastQueryRequest, initQueryRequest, shouldDeepLink])
```
nickgros, line 242: Handle `shouldDeepLink` in this effect.

If the `initQueryRequest` matches the current query, pass `null` to  `updateUrlWithNewSearchParam`, which will cause the search param to be dropped.

Otherwise, update the URL as desired.

packages/synapse-react-client/src/components/widgets/facet-nav/SelectionCriteriaPills.tsx
```diff
@@ -96,7 +94,7 @@ function getPillPropsFromTextMatchesQueryFilter(
 }
 
 function getPillPropsFromQueryFilters(
-  queryFilters: QueryFilter[],
+  queryFilters: ReadonlyDeep<QueryFilter[]>,
```
nickgros, line 97: Type change was not supposed to be committed yet, but it's not a big deal

packages/synapse-react-client/src/components/widgets/facet-nav/SelectionCriteriaPills.tsx
```diff
@@ -105,7 +103,10 @@ function getPillPropsFromQueryFilters(
       isColumnSingleValueQueryFilter(queryFilter) ||
       isColumnMultiValueFunctionQueryFilter(queryFilter)
     ) {
-      if (queryFilter.columnName === queryContext.lockedColumn?.columnName) {
+      if (
+        queryFilter.columnName.toLowerCase() ===
+        queryContext.lockedColumn?.columnName?.toLowerCase()
+      ) {
```
nickgros, line 109: Fix https://sagebionetworks.jira.com/browse/PORTALS-2758

packages/synapse-react-client/src/components/widgets/facet-nav/SelectionCriteriaPills.tsx
```diff
@@ -116,22 +117,22 @@ function getPillPropsFromQueryFilters(
     } else if (isTextMatchesQueryFilter(queryFilter)) {
       return [getPillPropsFromTextMatchesQueryFilter(queryFilter, queryContext)]
     } else {
-      console.log(
-        'Unknown query filter type',
-        (queryFilter as any).concreteType,
-      )
```
nickgros, line 122: Can drop the cast if we just log the whole object

packages/synapse-react-client/src/components/widgets/facet-nav/SelectionCriteriaPills.tsx
```diff
@@ -116,22 +117,22 @@ function getPillPropsFromQueryFilters(
...
+  selectedFacets: ReadonlyDeep<FacetColumnRequest[]>,
   queryContext: QueryContextType,
   queryVisualizationContext: QueryVisualizationContextType,
 ): SelectionCriteriaPillProps[] {
   return selectedFacets.flatMap(selectedFacet => {
-    if (selectedFacet.columnName === queryContext.lockedColumn?.columnName) {
+    if (
+      selectedFacet.columnName.toLowerCase() ===
+      queryContext.lockedColumn?.columnName?.toLowerCase()
+    ) {
```
nickgros, line 135: https://sagebionetworks.jira.com/browse/PORTALS-2758

packages/synapse-react-client/src/utils/functions/deepLinkingUtils.ts
```diff
@@ -9,92 +9,48 @@ function getComponentSearchHashId(
   return `${componentName}${componentIndex}`
 }
 
-//returns updated search string with the component's info
-function patchSearchString(
```
nickgros, line 13: Use URLSearchParams instead of this custom function

packages/synapse-react-client/src/utils/functions/deepLinkingUtils.ts
```diff
@@ -9,92 +9,48 @@ function getComponentSearchHashId(
...
-  }${location.pathname}?${searchString}`
 
-  window.history.replaceState('object or string', 'Title', newURL)
+  const currentSearch = new URLSearchParams(window.location.search)
+
+  if (stringifiedQuery) {
+    currentSearch.set(componentSearchHashId, stringifiedQuery)
+  } else {
+    currentSearch.delete(componentSearchHashId)
+  }
```
nickgros, line 35: New behavior -- if stringifiedQuery is null (or empty string), then delete the entry from the search params.

packages/synapse-react-client/src/components/useImmutableTableQuery.ts
nickgros: Also, no longer use `encodeURIComponent` because the string is now passed to URLSearchParams, which handles encoding internally.

I have verified that an encoded string created before these changes can be decoded after these changes, so links in portals including an encoded query will not be broken.

packages/synapse-react-client/test/utils/hooks/useImmutableTableQuery.test.tsx
```diff
@@ -176,8 +177,14 @@ describe('useImmutableTableQuery tests', () => {
...
       4,
-      encodeURIComponent(JSON.stringify(newQuery.query)),
+      JSON.stringify(newQuery.query),
     )
+
+    // Change the query back to the initial query, and the parameter should be removed
+    act(() => {
+      result.current.setQuery(options.initQueryRequest)
+    })
+    expect(mockUpdateUrl).toHaveBeenLastCalledWith('QueryWrapper', 4, null)
```
nickgros, line 187: Include a check for the new behavior

packages/synapse-react-client/src/components/useImmutableTableQuery.ts
jay-hodgson: Nice catch (most users will benefit from this performance improvement)

PR #384, opened 2023-07-25
PORTALS-2750, PORTALS-2742: In NF, use Dataset Collection

jay-hodgson:
And truncate in the facet chart legend via css rather than truncating the value

### DIFF
--- a/apps/portals/src/configurations/nf/routesConfig.ts
+++ b/apps/portals/src/configurations/nf/routesConfig.ts
@@ -39,8 +39,8 @@ import {
   studiesSql,
 } from './resources'
 import { toolsDetailsPage } from './synapseConfigs/tools'
-import datasetsV2, { datasetsDetailsPage } from './synapseConfigs/datasetsV2'
 import { ColumnSingleValueFilterOperator } from '@sage-bionetworks/synapse-types'
+import { datasetsDetailsPage } from './synapseConfigs/datasets'
 
 const limit = 3
 
@@ -290,22 +290,7 @@ const routes: GenericRoute[] = [
         ],
       },
       {
-        exact: true,
         path: 'Datasets',
-        synapseConfigArray: [
-          {
-            name: 'RouteControlWrapper',
-            isOutsideContainer: true,
-            props: {
-              ...routeControlWrapperProps,
-              synapseConfig: datasets,
-            },
-          },
-        ],
-      },
-      {
-        path: 'DatasetsV2',
-        hideRouteFromNavbar: true,
         routes: [
           {
             path: '',
@@ -316,7 +301,7 @@ const routes: GenericRoute[] = [
                 isOutsideContainer: true,
                 props: {
                   ...routeControlWrapperProps,
-                  synapseConfig: datasetsV2,
+                  synapseConfig: datasets,
                 },
               },
             ],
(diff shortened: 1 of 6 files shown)


jay-hodgson:
Screen Shot 2023-07-25 at 3 18 29 PM

apps/portals/src/configurations/nf/routesConfig.ts
```diff
@@ -290,22 +290,7 @@ const routes: GenericRoute[] = [
...
-            isOutsideContainer: true,
-            props: {
-              ...routeControlWrapperProps,
-              synapseConfig: datasets,
-            },
-          },
-        ],
-      },
-      {
-        path: 'DatasetsV2',
```
jay-hodgson, line 307: Removed DatasetsV2 path that was not referenced elsewhere

apps/portals/src/configurations/nf/synapseConfigs/datasets.ts
```diff
@@ -1,33 +1,51 @@
...
 const rgbIndex = 8
 
+export const datasetCardConfiguration: CardConfiguration = {
+  type: SynapseConstants.GENERIC_CARD,
+  genericCardSchema: {
+    type: SynapseConstants.DATASET,
+    title: 'title',
+    description: 'description',
+    secondaryLabels: [
+      // 'assay',
```
jay-hodgson, line 18: I've been told that assay might be an available column soon

packages/synapse-react-client/src/components/GenericCard/GenericCard.tsx
```diff
@@ -688,8 +692,15 @@ export default class GenericCard extends React.Component<
...
       if (value) {
-        // SWC-6115: special rendering of the version column (for Views)
-        if (isView && columnName === 'currentVersion') {
+        // PORTALS-2750: special rendering of the datasetSizeInBytes (for Dataset Collections)
+        if (
+          isDatasetCollection(table as Entity) &&
+          columnName === 'datasetSizeInBytes'
+        ) {
+          columnDisplayName = 'Size'
+          value = calculateFriendlyFileSize(parseInt(value))
```
jay-hodgson, line 701: If this is a Dataset Collection and the column is named "datasetSizeInBytes", then use a special renderer.  Output the display column "Size" and a friendly file size

packages/synapse-react-client/src/components/widgets/facet-nav/FacetNav.tsx
```diff
@@ -203,8 +203,10 @@ const FacetNav: React.FunctionComponent<FacetNavProps> = ({
             <div className="FacetNav__row" role="list">
               {facets.map(facet => (
                 <div
-                  className="col-sm-12 col-md-4"
```
jay-hodgson, line 206: These bootstrap grid styles were effectively only adding a padding left and right (because the responsive behavior is due to the display flex wrap styles!)

packages/synapse-react-client/src/components/widgets/facet-nav/FacetPlotLegendList.tsx
```diff
@@ -23,13 +19,19 @@ export function FacetPlotLegendList(props: FacetPlotLegendProps) {
...
         return (
           <ElementWithTooltip
             tooltipText={facetValue.label}
             key={facetValue.label}
           >
-            <Box sx={{ display: 'flex', alignItems: 'center', gap: '10px' }}>
+            <Box
+              sx={{
+                display: 'grid',
+                gridTemplateColumns: '20px auto 30px',
```
jay-hodgson, line 30: Change into a Grid layout, with the facet value label taking up the max space of an auto column, then being truncated via css.  This makes the truncation dynamic based on the available width.

nickgros (approved):

PR #386, opened 2023-07-26
PORTALS-2752 - Facet filter form can be reset on cancellation

nickgros:
Separate query state into "current" and "next", allowing us to fully reset the query if a change is canceled during confirmation

useImmutableTableQuery:
 - Track query state in useCommittedState
 - Expose commit options. No option commits changes immediately. Can also commit after debounce delay, or no commit.
 - Add `addValueToSelectedFacet` to simplify selected facet manipulation
 - Rename lastQueryRequest to currentQueryRequest
 - Expose readonly instances of currentQueryRequest and nextQueryRequest, and pass through QueryContext in QueryWrapper
 - Unit tests to reach 98% line coverage

EnumFacetFilter:
 - Replace onChange with add/remove filter from selected facet via query context
 - Remove debounce on change, now handled within useImmutableTableQuery

General table query components
 - Use executeQueryRequest function instead of object where change is easy
 - Fix tests


packages/synapse-react-client/src/components/FullTextSearch.tsx
```diff
@@ -33,31 +33,31 @@ export const FullTextSearch: React.FunctionComponent<FullTextSearchProps> = ({
         `Search term must have a minimum of ${MIN_SEARCH_QUERY_LENGTH} characters`,
       )
     } else {
-      const lastQueryRequestDeepClone = getLastQueryRequest()
+      executeQueryRequest(request => {
```
nickgros, line 36: In general, make changes to pass a transforming function to `executeQueryRequest` instead of passing a (possibly stale) object.

packages/synapse-react-client/src/components/QueryContext/QueryContext.tsx
```diff
@@ -49,14 +48,17 @@ export type QueryContextType<
           : never
       >
     | undefined
+  currentQueryRequest: ReadonlyDeep<QueryBundleRequest>
+  nextQueryRequest: ReadonlyDeep<QueryBundleRequest>
   /** Returns a deep clone of the current query bundle request */
-  getLastQueryRequest: () => QueryBundleRequest
+  getCurrentQueryRequest: () => QueryBundleRequest
```
nickgros, line 54: Rename `lastQueryRequest` to `currentQueryRequest`

packages/synapse-react-client/src/components/QueryContext/QueryContext.tsx
```diff
@@ -49,14 +48,17 @@ export type QueryContextType<
           : never
       >
     | undefined
+  currentQueryRequest: ReadonlyDeep<QueryBundleRequest>
+  nextQueryRequest: ReadonlyDeep<QueryBundleRequest>
```
nickgros, line 52: If consuming components just need to reference the query object, they can access this readonly object instead of calling `getCurrentQueryRequest`.

I think it's a bit more intuitive to use objects when thinking about the React component lifecycle, since it's more natural to consider cases where an object updates and triggers a re-render, vs a `get` function updating.

packages/synapse-react-client/src/components/SynapseTable/SynapseTable.tsx
```diff
@@ -843,48 +832,30 @@ export class SynapseTable extends React.Component<
...
-            facetNamesMap,
-          )
-        }}
-        onClear={() => {
-          applyChangesToValuesColumn(
-            lastQueryRequest,
-            facetColumnResult,
-            this.applyChangesFromQueryFilter,
-          )
-        }}
```
nickgros, line 877: EnumFacetFilter now directly uses the context instead of this complicated chain of callback props

packages/synapse-react-client/src/components/useImmutableTableQuery.ts
```diff
@@ -18,18 +19,36 @@ import {
...
+
+type QueryChangeCommitOptions =
+  | {
+      // This and future changes including debounce will not be committed until the debounceDelay has elapsed
+      debounce: true
+    }
+  | {
+      // The change will not be committed until the commit function is invoked.
+      noCommit: true
+    }
```
nickgros, line 34: Used an object instead of a string literal in case we wanted to add more options, like a way to manually control the debounce duration.

packages/synapse-react-client/src/components/useImmutableTableQuery.ts
```diff
@@ -102,20 +130,17 @@ export default function useImmutableTableQuery(
...
+    commit,
+    setUncommittedState: setNextQueryRequest,
+  } = useCommittedState<QueryBundleRequest>(initQueryRequest, queryRequest => {
     if (onQueryChange) {
-      onQueryChange(JSON.stringify(request.query))
+      const queryJsonString = JSON.stringify(queryRequest.query)
+      onQueryChange(queryJsonString)
     }
-  }
+  })
```
nickgros, line 143: Track two versions of state so we can easily roll back if the user cancels a filter change when we ask for confirmation

packages/synapse-react-client/src/components/useImmutableTableQuery.ts
```diff
@@ -218,41 +280,46 @@ export default function useImmutableTableQuery(
...
       }
     }
-  }, [componentIndex, lastQueryRequest, initQueryRequest, shouldDeepLink])
+  }, [componentIndex, currentQueryRequest, initQueryRequest, shouldDeepLink])
 
   const onCancelChangeQuery = useCallback(() => {
     setIsConfirmingChange(false)
-  }, [])
+    setNextQueryRequest(currentQueryRequest)
+  }, [currentQueryRequest, setNextQueryRequest])
```
nickgros, line 312: Add rollback behavior for cancellation

packages/synapse-react-client/src/components/useImmutableTableQuery.ts
```diff
@@ -281,11 +348,46 @@ export default function useImmutableTableQuery(
...
+            columnName: columnName,
+            facetValues: [value],
+          })
+        }
+        request.query.selectedFacets = newFacets
+        return request
+      }, commitOptions)
+    },
+    [setQueryOrPromptConfirmation],
+  )
```
nickgros, line 389: Add an easy, centralized way to add a filter on an enumerated facet

packages/synapse-react-client/src/components/useImmutableTableQuery.ts
```diff
@@ -302,7 +404,13 @@ export default function useImmutableTableQuery(
   )
 
   const removeValueFromSelectedFacet = useCallback(
-    (facet: FacetColumnRequest, value: string) => {
+    (
+      facet: FacetColumnRequest | string,
```
nickgros, line 408: There's at most one facet per column, so allow passing a column name to make it easier.

The motivation for this change is that the facet filter controls are populated using the facet statistics response, not the `FacetColumnRequest` objects. Within the control components, we can make facet manipulation very simple by just performing operations against a column name--the complexity of the `FacetColumnRequest` objects can be abstracted.

packages/synapse-react-client/src/components/widgets/query-filter/EnumFacetFilter.tsx
```diff
@@ -81,17 +83,42 @@ function formatFacetValuesForDisplay(
...
+  nextQueryRequest: ReadonlyDeep<QueryBundleRequest>,
+  columnModel: SelectColumn,
+) {
+  return !nextQueryRequest.query.selectedFacets?.find(
+    facetRequest =>
+      facetRequest.concreteType ===
+        'org.sagebionetworks.repo.model.table.FacetColumnValuesRequest' &&
+      facetRequest.columnName === columnModel.name,
+  )
+}
```
nickgros, line 105: Extracted from component. Logic was changed to use the _request_ object rather than the _response_ object, so it should more accurately track with what the user is about to do, instead of what the system most recently did.

packages/synapse-react-client/src/components/widgets/query-filter/EnumFacetFilter.tsx
```diff
@@ -7,23 +7,25 @@ import { SynapseConstants } from '../../../utils'
...
+import { useQueryContext } from '../../QueryContext'
+import { ReadonlyDeep } from 'type-fest'
+import { isFacetColumnValuesRequest } from '../../../utils/types/IsType'
+import { cloneDeep } from 'lodash-es'
 
 export type EnumFacetFilterProps = {
   facetValues: FacetColumnResultValueCount[]
   columnModel: SelectColumn
-  onChange: (facetNamesMap: Record<string, boolean>) => void
-  onClear: () => void
```
nickgros, line 26: Now use the query context instead of these methods.

Ideally I would like to split this into a "View" component that has these props and a wrapper component that passes business logic into the View components' props, but I held off since that would be an even larger change.

packages/synapse-react-client/test/components/SynapseTable/SynapseTable.integration.test.tsx
```diff
@@ -315,15 +317,17 @@ describe('SynapseTable tests', () => {
         direction: SORT_STATE[1],
       }
       // below we match only the part of the object that we expect to have changed
-      await waitFor(() =>
-        expect(executeQueryRequest).toHaveBeenCalledWith(
+      await waitFor(() => {
+        expect(executeQueryRequest).toHaveBeenCalled()
+        expect(typeof executeQueryRequest.mock.lastCall[0]).toBe('function')
+        expect(executeQueryRequest.mock.lastCall[0](lastQueryRequest)).toEqual(
```
nickgros, line 323: Many test changes account for the fact that `executeQueryRequest` is now called with a function

packages/synapse-react-client/test/containers/widgets/Checkbox.test.tsx
```diff
@@ -34,20 +37,11 @@ describe('basic function', () => {
...
   })
 
-  it('should call callbackFn on change with correct params and change the value', async () => {
-    render(<Checkbox {...props} checked={false} />)
-
-    await userEvent.click(screen.getByRole('checkbox'))
-    expect(mockCallback).toHaveBeenCalledWith(true)
-
-    expect(screen.getByRole<HTMLInputElement>('checkbox').checked).toBe(true)
-  })
```
nickgros, line 44: This basically tested that the Checkbox had its own internal state. This is no longer true, so test removed!

packages/synapse-react-client/test/containers/widgets/query-filter/FacetFilterControls.test.tsx
```diff
@@ -161,57 +164,6 @@ describe('FacetFilterControls tests', () => {
   })
 
   describe('handling child component callbacks', () => {
-    it('should propagate enum update correctly', async () => {
```
nickgros, line 164: These tested the callbacks passed to a mocked `EnumFacetFilter`. Those callbacks are no longer passed, so I just removed the tests.

packages/synapse-react-client/src/components/useImmutableTableQuery.ts
```diff
@@ -148,56 +184,82 @@ export default function useImmutableTableQuery(
...
+      nextQueryRequest.query = removeEmptyQueryParams(nextQueryRequest.query)
 
-      setLastQueryRequest(newQueryRequest)
+      setNextQueryRequest(nextQueryRequest)
+      if (commitOptions) {
+        if ('debounce' in commitOptions && commitOptions.debounce) {
+          setCommitAfterDebounce(true)
+        }
+      } else {
+        commit()
```
jay-hodgson, line 225: If there are `commitOptions` but `debounce` is either `undefined` or `false`, will it ever call `commit()`?

packages/synapse-react-client/src/components/useImmutableTableQuery.ts
jay-hodgson: Excellent

packages/synapse-react-client/src/components/widgets/query-filter/EnumFacetFilter.tsx
jay-hodgson: I would think that this logic would fix https://sagebionetworks.jira.com/browse/PORTALS-2680.  Is this accurate?

jay-hodgson:
Looks like there's an unstable test (SchemaDrivenAnnotationsEditor.integration.test.tsx failure)

packages/synapse-react-client/src/components/widgets/query-filter/EnumFacetFilter.tsx
nickgros: I think it will! I'll add the issue to my sprint and double-check to confirm

packages/synapse-react-client/src/components/useImmutableTableQuery.ts
nickgros: Nope, but you would also get a compile-time error

jay-hodgson (approved):

PR #389, opened 2023-07-27
PORTALS-2748 - mocks, test utils are type-checked

nickgros:
In order to type-check tests, test dependencies need to be type-checked. In most tests, that means we need to type-check our mock data.

- Move `mocks` to `src/mocks`
- Move most test utils from `test/testutils` to `src/testutils`
- Add missing properties to types
- Add type guards and casts when appropriate
- Fix `ColumnModel` type which was missing a property


packages/synapse-types/src/Table/ColumnModel.ts
```diff
@@ -2,11 +2,18 @@ import { ColumnType } from './ColumnType'
...
+  name: string
+  /** The default value for this column. Columns of type FILEHANDLEID and ENTITYID are not allowed to have default values. */
+  defaultValue?: string
   columnType: ColumnType
-  maximumSize?: number // A parameter for columnTypes with a maximum size. For example, ColumnTypeEnum.strings have a default maximum size of 50 characters, but can be set to a maximumSize of 1 to 1000 characters.
-  enumValues?: string[] // Columns of type string can be constrained to an enumeration values set on this list. The maximum number of entries for an enum is 100
+  /** A parameter for columnTypes with a maximum size. For example, ColumnTypeEnum.strings have a default maximum size of 50 characters, but can be set to a maximumSize of 1 to 1000 characters. */
+  maximumSize?: number
+  /** Required if using a columnType with a "_LIST" suffix. Describes the maximum number of values that will appear in that list. Value range 1-100 inclusive. Default 100 */
+  maximumListLength?: number
```
jay-hodgson, line 15: This is a difficult diff to make sense of, but did you add the maximumListLength field?

jay-hodgson (approved):

packages/synapse-types/src/Table/ColumnModel.ts
nickgros: Yes

PR #395, opened 2023-07-31
PORTALS-2634, PORTALS-2636 - SDS for FullTextSearch, SqlEditor, Explore alignment fixes

nickgros:
Changed FullTextSearch and SqlEditor components to match SDS style and put info button inside the input. While these changes were not explicitly called out in designs, it supports PORTALS-2636 by having the end of the input align with the right edge of the components

Also updated CSS margins for many components

|Component|Before|After|
|---|---|---|
|FullTextSearch|image |image |
| SqlEditor | image| image |

### DIFF
--- a/packages/synapse-react-client/src/components/FullTextSearch.tsx
+++ b/packages/synapse-react-client/src/components/FullTextSearch.tsx
@@ -1,7 +1,7 @@
-import { Collapse } from '@mui/material'
-import React, { useRef, useState } from 'react'
+import { Collapse, IconButton, TextField } from '@mui/material'
+import React, { ChangeEvent, useRef, useState } from 'react'
 import { TextMatchesQueryFilter } from '@sage-bionetworks/synapse-types'
-import { useQueryContext } from './QueryContext/QueryContext'
+import { useQueryContext } from './QueryContext'
 import { useQueryVisualizationContext } from './QueryVisualizationWrapper'
 import { HelpPopover } from './HelpPopover/HelpPopover'
 import IconSvg from './IconSvg/IconSvg'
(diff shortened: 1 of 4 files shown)


packages/synapse-react-client/src/style/components/_query-wrapper-text-input.scss
```diff
@@ -1,7 +1,6 @@
 @use '../abstracts/variables' as SRC;
 
-.QueryWrapperSearchInput,
-.QueryWrapperSqlEditorInput {
+.QueryWrapperSearchInput {
```
nickgros, line 3: These styles now only apply to the regular (per-column) search component, which is a little more complicated so I didn't touch it for now

packages/synapse-react-client/src/style/components/_query-wrapper-plot-nav.scss
```diff
@@ -85,14 +90,23 @@
...
     }
+    .FacetNav,
+    .FilterAndView,
+    .ErrorBannerWrapper,
+    .QueryWrapperFullTextSearchInput,
+    .QueryWrapperSearchInput,
+    .QueryWrapperSqlEditorInput,
+    .download-confirmation {
+      margin-left: 0px;
+    }
```
nickgros, line 101: If hiding facet filters, these should not have left margin

packages/synapse-react-client/src/style/components/_query-wrapper-plot-nav.scss
```diff
@@ -12,6 +12,10 @@
       grid-column: 1 / span 2;
     }
 
+    .ErrorBannerWrapper {
+      grid-column: 1 / span 2;
+    }
```
nickgros, line 17: Change to take up full width (if you get this error wrapper we probably don't have facet filters to show, so taking up full width is probably best)

packages/synapse-react-client/src/components/FullTextSearch.tsx
```diff
@@ -60,50 +60,58 @@ export const FullTextSearch: React.FunctionComponent<FullTextSearchProps> = ({
...
-          </form>
-          <div className="QueryWrapperSearchInput__helppopover">
-            <HelpPopover
-              markdownText={helpMessage}
-              helpUrl={helpUrl}
-              placement="left"
-            />
-          </div>
-        </div>
+        <form onSubmit={search}>
```
nickgros, line 71: For both of these components, use MUI TextField and put the popover in the `endAdornment`. This lets the input field span the full width of the column, and matches the nice clean edge line that we're trying to establish in PORTALS-2636

jay-hodgson: "`endAdornment`"?  Ok

jay-hodgson:
Very slick, nice work

jay-hodgson (approved):

PR #399, opened 2023-08-01
PORTALS-2716

jay-hodgson:


packages/synapse-react-client/src/components/QueryWrapper/QueryWrapper.tsx
```diff
@@ -177,6 +180,7 @@ export function QueryWrapper(props: QueryWrapperProps) {
     selectedRows,
     setSelectedRows,
     addValueToSelectedFacet,
+    combineRangeFacetConfig,
```
jay-hodgson, line 183: Add to query context

packages/synapse-react-client/src/components/QueryWrapperPlotNav/QueryWrapperPlotNav.stories.tsx
```diff
@@ -213,6 +213,18 @@ export const SendToCavatica: Story = {
   },
 }
 
+export const CombinedRangeFacetsDemo: Story = {
```
jay-hodgson, line 216: Demo!

packages/synapse-react-client/src/components/widgets/query-filter/CombinedRangeFacetFilter.test.tsx
```diff
@@ -0,0 +1,318 @@
+import { Collapse as MockCollapse } from '@mui/material'
```
jay-hodgson, line 1: Copied RangeFacetFilter tests, and modified to have 2 faceted columns and test for the expected callback values array

packages/synapse-react-client/src/components/widgets/query-filter/CombinedRangeFacetFilter.tsx
```diff
@@ -0,0 +1,157 @@
+import { Collapse } from '@mui/material'
```
jay-hodgson, line 1: Core of the change.  Originally tried to change RangeFacetFilter to support both cases, but this is so different that I'd like to isolate the logic.

packages/synapse-react-client/src/components/widgets/query-filter/CombinedRangeFacetFilter.tsx
```diff
@@ -0,0 +1,157 @@
...
+    }
+  }
+
+  const [radioValue, setRadioValue] = useState(
+    getRadioValue(selectedMin, isAnyValue),
+  )
+
+  useEffect(() => {
+    setRadioValue(getRadioValue(selectedMin, isAnyValue))
+  }, [selectedMin, isAnyValue])
```
jay-hodgson, line 88: Bug fix for both the RangeFacetFilter and this Combined version.  In the old implementation, after you specified a range the UI would often be initialized with the radio checkbox on "Any" (but with the correct range slider if you selected the Range radio!).  Be sure to update the radio whenever the input changes (cannot rely on the default value on mount only)

packages/synapse-react-client/src/components/widgets/query-filter/CombinedRangeFacetFilter.tsx
```diff
@@ -0,0 +1,157 @@
...
+                  {'>'}
+                </RangeSlider>
+              )}
+              {columnType === 'DATE' && (
+                <Range
+                  key="Range"
+                  initialValues={{
+                    // From the backend, selectedMin is a formatted date (like "2021-06-15"), but columnMin is a unix timestamp in millis (like "1624651794856")
+                    min: col2SelectedMin ?? dayjs(parseInt(col1Min)).toString(),
+                    max: col1SelectedMax ?? dayjs(parseInt(col2Max)).toString(),
```
jay-hodgson, line 131: Another bug fix made to both the RangeFacetFilter and this new version.

jay-hodgson: Use the selected value (since it's a value that the calendar widget recognizes).  But if it's a unix timestamp (columnMin or columnMax), then parse it using dayjs for the range widget.

packages/synapse-react-client/src/components/widgets/query-filter/FacetFilterControls.tsx
```diff
@@ -49,19 +50,22 @@ const convertFacetColumnRangeRequest = (
...
-  const selections = lastRequest?.query?.selectedFacets ?? []
   const changedFacetIndex = selections.findIndex(
     facet => facet.columnName === changedFacet.columnName,
   )
-
   const isEmptyValuesFacet =
     changedFacet.concreteType ===
       'org.sagebionetworks.repo.model.table.FacetColumnValuesRequest' &&
     (!changedFacet.facetValues || !changedFacet.facetValues.length)
+  const isEmptyRangesFacet =
```
jay-hodgson, line 62: Added a corresponding check for an empty facet column range request

packages/synapse-react-client/src/components/widgets/query-filter/FacetFilterControls.tsx
```diff
@@ -107,10 +114,33 @@ export const applyChangesToRangeColumn = (
...
   const changedFacet = convertFacetColumnRangeRequest(facet)
-  const result = patchRequestFacets(changedFacet, lastRequest)
+  const result = patchRequestFacets(
+    changedFacet,
+    lastRequest?.query?.selectedFacets,
+  )
   onChangeFn(result)
 }
 
+export const applyCombinedChangesToRangeColumn = (
```
jay-hodgson, line 124: Update the ranges to both facet filters based on the values from the CombinedRangeFacetFilter component

packages/synapse-react-client/src/components/widgets/query-filter/FacetFilterControls.tsx
```diff
@@ -227,9 +272,30 @@ function FacetFilterControls(props: FacetFilterControlsProps) {
     },
     [facetFiltersShown],
   )
+  const combinedRangeFacetsColumnModelType = combineRangeFacetConfig
```
jay-hodgson, line 275: Assumption here that both facet columns are of the same type.  The combination range facet filter component would not work otherwise (and since this is set in the portal configuration, it's something that we control).

packages/synapse-react-client/src/components/widgets/query-filter/RangeFacetFilter.tsx
```diff
@@ -1,62 +1,64 @@
...
   const hasAnyValue = !selectedMin && !selectedMax
 
-  selectedMin = selectedMin || columnMin
-  selectedMax = selectedMax || columnMax
+  const currentMin = selectedMin || columnMin
+  const currentMax = selectedMax || columnMax
 
-  const rangeType = columnModel.columnType === 'DOUBLE' ? 'number' : 'date'
-
-  const getRadioValue = (min: string, isAnyValue: boolean) => {
```
jay-hodgson, line 52: Refactored and shared

packages/synapse-react-client/src/components/widgets/query-filter/RangeFacetFilter.tsx
```diff
@@ -96,28 +101,44 @@ export const RangeFacetFilter: React.FunctionComponent<
...
                   }
                 >
-                  ) {'>'}
+                  {'>'}
                 </RangeSlider>
               )}
 
-              {(columnModel.columnType === 'DATE' ||
-                columnModel.columnType === 'DOUBLE') && (
+              {columnType === 'DATE' && (
```
jay-hodgson, line 122: Treat DATE and DOUBLE differently, since DATE sometimes requires extra parsing.

packages/synapse-react-client/src/utils/hooks/useImmutableTableQuery/useImmutableTableQuery.ts
```diff
@@ -64,7 +64,9 @@ export type ImmutableTableQueryResult = {
     commitOptions?: QueryChangeCommitOptions,
   ) => void
   /** Removes a particular selected facet from the query */
-  removeSelectedFacet: (facet: FacetColumnRequest) => void
+  removeSelectedFacet: (
+    facet: FacetColumnRequest | FacetColumnRequest[],
+  ) => void
```
jay-hodgson, line 69: Now we might want to remove multiple selected facets at once

packages/synapse-react-client/src/components/widgets/query-filter/CombinedRangeFacetFilter.test.tsx
```diff
@@ -0,0 +1,339 @@
...
+  }
+}
+
+describe('CombinedRangeFacetFilter tests', () => {
+  let props: CombinedRangeFacetFilterProps
+  function init(overrides?: CombinedRangeFacetFilterProps) {
+    props = createTestProps(overrides)
+    return render(
+      <QueryVisualizationContextProvider
+        queryVisualizationContext={{
```
jay-hodgson, line 108: Type check enforced here led me to adding mocks for all of the required properties (which the RangeFacetFilter test does not need because it's still in the /test/ folder, I think).  Is there another way to do this, since this seems excessive.

packages/synapse-react-client/src/components/QueryContext/QueryContext.tsx
```diff
@@ -198,4 +200,10 @@ export function useInfiniteQueryContext(): InfiniteQueryContextType {
   return context as InfiniteQueryContextType
 }
 
+export type CombineRangeFacetConfig = {
+  minFacetColumn: string
+  maxFacetColumn: string
+  label: string
+}
+
```
nickgros, line 208: Do you think it makes more sense to have this here in `QueryContext`/`QueryWrapper` or `QueryVisualizationContext`/`QueryVisualizationWrapper`? I'm starting to think that this is probably not the best way to separate responsibilities, since many UI config tweaks may actually need to modify the query itself.

packages/synapse-react-client/src/components/QueryContext/QueryContext.tsx
```diff
@@ -96,6 +96,8 @@ export type QueryContextType<
   rowSelectionPrimaryKey?: string[]
   /** Whether the user has selected any rows */
   hasSelectedRows: boolean
+  /** Combines two faceted columns into a single inclusive range selector */
+  combineRangeFacetConfig?: ReadonlyDeep<CombineRangeFacetConfig>
```
nickgros, line 100: Like the ReadonlyDeep

packages/synapse-react-client/src/components/widgets/query-filter/CombinedRangeFacetFilter.tsx
nickgros: Worth filing a PLFM bug? I guess it could be a breaking change so I dunno if they would fix it

packages/synapse-react-client/src/components/widgets/query-filter/CombinedRangeFacetFilter.test.tsx
nickgros: I want to add a mock helper for this context

packages/synapse-react-client/src/components/widgets/query-filter/CombinedRangeFacetFilter.test.tsx
```diff
@@ -0,0 +1,339 @@
...
+        facetResults: [minRangeFacetResult, maxRangeFacetResult],
+        columnType: dateColumnModel.columnType,
+        label: dateColumnModel.name,
+      }
+      init(updatedProps)
+      await waitFor(() => expect(Range).toHaveBeenCalled())
+      await waitFor(() => expect(capturedOnChange).toBeDefined())
+      capturedOnChange!({ min: '40', max: '50' })
+      // Call back with column1 min to range selector max, and range selector min to column2 max.
+      expect(mockCallback).toHaveBeenCalledWith(['10', '50', '40', '70'])
```
nickgros, line 322: Nice!

nickgros (approved):
This is great, fantastically simple compared to how complicated I thought it might be!

packages/synapse-react-client/src/components/widgets/query-filter/FacetFilterControls.tsx
```diff
@@ -227,9 +272,30 @@ function FacetFilterControls(props: FacetFilterControlsProps) {
...
+          onChange={(values: (string | number | undefined)[]) =>
+            applyCombinedChangesToRangeColumn(
+              lastRequest,
+              combinedRangeFacets as FacetColumnResultRange[],
+              applyChanges,
+              values as string[],
+            )
+          }
+        ></CombinedRangeFacetFilter>
+      )}
```
nickgros, line 298: So will this always be first? If that's not desired it should be an easy future change

PR #400, opened 2023-08-02
PORTALS-2634, PORTALS-2636: TopLevelControls/TotalQueryResults appears to be full width in portals

nickgros:
image


jay-hodgson (approved):

jay-hodgson:
Might want to look at the overall portal logo in the header as well, to see if it can be aligned as well

nickgros:
> Might want to look at the overall portal logo in the header as well, to see if it can be aligned as well

It varies by logo -- some have "padding" in the image that will need to be corrected

PR #404, opened 2023-08-03
ELITE portal: hide Explore -> Data by Participants (again), and point cohort builder source to V…

jay-hodgson:
…irtual Tables


nickgros (approved):
It's so cool to see this come together

jay-hodgson:
The cohort builder Virtual Tables are not functional as it stands today, but we're getting closer!

PR #409, opened 2023-08-04
PORTALS-2634, PORTALS-2636

nickgros:
- Sign In button should be right-aligned with other content
- Crop ADKP logo


apps/portals/src/Navbar.tsx
```diff
@@ -208,7 +208,7 @@ function Navbar() {
             )}
           {!isSignedIn &&
             isSynapseSubdomainOrLocal && ( // desktop sign in
-              <div className="center-content nav-button nav-button-signin">
+              <div className="center-content nav-button-signin">
```
nickgros, line 211: `.nav-button` gave this undesirable side padding

apps/portals/src/configurations/adknowledgeportal/style/logo.svg
nickgros, line 1: Just cropped using Inkscape. Not sure why GitHub thinks it's smaller, but it's an svg so it doesn't matter.

jay-hodgson (approved):

PR #412, opened 2023-08-14
PORTALS-2771 - primary key equality check to determine row selection

nickgros:
PORTALS-2771 references a case where there is no `rowId`, which breaks 'selected' checkbox state. Add a new helper to determine this state. When defined, we should reference the `rowSelectionPrimaryKey` (so add it for the corresponding EL Portal Cohort Builder case). If there's no specified primary key, gracefully fall back to using rowId or the `id` column when it is appropriate, using deep comparison of Row objects as a last resort.

One goal that is considered, though not yet attained, is to allow users to maintain a cohort selection between sessions. Since data in the portal may change over time, using a primary key is more resilient than using deep equality. However, this is still not perfect; we don't save column models, so this would still break if the index of the primary key column(s) were to change.

### DIFF
--- a/packages/synapse-react-client/src/components/QueryWrapper/useTableRowSelection.test.tsx
+++ b/packages/synapse-react-client/src/components/QueryWrapper/useTableRowSelection.test.tsx
@@ -1,0 +1,263 @@
+import useTableRowSelection from './useTableRowSelection'
+import { act, renderHook } from '@testing-library/react'
+import { mockTableEntity } from '../../mocks/entity/mockTableEntity'
+import { mockQueryResultBundle } from '../../mocks/mockFileViewQuery'
+import { mockFileViewEntity } from '../../mocks/entity/mockEntity'
+import { QueryResultBundle, Row } from '@sage-bionetworks/synapse-types'
+import { cloneDeep } from 'lodash-es'
+
+describe('useTableRowSelection tests', () => {
+  it('hides rowSelection by default', () => {
+    const { result } = renderHook(() =>
+      useTableRowSelection({
+        isRowSelectionVisible: undefined, // !
+        entity: mockTableEntity,
+        columnModels: mockQueryResultBundle.columnModels,
+      }),
+    )
+
+    expect(result.current.isRowSelectionVisible).toBe(false)
+    expect(result.current.hasSelectedRows).toBe(false)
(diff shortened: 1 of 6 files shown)


packages/synapse-react-client/src/components/QueryWrapper/useTableRowSelection.ts
```diff
@@ -18,6 +20,49 @@ type UseTableRowSelectionReturn = {
...
+      )
+      return isEqual(r1PrimaryKeyValues, r2PrimaryKeyValues)
+    }
+  } else if (row.rowId) {
+    // If there's a rowId, we can use that as a fallback
+    comparator = (r1: Row, r2: Row) => r1.rowId === r2.rowId
+  }
+
+  return comparator
+}
```
nickgros, line 52: I'm not sure if we've committed to persisting selection between sessions, but that was a consideration when deciding to use primary key instead of just naively using a deep equality check.

packages/synapse-react-client/src/components/QueryWrapper/useTableRowSelection.test.tsx
```diff
@@ -0,0 +1,263 @@
+import useTableRowSelection from './useTableRowSelection'
+import { act, renderHook } from '@testing-library/react'
+import { mockTableEntity } from '../../mocks/entity/mockTableEntity'
+import { mockQueryResultBundle } from '../../mocks/mockFileViewQuery'
+import { mockFileViewEntity } from '../../mocks/entity/mockEntity'
+import { QueryResultBundle, Row } from '@sage-bionetworks/synapse-types'
+import { cloneDeep } from 'lodash-es'
+
+describe('useTableRowSelection tests', () => {
```
nickgros, line 9: 100% branch coverage for the corresponding utility

packages/synapse-react-client/src/components/QueryWrapper/useTableRowSelection.ts
jay-hodgson: Thank for falling back to the rowId as well.  Since CRC no longer supports joining the study, I think the CRC Researcher portal will no longer be used.  But I think it would still work properly with this logic

jay-hodgson (approved):

PR #413, opened 2023-08-14
PORTALS-2774 - Reorder routes so OrientationBanner appears in the correct location

nickgros:
PORTALS-2774

### DIFF
--- a/apps/portals/src/configurations/elportal/routesConfig.tsx
+++ b/apps/portals/src/configurations/elportal/routesConfig.tsx
@@ -185,28 +185,6 @@ const routes: GenericRoute[] = [
     path: 'Explore',
     routes: [
       {
-        path: ':slug/',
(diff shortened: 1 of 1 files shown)


apps/portals/src/configurations/elportal/routesConfig.tsx
```diff
@@ -184,28 +184,6 @@ const routes: GenericRoute[] = [
   {
     path: 'Explore',
     routes: [
-      {
-        path: ':slug/',
-        hideRouteFromNavbar: true,
-        exact: true,
-        synapseConfigArray: [
-          {
-            name: 'RouteControlWrapper',
```
nickgros, line 193: RouteControlWrapper should appear below OrientationBanner, so change the route order so it comes after

jay-hodgson (approved):

PR #414, opened 2023-08-15
PORTALS-2773 - fix cohort builder applied filters

nickgros:
PORTALS-2773

Also needed to modify SelectionCriteriaPills to handle cases where we have a filter where a column model is not present, since these filters are applied to columns on the table/view referenced in the definingSql.

### DIFF
--- a/packages/synapse-react-client/src/components/widgets/facet-nav/SelectionCriteriaPills.tsx
+++ b/packages/synapse-react-client/src/components/widgets/facet-nav/SelectionCriteriaPills.tsx
@@ -35,10 +35,13 @@ function getPillPropsFromColumnQueryFilter(
   queryVisualizationContext: QueryVisualizationContextType,
 ): SelectionCriteriaPillProps[] {
   const { getColumnDisplayName } = queryVisualizationContext
-  const columnModel = queryContext.getColumnModel(queryFilter.columnName)!
+  const columnModel = queryContext.getColumnModel(queryFilter.columnName)
   // ColumnSingleValueQueryFilter and ColumnMultiValueQueryFilter both allow for a list of values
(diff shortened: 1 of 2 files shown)


packages/synapse-react-client/src/components/widgets/facet-nav/SelectionCriteriaPills.tsx
```diff
@@ -35,10 +35,13 @@ function getPillPropsFromColumnQueryFilter(
...
 ): SelectionCriteriaPillProps[] {
   const { getColumnDisplayName } = queryVisualizationContext
-  const columnModel = queryContext.getColumnModel(queryFilter.columnName)!
+  const columnModel = queryContext.getColumnModel(queryFilter.columnName)
   // ColumnSingleValueQueryFilter and ColumnMultiValueQueryFilter both allow for a list of values
   // If there are more than _n_ values, consolidate to one pill
-  if (queryFilter.values.length > MAX_VALUES_IN_FILTER_FOR_INDIVIDUAL_PILLS) {
+  if (
+    queryFilter.values.length > MAX_VALUES_IN_FILTER_FOR_INDIVIDUAL_PILLS ||
+    !columnModel
```
nickgros, line 43: We don't have a column model because the filter is applied to a column on the table referenced in `definingSql`. 

For now it will look like this, I will open a new ticket to improve this:

image

nickgros: Ticket is intentionally vague because it's not clear to me how we want to approach this w.r.t. the design: https://sagebionetworks.jira.com/browse/PORTALS-2775

jay-hodgson (approved):

PR #415, opened 2023-08-16
PORTALS-2782: ability to customize the file ID column name, file name…

jay-hodgson:
… column name, and file (entity) version column name.  Needed for the Virtual Table export


apps/portals/src/configurations/elportal/synapseConfigs/cohortbuilder.ts
```diff
@@ -80,6 +80,9 @@ export const filesView: SynapseConfig = {
     rgbIndex,
     name: 'Data',
     showExportToCavatica: true,
+    fileIdColumnName: 'fileId',
+    fileNameColumnName: 'fileId',
```
nickgros, line 84: Is ID as name expected or should we use the name (if that's currently available in the virtual table--maybe it needs to be updated)?

packages/synapse-react-client/src/synapse-queries/entity/useExportToCavatica.ts
```diff
@@ -21,7 +24,7 @@ export function useExportToCavatica(
     try {
       // add drs_uri to select
       const selectColumnsList = selectColumns?.map(col => col.name).join(',')
-      const sql = `SELECT CONCAT('drs://repo-prod.prod.sagebase.org/syn', id, '.', currentVersion) AS drs_uri, name as file_name, ${selectColumnsList} FROM ${originalSql.slice(
+      const sql = `SELECT CONCAT('drs://repo-prod.prod.sagebase.org/syn', ${fileIdColumnName}, '.', ${fileVersionColumnName}) AS drs_uri, ${fileNameColumnName} as file_name, ${selectColumnsList} FROM ${originalSql.slice(
```
nickgros, line 27: might be worth checking this SQL in useExportToCavatica.test.ts under the default case and custom parameters case

`expect(mockGetDownloadFromTableRequest.mock.calls[0][0]).equals('SELECT ...")`

apps/portals/src/configurations/elportal/synapseConfigs/cohortbuilder.ts
jay-hodgson: Rather use the name once it's available in the Virtual Table (VT).  Currently this functionality is broken because the entity version has not been selected in the VT!

packages/synapse-react-client/src/synapse-queries/entity/useExportToCavatica.ts
jay-hodgson: Good idea

nickgros (approved):

PR #419, opened 2023-08-21
SWC-6484 - SynapseTable rewrite and performance improvements 

nickgros:
Work in progress


packages/synapse-react-client/package.json
```diff
@@ -111,7 +112,8 @@
     "ua-parser-js": "^1.0.35",
     "universal-cookie": "^4.0.4",
     "use-deep-compare-effect": "^1.8.1",
-    "xss": "^1.0.14"
+    "xss": "^1.0.14",
+    "zustand": "^4.4.1"
```
nickgros, line 116: [zustand](https://github.com/pmndrs/zustand) is a state management library that provides far superior performance to React Context for write-many workloads.

There are many options in this space, I chose zustand because it's small and already included in our bundle (used by reactflow, which powers the ProvenanceGraph)

packages/synapse-react-client/src/components/ChallengeDataDownload/Renderers.tsx
```diff
@@ -34,10 +38,19 @@ export function SizeRenderer<T extends EntityIdAndVersionNumber>(
...
+    props.rowData.versionNumber,
+  )
+  if (!entity) {
+    return <></>
+  }
   return (
     <DirectDownload
+      fileHandleId={entity?.dataFileHandleId}
       associatedObjectId={props.rowData.entityId}
+      associatedObjectType={FileHandleAssociateType.FileEntity}
```
nickgros, line 52: Changed `DirectDownload` to no longer accept an entity ID, you must now provide the full FileHandleAssociation tuple.

So this is one of a few places where we must now fetch the FileEntity beforehand.

packages/synapse-react-client/src/components/DirectDownload/DirectDownload.tsx
```diff
@@ -1,44 +1,38 @@
...
 import { useSynapseContext } from '../../utils/context/SynapseContext'
 import { TOOLTIP_DELAY_SHOW } from '../SynapseTable/SynapseTableConstants'
 import { UAParser } from 'ua-parser-js'
 import { Tooltip } from '@mui/material'
 import { implementsExternalFileHandleInterface } from '../../utils/types/IsType'
+import { useGetFileBatch } from '../../synapse-queries/file/useFiles'
 
 export type DirectFileDownloadProps = {
   associatedObjectId: string
-  entityVersionNumber?: string
```
nickgros, line 24: For this component I wanted to use react-query so I could pre-fill the cache in the SynapseTable case. However, we had complicated branching logic because you could either pass a full file handle association, or just pass an entity ID and version number and the component would magically figure out the file handle association for you.

IMO supporting both configurations in one component was confusing, especially when trying to integrate with react-query, so this was simplified to only accept a file handle association.

I don't think the FileEntity-specific case was adding much value, but if we think it's useful, we could add a custom component that uses this one, rather than intertwining logic for identifying the file handle association and using it.

packages/synapse-react-client/src/components/EntityFinder/tree/EntityTree.tsx
```diff
@@ -206,22 +206,20 @@ export function EntityTree(props: EntityTreeProps) {
...
     FinderScope.CURRENT_PROJECT === scope &&
     projectId &&
-    fetchProjectError &&
-    fetchProjectError.status === 403
+    !isLoadingProjectHeader &&
+    !project
   ) {
-    // The user doesn't have access to the project, so let's change the scope to something else
+    // The header wasn't returned, so the user doesn't have access to the current project
+    // Let's change the scope to something else
```
nickgros, line 222: The implementation was changed to use [POST /entity/header](https://rest-docs.synapse.org/rest/POST/entity/header.html), which is subtly different because you CAN fetch a version, but it's a batch API so it doesn't return a 403 if unauthorized.

So we have to make the assumption that if it's not present in the response body, it's because the user doesn't have access. In theory this ID came from an existing project, so assuming 403 (vs 404) seems safe.

Tracking a request for the ideal service in https://sagebionetworks.jira.com/browse/PLFM-7989

packages/synapse-react-client/src/components/EntityLink.tsx
```diff
@@ -38,9 +38,13 @@ export const EntityLink = (props: EntityLinkProps) => {
...
-  const { data: fetchedEntity, error } = useGetEntity(entityId, versionNumber, {
-    enabled: !!entityId && typeof entityOrId === 'string',
-  })
+  const { data: fetchedEntity, error } = useGetEntityHeader(
+    entityId,
+    versionNumber,
+    {
+      enabled: !!entityId && typeof entityOrId === 'string',
+    },
+  )
```
nickgros, line 47: No need to fetch the entire entity. We can just fetch the header (which can also be pre-fetched in bulk for SynapseTable!)

packages/synapse-react-client/src/components/QueryContext/QueryContext.tsx
```diff
@@ -84,20 +83,6 @@ export type QueryContextType<
...
-  /** State updater for `selectedRows`  */
-  setSelectedRows: React.Dispatch<React.SetStateAction<Row[]>>
-  /** The set of columns that defines a uniqueness constraint on the table for the purposes of filtering based on row selection.
-   * Note that Synapse tables have no internal concept of a primary key.
-   */
-  rowSelectionPrimaryKey?: string[]
-  /** Whether the user has selected any rows */
-  hasSelectedRows: boolean
-  /** Method used to determine if an individual row is selected */
-  getIsRowSelected: (row: Row) => boolean
```
nickgros, line 100: Selection fields removed from QueryContext and now included in a Zustand store

PR #425, opened 2023-08-22
PORTALS-2772, PORTALS-2276

jay-hodgson:


packages/synapse-react-client/src/components/SynapseTable/SendToCavaticaConfirmationDialog.tsx
```diff
@@ -78,7 +78,6 @@ export default function SendToCavaticaConfirmationDialog(
   const { data: actions, isLoading } = useGetActionsRequiredForTableQuery(
     cavaticaQueryRequest,
     data?.columnModels as ColumnModel[],
-    undefined,
```
jay-hodgson, line 81: We never surfaced a param in the config to set this value (the file column model id), and we may never need to.

jay-hodgson:
getFileColumnModelId() is well tested, we just need to use it!

packages/synapse-react-client/src/components/SynapseTable/SendToCavaticaConfirmationDialog.tsx
nickgros: Oh I see, we may never need this because our heuristic for picking a column in `getFileColumnModelId` might always be good enough

nickgros (approved):
Looks great!

PR #426, opened 2023-08-24
SWC-6484 - QueryWrapper/SynapseTable refactor prework

nickgros:
- Fix a case where Project Views would show "add to download cart" option
- Rewrite many tests to rely less on QueryWrapper/QueryContext implementation details and to be type-checked
- Pull SynapseCardLabel and tests out of GenericCard
- Move QueryVisualizationWrapper.tsx into its own folder
- Flesh out a few entity mocks, add a team mock
- Add stories for SynapseTable, UserOrTeamBadge
- Hash the access token in the React Query cache to make devtools easier to use


packages/synapse-react-client/src/components/CardContainer/CardContainer.integration.test.tsx
```diff
@@ -0,0 +1,147 @@
...
+        {queryContext => {
+          capturedQueryContext = queryContext as InfiniteQueryContextType
+          return (
+            <QueryVisualizationWrapper unitDescription={unitDescription}>
+              <CardContainer {...props} />
+            </QueryVisualizationWrapper>
+          )
+        }}
+      </QueryContextConsumer>
+    </QueryWrapper>,
```
nickgros, line 65: Most of these test file changes have the same tests, but now
1. Use QueryWrapper and mock table query endpoints using mock-service-worker instead of creating our own QueryContext object, which exposed too many implementation details that broke on a refactor
2. Move test file to `src/` folder to be fully type-checked

packages/synapse-react-client/src/components/GenericCard/GenericCard.tsx
```diff
@@ -211,208 +206,6 @@ export const VersionLabel: React.FC<{
   )
 }
 
-type SynapseCardLabelProps = {
```
nickgros, line 214: SynapseCardLabel component moved to its own file, no changes

packages/synapse-react-client/src/components/GenericCard/SynapseCardLabel.test.tsx
```diff
@@ -0,0 +1,204 @@
+import { LabelLinkConfig } from '../CardContainerLogic'
+import { ColumnTypeEnum, SelectColumn } from '@sage-bionetworks/synapse-types'
+import { render, screen, within } from '@testing-library/react'
+import { SynapseCardLabel } from './SynapseCardLabel'
+import { createWrapper } from '../../testutils/TestingLibraryUtils'
+import React from 'react'
+
+describe('SynapseCardLabel tests', () => {
```
nickgros, line 8: Tests extracted from GenericCard.integration.test with minimal changes

packages/synapse-react-client/src/components/GenericCard/SynapseCardLabel.tsx
```diff
@@ -0,0 +1,222 @@
...
+  columnName: string
+  labelLink: CardLink | MarkdownLink | ColumnSpecifiedLink | undefined
+  selectColumns: SelectColumn[] | undefined
+  columnModels: ColumnModel[] | undefined
+  isHeader: boolean
+  className?: string
+  rowData: Row['values']
+  rowId?: number
+}
+export const SynapseCardLabel: React.FC<SynapseCardLabelProps> = props => {
```
nickgros, line 34: Pulled out from GenericCard.tsx, no changes

packages/synapse-react-client/src/components/QueryVisualizationWrapper/QueryVisualizationWrapper.integration.test.tsx
```diff
@@ -112,7 +103,7 @@ describe('QueryVisualizationWrapper', () => {
         }),
       )
       firstSetOfVisibleColumns =
-        onContextReceived.mock.lastCall[0].columnsToShowInTable
```
nickgros, line 115: Many changes to make TypeScript happy

packages/synapse-react-client/src/components/SynapseTable/SynapseTable.stories.tsx
```diff
@@ -0,0 +1,90 @@
...
+import { MOCK_REPO_ORIGIN } from '../../utils/functions/getEndpoint'
+import queryResultBundleJson from '../../mocks/query/syn16787123'
+import SynapseTable from './SynapseTable'
+import { QueryWrapper } from '../QueryWrapper'
+import { QueryVisualizationWrapper } from '../QueryVisualizationWrapper/QueryVisualizationWrapper'
+import { SynapseConstants } from '../../utils'
+import { DEFAULT_PAGE_SIZE } from '../../utils/SynapseConstants'
+import { mockTableEntity } from '../../mocks/entity/mockTableEntity'
+
+const meta: Meta = {
```
nickgros, line 14: I don't think these stories work yet, but if not, they will in a few PRs

packages/synapse-react-client/src/components/SynapseTable/SynapseTableUtils.ts
```diff
@@ -114,3 +115,22 @@ export function isSortableColumn(column: ColumnType) {
       return false
   }
 }
+
+export function getDefaultPrimaryKey(
```
nickgros, line 119: Extracted behavior from useTableRowSelection. In a future commit, this will be used, and the logic that this duplicates will be removed from useTableRowSelection.

packages/synapse-react-client/src/components/SynapseTable/table-top/DownloadOptions.integration.test.tsx
```diff
@@ -0,0 +1,262 @@
...
+      expect(
+        screen.queryByRole('button', { name: ADD_ALL_FILES_TO_DOWNLOAD_CART }),
+      ).not.toBeInTheDocument()
+    })
+
+    // Verify that programmatic download isn't disabled
+    expect(programmaticOptionsMenuItem.classList.contains('disabled')).toBe(
+      false,
+    )
+  })
```
nickgros, line 177: By turning this into an integration test (fetching a full table) I discovered and fixed a new bug caused by the new download cart feature -- Project Views should probably never show "add to download cart", but they do before this PR

packages/synapse-react-client/src/components/SynapseTable/table-top/DownloadOptions.tsx
```diff
@@ -37,9 +41,13 @@ export const DownloadOptions: React.FunctionComponent<
...
   const fileColumnId = getFileColumnModelId(queryResultBundle?.columnModels)
-  const showAddQueryToDownloadList =
-    fileColumnId ?? canTableQueryBeAddedToDownloadList(entity)
 
+  const isEntityViewWithoutFiles =
+    entity && isEntityView(entity) && !hasFilesInView(entity)
+
+  const showAddQueryToDownloadList =
+    !isEntityViewWithoutFiles &&
+    (fileColumnId || canTableQueryBeAddedToDownloadList(entity))
```
nickgros, line 50: If this is an EntityView without files, _never_ show "Add to download list". The likely ENTITY_ID column will never have a value that can be added to the download list.

packages/synapse-react-client/src/components/TeamBadge.tsx
```diff
@@ -9,12 +13,28 @@ export type TeamBadgeProps = {
...
+  if (teamId == AUTHENTICATED_PRINCIPAL_ID) {
+    icon = 'public'
+    teamName = 'All registered Synapse users'
+    disableHref = true
+  }
+  if (teamId == PUBLIC_PRINCIPAL_ID) {
+    icon = 'public'
+    teamName = 'Anyone on the web'
+    disableHref = true
+  }
```
nickgros, line 30: Special cases for these teams--SynapseTableCell does this. In a future change, SynapseTableCell will use this component, so this brings parity.

packages/synapse-react-client/src/components/TeamBadge.tsx
```diff
@@ -9,12 +13,28 @@ export type TeamBadgeProps = {
...
+
+  const Tag = disableHref ? 'span' : 'a'
```
nickgros, line 32: Use `span` if there's no href

nickgros, line 32: If there's no `href`, don't use `a` because we don't want link styling

packages/synapse-react-client/src/mocks/msw/handlers/tableQueryHandlers.ts
```diff
@@ -54,14 +54,15 @@ export function getHandlersForTableQuery(
   response: QueryResultBundle,
   backendOrigin = getEndpoint(BackendDestinationEnum.REPO_ENDPOINT),
 ) {
-  const asyncJobId = uniqueId()
-  let queryBundleRequest: QueryBundleRequest | undefined
+  // We must map generated async job IDs to requests to disambiguate multiple calls to these endpoints
+  const mapOfRequests = new Map<string, QueryBundleRequest>()
```
nickgros, line 58: If we have multiple table query requests, keep track of IDs so the correct response is returned.

packages/synapse-react-client/src/synapse-queries/KeyFactory.ts
```diff
@@ -192,7 +193,12 @@ export class KeyFactory {
    * @private
    */
   private getKey(...args: any[]): QueryKey {
-    return [this.accessToken, ...removeTrailingUndefinedElements(args)]
+    return [
+      this.accessToken === undefined
+        ? this.accessToken
+        : btoa(String(hashCode(this.accessToken))),
+      ...removeTrailingUndefinedElements(args),
+    ]
```
nickgros, line 201: All react-query cache keys include the current access token. In the react-query devtools, the entire JWT is visible, which makes it hard to see the actual data and relevant key values. So, hash the token to shorten it and make the devtools usable.

packages/synapse-react-client/src/components/SynapseTable/SynapseTableUtils.ts
```diff
@@ -114,3 +115,22 @@ export function isSortableColumn(column: ColumnType) {
...
+  entity?: Table,
+  columnModels?: ColumnModel[],
+): string[] | undefined {
+  if (
+    isFileViewOrDataset(entity) &&
+    columnModels?.find(cm => cm.name === 'id')
```
jay-hodgson, line 125: Also verify that the id column model type is entity id?

packages/synapse-react-client/src/components/SynapseTable/table-top/DownloadOptions.integration.test.tsx
jay-hodgson: Nice catch!

packages/synapse-react-client/src/utils/functions/StringUtils.ts
```diff
@@ -16,3 +16,19 @@ export function hex2ascii(inputString: string): string {
...
+ * Returns a hash code from a string
+ * @param  {String} str The string to hash.
+ * @return {Number}    A 32bit integer
+ * @see https://stackoverflow.com/a/8831937
+ */
+export function hashCode(str: string) {
+  let hash = 0
+  for (let i = 0, len = str.length; i < len; i++) {
+    const chr = str.charCodeAt(i)
+    hash = (hash << 5) - hash + chr
```
jay-hodgson, line 30: May want to include the [algorithm](https://devdocs.io/openjdk~8/java/lang/string#hashCode--) in the comments for posterity (but then again, I didn't verify that this is the actual result from this function!).
`s[0]*31^(n-1) + s[1]*31^(n-2) + ... + s[n-1]`

jay-hodgson (approved):

packages/synapse-react-client/src/components/SynapseTable/SynapseTableUtils.ts
nickgros: Since it's a view I think the data in this column has to be the unique synID, so it shouldn't matter if it's entity ID or string (or number, if that cast works?) because this is used for row selection.

PR #428, opened 2023-08-24
SWC-6484 - SynapseTable rewrite

nickgros:
SynapseTable:
- Almost completely rewritten
- Use @tanstack/react-table to transform data and render the table
- Create columns and custom renderers for synthetic columns in SynapseTableRenderers
- Remove use of custom in-memory cache (use of react-query cache will be added in a later commit).

DirectDownload:
- Simplify logic, no longer handles file entities. Only accepts a complete FileHandleAssociation
- Created FileEntityDirectDownload to fetch file data and render DirectDownload

useGetEntityHeaders, useGetFileBatch:
- After fetching a batch of data, insert the atomic results into the cache to increase the likelihood of a cache hit. This will allow "pre-fetching" data for SynapseTable in a future commit.

EntityHeader fetching:
- SynapseClient.getEntityHeader now supports fetching a versioned header by using the batch API.


packages/synapse-react-client/src/components/ChallengeDataDownload/Renderers.tsx
```diff
@@ -35,9 +35,10 @@ export function DownloadRenderer<T extends EntityIdAndVersionNumber>(
...
 ) {
   return (
-    <DirectDownload
-      associatedObjectId={props.rowData.entityId}
+    <FileEntityDirectDownload
+      entityId={props.rowData.entityId}
+      entityVersionNumber={props.rowData.versionNumber}
       stopPropagation={true}
-    ></DirectDownload>
+    />
```
nickgros, line 42: `DirectDownload` was changed to no longer accept entity IDs to reduce complexity. `FileEntityDirectDownload` was created to handle this case, so use it here.

packages/synapse-react-client/src/components/DirectDownload/DirectDownload.tsx
```diff
@@ -1,44 +1,38 @@
...
 import { useSynapseContext } from '../../utils/context/SynapseContext'
 import { TOOLTIP_DELAY_SHOW } from '../SynapseTable/SynapseTableConstants'
 import { UAParser } from 'ua-parser-js'
 import { Tooltip } from '@mui/material'
 import { implementsExternalFileHandleInterface } from '../../utils/types/IsType'
+import { useGetFileBatch } from '../../synapse-queries/file/useFiles'
 
 export type DirectFileDownloadProps = {
   associatedObjectId: string
-  entityVersionNumber?: string
```
nickgros, line 24: No longer accept `entityVersionNumber`. The FileHandleAssociation tuple of (file handle ID, object ID, object type) must now always be passed by the caller.

packages/synapse-react-client/src/components/DirectDownload/DirectDownload.tsx
```diff
@@ -93,36 +87,42 @@ function DirectDownloadIcon(props: DirectDownloadIconProps) {
...
+  }
+
+  // Get file handle data to ensure we have access / display information
+  const { data } = useGetFileBatch({
+    requestedFiles: [fileHandleAssociation],
+    includeFileHandles: true,
+    // don't get the presigned URL with this query, it may expire before the user clicks the download button
+    includePreSignedURLs: false,
+    includePreviewPreSignedURLs: false,
+  })
```
nickgros, line 115: Use react-query to get the file handle information (not the presigned URL, which will still be fetched on click)

packages/synapse-react-client/src/components/DirectDownload/DirectDownload.tsx
```diff
@@ -199,49 +199,6 @@ const DirectDownload: React.FunctionComponent<
     return getFiles(batchFileRequest, accessToken)
   }
 
-  const getFileEntityFileHandle = () => {
```
nickgros, line 202: Removed, this component no longer is concerned with the `Entity` object

packages/synapse-react-client/src/components/DirectDownload/FileEntityDirectDownload.tsx
```diff
@@ -0,0 +1,42 @@
...
+
+export type FileEntityDirectDownloadProps = Omit<
+  DirectFileDownloadProps,
+  'associatedObjectId' | 'associatedObjectType' | 'fileHandleId'
+> & {
+  entityId: string
+  entityVersionNumber?: number
+}
+
+export default function FileEntityDirectDownload(
```
nickgros, line 16: New component to handle the FileEntity behavior removed from `DirectDownload`. IMO much simpler to do it this way

packages/synapse-react-client/src/components/DownloadCart/RequestDownloadCard.tsx
```diff
@@ -24,10 +23,13 @@ export const RequestDownloadCard: React.FunctionComponent<
...
-    useErrorBoundary: true,
-  })
-  const entityHeader: EntityHeader | undefined = data?.results[0]
+  const { data: entityHeader, isLoading } = useGetEntityHeader(
+    entityId,
+    undefined,
+    {
+      useErrorBoundary: true,
+    },
+  )
```
nickgros, line 32: Only fetching one header, so use the singular version of the query

packages/synapse-react-client/src/components/EntityFinder/tree/EntityTree.tsx
```diff
@@ -206,22 +206,20 @@ export function EntityTree(props: EntityTreeProps) {
...
     FinderScope.CURRENT_PROJECT === scope &&
     projectId &&
-    fetchProjectError &&
-    fetchProjectError.status === 403
+    !isLoadingProjectHeader &&
+    !projectHeader
   ) {
-    // The user doesn't have access to the project, so let's change the scope to something else
+    // The header wasn't returned, so the user doesn't have access to the current project
+    // Let's change the scope to something else
```
nickgros, line 222: `useGetEntityHeader` was changed to allow requesting a version number (which doesn't matter here because this is a project). However, the actual service we use to fetch a versioned entity header doesn't return a 4XX status on unauthorized, because it's a batch API.

I chose not to have two functions that fetch the same information using different services, so we just need to change how we handle the failure case here.

I filed a PLFM ticket to close the existing gap here: https://sagebionetworks.jira.com/browse/PLFM-7989

packages/synapse-react-client/src/components/EntityLink.tsx
```diff
@@ -38,9 +38,13 @@ export const EntityLink = (props: EntityLinkProps) => {
...
-  const { data: fetchedEntity, error } = useGetEntity(entityId, versionNumber, {
-    enabled: !!entityId && typeof entityOrId === 'string',
-  })
+  const { data: fetchedEntity, error } = useGetEntityHeader(
+    entityId,
+    versionNumber,
+    {
+      enabled: !!entityId && typeof entityOrId === 'string',
+    },
+  )
```
nickgros, line 47: We only need EntityHeader data here.

In a future change, this query's cache data will be pre-loaded by a batch request before SynapseTable renders many instances of this component.

packages/synapse-react-client/src/components/GenericCard/GenericCard.integration.test.tsx
```diff
@@ -1,37 +1,60 @@
 import { render, screen, within } from '@testing-library/react'
```
nickgros, line 1: Test file was moved into `src/` and is now type-checked

packages/synapse-react-client/src/components/GenericCard/GenericCard.integration.test.tsx
```diff
@@ -1,37 +1,60 @@
...
-const renderComponent = (props: GenericCardProps) => {
   return render(
-    <GenericCard
-      {...props}
-      queryVisualizationContext={{
-        getColumnDisplayName: jest.fn().mockImplementation(col => col),
-      }}
-    />,
+    <QueryWrapper initQueryRequest={request}>
+      <QueryVisualizationWrapper>
```
nickgros, line 54: Use real `QueryWrapper`, just like most of the test changes in #426

packages/synapse-react-client/src/components/GenericCard/GenericCard.integration.test.tsx
```diff
@@ -1,37 +1,60 @@
...
+  tableType: 'TableEntity' | 'EntityView',
+) => {
+  const request = cloneDeep(mockQueryBundleRequest)
+  if (tableType === 'TableEntity') {
+    request.entityId = mockTableEntity.id
+    request.query.sql = `SELECT * FROM ${mockTableEntity.id}`
+  } else if (tableType === 'EntityView') {
+    request.entityId = mockFileViewEntity.id
+    request.query.sql = `SELECT * FROM ${mockFileViewEntity.id}`
+  }
```
nickgros, line 50: Added a quick way to change between the data we load for this component

packages/synapse-react-client/src/components/GenericCard/GenericCard.tsx
```diff
@@ -716,3 +723,15 @@ export default class GenericCard extends React.Component<
...
+  const { entity: table } = useQueryContext()
+  const queryVisualizationContext = useQueryVisualizationContext()
+  return (
+    <_GenericCard
+      {...props}
+      table={table}
+      queryVisualizationContext={queryVisualizationContext}
+    />
+  )
+}
```
nickgros, line 737: Just added a function component wrapper that can get the context values and pass into the internal component.

This pattern will also work for the next state management solution coming in a later PR.

packages/synapse-react-client/src/components/QueryWrapper/QueryWrapper.test.tsx
```diff
@@ -5,53 +5,52 @@ import {
   LockedColumn,
   QueryContextType,
```
nickgros, line 6: Test suite is now typechecked

packages/synapse-react-client/src/components/QueryWrapper/QueryWrapper.test.tsx
```diff
@@ -5,53 +5,52 @@ import {
...
+let setSelectedRows: QueryContextType['setSelectedRows'] | undefined
 
-const QueryContextReceiver = jest.fn((props: any) => {
+const QueryContextReceiver = () => {
   // An error would be thrown if context was not provided by QueryWrapper
   const context = useQueryContext()
+  isLoadingNewBundleValue = context.isLoadingNewBundle
+  currentQueryDataValue = context.data
+  selectedRows = context.selectedRows
+  setSelectedRows = context.setSelectedRows
```
nickgros, line 45: Having individual variables for different state items will make more sense when the new state manager is pulled in. This is sort of "intermediary" code that works with the existing context but can be easily changed

packages/synapse-react-client/src/components/QueryWrapperPlotNav/QueryWrapperPlotNav.tsx
```diff
@@ -93,14 +90,30 @@ export type QueryWrapperPlotNavProps = SearchParams &
...
+  | 'customControls'
+  | 'fileIdColumnName'
+  | 'fileNameColumnName'
+  | 'fileVersionColumnName'
+> & {
+  isFullTextSearchEnabled: boolean
+  remount: () => void
+}
+
+function QueryWrapperPlotNavContents(props: QueryWrapperPlotNavContentsProps) {
```
nickgros, line 115: Contents of `QueryWrapperPlotNav` were extracted into an individual component because we won't be able to use the `Context.Consumer` pattern with the new state management tool.

packages/synapse-react-client/src/components/SynapseTable/RowSelection/RowSelectionUI.tsx
```diff
@@ -54,7 +55,9 @@ export function RowSelectionUI(props: RowSelectionUIProps) {
             color="primary"
             max={Infinity}
           />
-          <Typography variant="body1">Rows Selected</Typography>
+          <Typography variant="body1">
+            {pluralize('Row', selectedRowCount)} Selected
+          </Typography>
```
nickgros, line 60: Fix "1 Rows Selected"

packages/synapse-react-client/src/components/SynapseTable/SendToCavaticaConfirmationDialog.integration.test.tsx
```diff
@@ -0,0 +1,183 @@
+import React from 'react'
```
nickgros, line 1: Last PR accidentally deleted this test suite. It was moved into `src/` to be type checked and converted into an integration test that uses a real `QueryWrapper`. The tests are otherwise mostly unchanged.

packages/synapse-react-client/src/components/SynapseTable/SynapseTable.integration.test.tsx
```diff
@@ -0,0 +1,611 @@
+import { render, screen, waitFor } from '@testing-library/react'
```
nickgros, line 1: New integration test suite mostly based on old tests. There's still probably a few cases to backfill

packages/synapse-react-client/src/components/SynapseTable/SynapseTable.integration.test.tsx
```diff
@@ -0,0 +1,611 @@
...
+          }
+          return res(ctx.status(200), ctx.json(responseBody))
+        },
+      ),
+    )
+  })
+  afterEach(() => server.restoreHandlers())
+  afterAll(() => server.close())
+
+  describe('Properly renders supplemental UI columns', () => {
```
nickgros, line 188: Not sure if I should parameterize this with `.each` and pass along the table type and expected values, or have duplicate code. Thoughts?

packages/synapse-react-client/src/components/SynapseTable/SynapseTable.integration.test.tsx
```diff
@@ -0,0 +1,611 @@
...
+      ...getHandlersForTableQuery(queryResultBundleWithoutRowIds),
+    )
+
+    renderTable({ showDirectDownloadColumn: true })
+    mockAllIsIntersecting(true)
+
+    expect(
+      screen.queryByTestId('AddToDownloadListCell'),
+    ).not.toBeInTheDocument()
+  })
```
nickgros, line 610: This test might not make much sense anymore since we can add an arbitrary column

packages/synapse-react-client/src/components/SynapseTable/SynapseTable.tsx
```diff
@@ -1,889 +1,275 @@
...
+  /** If true and entity is a file view or dataset, renders a column that allows the logged-in user to directly download the file, if they have permission */
+  showDirectDownloadColumn?: boolean
+  /** If true, the add to download list column will be hidden */
+  hideAddToDownloadListColumn?: boolean
+  /** Configuration to override cell renderers with e.g. a link to a portals detail page */
   columnLinks?: LabelLinkConfig
-  hideDownload?: boolean
 }
 
-export class SynapseTable extends React.Component<
```
nickgros, line 88: IMO not worth looking at this side of the diff, since this was a complicated class component with a bunch of unused code.

packages/synapse-react-client/src/components/SynapseTable/SynapseTable.tsx
```diff
@@ -1,889 +1,275 @@
...
+    showDirectDownloadColumn = showDownloadColumn,
+    hideAddToDownloadListColumn,
+    columnLinks,
+  } = props
+  const { getCurrentQueryRequest } = useQueryContext()
+  const { data, entity, isLoadingNewBundle, isRowSelectionVisible } =
+    useQueryContext()
+  const { columnsToShowInTable, NoContentPlaceholder } =
+    useQueryVisualizationContext()
+  const synapseTableContext = useMemo(() => ({ columnLinks }), [columnLinks])
```
nickgros, line 68: `columnLinks` needs to be passed to `SynapseTableCell` and there's no longer an easy way to do that, so just use a context object.

packages/synapse-react-client/src/components/SynapseTable/SynapseTable.tsx
```diff
@@ -1,889 +1,275 @@
...
-    } else {
-      // TODO: use event driven solution instead of timeout.
-      // We need to give SynapseTableCell time to render (which will change the column size).
-      // Alternatively, we may be able to use SizeMe (or withSize() on Cell component) to respond to width change.
-      setTimeout(() => {
-        this.resizer.reset(RESIZER_OPTIONS)
-      }, 1000)
-    }
-  }
+  const { sort, setSort } = useTableSort()
```
nickgros, line 77: Custom hook to map between our table query sorting logic and `@tanstack/react-table`'s sorting logic (just for keeping track of which columns are sorted, no client-side data sorting happening here)

packages/synapse-react-client/src/components/SynapseTable/SynapseTable.tsx
```diff
@@ -1,889 +1,275 @@
...
+      !hideAddToDownloadListColumn &&
+      !isRowSelectionVisible,
+  )
+
+  const columns: ColumnDef<Row, string | null>[] = useMemo(
+    () => [
+      rowSelectionColumn,
+      addToDownloadListColumn,
+      directDownloadColumn,
+      accessColumn,
```
nickgros, line 104: We define these columns and their cell components in a separate file (SynapseTableRenderers.tsx)

packages/synapse-react-client/src/components/SynapseTable/SynapseTable.tsx
```diff
@@ -1,889 +1,275 @@
...
+      ...(data?.columnModels?.map((columnModel, index) => {
```
nickgros, line 105: Create one react-table column for each ColumnModel

packages/synapse-react-client/src/components/SynapseTable/SynapseTable.tsx
```diff
@@ -1,889 +1,275 @@
...
+        return columnHelper.accessor(row => row.values[index], {
+          id: columnModel.name,
+          enableSorting: isSortableColumn(columnModel.columnType),
+          enableResizing: true,
+          header: TableDataColumnHeader,
+          cell: TableDataCell,
```
nickgros, line 111: Components for column headers and cells are defined in SynapseTableRenderers.tsx

packages/synapse-react-client/src/components/SynapseTable/SynapseTable.tsx
```diff
@@ -1,889 +1,275 @@
...
+  const columnVisibility: VisibilityState = useMemo(
+    () =>
+      (data?.columnModels ?? []).reduce((prev: VisibilityState, curr) => {
+        return {
+          ...prev,
+          [curr.name]: columnsToShowInTable.includes(curr.name),
+        }
+      }, prependColumnVisibility),
+    [columnsToShowInTable, data?.columnModels, prependColumnVisibility],
+  )
```
nickgros, line 144: And here compute the visibility state of the rest of the columns

packages/synapse-react-client/src/components/SynapseTable/SynapseTable.tsx
```diff
@@ -1,889 +1,275 @@
...
+
+  const table = useReactTable({
+    data: data?.queryResult?.queryResults.rows ?? [],
+    columns,
+    getCoreRowModel: getCoreRowModel(),
+    manualSorting: true,
+    manualFiltering: true,
```
nickgros, line 151: They have a built-in filtering/facet API but I'm not using it here.

packages/synapse-react-client/src/components/SynapseTable/SynapseTable.tsx
```diff
@@ -1,889 +1,275 @@
...
-  }
+  const { sort, setSort } = useTableSort()
 
-  disableResize() {
-    if (this.resizer) {
-      this.resizer.reset({ disable: true })
-    }
-  }
-
-  public async getEntityHeadersInData(forceRefresh: boolean) {
```
nickgros, line 210: I ripped out this custom cache and will add functionality to "pre-load" the react-query cache in the next PR

packages/synapse-react-client/src/components/SynapseTable/SynapseTableCell/EntityIdList.tsx
```diff
@@ -1,51 +1,28 @@
...
-      getEntityTypes()
-    }
-    return () => {
-      mounted = false
-    }
-  }, [entityIdList, inView])
+  const { data: entityHeaders } = useGetEntityHeaders(
+    entityIdList.map(id => ({ targetId: id })),
+    {},
+  )
```
nickgros, line 15: Use react-query cache

packages/synapse-react-client/src/components/SynapseTable/SynapseTableCell/SynapseTableCell.tsx
```diff
@@ -157,22 +146,29 @@ export function SynapseTableCell(props: SynapseTableCellProps) {
...
       )
     }
-    case ColumnTypeEnum.FILEHANDLEID:
+    case ColumnTypeEnum.FILEHANDLEID: {
+      let associatedObjectId = entity!.id!
+      let associatedObjectType = FileHandleAssociateType.TableEntity
+      if (isFileViewOrDataset(entity) && columnName === 'dataFileHandleId') {
+        associatedObjectId = String(rowId)
+        associatedObjectType = FileHandleAssociateType.FileEntity
+      }
```
nickgros, line 155: I think I fixed a bug here where users could not directly download `dataFileHandleId` columns in views (unless they were the file handle uploader)

packages/synapse-react-client/src/components/SynapseTable/SynapseTableCell/SynapseTableCell.tsx
```diff
@@ -299,3 +260,5 @@ export function SynapseTableCell(props: SynapseTableCellProps) {
   // TODO: If we don't have a id:data mapping, we should render a component that can fetch the required data, rather than breaking from the case.
   return <p className={isBold}>{columnValue}</p>
 }
+
+export default React.memo(SynapseTableCell)
```
nickgros, line 264: May slightly help performance

packages/synapse-react-client/src/components/SynapseTable/TopLevelControls/TopLevelControlsUtils.ts
```diff
@@ -17,9 +18,9 @@ const TO_DOWNLOAD_CART = 'to Download Cart'
  * @param data
  */
 export function getNumberOfResultsToInvokeAction(
-  hasSelectedRows: QueryContextType['hasSelectedRows'],
-  selectedRows: QueryContextType['selectedRows'],
-  data: QueryContextType['data'],
```
nickgros, line 22: These are going to change when this data is removed from the context object, so update the types now.

packages/synapse-react-client/src/components/widgets/facet-nav/FacetNav.tsx
```diff
@@ -67,26 +67,24 @@ const FacetNav: React.FunctionComponent<FacetNavProps> = ({
...
+        facets.map((item, index) => ({
           name: item.columnName,
           isHidden: index >= DEFAULT_VISIBLE_FACETS,
           plotType: 'PIE',
         })),
       )
-      setIsFirstTime(false)
     }
-  }, [data, isFirstTime, facetsToPlot])
+  }, [facetUiStateArray.length, facets])
```
nickgros, line 87: Refactored to get rid of `isFirstTime` state variable.

packages/synapse-react-client/src/components/widgets/facet-nav/FacetNavPanel.tsx
```diff
@@ -326,7 +326,11 @@ const FacetNavPanel: React.FunctionComponent<FacetNavPanelProps> = (
     </div>
   )
 
-  if ((!data && isLoadingNewBundle) || !facetToPlot) {
+  const columnModel = data?.columnModels?.find(
+    columnModel => columnModel.name === facetToPlot.columnName,
+  )
```
nickgros, line 331: Pull out the column model first, which just made debugging easier.

packages/synapse-react-client/src/synapse-queries/entity/useGetEntityHeaders.ts
```diff
@@ -13,23 +14,76 @@ export function useGetEntityHeaders(
...
+            keyFactory.getEntityHeadersQueryKey([requestWithoutVersionNumber]),
+            { results: [header], totalNumberOfResults: 1 },
+          )
+
+          queryClient.setQueryData(
+            keyFactory.getEntityHeaderQueryKey(requestedItem.targetId),
+            header,
+          )
+        }
+      }
```
nickgros, line 64: This is how we're going to pre-load the react-query cache with all of the entity headers

packages/synapse-react-client/src/synapse-queries/file/useFiles.ts
```diff
@@ -125,7 +126,29 @@ export function useGetFileBatch(
...
+        requestedFiles: [requestedItem],
+      }
+      const atomicResult: BatchFileResult = {
+        requestedFiles: [fileResult],
+      }
+      queryClient.setQueryData(
+        keyFactory.getBatchOfFiles(atomicRequest),
+        atomicResult,
+      )
+    })
```
nickgros, line 149: Pre-load the cache with all of the retrieved file handles.

packages/synapse-react-client/test/containers/SynapseTable/SynapseTableCell.integration.test.tsx
```diff
@@ -0,0 +1,348 @@
+import { render, screen } from '@testing-library/react'
```
nickgros, line 1: This were previously in the SynapseTable test suite, just pulled them out to their own file. No test changes.

apps/portals/src/configurations/adknowledgeportal/synapseConfigs/target_enabling_resources.ts
```diff
@@ -39,7 +39,6 @@ export const targetEnablingResourcesCardConfiguration: CardConfiguration = {
 
 export const targetEnablingResourcesTableConfiguration: QueryWrapperPlotNavProps['tableConfiguration'] =
   {
-    title: 'Target Enabling Resources',
```
jay-hodgson, line 42: Removes the title from the Synapse Table?

jay-hodgson: Seems fine for this case (where the tab name indicates the title), but may be confusing in a Details page (where there may be multiple tables)

packages/synapse-react-client/src/components/DirectDownload/DirectDownload.tsx
```diff
@@ -93,36 +87,42 @@ function DirectDownloadIcon(props: DirectDownloadIconProps) {
...
+    includePreviewPreSignedURLs: false,
+  })
+
+  const fileHandle = data?.requestedFiles[0]?.fileHandle
+  const hasFileAccess = Boolean(fileHandle)
+  const fileName = fileHandle?.fileName
+  const isExternalFile = fileHandle
+    ? implementsExternalFileHandleInterface(fileHandle)
+    : undefined
+  const externalURL: string | undefined = fileHandle
```
jay-hodgson, line 123: Could this be conditional based on `isExternalFile` and avoid the cast?

packages/synapse-react-client/src/components/DirectDownload/FileEntityDirectDownload.tsx
jay-hodgson: Agreed, nice refactor

packages/synapse-react-client/src/components/GenericCard/GenericCard.integration.test.tsx
```diff
@@ -332,29 +321,25 @@ describe('GenericCard tests', () => {
...
+          titleLinkConfig: undefined,
         },
-        columnModels: columnModelWithFileHandle,
-        titleLinkConfig: undefined,
-        tableId,
-      })
-      screen.getByTestId('ImageFileHandle')
+        'TableEntity',
+      )
+      await screen.findByTestId('ImageFileHandle')
```
jay-hodgson, line 337: I appreciate your changes to use async/await to improve test robustness as well

packages/synapse-react-client/src/components/SynapseTable/SearchV2.tsx
```diff
@@ -327,4 +331,7 @@ class Search extends React.Component<InternalSearchProps, SearchState> {
   }
 }
 
-export default Search
+export default function Search(props: SearchV2Props) {
+  const { data } = useQueryContext()
```
jay-hodgson, line 335: Assume same motivation here, to isolate the context to prep for removal?

packages/synapse-react-client/src/components/SynapseTable/SynapseTable.tsx
```diff
@@ -1,889 +1,275 @@
...
+    () => [
+      rowSelectionColumn,
+      addToDownloadListColumn,
+      directDownloadColumn,
+      accessColumn,
+      ...(data?.columnModels?.map((columnModel, index) => {
+        return columnHelper.accessor(row => row.values[index], {
+          id: columnModel.name,
+          enableSorting: isSortableColumn(columnModel.columnType),
+          enableResizing: true,
```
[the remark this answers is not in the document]
jay-hodgson: You don't know how happy this line makes me!

packages/synapse-react-client/src/components/SynapseTable/SynapseTable.tsx
```diff
@@ -1,889 +1,277 @@
...
+  )
+
+  const table = useReactTable({
+    data: data?.queryResult?.queryResults.rows ?? [],
+    columns,
+    getCoreRowModel: getCoreRowModel(),
+    manualSorting: true,
+    manualFiltering: true,
+    onSortingChange: setSort,
+    enableMultiSort: true,
```
jay-hodgson, line 155: In one design, I remember someone on the design team indicating that they wanted to remove the ability to sort multiple columns at once, but this doesn't need to be verified/implemented in this PR.

packages/synapse-react-client/src/components/SynapseTable/SynapseTableCell/SynapseTableCell.tsx
```diff
@@ -29,21 +24,19 @@ import EntityIdList from './EntityIdList'
...
+import { isFileViewOrDataset } from '../SynapseTableUtils'
+import { useQueryContext } from '../../QueryContext'
 
 export type SynapseTableCellProps = {
   columnType: ColumnType
   columnValue: string | null
   isBold: string
   columnLinkConfig?: CardLink | MarkdownLink | ColumnSpecifiedLink
-  mapEntityIdToHeader: Record<string, EntityHeader>
-  mapUserIdToHeader: Record<string, UserGroupHeader | UserProfile>
```
jay-hodgson, line 46: Yay!

packages/synapse-react-client/src/components/SynapseTable/SynapseTableCell/SynapseTableCell.tsx
jay-hodgson: Yes, I don't think accessing the FileHandle.dataFileHandle via the previous FHA (type=TableEntity) would work unless you were the file owner

jay-hodgson: Nice catch

packages/synapse-react-client/src/components/SynapseTable/SynapseTableCell/SynapseTableCell.tsx
```diff
@@ -210,43 +206,7 @@ export function SynapseTableCell(props: SynapseTableCellProps) {
...
-            <UserCard
-              userProfile={cachedData}
-              preSignedURL={cachedData.clientPreSignedURL}
-              size={'SMALL USER CARD'}
-            />
-          )
-        }
-      }
-      break
+      return <UserOrTeamBadge principalId={columnValue} />
```
jay-hodgson, line 209: I think this fixes https://sagebionetworks.jira.com/browse/PORTALS-2403, could you verify?

jay-hodgson: Oh wait, UserIdList would also need to use UserOrTeamBadge.

jay-hodgson (approved):

packages/synapse-react-client/src/components/EntityFinder/tree/EntityTree.tsx
nickgros: Can't be a 404 since this project ID is determined by looking at the path of an entity, but 401 is definitely possible. Will have to think about this a bit more

apps/portals/src/configurations/adknowledgeportal/synapseConfigs/target_enabling_resources.ts
nickgros: I removed this `title` prop from `SynapseTableProps` because it was unused prior to my changes. We have to remove it here or else there's a type error. I think in most of these cases, the title is set in a different place.

packages/synapse-react-client/src/components/SynapseTable/SearchV2.tsx
nickgros: Yep, Jotai (the new state manager) uses hooks so the change becomes a simple hook swap here.

packages/synapse-react-client/src/components/EntityFinder/tree/EntityTree.tsx
nickgros: I don't think we expose the Entity Finder anywhere to unauthenticated users. Even if we do, I think they will likely hit a 401 elsewhere before they hit it here. So while not ideal I think this is acceptable risk

packages/synapse-react-client/src/components/SynapseTable/SynapseTableCell/SynapseTableCell.tsx
nickgros: I went ahead and updated UserIdList in [this commit](https://github.com/Sage-Bionetworks/synapse-web-monorepo/pull/428/commits/9b1e7ea22dff3f0d9f30d21672700880d2497391)

PR #429, opened 2023-08-24
SWC-6484 - Use Jotai to manage many aspects of QueryWrapper/Table row selection state

nickgros:
[Jotai](https://jotai.org/) is an atomic state management tool that can provide better performance than React Context for frequently changing state values

Some benchmarks, tested against Portals built for production, while logged-in.

| | main | feature |
|---|---|---|
| ADKP Data, POST /entity/header calls | 25 (1 object per req) | 1 (30 objects in req)* |
| ADKP Data, POST /fileHandle/batch calls | 25 (1 object per req) | 1 (25 objects in req) |
| ELITE Portal Data By Files, POST /entity/header calls | 25 (1 object per req) | 1 (25 objects in req) |
| ELITE Portal Data By Files, POST /fileHandle/batch calls | 25 (1 object per req) | 25 (1 object per req)** |

\* The difference in number of entity headers fetched is accounted for by the entities represented in the benefactorId and projectId columns, which are not visible by default. In `main`, these are fetched when the column becomes visible. In the feature branch, these are fetched in the pre-fetch batch step.

\*\* The ELITE Portal file view doesn't include the `dataFileHandleId` column, so we cannot prefetch file handles using table data. The component falls back to fetching the entity, then fetching the individual file handle in series.

## Table row selection
Screen captures highlighting component re-renders using React Dev Tools extension. Note that on the `main` branch, I am throttled when clicking rapidly because React is doing so much work to re-render all of the table cells. On the feature branch, other than when making the initial selection, the entire QueryWrapperPlotNav component does not re-render on select/deselect.

Main:

https://github.com/Sage-Bionetworks/synapse-web-monorepo/assets/17580037/b54c69eb-b6a0-4312-a688-b7a32b7fd9a6 

Feature:

https://github.com/Sage-Bionetworks/synapse-web-monorepo/assets/17580037/aa4bd6dc-e24a-4082-9598-1ac12ee4b742


packages/synapse-react-client/package.json
```diff
@@ -58,6 +58,7 @@
     "downshift": "^6.1.12",
     "history": "^5.3.0",
     "immutable": "4.1.0",
+    "jotai": "^2.3.1",
```
nickgros, line 61: [Jotai](https://jotai.org/) is an atomic state library that we can use to get better performance than React Context for frequently-changing state.

packages/synapse-react-client/src/components/CardContainer/CardContainer.tsx
```diff
@@ -63,8 +65,8 @@ export function CardContainer(props: CardContainerProps) {
   } = props
   const infiniteQueryContext = useInfiniteQueryContext()
   const { NoContentPlaceholder } = useQueryVisualizationContext()
-  const { data, appendNextPageToResults, hasNextPage } = infiniteQueryContext
-
+  const { appendNextPageToResults, hasNextPage } = infiniteQueryContext
+  const data = useAtomValue(tableQueryDataAtom)
```
nickgros, line 69: `useAtomValue` is how we can access the value of a state variable.

You'll notice I did not completely remove `QueryContext`. In this pass I just focused on state referenced within SynapseTable and its child components because updates to these components were causing the largest performance hit.

packages/synapse-react-client/src/components/QueryVisualizationWrapper/QueryVisualizationWrapper.tsx
```diff
@@ -117,18 +120,19 @@ export function QueryVisualizationWrapper(
...
-  const {
-    data,
-    getCurrentQueryRequest,
-    isFacetsAvailable,
-    hasResettableFilters,
-  } = useQueryContext()
+  const columnAliases = useMemo(
+    () => props.columnAliases ?? {},
+    [props.columnAliases],
+  )
```
nickgros, line 131: The previous initializer (default value `{}`) was re-created every render, which triggered undesirable rerenders.

packages/synapse-react-client/src/components/QueryWrapper/QueryWrapper.tsx
```diff
@@ -37,11 +51,26 @@ export type QueryWrapperProps = React.PropsWithChildren<{
...
+export const tableQueryEntityAtom = atom<Table | undefined>(undefined)
+/** Returns true when loading a brand-new query result bundle. Will not be true when just loading the next page of query results. */
+export const isLoadingNewBundleAtom = atom<boolean>(false)
+
+/**
+ * A column name may be "locked" so that it is both (1) not shown to the user that the filter is active, and (2) not modifiable by the user.
+ * For example, we may show only the data matching a particular facet value on a Details Page without implying that the shown data is part of a larger table.
+ * The presence of a locked filter will result in a client-side modification of the active query and result bundle data.
+ */
+export const lockedColumnAtom = atom<LockedColumn | undefined>(undefined)
```
nickgros, line 67: Atoms that encapsulate state are created here.

packages/synapse-react-client/src/components/QueryWrapper/QueryWrapper.tsx
```diff
@@ -56,17 +85,14 @@ export function QueryWrapper(props: QueryWrapperProps) {
     combineRangeFacetConfig,
   } = props
 
-  // Must store requireConfirmationOnChange in state to break dependency cycle between useImmutableTableQuery and useTableRowsSelection
-  // TODO: consider combining these two hooks, or decoupling them in some other way
-  const [requireConfirmationOnChange, setRequireConfirmationOnChange] =
-    useState(false)
```
nickgros, line 62: These changes basically break this dependency cycle, so we can remove this state variable and derive it from the atom.

packages/synapse-react-client/src/components/QueryWrapper/QueryWrapper.tsx
```diff
@@ -135,55 +182,49 @@ export function QueryWrapper(props: QueryWrapperProps) {
...
-    hasSelectedRows,
-    selectedRows,
-    setSelectedRows,
-    getIsRowSelected,
-  } = useTableRowSelection({
-    entity: entity,
-    columnModels: data?.columnModels,
-    isRowSelectionVisible: isRowSelectionVisibleFromProps,
-    rowSelectionPrimaryKey: rowSelectionPrimaryKeyFromProps,
-  })
```
nickgros, line 150: Removed `useTableRowSelection`. It's all just atomic state now.

packages/synapse-react-client/src/components/QueryWrapper/QueryWrapper.tsx
```diff
@@ -210,3 +251,13 @@ export function QueryWrapper(props: QueryWrapperProps) {
...
 }
+
+export function QueryWrapper(props: QueryWrapperProps) {
+  // Wrap in a Jotai provider to ensure the Jotai atomic state is unique to this component tree
+  // i.e. other instances of QueryWrapper will not share state with this instance
+  return (
+    <Provider>
+      <_QueryWrapper {...props} />
+    </Provider>
+  )
```
nickgros, line 262: This `Provider` is essential to keep state local to this subtree

packages/synapse-react-client/src/components/QueryWrapper/QueryWrapper.tsx
```diff
@@ -210,3 +251,13 @@ export function QueryWrapper(props: QueryWrapperProps) {
...
+}
```
nickgros, line 263: While these atoms are globally accessible, it's not really global state thanks to this provider

packages/synapse-react-client/src/components/QueryWrapper/TableRowSelectionState.ts
```diff
@@ -0,0 +1,68 @@
+import { ColumnModel, Row } from '@sage-bionetworks/synapse-types'
+import { isEqual } from 'lodash-es'
+import { atom } from 'jotai'
+import { tableQueryDataAtom } from './QueryWrapper'
+
+export function getRowSelectionEqualityComparator(
```
nickgros, line 6: This function has not changed

packages/synapse-react-client/src/components/QueryWrapper/TableRowSelectionState.ts
```diff
@@ -0,0 +1,68 @@
+import { ColumnModel, Row } from '@sage-bionetworks/synapse-types'
```
nickgros, line 1: This file replaces the `useTableRowSelection`. It contains the table row selection atoms

packages/synapse-react-client/src/components/SynapseTable/SynapseTable.tsx
```diff
@@ -160,6 +170,8 @@ export function SynapseTable(props: SynapseTableProps) {
     },
   })
 
+  const { dataHasBeenPrefetched } = usePrefetchTableData()
```
nickgros, line 173: Delay rendering the table cells until we have prefetched all the data to prevent the waterfall effect

packages/synapse-react-client/src/components/SynapseTable/SynapseTable.tsx
```diff
@@ -238,16 +250,24 @@ export function SynapseTable(props: SynapseTableProps) {
...
-                          cell.getContext(),
+                        {renderPlaceholder ? (
+                          <p>
+                            <Skeleton width={'80%'} height={'20px'} />
+                          </p>
+                        ) : (
+                          flexRender(
+                            cell.column.columnDef.cell,
+                            cell.getContext(),
+                          )
```
nickgros, line 270: Delay rendering the table data cells until we have prefetched all the data to prevent the waterfall effect

packages/synapse-react-client/src/components/SynapseTable/usePrefetchTableData.ts
```diff
@@ -0,0 +1,158 @@
...
+ * fetch and de-normalize batch requests, so when the SynapseTableCell component renders, the data is already loaded
+ * in the cache. In practice, this avoids Synapse backend throttling and 'waterfall' content rendering.
+ */
+export function usePrefetchTableData(): { dataHasBeenPrefetched: boolean } {
+  const { isLoading: isLoadingEntityData } = usePrefetchEntityData()
+  const { isLoading: isLoadingFileHandleData } = usePrefetchFileHandleData()
+
+  // TODO: Add user profile prefetch
+  // User profile prefetch may not be easy/possible because the user card component fetches the UserProfile,
+  // But only UserGroupHeaders can be fetched in a batch request.
```
nickgros, line 153: Will file a ticket to follow up on this.

packages/synapse-react-client/src/components/QueryWrapper/QueryWrapper.tsx
nickgros, line 262: I verified that multiple instances of QueryWrapper on the same page don't share state (e.g. ADKP Study Details Page has multiple QueryWrapperPlotNav instances)

packages/synapse-react-client/package.json
nickgros: I chose Jotai over [Zustand](https://github.com/pmndrs/zustand) because it's a more natural shift from React Context (in that state is tied to a component tree vs a global store).

I chose Jotai over [Recoil](https://github.com/facebookexperimental/Recoil), which has a similar API, because Meta brands Recoil as "experimental" so it seems fairly risky to choose it.

Either of these libraries (and more, including [Redux](https://redux.js.org/), [Valtio](https://github.com/pmndrs/valtio), [MobX](https://mobx.js.org/README.html), I could go on...), could have been used to get the same performance benefits. I don't feel that we're locked-in to Jotai compared to any other option, but we may want to re-evaluate our decision later if we feel we need to keep relying on these tools.

jay-hodgson (approved):

packages/synapse-react-client/src/components/SynapseTable/usePrefetchTableData.ts
nickgros: https://sagebionetworks.jira.com/browse/SWC-6539

PR #442, opened 2023-08-25
Fix null case for hashCode

nickgros:
SWC apparently sets `accessToken` to `null`, which is an unhandled case (only `undefined` was handled)

### DIFF
--- a/packages/synapse-react-client/src/utils/functions/StringUtils.ts
+++ b/packages/synapse-react-client/src/utils/functions/StringUtils.ts
@@ -23,7 +23,10 @@ export function hex2ascii(inputString: string): string {
  * @return {Number}    A 32bit integer
  * @see https://stackoverflow.com/a/8831937
  */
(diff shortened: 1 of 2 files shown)


jay-hodgson (approved):

packages/synapse-react-client/src/utils/functions/StringUtils.ts
```diff
@@ -23,7 +23,10 @@ export function hex2ascii(inputString: string): string {
  * @return {Number}    A 32bit integer
  * @see https://stackoverflow.com/a/8831937
  */
-export function hashCode(str: string) {
+export function hashCode(str?: string | null) {
+  if (str == null) {
```
jay-hodgson, line 27: `|| str == undefined` ?

nickgros: `null == undefined` -> `true` :)

PR #445, opened 2023-08-28
SWC-6486 - Fix case where cached annotations would not load into the schematized annotation editor

nickgros:
SWC-6486

State management errors in useGetJson and SchemaDrivenAnnotationEditor led to a case where

1. The cache was 'hot', but since `useGetJson`'s return `annotations` was stored in state instead of derived from state, it was undefined when the data was loaded; i.e. there was one instance where `annotations` was `undefined` but `isLoading` was `false`
2. SchemaDrivenAnnotationEditor's `formData` was initialized with the undefined `annotations` value
3. RJSF takes the undefined `formData` value and transforms it to an empty object and somehow components become uncontrolled and get out of sync-this seems to be a bug in RJSF.

We can fix this by
1. Derive the value of `annotations` from the Entity JSON data, so it will never be undefined when `isLoading` is false
2. Initialize `formData` using `annotations`, so `undefined` will not passed into the `Form` component when the cache is hot
3. Prevent rendering the `Form` component when `formData` is undefined


packages/synapse-react-client/src/components/SchemaDrivenAnnotationEditor/SchemaDrivenAnnotationEditor.tsx
```diff
@@ -138,6 +133,11 @@ export function SchemaDrivenAnnotationEditor(
     useErrorBoundary: true,
   })
 
+  // Annotation fields fetched and modified via the form
+  const [formData, setFormData] = React.useState<
+    Record<string, unknown> | undefined
+  >(annotations)
```
nickgros, line 139: Note the initial value is now `annotations`, to prevent a case where this was `undefined` which broke the form component

packages/synapse-react-client/src/components/SchemaDrivenAnnotationEditor/SchemaDrivenAnnotationEditor.tsx
```diff
@@ -223,139 +223,146 @@ export function SchemaDrivenAnnotationEditor(
...
+              <Box display={'flex'} alignItems={'center'} gap={0.5}>
+                <Typography variant={'smallText1'}>
+                  <b>{entityJson.name}</b> has no annotations. Click the{' '}
+                </Typography>
+                <AddToList />
+                <Typography variant={'smallText1'}>
+                  button to annotate.
+                </Typography>
+              </Box>
+            </Alert>
```
nickgros, line 237: Added this flex container to fix a vertical alignment issue

packages/synapse-react-client/src/components/SchemaDrivenAnnotationEditor/SchemaDrivenAnnotationEditor.tsx
```diff
@@ -223,139 +223,146 @@ export function SchemaDrivenAnnotationEditor(
...
+          )}
+          {formData != undefined && (
```
nickgros, line 239: Added this check to not render if `undefined`. Not a problem for entities with no annotations, `formData` will be an empty object in that case. No other changes in this file--the remaining diff is from adding this conditional wrapper

packages/synapse-react-client/src/components/SchemaDrivenAnnotationEditor/SchemaDrivenAnnotationsEditor.integration.test.tsx
```diff
@@ -348,7 +347,85 @@ describe('SchemaDrivenAnnotationEditor tests', () => {
...
     )
-    await renderComponent()
+    const { container } = await renderComponent()
+    await screen.findByText('requires scientific annotations', { exact: false })
+
+    await waitFor(() => {
+      const countryInput = container.querySelector(
+        '#root_country',
+      ) as HTMLInputElement
+      expect(countryInput.value).toBe('USA')
```
nickgros, line 357: The `@testing-library/react` selectors do not work well with RJSF+MUI

packages/synapse-react-client/src/components/SchemaDrivenAnnotationEditor/SchemaDrivenAnnotationsEditor.integration.test.tsx
```diff
@@ -348,7 +347,85 @@ describe('SchemaDrivenAnnotationEditor tests', () => {
...
+    // Saving the form should maintain the existing annotations
+    await clickSave()
+    await waitFor(() =>
+      expect(updatedJsonCaptor).toHaveBeenCalledWith(
+        expect.objectContaining({ country: 'USA', state: 'Washington' }),
+      ),
+    )
+  })
+
+  it('Reloads the form data properly after remounting the component (SWC-6486)', async () => {
```
nickgros, line 369: This new test failed before adding changes, so I think this is a valid reproduction

packages/synapse-react-client/src/components/SchemaDrivenAnnotationEditor/SchemaDrivenAnnotationsEditor.integration.test.tsx
```diff
@@ -348,7 +347,85 @@ describe('SchemaDrivenAnnotationEditor tests', () => {
...
+      rerender(<SchemaDrivenAnnotationEditor {...defaultProps} />)
+    })
+    await screen.findByText('requires scientific annotations', { exact: false })
+
+    await waitFor(() => {
+      const countryInput = container.querySelector(
+        '#root_country',
+      ) as HTMLInputElement
+      expect(countryInput.value).toBe('USA')
+    })
```
nickgros, line 407: Now the cache is hot. When the bug occurred, the input field never got the value

packages/synapse-react-client/src/synapse-queries/entity/useEntity.ts
```diff
@@ -221,12 +217,21 @@ export function useGetJson(
...
+    [query.data],
+  )
+
+  const annotations = useMemo(
+    () =>
+      query?.data == undefined
+        ? undefined
+        : removeStandardEntityFields(query.data),
+    [query.data],
+  )
```
nickgros, line 234: These were previously state variables, so there was one "frame" where our data was loaded, so the returned `isLoading` was false, but these state variables hadn't updated yet, so they were still undefined. We can fix this by deriving the values from the `useQuery` result using `useMemo`.

packages/synapse-react-client/package.json
```diff
@@ -43,10 +43,10 @@
...
     "@react-hook/resize-observer": "^1.2.6",
     "@react-hookz/web": "^23.1.0",
-    "@rjsf/core": "^5.9.0",
-    "@rjsf/mui": "^5.9.0",
-    "@rjsf/utils": "^5.9.0",
-    "@rjsf/validator-ajv8": "^5.9.0",
+    "@rjsf/core": "5.12.1",
+    "@rjsf/mui": "5.12.1",
+    "@rjsf/utils": "5.12.1",
+    "@rjsf/validator-ajv8": "5.12.1",
```
nickgros, line 49: Bump while I'm here, and remove the `^` marker because we're relying on an experimental prop (for derived annotations)

jay-hodgson (approved):

PR #446, opened 2023-08-28
SWC-6540 - Use `selectColumns` in SynapseTable, not `columnModels`, which may not match the current query

nickgros:


packages/synapse-react-client/src/components/SynapseTable/SynapseTable.integration.test.tsx
```diff
@@ -447,4 +451,48 @@ describe('SynapseTable tests', () => {
...
+          concreteType: 'org.sagebionetworks.repo.model.table.RowSet',
+          tableId: MOCK_TABLE_ENTITY_ID,
+          etag: '53e1e27a-dbf3-4db3-acd1-dafca30b894c',
+          headers: [{ name: 'Number of Studies', columnType: 'INTEGER' }],
+          rows: [{ values: ['123'] }],
+        },
+      },
+    }
+
+    // Return a file view entity, so the download column would be shown if not for the missing row IDs
```
nickgros, line 478:
```suggestion
```

packages/synapse-react-client/src/components/SynapseTable/SynapseTable.tsx
```diff
@@ -172,6 +172,7 @@ export function SynapseTable(props: SynapseTableProps) {
 
   const { dataHasBeenPrefetched } = usePrefetchTableData()
 
+  console.log(columns, data)
```
nickgros, line 175:
```suggestion
```

PR #450, opened 2023-08-29
PORTALS-2732 - Center align checkbox column

nickgros:
Use table `meta` to define alignment per-column

### DIFF
--- a/packages/synapse-react-client/src/components/SynapseTable/SynapseTable.tsx
+++ b/packages/synapse-react-client/src/components/SynapseTable/SynapseTable.tsx
@@ -258,6 +258,7 @@ export function SynapseTable(props: SynapseTableProps) {
                         style={{
                           width: cell.column.getSize(),
                         }}
+                        align={(cell.column.columnDef.meta as any)?.align}
                       >
                         {renderPlaceholder ? (
                           <p>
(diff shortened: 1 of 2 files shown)


packages/synapse-react-client/src/components/SynapseTable/SynapseTableRenderers.tsx
```diff
@@ -45,43 +45,45 @@ function RowSelectionCell(props: CellContext<Row, unknown>) {
...
-        const cloneSelectedRows = [...selectedRows]
-        if (checked) {
-          cloneSelectedRows.push(row.original)
-        } else {
-          const index = cloneSelectedRows.findIndex(selectedRow =>
-            isEqual(selectedRow, row.original),
-          )
-          if (index > -1) {
-            cloneSelectedRows.splice(index, 1)
+    <div>
```
nickgros, line 48: wrap in a div so the CSS that defines padding for all table elements adds to the `.checkbox` class padding instead of replacing it

packages/synapse-react-client/src/components/SynapseTable/SynapseTable.tsx
```diff
@@ -258,6 +258,7 @@ export function SynapseTable(props: SynapseTableProps) {
                         style={{
                           width: cell.column.getSize(),
                         }}
+                        align={(cell.column.columnDef.meta as any)?.align}
```
nickgros, line 261: Read column alignment from the `meta` field, which can contain any column-related data we want to specify, and apply it to the `td` element

jay-hodgson: Useful and generalizable, I like it

PR #455, opened 2023-08-30
PORTALS-2793 - Use a larger runner, add a 1hr timeout, and remove the existing parallel job limit

nickgros:
Using a larger runner brings our `test` job worst-case time to around 17m: https://github.com/Sage-Bionetworks/synapse-web-monorepo/actions/runs/6030039839/job/16360992278?pr=455

With the regular runner we were consistently seeing times >30min

### DIFF
--- a/.github/workflows/build-test.yml
+++ b/.github/workflows/build-test.yml
@@ -11,7 +11,10 @@ on:
 
 jobs:
   build:
-    runs-on: ubuntu-latest
+    runs-on:
+      labels: ubuntu-22.04-4core-16GBRAM-150GBSSD
+    # Job should not last longer than 60 minutes--and the larger runner is billed per-minute ;)
+    timeout-minutes: 60
     # See supported Node.js release schedule at https://nodejs.org/en/about/releases/
     # We should test all active LTS releases, but not until our test suite is less flaky
     strategy:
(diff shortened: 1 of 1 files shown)


.github/workflows/build-test.yml
```diff
@@ -25,17 +25,20 @@ jobs:
...
         # We may want to invert the model and give each portal its own project in the workspace, so we can build the portals in parallel.
       - run: pnpm nx run portals:copy-test-configuration
-      - run: pnpm nx affected --target=build --parallel=2 --base=remotes/origin/main
+      - run: pnpm nx affected --target=build --base=remotes/origin/main
   test:
-    runs-on: ubuntu-latest
+    runs-on:
+      labels: ubuntu-22.04-4core-16GBRAM-150GBSSD
+    # Job should not last longer than 60 minutes--and the larger runner is billed per-minute ;)
+    timeout-minutes: 60
```
hallieswan, line 33: Very cool that this is so much faster! And good idea on the timeout :) Do we need to keep an eye on cost over time? This workflow runs more frequently (per PR to main and push to main) than the e2e tests (per push to develop), so may be higher monthly cost than the [estimates](https://docs.google.com/document/d/1m1I9SMOQw0wrUjYHYrYTWnz8wZez_W9Ljz5xGm2AGpw/edit?usp=sharing) for the e2e workflow.

hallieswan (approved):

.github/workflows/build-test.yml
nickgros: Good idea! I'll ballpark some numbers before merging

nickgros: I counted ~60 build/test runs in the last 2 weeks that ran for 20+ min, so let's estimate that we'll have ~120 worst-case (entire build) jobs per month

120 jobs/month * 18min avg job time * 60sec/min * $0.016 / sec = $2073/month, not including shorter test jobs that just run against portals. Might be worth comparing to monthly costs of a runner we maintain on a service like EC2.

jay-hodgson (approved):

PR #456, opened 2023-08-31
SWC-6531: expose HttpClient and BackendDestinationEnum in UMD, allow BackendDestinationEnum to be used as value

hallieswan:
Allows HttpClient and BackendDestinationEnum to be called when evaluating JS in the browser page in a Playwright test, e.g. `window.SRC.HttpClient.doGet(...)`

### DIFF
--- a/packages/synapse-react-client/src/utils/functions/index.ts
+++ b/packages/synapse-react-client/src/utils/functions/index.ts
@@ -8,7 +8,7 @@ import {
 } from './SqlFunctions'
 import { hex2ascii } from './StringUtils'
 import type { SQLOperator } from './SqlFunctions'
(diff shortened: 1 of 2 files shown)


packages/synapse-react-client/src/utils/functions/index.ts
```diff
@@ -8,7 +8,7 @@ import {
 } from './SqlFunctions'
 import { hex2ascii } from './StringUtils'
 import type { SQLOperator } from './SqlFunctions'
-import type { BackendDestinationEnum } from './getEndpoint'
+import { BackendDestinationEnum } from './getEndpoint'
```
hallieswan, line 11: Allow `BackendDestinationEnum` to be used as a value

nickgros (approved):

PR #457, opened 2023-08-31
SWC-6545: Bump react-query v3 to @tanstack/react-query v4

nickgros:
SWC-6545


packages/synapse-react-client/src/testutils/ReactQueryMockUtils.ts
```diff
@@ -31,6 +31,50 @@ export function getUseQuerySuccessMock<TData>(
...
     remove: jest.fn(),
     status: 'success',
+    failureReason: null,
+    isInitialLoading: false,
+    isPaused: false,
+    fetchStatus: 'idle',
+  }
+}
+
+export function getUseInfiniteQuerySuccessMock<TData>(
```
nickgros, line 41: Add mock for useInfiniteQuery, and use it in a couple of tests

PR #461, opened 2023-09-06
PORTALS-2794

jay-hodgson:


apps/portals/src/configurations/elportal/resources.ts
```diff
@@ -8,7 +8,7 @@ export const studiesSql = 'SELECT * FROM syn51209885 ORDER BY isFeatured DESC'
 export const speciesSql = 'SELECT * FROM syn51449110 ORDER BY ItemOrder'
 export const computationalSql = 'SELECT * FROM syn51469335'
 export const upsetPlotSql = 'SELECT files, assay FROM syn51581771'
-export const cohortBuilderSql = 'SELECT * FROM syn52239906'
+export const cohortBuilderSql = 'SELECT * FROM syn52234652'
```
jay-hodgson, line 11: syn52239906 deleted.  Looks like syn52234652 should be right, but have yet to test

packages/synapse-react-client/src/components/widgets/query-filter/CombinedRangeFacetFilter.test.tsx
```diff
@@ -33,6 +43,28 @@ jest.mock('../RangeSlider/RangeSlider', () => ({
   }),
 }))
 
+const mockQueryResponseData: QueryResultBundle = {
```
jay-hodgson, line 46: Mock query response when asking for facet stats (with no selected facets, to ensure full column ranges)

packages/synapse-react-client/src/components/widgets/query-filter/CombinedRangeFacetFilter.test.tsx
```diff
@@ -104,51 +147,65 @@ function createTestProps(
...
+  beforeEach(() => {
+    jest.clearAllMocks()
+    server.use(...getHandlersForTableQuery(mockQueryResponseData))
+  })
+  afterEach(() => server.restoreHandlers())
+  afterAll(() => server.close())
+
   let props: CombinedRangeFacetFilterProps
-  function init(overrides?: CombinedRangeFacetFilterProps) {
+  async function init(overrides?: CombinedRangeFacetFilterProps) {
```
jay-hodgson, line 159: Now that we run a query, change initialization to wait for loading state to finish before looking for the expected state

packages/synapse-react-client/src/components/widgets/query-filter/CombinedRangeFacetFilter.tsx
```diff
@@ -45,24 +48,50 @@ export const CombinedRangeFacetFilter: React.FunctionComponent<
   onChange,
   collapsed = false,
 }: CombinedRangeFacetFilterProps) => {
+  const { getCurrentQueryRequest } = useQueryContext()
+
   const [isCollapsed, setIsCollapsed] = useState<boolean>(collapsed)
   const {
-    columnMin: col1Min,
+    columnName: col1Name,
```
jay-hodgson, line 55: Do not use the column min from the current facet result, instead run a query asking for the facet stats without any selected facets.  Look for the facet associated to this facet column name

packages/synapse-react-client/src/components/widgets/query-filter/CombinedRangeFacetFilter.tsx
```diff
@@ -45,24 +48,50 @@ export const CombinedRangeFacetFilter: React.FunctionComponent<
...
   } = facetResults[1]
 
+  // Also query for the facet columns full range (without any selected facets), to set the range selector range
+  const queryBundleRequest = useMemo(() => {
+    const requestCopy = getCurrentQueryRequest()
+    requestCopy.partMask = SynapseConstants.BUNDLE_MASK_QUERY_FACETS
+    const { selectedFacets } = requestCopy.query
+    requestCopy.query.selectedFacets = selectedFacets
+      ? selectedFacets.filter(
+          facetRequest =>
```
jay-hodgson, line 74: Let's keep the other selected facets (so the range could change based on other facet selections, but these two columns should not impact each other!)

packages/synapse-react-client/src/components/widgets/query-filter/CombinedRangeFacetFilter.tsx
```diff
@@ -104,19 +136,19 @@ export const CombinedRangeFacetFilter: React.FunctionComponent<
...
+            <label>{col2MaxMax}</label>
           ) : (
             <>
               {columnType === 'INTEGER' && (
                 <RangeSlider
-                  key="RangeSlider"
-                  domain={[col1Min, col2Max]}
+                  key={`RangeSlider-${selectedMin}-${selectedMax}`}
+                  domain={[col1MinMin, col2MaxMax]}
                   initialValues={{ min: selectedMin, max: selectedMax }}
```
jay-hodgson, line 147: Set the key to force re-rendering when selectedMin and selectedMax changes.  At times it did not re-render, even though the initialValues were updated to the new selected values!

nickgros (approved):
Just have suggestions for a couple of names, looks good

packages/synapse-react-client/src/components/widgets/query-filter/CombinedRangeFacetFilter.tsx
```diff
@@ -45,24 +48,50 @@ export const CombinedRangeFacetFilter: React.FunctionComponent<
...
+
+  const { data: queryResultFacetResponse, isLoading: isLoadingFacetStats } =
+    useGetQueryResultBundleWithAsyncStatus(queryBundleRequest)
+
+  const fullFacetStats = queryResultFacetResponse?.responseBody?.facets
+
+  const col1Facet = fullFacetStats?.find(facet => facet.columnName === col1Name)
+  const col2Facet = fullFacetStats?.find(facet => facet.columnName === col2Name)
+  const col1MinMin = (col1Facet as FacetColumnResultRange)?.columnMin
+  const col2MaxMax = (col2Facet as FacetColumnResultRange)?.columnMax
```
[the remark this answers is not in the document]
jay-hodgson: It's just the column minimum/maximum, but ignores any facet selection from either column1 or column2.  Instead of col1MinMin, what about col1GlobalMin (to suggest this idea)?

PR #466, opened 2023-09-11
SWC-6552 - Fix UserOrTeamBadge showing stale data

nickgros:
`UserOrTeamBadge` provides two 'interfaces' so the userGroupHeader data can be either provided or fetched (via react-query) using a principalId. This bug only occurred when data was fetched, but the faulty logic existed to manage both scenarios.

The data was stored in `useState`. Due to incorrect logic in an effect, the data was NOT updated when the fetched data changed due to the `principalId` changing.

We can remove the effect and just pick the correct object inline and the bug is resolved.

### DIFF
--- a/packages/synapse-react-client/src/components/UserOrTeamBadge/UserOrTeamBadge.tsx
+++ b/packages/synapse-react-client/src/components/UserOrTeamBadge/UserOrTeamBadge.tsx
@@ -1,7 +1,6 @@
 import { Skeleton } from '@mui/material'
-import React, { useEffect, useState } from 'react'
-import { useGetUserGroupHeader } from '../../synapse-queries/user/useUserGroupHeader'
-import { useSynapseContext } from '../../utils/context/SynapseContext'
+import React from 'react'
+import { useGetUserGroupHeader } from '../../synapse-queries'
 import { UserGroupHeader } from '@sage-bionetworks/synapse-types'
 import TeamBadge from '../TeamBadge'
 import { UserBadge } from '../UserCard/UserBadge'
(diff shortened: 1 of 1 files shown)


packages/synapse-react-client/src/components/UserOrTeamBadge/UserOrTeamBadge.tsx
```diff
@@ -29,23 +28,14 @@ export default function UserOrTeamBadge(props: UserOrTeamBadgeProps) {
...
     {
       enabled: !providedUserGroupHeader,
     },
   )
 
-  useEffect(() => {
-    if (principalId && userGroupHeader == undefined && fetchedUserGroupHeader) {
-      setUserGroupHeader(fetchedUserGroupHeader)
-    }
-  }, [accessToken, principalId, userGroupHeader, fetchedUserGroupHeader])
```
nickgros, line 48: In particular, this effect would never re-run when the `principalId` changed, so the data remained stale.

packages/synapse-react-client/src/components/UserOrTeamBadge/UserOrTeamBadge.tsx
```diff
@@ -29,23 +28,14 @@ export default function UserOrTeamBadge(props: UserOrTeamBadgeProps) {
...
+  const userGroupHeader = providedUserGroupHeader ?? fetchedUserGroupHeader
```
nickgros, line 38: `fetchedUserGroupHeader` has a stable identity, so this won't trigger excess re-renders and doesn't need a `useMemo`

PR #469, opened 2023-09-11
PORTALS-2367: add a breakpoint for categories, and flex-wrap the feat…

jay-hodgson:
…ured tools to fix x-overflow


jay-hodgson:
This is all code specific to the NF portal (Tools Database).  Pulling in and letting Robert validate

PR #472, opened 2023-09-12
Update tools.ts

cconrad8:
this is the addition of dateModified for this Jira ticket. 
https://sagebionetworks.jira.com/browse/NFINT-906

Is this correct? 
Thanks!


cconrad8:
@jay-hodgson @nickgros

jay-hodgson:
It must not be referencing the column model type when looking for a renderer for the secondary label, because I don't think this is what you were hoping for:
Screen Shot 2023-09-12 at 2 01 42 PM

PR #477, opened 2023-09-19
PORTALS-2799

jay-hodgson:


packages/synapse-react-client/src/components/FeaturedDataTabs/FeaturedDataTabs.test.tsx
```diff
@@ -16,7 +16,7 @@ const renderComponent = (props: FeaturedDataTabsProps) => {
 describe('FeaturedDataTabs tests', () => {
   const props: FeaturedDataTabsProps = {
     rgbIndex: 0,
-    sql: 'select * from syn11346063',
+    sql: 'select * from syn50000',
```
jay-hodgson, line 19: replaced test references of syn11346063 to the mock file view syn50000 due to a flaky test that was failing when it attempted to get the file view entity

packages/synapse-react-client/src/components/SynapseTable/SendToCavaticaConfirmationDialog.integration.test.tsx
```diff
@@ -25,8 +25,10 @@ import {
   getUseQueryLoadingMock,
   getUseQuerySuccessMock,
 } from '../../testutils/ReactQueryMockUtils'
+import { getEntityHandlers } from '../../mocks/msw/handlers/entityHandlers'
+import { MOCK_REPO_ORIGIN } from '../../utils/functions/getEndpoint'
 
-const onExportToCavatica = jest.fn()
+const onExportToCavatica = jest.fn().mockImplementation(() => Promise.resolve())
```
jay-hodgson, line 31: Fix a bug I introduced in a previous PR.  On success, the export dialog is closed

jay-hodgson:
Screen Shot 2023-09-20 at 11 25 32 AM copy

packages/synapse-react-client/src/components/TimelinePlot/TimelinePhase.tsx
```diff
@@ -0,0 +1,219 @@
...
+import createPlotlyComponent from 'react-plotly.js/factory'
+import dayjs, { ManipulateType } from 'dayjs'
+import { Dialog } from '@mui/material'
+import {
+  ObservationCard,
+  ObservationCardSchema,
+} from '../row_renderers/ObservationCard'
+import { Row } from '@sage-bionetworks/synapse-types'
+import pluralize from 'pluralize'
+const Plot = createPlotlyComponent(Plotly)
```
jay-hodgson, line 12: All of the plot.ly specific configuration is found in this component.

packages/synapse-react-client/src/components/TimelinePlot/TimelinePhase.tsx
```diff
@@ -0,0 +1,219 @@
...
+const getLayout = (
+  start: dayjs.Dayjs,
+  end: dayjs.Dayjs,
+  color: string,
+  annotateTime?: number,
+  annotateTimeUnits?: ManipulateType,
+): Partial<Layout> => {
+  const annotations: Partial<Plotly.Annotations>[] = [
+    {
+      x: start.add(1, 'day').format(),
```
jay-hodgson, line 82: Without this annotation, the individual phase plot shifts when we show another annotation on hover

packages/synapse-react-client/src/components/TimelinePlot/TimelinePhase.tsx
```diff
@@ -0,0 +1,219 @@
...
+
+const TimelinePhase = ({
+  color,
+  rowData,
+  schema,
+  widthPx,
+}: TimelinePhaseProps) => {
+  const [clickEvent, setClickEvent] = useState<Plotly.PlotMouseEvent>()
+  const [hoverEvent, setHoverEvent] = useState<Plotly.PlotHoverEvent>()
+  const start = dayjs()
```
jay-hodgson, line 161: The plot x axis are dates.  The observation data are all durations, so we just add the duration to the current date to find where each point should be on the x-axis.

packages/synapse-react-client/src/components/TimelinePlot/TimelinePlot.integration.test.tsx
```diff
@@ -0,0 +1,69 @@
...
+    queryResultBundleJson.queryResult!.queryResults.rows = newRows!
+    await renderTimeline()
+    expect(await screen.findAllByText('PRENATAL')).toHaveLength(1)
+    expect(await screen.findAllByText('NEONATAL')).toHaveLength(1)
+    expect(await screen.findAllByText('WEANLING')).toHaveLength(1)
+    expect(await screen.findAllByText('ADOLESCENT')).toHaveLength(1)
+    expect(await screen.findAllByText('ADULT')).toHaveLength(1)
+
+    // test
+    let notExist = false
```
jay-hodgson, line 61: Is there a better way to do this (test that an element does not exist)?

packages/synapse-react-client/src/components/TimelinePlot/TimelinePlot.stories.tsx
```diff
@@ -0,0 +1,45 @@
...
+type Story = StoryObj<typeof meta>
+
+export const Demo: Story = {
+  args: {
+    sql: `SELECT observationId as "id", observationPhase as "phase", observationSubmitterName as "submitterName", synapseId as "submitterUserId", observationTime as "time", observationTimeUnits as "timeUnits", observationText as "text", observationType as "tag" FROM syn51735464`,
+    searchParams: {
+      resourceId: '9971e47e-976a-4631-8edd-5cae04304b01',
+    },
+    sqlOperator: ColumnSingleValueFilterOperator.EQUAL,
+    defaultSpecies: 'Rattus norvegicus',
```
jay-hodgson, line 26: I set the species because I didn't see a way to easily independently mock multiple table query results (one for the event data, and the other that the species selector component executes to find all of the possible species for the resource id)

packages/synapse-react-client/src/components/TimelinePlot/TimelinePlot.tsx
```diff
@@ -0,0 +1,243 @@
...
+          },
+        ],
+        additionalFilters: additionalFilters,
+      },
+
+      partMask: BUNDLE_MASK_QUERY_RESULTS,
+      concreteType: 'org.sagebionetworks.repo.model.table.QueryBundleRequest',
+    },
+    {
+      enabled: !!species,
```
jay-hodgson, line 86: I don't know if this actually works.  It should be limited to the resource ID anyway, so this should not be a large result if it doesn't work.  Ideally it would only execute when it could apply the filters by both species and resourceId

packages/synapse-react-client/src/components/TimelinePlot/TimelinePlot.tsx
```diff
@@ -0,0 +1,243 @@
...
+    eventsData?.queryResult?.queryResults.headers.findIndex(
+      header => header.name.toLowerCase() === OBSERVATION_TYPE_COLUMN_NAME,
+    )!
+  const submitterUserIdIndex =
+    eventsData?.queryResult?.queryResults.headers.findIndex(
+      header =>
+        header.name.toLowerCase() === OBSERVATION_SUBMITTER_USER_ID_COLUMN_NAME,
+    )!
+
+  const schema: ObservationCardSchema = {
```
jay-hodgson, line 127: For challenges, some of this will need to change.  We'll probably need to support a generic card rather than this special ObservationCard that is being used by NF.

packages/synapse-react-client/src/components/TimelinePlot/phasesQueryResponseData.ts
```diff
@@ -0,0 +1,181 @@
+import { QueryResultBundle } from '@sage-bionetworks/synapse-types'
```
jay-hodgson, line 1: Hard coded phase data from a table that Christina created (syn52408661).  We don't really need the observationTimeMax or observationTimeUnits columns (these are just estimates for the max time in a phase, and they want the x-axis range to be based on the observational data instead).

packages/synapse-react-client/src/style/base/_core.scss
```diff
@@ -21,6 +21,16 @@
 a:hover {
   cursor: pointer;
 }
+.forcePlotlyDefaultCursor {
+  .js-plotly-plot {
+    .plotly {
+      .drag {
+        cursor: default !important;
```
jay-hodgson, line 28: For some reason, even when we tell plot.ly to not be interactive, it changes the cursor as if it still is interactive!  Override via css

packages/synapse-react-client/src/components/TimelinePlot/TimelinePhase.tsx
```diff
@@ -0,0 +1,219 @@
...
+        },
+      },
+    ],
+    margin: {
+      l: 0,
+      r: 0,
+      t: 60,
+      b: 60,
+    },
+    // autosize: false,
```
nickgros, line 141:
```suggestion
```

packages/synapse-react-client/src/components/TimelinePlot/TimelinePlot.integration.test.tsx
```diff
@@ -0,0 +1,69 @@
...
+}
+
+describe('TimelinePlot tests', () => {
+  jest
+    .spyOn(SynapseClient, 'getFullQueryTableResults')
+    .mockResolvedValue(queryResultBundleJson)
+
+  it('renders all phases of the timeline plot', async () => {
+    await renderTimeline()
+    expect(SynapseClient.getFullQueryTableResults).toHaveBeenCalledTimes(1),
```
nickgros, line 37: I'm not convinced that `await act` reliably waits until all async side effects are done. Recommend changing this line to
`await waitFor(() => expect(SynapseClient.getFullQueryTableResults).toHaveBeenCalledTimes(1))`

packages/synapse-react-client/src/components/TimelinePlot/TimelinePlot.integration.test.tsx
nickgros: `expect(screen.queryByText("JUVENILE")).not.toBeInTheDocument()`

Make sure you don't use `queryAllByText` because that will return `[]`, and I think `toBeInTheDocument` just checks if the argument is not null

packages/synapse-react-client/src/components/TimelinePlot/TimelinePlot.tsx
```diff
@@ -0,0 +1,243 @@
...
+
+  const { data: eventsData, isLoading } = eventTableQuery
+
+  if (isLoading) {
+    return <LoadingTimelinePlot />
+  }
+  const observationPhaseIndex =
+    eventsData?.queryResult?.queryResults.headers.findIndex(
+      header => header.name.toLowerCase() === OBSERVATION_PHASE_COLUMN_NAME,
+    )!
```
nickgros, line 98: we have `getFieldIndex` in `queryUtils` but you'd have to change the part mask to get the `selectColumns`. Maybe worth adding another utility that does the same thing with the `queryResults.headers` (is that data ever different)?

packages/synapse-react-client/src/components/TimelinePlot/phasesQueryResponseData.ts
nickgros: Are all aware that this is hard-coded (so changes they make to the table won't affect the code unless we update this entire object)?

What's the motivation for hard-coding this?

packages/synapse-react-client/src/components/TimelinePlot/TimelinePlotSpeciesSelector.test.tsx
```diff
@@ -0,0 +1,87 @@
...
+  })
+  return component
+}
+
+describe('TimelinePlotSpeciesSelector tests', () => {
+  jest
+    .spyOn(SynapseClient, 'getFullQueryTableResults')
+    .mockResolvedValueOnce(queryResultBundleJson)
+    .mockResolvedValueOnce(singleRowQueryResultBundle)
+    .mockResolvedValueOnce(emptyQueryResultBundle)
```
nickgros, line 53: Personally I would prefer setting the appropriate mock in each test. Implicitly relying on test order kinda scares me

packages/synapse-react-client/src/components/TimelinePlot/TimelinePlotSpeciesSelector.test.tsx
```diff
@@ -0,0 +1,87 @@
...
+    expect(setSpecies).toHaveBeenCalledWith('Mus musculus')
+
+    let notExist = false
+    try {
+      await screen.findByRole('button')
+    } catch (error) {
+      notExist = true
+    }
+    expect(notExist).toEqual(true)
+  })
```
nickgros, line 79: `expect(screen.queryByRole('button')).not.toBeInTheDocument()`

nickgros (approved):
Code looks great, had a few questions and very minor suggestions

packages/synapse-react-client/src/components/TimelinePlot/TimelinePhase.tsx
jay-hodgson: It's expected.  When we draw a new annotation on the plot (in the place that we want, under the x-axis) we introduce more content to fit in the same size viewport (so elements shift to accommodate)

packages/synapse-react-client/src/components/TimelinePlot/TimelinePlotSpeciesSelector.test.tsx
jay-hodgson: I agree, this was based on an example.  I'll updated...

nickgros (approved):

PR #481, opened 2023-09-22
PORTALS-2799: support multiple observation events at the same time point

jay-hodgson:


jay-hodgson:
Multiple rows, hover annotation:
Screen Shot 2023-09-25 at 10 46 40 AM

Popup with multiple cards:
Screen Shot 2023-09-25 at 10 46 49 AM

Single row annotation still works:
Screen Shot 2023-09-25 at 10 46 59 AM

Popup with single card:
Screen Shot 2023-09-25 at 10 47 05 AM

packages/synapse-react-client/src/components/TimelinePlot/TimelinePhase.tsx
```diff
@@ -87,12 +104,17 @@ const getLayout = (
...
+      annotateEventCount > 1 ? `(${annotateEventCount})` : ''
     const x = start.add(annotateTime, annotateTimeUnits)
     annotations.push({
       x: x.format(),
       y: -1,
-      text: `${annotateTime} ${pluralize(annotateTimeUnits, annotateTime)}`,
+      text: `${annotateTime} ${pluralize(
+        annotateTimeUnits,
+        annotateTime,
+      )} ${eventCountText}`,
```
hallieswan, line 117: I think `text` will have a trailing space when `eventCountText` is an empty string, e.g. `1 month ` -- just wondering if that was intentional, e.g. to adjust the spacing of the label to the axis tick mark?

hallieswan (approved):
Looks good to me!

packages/synapse-react-client/src/components/TimelinePlot/TimelinePhase.tsx
jay-hodgson: Noted!  The extra space will not shift the plot (since we are explicitly positioning the annotations), so I don't think it will have any impact on the result.  Thanks for pointing this out!

PR #486, opened 2023-09-28
PLFM-8058: Add handlers for cohort builder custom commands.  When ask…

jay-hodgson:
…ing for all table data, clear our the limit from the incoming request


apps/portals/src/configurations/elportal/routesConfig.tsx
```diff
@@ -200,9 +200,11 @@ const routes: GenericRoute[] = [
               text: (
                 <>
                   This page helps you find data related to anonymized
-                  Participants. Add filters to build a list of participants.
-                  Once you’ve built your cohort, you can download the related
-                  data files or <strong>Send to CAVATICA</strong> for analysis.
+                  Participants in the ELITE Portal. Add filters to build a list
```
jay-hodgson, line 203: Update copy

apps/portals/src/configurations/elportal/synapseConfigs/cohortbuilder.ts
```diff
@@ -44,29 +47,17 @@ export const individualsView: SynapseConfig = {
       {
         buttonText: 'View files in selection',
         onClick: (event) => {
```
jay-hodgson, line 49: Refactor, move to new file (handleParticipantsToFiles)

apps/portals/src/configurations/elportal/synapseConfigs/cohortbuilder.ts
```diff
@@ -106,29 +97,17 @@ export const filesView: SynapseConfig = {
       {
         buttonText: 'View participants in selection',
         onClick: (event) => {
-          // add filter for files perspective, to show participants associated to the selected files only.
```
jay-hodgson, line 109: Refactor, move to new file (handleFilesToParticipants)

packages/synapse-react-client/src/components/LoadingScreen/index.ts
```diff
@@ -0,0 +1,2 @@
+import { SynapseSpinner, BlockingLoader } from './LoadingScreen'
+export { SynapseSpinner, BlockingLoader }
```
jay-hodgson, line 2: Should show some kind of loading while the command is executing, but I'm not sure how we would do this today given the current architecture.  Any ideas, @nickgros ?

apps/portals/src/configurations/elportal/synapseConfigs/cohortbuilder.ts
```diff
@@ -18,7 +21,7 @@ export const individualsView: SynapseConfig = {
     rgbIndex,
     name: 'Participants',
     visibleColumnCount: 10,
-    facetsToPlot: ['Diagnosis'],
+    facetsToPlot: ['Sex', 'dataTypes', 'Assays', 'Diagnosis', 'fileFormat'],
```
jay-hodgson, line 24: Milan asked for other facets to be plotted

apps/portals/src/configurations/elportal/synapseConfigs/handleFilesToParticipants.tsx
```diff
@@ -0,0 +1,65 @@
...
+  event: CustomControlCallbackData,
+) => {
+  // add filter for files perspective, to show participants associated to the selected files only.
+  const idColIndex = event.data?.columnModels?.findIndex(
+    (cm) => cm.name === 'fileId',
+  )
+
+  const token = await SynapseClient.getAccessTokenFromCookie()
+  const queryResultBundle = await SynapseClient.getFullQueryTableResults(
+    event.request!,
```
nickgros, line 18: Strongly recommend overriding partMask to be 1 (row data only), unless that happens in an earlier step

apps/portals/src/configurations/elportal/synapseConfigs/handleParticipantsToFiles.tsx
```diff
@@ -0,0 +1,63 @@
...
+export const handleParticipantsToFiles = async (
+  event: CustomControlCallbackData,
+) => {
+  // add filter for files perspective, to show files associated to all participant rows.
+  const idColIndex = event.data?.columnModels?.findIndex(
+    (cm) => cm.name === 'individualID',
+  )
+  const token = await SynapseClient.getAccessTokenFromCookie()
+  const queryResultBundle = await SynapseClient.getFullQueryTableResults(
+    event.request!,
```
nickgros, line 17: Same note as above for the partMask

packages/synapse-react-client/src/components/LoadingScreen/index.ts
nickgros: This is an interesting problem. Since this is just dealing with an external data source, my gut is telling me that we should just somehow use react-query since that will let us neatly handle loading/error cases.

Is the loading behavior generalizable to all of the custom buttons? Maybe instead of the synchronous `onClick` prop, we just pass in arguments that match the arguments to `useMutation`? So we would have 

* a `mutationFn` prop, where the passed prop would be like:
```
async (e: CustomControlCallbackData) => {
  await SynapseClient.getFullQueryTableResults(/* ... */ );
  localStorage.setItem(/* ... */)
  return
}
```
* an `onSuccess` prop, where the passed prop would be like:
```
() => {
  window.location.href = '...'
}
```

Optionally have an `onError` prop, but I'm not sure what to do in this specific case.

The button component itself then manages loading state based on the return value of `useMutation`.

If we wanted different loading behavior, that could be an enum/boolean prop too I guess?

I feel like this idea might be overcomplicated.

nickgros (changes requested):
I like the changes, couple notes, and CI is showing typecheck failure

apps/portals/src/configurations/elportal/synapseConfigs/cohortbuilder.ts
```diff
@@ -79,6 +79,11 @@ export const filesView: SynapseConfig = {
     isRowSelectionVisible: true,
     rowSelectionPrimaryKey: ['fileId'],
     additionalFiltersLocalStorageKey: 'cohort-builder-files-perspective',
+    combineRangeFacetConfig: {
```
jay-hodgson, line 82: Looks like they added the minAge and maxAge column to the file perspective of the Cohort Builder!

apps/portals/src/configurations/elportal/synapseConfigs/handleFilesToParticipants.tsx
```diff
@@ -15,7 +19,10 @@ export const handleFilesToParticipants = async (
 
   const token = await SynapseClient.getAccessTokenFromCookie()
   const queryResultBundle = await SynapseClient.getFullQueryTableResults(
-    event.request!,
+    {
+      ...event.request!,
+      partMask: SynapseConstants.BUNDLE_MASK_QUERY_RESULTS,
```
jay-hodgson, line 24: From code review

apps/portals/src/configurations/elportal/synapseConfigs/handleParticipantsToFiles.tsx
```diff
@@ -14,7 +18,10 @@ export const handleParticipantsToFiles = async (
   )
   const token = await SynapseClient.getAccessTokenFromCookie()
   const queryResultBundle = await SynapseClient.getFullQueryTableResults(
-    event.request!,
+    {
+      ...event.request!,
+      partMask: SynapseConstants.BUNDLE_MASK_QUERY_RESULTS,
```
jay-hodgson, line 23: From code review

packages/synapse-react-client/src/components/SynapseTable/TopLevelControls/CustomControlButton.tsx
```diff
@@ -0,0 +1,37 @@
...
+    <Button
+      {...props}
+      disabled={isDisabled}
+      onClick={() => {
+        setIsLoading(true)
+        onClick({
+          ...callbackData,
+          refresh: () => {
+            setIsLoading(false)
+            callbackData.refresh()
```
jay-hodgson, line 26: In addition to the parent refresh function, also clear the loading state

nickgros (approved):

PR #490, opened 2023-10-02
PORTALS-2806, add a DOI link if available

jay-hodgson:
Screen Shot 2023-10-02 at 12 51 05 PM

Screen Shot 2023-10-02 at 12 51 22 PM


jay-hodgson:
This may not be right, but this is limited to the ObservationCard, which is only used by the NF portal by utilizing the Timeline plot.

jay-hodgson:
Will ask Robert to evaluate on the live staging site.

PR #498, opened 2023-10-12
PORTALS-2808

jay-hodgson:


packages/synapse-react-client/src/components/Plot/SynapsePlot.tsx
```diff
@@ -1,135 +1,164 @@
...
-import { SynapseContext } from '../../utils/context/SynapseContext'
+import { useGetFullTableQueryResults } from '../../synapse-queries'
+import { Skeleton } from '@mui/material'
+import { QueryWrapperSynapsePlotRowClickEvent } from '../QueryWrapperPlotNav/QueryWrapperSynapsePlot'
+import { QueryContextType } from '../QueryContext'
 const Plot = createPlotlyComponent(Plotly)
 
-export type SynapsePlotProps = {
-  ownerId?: string
-  wikiId?: string
```
jay-hodgson, line 16: removed ownerId and wikiId since they were not used

packages/synapse-react-client/src/components/Plot/SynapsePlot.tsx
```diff
@@ -1,135 +1,164 @@
...
+  horizontal?: string // sets the if a bar chart should be horizontal or vertical ('true' | 'false')
+  barmode?: string // Plotly barmode ('stack' | 'group' | 'overlay' | 'relative')
+  displayModebar?: string // sets the modebar visibility ('true' | 'false')
 }
 
-type SynapsePlotState = {
-  isLoaded: boolean
-  queryData: QueryResultBundle
+// QueryWrapperPlotNav customPlot parameters, undefined otherwise
+export type QueryWrapperPlotNavCustomPlotParams = {
```
jay-hodgson, line 32: Added a new optional parameter, customPlotParams, to enhance the SynapsePlot in the context of a query wrapper plot nav

packages/synapse-react-client/src/components/Plot/SynapsePlot.tsx
```diff
@@ -1,135 +1,164 @@
...
+  additionalFilters: QueryFilter[]
+  onCustomPlotClick?: (event: QueryWrapperSynapsePlotRowClickEvent) => void
+  queryContext?: QueryContextType
+}
+export type SynapsePlotProps = {
+  widgetparamsMapped: SynapsePlotWidgetParams
+  customPlotParams?: QueryWrapperPlotNavCustomPlotParams
 }
 
-export class SynapsePlot extends React.Component<
```
jay-hodgson, line 25: Rewrote class component to be a functional component

packages/synapse-react-client/src/components/Plot/SynapsePlot.tsx
```diff
@@ -1,135 +1,164 @@
...
-
-  static contextType = SynapseContext
-  declare context: NonNullable<React.ContextType<typeof SynapseContext>>
-
-  public componentDidMount() {
-    this.fetchPlotlyData()
+  const { data: queryData, isLoading } =
+    useGetFullTableQueryResults(queryRequest)
+  if (isLoading) {
+    return <Skeleton width={'100%'} height={'200px'} />
```
jay-hodgson, line 63: Added a loading UI

packages/synapse-react-client/src/components/Plot/SynapsePlot.tsx
```diff
@@ -1,135 +1,164 @@
...
+  const headers = queryData.queryResult?.queryResults.headers ?? []
+  for (let i = 0; i < headers.length - 1; i += 1) {
+    // make an entry for each set of data points
+    plotData[i] = {
+      orientation,
+      name: headers[i + 1].name,
+      type: type.toLowerCase() as PlotType,
+      x: [],
+      y: [],
+      customdata: [],
```
jay-hodgson, line 121: Added customdata for each point, to pass Row back to click handler (if specified)

packages/synapse-react-client/src/components/Plot/SynapsePlot.tsx
```diff
@@ -1,135 +1,164 @@
...
+  for (const row of queryData.queryResult?.queryResults.rows ?? []) {
+    for (let j = 1; j < row.values.length; j += 1) {
+      // create pairs of data
+      const rowValues = row.values
+      const xArray = plotData[j - 1]!.x as Plotly.Datum[]
+      const yArray = plotData[j - 1]!.y as Plotly.Datum[]
+      const customdata = plotData[j - 1]!.customdata as Plotly.Datum[]
+      xArray.push(rowValues[0])
+      yArray.push(rowValues[j])
+      customdata.push(JSON.stringify(row))
```
jay-hodgson, line 134: Hold the Row as a string in this Plotly Datum object

packages/synapse-react-client/src/components/Plot/SynapsePlot.tsx
```diff
@@ -1,135 +1,164 @@
...
   }
 
-  public render() {
-    if (!this.state.isLoaded) {
-      return null
+  let onPlotClick:
+    | ((event: Readonly<Plotly.PlotMouseEvent>) => void)
+    | undefined
+  if (onCustomPlotClick && queryContext) {
+    onPlotClick = eventData => {
```
jay-hodgson, line 142: On plot point click, pass the selected Row as well as the entire query context that was passed to us (from the query  wrapper plot nav)!

packages/synapse-react-client/src/components/QueryWrapperPlotNav/QueryWrapperPlotNav.stories.tsx
```diff
@@ -36,11 +37,35 @@ const meta = {
...
+      values: [event.row.values[0] as string],
+    },
+  ]
+  executeQueryRequest(requestCopy)
+}
 export const Cards: Story = {
   args: {
     name: 'Tools',
     sql: 'SELECT * FROM syn26438037',
+    customPlots: [
```
jay-hodgson, line 60: This is a silly demo, since these particular statistics are returned in the resourceType Values facet, but it demonstrates and exercises the new functionality.

apps/portals/src/configurations/elportal/synapseConfigs/cohortbuilder.ts
```diff
@@ -12,9 +12,62 @@ import {
...
+    newMin = values[0]
+    if (values.length > 1) {
+      newMax = values[1]
+    }
+  }
+  // need to set both for the combined range filter
+  setRangeFacetValue({ columnName: 'minAge' }, newMin, newMax)
+  setRangeFacetValue({ columnName: 'maxAge' }, newMin, newMax)
+}
+const getPlotConfig = (tableId: string) => {
```
jay-hodgson, line 40: ELITE portal (complex) custom interactive plot definition, with custom click handler.  May want to review this last!

packages/synapse-react-client/src/style/components/query_filter/_query-wrapper-synapse-plot.scss
```diff
@@ -0,0 +1,14 @@
...
+  padding-top: 5px;
+  > .SynapsePlot {
+    // Hack around Plotly.  If modebar is hidden, it still takes up space for some reason.
+    // Move plot up (closer to title), and make sure the title is on top.
+    .plot-container {
+      margin-top: -19px;
+    }
+    p {
+      position: relative;
+      z-index: 1000;
```
nickgros, line 11: I think this is a little too aggressive ;)

image

The z-index for the dropdown is also 1000 but I think you can pick anything in [1, 999]

apps/portals/src/configurations/elportal/synapseConfigs/cohortbuilder.ts
```diff
@@ -12,9 +12,62 @@ import {
...
+    when maxAge >= 50 and maxAge < 60  then ' 50 - 59 '
+    when maxAge >= 60 and maxAge < 70  then ' 60 - 69 ' 
+    when maxAge >= 70 and maxAge < 80  then ' 70 - 79 '
+    when maxAge >= 80 and maxAge < 90  then ' 80 - 89 '
+    when maxAge >= 90 and maxAge < 100  then ' 90 - 99 '
+    else '100+' END) AS STRING) AS bucket,
+          count(*) AS totalWithinRange
+            FROM ${tableId}
+            WHERE maxAge IS NOT NULL
+            GROUP BY 1
```
nickgros, line 57: TIL about `GROUP BY 1`. I think I prefer `GROUP BY bucket` for clarity/resiliency but this isn't a hard statement to follow, so I think it's fine either way

apps/portals/src/configurations/elportal/synapseConfigs/cohortbuilder.ts
```diff
@@ -12,9 +12,62 @@ import {
...
+  const values = value.match(/\d+/g)
+  if (values && values.length > 0) {
+    newMin = values[0]
+    if (values.length > 1) {
+      newMax = values[1]
+    }
+  }
+  // need to set both for the combined range filter
+  setRangeFacetValue({ columnName: 'minAge' }, newMin, newMax)
+  setRangeFacetValue({ columnName: 'maxAge' }, newMin, newMax)
```
nickgros, line 38: I think this filter + the SQL only works in the participant table without extra logic to handle absolute min/max because
1. All of the rows represent a single participant that has a single age
2. The plot bin boundaries match the age bin boundaries in the data.

How is this supposed to work for the files table? You have a file with many participants, where many are under 90 and one is 90-94. I think now that file would be put in the 90-99 bin (since that's the maxAge), and it would be improperly excluded from the all of the other bins. 

I think there's a fundamental issue where we're flattening multiple data points per row to 1 data point and still displaying it like it represents many.

packages/synapse-react-client/src/components/Plot/SynapsePlot.tsx
```diff
@@ -1,135 +1,165 @@
...
-  queryData: QueryResultBundle
+// QueryWrapperPlotNav customPlot parameters, undefined otherwise
+export type QueryWrapperPlotNavCustomPlotParams = {
+  selectedFacets: FacetColumnRequest[]
+  additionalFilters: QueryFilter[]
+  onCustomPlotClick?: (event: QueryWrapperSynapsePlotRowClickEvent) => void
+  queryContext?: QueryContextType
+}
+export type SynapsePlotProps = {
+  widgetparamsMapped: SynapsePlotWidgetParams
```
nickgros, line 39: Consider creating a new component that wraps SynapsePlot that has the `widgetparamsMapped` interface (`SynapseWikiPlot`?) and is only used by the wiki. Then you could directly use the strict/accurate Plotly types in `SynapsePlotProps` and not have to use this fairly type-unsafe `widgetparamsMapped` intermediate data structure in places like the new QueryWrapperSynapsePlot

packages/synapse-react-client/src/components/QueryWrapperPlotNav/QueryWrapperSynapsePlot.tsx
```diff
@@ -0,0 +1,77 @@
...
+  const { title, onCustomPlotClick } = props
+
+  const widgetParamsMapped: SynapsePlotWidgetParams = useMemo(() => {
+    return {
+      ...props,
+      title: undefined, // we are handling the plot title here in this component
+      showlegend: 'false',
+      displayModebar: 'false',
+    }
+  }, [props])
```
nickgros, line 46: as suggested above, if the `widgetParamsMapped` interface was only used in the wiki, we wouldn't need to put booleans in strings.

You could also consider adding color/theming as is done in `FacetNavPanel`, but I'm not sure if there's gonna be a design pass on this after it gets into the staging portal.

packages/synapse-react-client/src/components/QueryWrapperPlotNav/QueryWrapperSynapsePlot.tsx
```diff
@@ -0,0 +1,77 @@
...
+        <Typography
+          variant="smallText1"
+          sx={{
+            fontWeight: 700,
+            borderBottom: '1px solid #dcdcdc',
+            paddingBottom: '8px',
+          }}
+        >
+          {title}
+        </Typography>
```
nickgros, line 69: You could probably use the `FacetFilterHeader` component with the `hideCollapsible` prop.

nickgros (changes requested):
Code looks great, called out these things in comments
* z-index issue
* data issue for the EL Portal files table

and a handful of optional suggestions

apps/portals/src/configurations/elportal/synapseConfigs/cohortbuilder.ts
jay-hodgson: You're absolutely right, Nick.  If a file has an age range of 85-95 (for example), it would only be represented in the 90-95 bin.  I think it would also be strange if the file was represented more than once in the plot.
Maybe this plot on the file level is not as important as the one for the participant table.   I'll ask Milan.

packages/synapse-react-client/src/components/QueryWrapperPlotNav/QueryWrapperSynapsePlot.tsx
jay-hodgson: Done

apps/portals/src/configurations/elportal/synapseConfigs/cohortbuilder.ts
jay-hodgson: Done

packages/synapse-react-client/src/style/components/query_filter/_query-wrapper-synapse-plot.scss
jay-hodgson: Done

packages/synapse-react-client/src/components/Plot/SynapsePlot.tsx
jay-hodgson: Great idea.  I think the existing "MarkdownSynapsePlot" wrapper component could service this purpose.  Moved string param resolution to that component.

apps/portals/src/configurations/elportal/synapseConfigs/cohortbuilder.ts
jay-hodgson: Confirmed with Milan, this should be on the participants view only.  "Even long term I think it only belongs on the Participant table. We will eventually get and surface animal data in the Files table, and ages won’t make sense across species"

nickgros (approved):
Looks great! Thanks @jay-hodgson !

PR #509, opened 2023-10-18
PORTALS-2820

jay-hodgson:


packages/synapse-react-client/src/components/QueryWrapperPlotNav/QueryWrapperPlotNav.stories.tsx
```diff
@@ -439,3 +439,20 @@ export const TableWithJSONColumns: Story = {
     shouldDeepLink: false,
   },
 }
+
+export const MaterializedViewOfFiles: Story = {
```
jay-hodgson, line 443: Example story demonstrating the access column and direct download column on a Materialized View

packages/synapse-react-client/src/components/SynapseTable/SynapseTableRenderers.tsx
```diff
@@ -130,14 +134,41 @@ export const directDownloadColumn = columnHelper.display({
   },
 })
 
+const getEntityOrRowId = (
```
jay-hodgson, line 137: These two functions, getEntityOrRowId() and getEntityOrRowVersion() are the meat of the PR.  Get the column index from the table meta.  If defined, then use that value from the Row.values[].   Otherwise return what it used to (the rowId or versionNumber?.toString()).

apps/portals/src/configurations/elportal/resources.ts
```diff
@@ -11,7 +11,7 @@ export const upsetPlotSql = 'SELECT files, assay FROM syn51581771'
 export const cohortBuilderSql =
   'SELECT individualID, inFiles, minAge, maxAge, Sex, Study, fileFormat, dataTypes, dataSubtypes, Assays, fileIds, fileVersions FROM syn52234652'
 export const cohortBuilderFilesSql =
-  'SELECT id, individualCount,Study,dataTypes,Assays,Species,fileFormat,minAge,maxAge,analysisType,Consortium,Grant,isModelSystem,isMultiSpecimen,metadataType,modelSystemType,Platform,Project,ResourceType,fileVersion FROM syn52234677'
+  'SELECT id, name, individualCount,Study,dataTypes,Assays,Species,fileFormat,minAge,maxAge,analysisType,Consortium,Grant,isModelSystem,isMultiSpecimen,metadataType,modelSystemType,Platform,Project,ResourceType,fileVersion FROM syn52234677'
```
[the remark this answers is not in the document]
jay-hodgson: Selecting "name" is important, because it has special treatment when the rowEntityIDColumnName is provided for Tables that are not inherently structured such that every Row represents an Entity.

packages/synapse-react-client/src/components/GenericCard/SynapseCardLabel.tsx
```diff
@@ -181,9 +181,7 @@ export const SynapseCardLabel: React.FC<SynapseCardLabelProps> = props => {
         <p>
           {split.map((el, index) => {
             const { baseURL, URLColumnName, wrapValueWithParens } = labelLink
-            const elOrRowId = labelLink.overrideValueWithRowID
-              ? `syn${rowId}`
-              : el
```
nickgros, line 186: It seems like we lose the `syn` here if this is a view. As far as I can tell, `overrideValueWithRowID` is only used for the ARK Portal Datasets explore page. It seems like these links will be different, but I think both the new and old links will still work. For example, these both seem to resolve to the same details page, making this a non-breaking change:

`https://staging.arkportal.synapse.org/Explore/Datasets/DetailsPage?id=syn29837062`

`https://staging.arkportal.synapse.org/Explore/Datasets/DetailsPage?id=29837062`

packages/synapse-react-client/src/components/SynapseTable/SynapseTableRenderers.tsx
nickgros: They are pretty self explanatory but a bit of context for each in a function doc wouldn't hurt--I'm finding the more custom configuration we add to Synapse Tables UI, the harder it is to remember how everything works!

nickgros (approved):

packages/synapse-react-client/src/components/GenericCard/SynapseCardLabel.tsx
jay-hodgson: The only time rowId is passed into the SynapseCardLabel is from SynapseTable, and that rowId always has the 'syn' prefix (if it exists)

packages/synapse-react-client/src/components/SynapseTable/SynapseTableCell/SynapseTableCell.tsx
```diff
@@ -75,28 +79,27 @@ function SynapseTableCell(props: SynapseTableCellProps) {
...
+        rowId={rowSynId}
       />
     )
   }
 
   // PORTALS-2095: Special case. If this is an EntityView, and we are rendering the 'id' or 'name' column,
   // and we have a rowId and rowVersionNumber (should always be the case), and our entityIdToHeader map
   // contains the row Synapse ID, then auto-link.
-  if (
+  const tableRowRepresentsFile =
```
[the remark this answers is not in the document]
jay-hodgson: Right you are

packages/synapse-react-client/src/components/SynapseTable/SynapseTableRenderers.tsx
jay-hodgson: Done

packages/synapse-react-client/src/components/GenericCard/SynapseCardLabel.tsx
jay-hodgson: Hold on, my assumption that the rowId is only a string when using the new configuration values is wrong.  Let me look into this...

PR #511, opened 2023-10-19
PORTALS-2825

jay-hodgson:
Update One Sage and OAuth sign-in app to use GTM


apps/synapse-oauth-signin/index.html
```diff
@@ -48,19 +48,16 @@
...
-      }
-      gtag('js', new Date())
-      gtag('config', 'UA-29804340-1')
-    </script>
+      <!-- Google Tag Manager -->
+      <script>(function(w,d,s,l,i){w[l]=w[l]||[];w[l].push({'gtm.start':
+        new Date().getTime(),event:'gtm.js'});var f=d.getElementsByTagName(s)[0],
+        j=d.createElement(s),dl=l!='dataLayer'?'&l='+l:'';j.async=true;j.src=
+        'https://www.googletagmanager.com/gtm.js?id='+i+dl;f.parentNode.insertBefore(j,f);
+        })(window,document,'script','dataLayer','GTM-KPW4KS62');
```
nickgros, line 57: Is it ok that the IDs are the same (GTM-KPW4KS62) ?

I dunno if that's the account ID or if it should be unique to each property

nickgros (approved):

apps/synapse-oauth-signin/index.html
jay-hodgson: We're currently sending all data to the same GA stream (and now the same GTM "container").  Historically we break apart by domain later.  We could certainly organize it differently (different GA property for each website), but this would be additional overhead (especially for each of the portal websites).  So this change is in line with how it has historically been done, for better or worse.

PR #515, opened 2023-10-23
SWC-6548 - Add utils and reducer for TableColumnSchemaEditor

nickgros:
This PR contains the bulk of the business logic for managing data in the Table Column Schema form in SWC-6548

### DIFF
--- a/packages/synapse-react-client/src/components/TableColumnSchemaEditor/TableColumnSchemaFormReducer.ts
+++ b/packages/synapse-react-client/src/components/TableColumnSchemaEditor/TableColumnSchemaFormReducer.ts
@@ -200,17 +200,21 @@ function changeColumnModelType(
   let newColumnModelValue: ColumnModelFormData | JsonSubColumnModelFormData
 
   // Create a copy of the selected column model
-  if (
-    prevState &&
-    prevState[columnModelIndex] &&
-    prevState[columnModelIndex].jsonSubColumns &&
-    jsonSubColumnModelIndex !== undefined
-  ) {
-    newColumnModelValue = cloneDeep(
-      prevState[columnModelIndex].jsonSubColumns![jsonSubColumnModelIndex],
-    )
+  if (prevState && prevState[columnModelIndex]) {
+    if (
+      prevState[columnModelIndex].jsonSubColumns &&
+      jsonSubColumnModelIndex !== undefined
+    ) {
+      newColumnModelValue = cloneDeep(
+        prevState[columnModelIndex].jsonSubColumns![jsonSubColumnModelIndex],
+      )
+    } else {
+      newColumnModelValue = cloneDeep(prevState[columnModelIndex])
+    }
   } else {
-    newColumnModelValue = cloneDeep(prevState[columnModelIndex])
+    throw new Error(
+      'Cannot change column model type for a column that does not exist',
+    )
   }
 
   // Update the column model. Remove fields that no longer make sense for the new column type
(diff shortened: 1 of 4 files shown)


packages/synapse-react-client/src/components/TableColumnSchemaEditor/TableColumnSchemaFormReducer.ts
```diff
@@ -0,0 +1,500 @@
...
+    default:
+      throw new Error(`Unexpected action`, action)
+  }
+  return [...prevState]
+}
+
+export const tableColumnSchemaFormDataAtom = atomWithReducer<
+  ColumnModelFormData[],
+  TableColumnSchemaFormReducerAction
+>([], reducer)
```
nickgros, line 500: This `atomWithReducer` will be passed to the form React component and each change will be mediated by the reducerFn.

packages/synapse-react-client/src/components/TableColumnSchemaEditor/TableColumnSchemaFormReducer.ts
```diff
@@ -0,0 +1,500 @@
...
+      toggleSelect(action, prevState)
+      break
+    }
+    case 'delete':
+      prevState = deleteColumnModel(prevState)
+      break
+    case 'move': {
+      prevState = moveColumnModel(action, prevState)
+      break
+    }
```
nickgros, line 468: `move` would handle Drag-N-Drop

packages/synapse-react-client/src/components/TableColumnSchemaEditor/TableColumnSchemaFormReducer.ts
```diff
@@ -0,0 +1,500 @@
...
+      prevState = moveSelectedItemsDown(prevState, cm => cm.isSelected)
+      prevState.forEach(cm => {
+        if (cm.jsonSubColumns) {
+          cm.jsonSubColumns = moveSelectedItemsDown(
+            cm.jsonSubColumns,
+            jsc => jsc.isSelected,
+          )
+        }
+      })
+      break
```
nickgros, line 490: `moveUp`/`moveDown` would handle moving selected items (i.e. the current paradigm)

packages/synapse-react-client/src/components/TableColumnSchemaEditor/TableColumnSchemaFormReducer.ts
```diff
@@ -0,0 +1,500 @@
...
+  action: {
+    type: 'setColumnModelValue'
+    columnModelIndex: number
+    jsonSubColumnModelIndex?: number
+    value: ColumnModelFormData | JsonSubColumnModelFormData
+  },
+  prevState: ColumnModelFormData[],
+) {
+  const { columnModelIndex, jsonSubColumnModelIndex, value } = action
+  if (
```
jay-hodgson, line 300: Should this be something like this (same comment for a subset of other utility functions)?
if (prevState && prevState[columnModelIndex]) {
  if (prevState[columnModelIndex].jsonSubColumns &&
    jsonSubColumnModelIndex !== undefined) { ...}
  else {...}
}

packages/synapse-react-client/src/components/TableColumnSchemaEditor/TableColumnSchemaFormReducer.ts
jay-hodgson: Great pattern

jay-hodgson (approved):

packages/synapse-react-client/src/components/TableColumnSchemaEditor/TableColumnSchemaFormReducer.ts
nickgros: Good idea, that would definitely be safer

PR #518, opened 2023-10-24
SWC-6594 - Fix grid blowout on review submission page

nickgros:
Fixes SWC-6594

Before change: 

image

After change: 

image

### DIFF
--- a/packages/synapse-react-client/src/style/components/_submission-page.scss
+++ b/packages/synapse-react-client/src/style/components/_submission-page.scss
@@ -29,7 +29,11 @@
     .SubmissionSummaryGrid {
       display: grid;
       column-gap: 10px;
-      grid-template-columns: auto auto;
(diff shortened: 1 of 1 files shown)


packages/synapse-react-client/src/style/components/_submission-page.scss
```diff
@@ -29,7 +29,11 @@
     .SubmissionSummaryGrid {
       display: grid;
       column-gap: 10px;
-      grid-template-columns: auto auto;
```
nickgros, line 32: I have no idea what caused this change. Maybe recent changes to UserCard styles caused its width to exceed the parent here? Needs further investigation. In any case, the grid columns should be bounded with [minmax](https://developer.mozilla.org/en-US/docs/Web/CSS/minmax).

See also: https://css-tricks.com/preventing-a-grid-blowout/

jay-hodgson (approved):

PR #528, opened 2023-10-27
DESIGN_1332: fix Close (confirmation) button after requesting to join a team

jay-hodgson:


jay-hodgson:
Change impacts the Challenge Portal only

PR #529, opened 2023-10-27
PORTALS-2838, PORTALS-2837: Add Close button to Access Requirement List, and fix Close button for step in Challenge Wizard

jay-hodgson:


packages/synapse-react-client/src/components/ChallengeTeamWizard/ChallengeTeamWizard.tsx
```diff
@@ -466,9 +466,16 @@ const ChallengeTeamWizard: React.FunctionComponent<
     onClose()
   }
 
-  const onConfirmHandlerMap: Record<string, () => Promise<void>> | void = {
+  const onConfirmHandlerMap: Record<
+    string,
+    (() => Promise<void>) | (() => undefined)
```
jay-hodgson, line 471: Downstream expects one of these signatures, so I'll just return undefined.

jay-hodgson:
In addition to the modal close button in the upper left, add a Close button in the Dialog actions:
Screen Shot 2023-10-27 at 2 26 52 PM

packages/synapse-react-client/src/components/ChallengeTeamWizard/ChallengeTeamWizard.tsx
```diff
@@ -466,9 +466,16 @@ const ChallengeTeamWizard: React.FunctionComponent<
...
+  > | void = {
     CREATE_NEW_TEAM: handleCreateTeam,
     JOIN_REQUEST_FORM: handleRequestMembership,
+    JOIN_REQUEST_SENT: () => {
```
jay-hodgson, line 475: The JOIN_REQUEST_SENT didn't have a confirmation handler.  It should just hide the dialog.

packages/synapse-react-client/src/components/ChallengeTeamWizard/ChallengeTeamWizard.tsx
nickgros: You could just have `() => Promise<void>` and `return Promise.resolve()`, if you think that might make it more simple.

nickgros (approved):

PR #537, opened 2023-10-31
remove esbuild-plugin-node-polyfill

jay-hodgson:
Screen Shot 2023-10-31 at 1 59 42 PM

Due to a package that has 31 downloads a week

Screen Shot 2023-10-31 at 2 02 10 PM


packages/synapse-react-client/esbuild.config.mjs
```diff
@@ -50,7 +50,7 @@ const esBuildOptions = {
       namedExport: 'ReactComponent',
       exportType: 'named',
     }),
-    ESBuildNodePolyfillsPlugin,
+    NodeModulesPolyfillPlugin(),
```
nickgros, line 53: Any issues with these React components in SWC after this change?

jay-hodgson: Using Chrome local overrides, I found that this likely breaks the SWC build.  So thanks for asking!

PR #539, opened 2023-11-01
SWC-6548 -  Implement ability to add default columns and annotation columns

nickgros:
SWC-6548

- Add functionality for appending default columns for views
- Add functionality for appending annotation columns for views
- Default columns are no longer editable, to match the original behavior
- Disable entire form while submitting changes
- Refactor useUpdateTableColumns to call createTableUpdateTransactionRequest instead of expecting the called to do so beforehand. The mutation will not capture errors called in this step, in addition to the actual update step.
- Move createTableUpdateTransactionRequest to utils
- Minor style updates
- Use disableInjectingGlobalStyles to reduce performance hit when adding many columns at once
 - Add types for retrieving column models based on view annotations

SynapseClient:
 - Add ability to get default column models
 - Add ability to get annotation column models
 - Add functionality to get all of paginated service that uses 'nextPageToken'

Mocks:
 - Refactor MSW handlers for async jobs so more than one async job type can be used at a time.

https://github.com/Sage-Bionetworks/synapse-web-monorepo/assets/17580037/c0bc5529-5dab-4a38-b930-b4e8e8fc6c96


packages/synapse-react-client/src/utils/functions/TableColumnSchemaUtils.ts
```diff
@@ -0,0 +1,85 @@
...
+
+/**
+ * Creates the column models for a change request and returns
+ * the change request to be passed to the Synapse backend.
+ * @param accessToken
+ * @param tableId
+ * @param oldSchema
+ * @param proposedSchema
+ */
+export async function createTableUpdateTransactionRequest(
```
nickgros, line 18: Moved without changes from TableColumnSchemaFormUtils.

I moved this into `src/utils` because I wanted to refactor to call this from inside the custom `useUpdateTable` mutation, but I didn't want the mutation to import from the `src/components` folder.

packages/synapse-react-client/src/utils/functions/TableColumnSchemaUtils.test.ts
```diff
@@ -0,0 +1,182 @@
+import {
```
nickgros, line 1: These tests were moved from `src/components` to `src/utils` without changes.

packages/synapse-react-client/src/synapse-queries/entity/useEntity.ts
```diff
@@ -370,23 +372,34 @@ export function useUpdateEntityACL(
...
+      // This call will create new column models as appropriate
+      const transactionRequest = await createTableUpdateTransactionRequest(
+        accessToken!,
+        request.entityId,
+        request.originalColumnModels,
+        request.newColumnModels,
+      )
+
+      return SynapseClient.updateTable(transactionRequest, accessToken)
+    },
```
nickgros, line 402: Before the refactor, the caller of this mutation had to 
1. Create column models using `createTableUpdateTransactionRequest`
2. Call `mutate`, which called `updateTable`.

I refactored so the mutation also creates the update, because this is always going to be an atomic step. We can handle loading/error states consistently while we wait for this sequence of requests to be made.

packages/synapse-react-client/src/synapse-client/SynapseClient.ts
```diff
@@ -3056,6 +3059,34 @@ export const getAllOfPaginatedService = async <T>(
   return results
 }
 
+export type FunctionReturningNextPageToken<T> = (
+  nextPageToken?: string | null,
+) => Promise<{ results: T[]; nextPageToken?: string | null }>
+
+export async function getAllOfNextPageTokenPaginatedService<T>(
```
nickgros, line 3066: See lines immediately above this for `getAllOfPaginatedService`, which pretty much works the same but with a different response body interface

packages/synapse-react-client/src/synapse-client/SynapseClient.ts
```diff
@@ -4720,3 +4751,56 @@ export function createColumnModels(
...
+      BackendDestinationEnum.REPO_ENDPOINT,
+    )
+    return getAsyncResultBodyFromJobId(
+      asyncJobId.token,
+      `/repo/v1/column/view/scope/async/get/${asyncJobId.token}`,
+      accessToken,
+    )
+  }
+
+  return getAllOfNextPageTokenPaginatedService(fn)
```
nickgros, line 4805: Use new utility to get all results.

packages/synapse-react-client/src/components/TableColumnSchemaEditor/ColumnModelForm.tsx
```diff
@@ -123,140 +142,176 @@ export default function ColumnModelForm(props: ColumnOrSubcolumnFormProps) {
...
-          }}
-          fullWidth
-        />
-      </Box>
-      <Box>
-        <FormControl fullWidth>
-          <Select
-            label="Column Type"
-            value={columnModel.columnType}
+        {isDefaultColumn ? (
```
nickgros, line 153: Most of these changes are adding ternary expressions to show the value instead of an input if it's a default column

packages/synapse-react-client/src/components/TableColumnSchemaEditor/TableColumnSchemaEditor.tsx
```diff
@@ -1,65 +1,42 @@
...
-import { SetOptional } from 'type-fest'
-import { createTableUpdateTransactionRequest } from './TableColumnSchemaEditorUtils'
-import { useSynapseContext } from '../../utils'
+import { Provider } from 'jotai'
 
 export type TableColumnSchemaEditorProps = {
   entityId: string
 }
 
-function transformFormDataToColumnModels(formData: ColumnModelFormData[]) {
```
nickgros, line 29: Moved to util file

packages/synapse-react-client/src/components/TableColumnSchemaEditor/TableColumnSchemaEditor.tsx
```diff
@@ -81,32 +58,65 @@ export default function TableColumnSchemaEditor(
       },
     )
 
-  const { mutate, isLoading: isMutating } = useUpdateTable()
+  const { mutate, isLoading: isMutating, error } = useUpdateTableColumns()
 
-  const isLoading = isLoadingEntity || isLoadingColumnModels || isMutating
```
nickgros, line 86: Inappropriate to show loading screen when `isMutating`. We should disable the form, not hide it!

packages/synapse-react-client/src/components/TableColumnSchemaEditor/TableColumnSchemaEditor.tsx
```diff
@@ -81,32 +58,65 @@ export default function TableColumnSchemaEditor(
...
+      return {
+        scope: entity.scopeIds,
+        viewTypeMask: undefined,
+        viewEntityType: convertToEntityType(
+          entity.concreteType,
+        ) as ViewEntityType,
+      }
+    }
+    return undefined
+  }, [entity])
```
nickgros, line 118: Should probably extract this hook

packages/synapse-react-client/src/components/TableColumnSchemaEditor/TableColumnSchemaEditorUtils.test.ts
```diff
@@ -287,175 +278,4 @@ describe('TableColumnSchemaEditorUtils', () => {
       })
     })
   })
-
-  describe('getTableUpdateTransactionRequest tests', () => {
```
nickgros, line 291: Tests were moved to `src/utils` folder

packages/synapse-react-client/src/components/TableColumnSchemaEditor/TableColumnSchemaEditorUtils.ts
```diff
@@ -245,78 +246,61 @@ export function canHaveRestrictedValues(
...
- * Creates the column models for a change request and returns
- * the change request to be passed to the Synapse backend.
- * @param accessToken
- * @param tableId
- * @param oldSchema
- * @param proposedSchema
+ * Transform the form data for the TableColumnSchemaForm to ColumnModels to be sent to Synapse
+ * @param formData the form data to transform
  */
-export async function createTableUpdateTransactionRequest(
```
nickgros, line 255: Function was moved to `src/utils`

packages/synapse-react-client/src/components/TableColumnSchemaEditor/TableColumnSchemaEditorUtils.ts
```diff
@@ -245,78 +246,61 @@ export function canHaveRestrictedValues(
...
-    const copy: SetOptional<ColumnModel, 'id'> = {
-      ...m,
-    }
-    if (copy.id != null) {
-      // any changes to the existing column model?
-      const oldColumnModel: ColumnModel | undefined =
-        oldColumnModelId2Model.get(copy.id)
-      if (oldColumnModel != undefined && !isEqual(oldColumnModel, copy)) {
-        delete copy.id
+export function transformFormDataToColumnModels(
```
nickgros, line 252: function was moved from component files

packages/synapse-react-client/src/components/TableColumnSchemaEditor/TableColumnSchemaEditorUtils.ts
```diff
@@ -245,78 +246,61 @@ export function canHaveRestrictedValues(
...
-    const oldColumnId: string | null = oldColumnModel.id
-    if (!columnIds.has(oldColumnId)) {
-      changes.push({ oldColumnId, newColumnId: null })
+/**
+ * Transform ColumnModels returned by Synapse into the form data for the TableColumnSchemaForm.
+ * Default column models, if available, are used to determine which columns should not be editable in the form.
+ * @param columnModels
+ * @param defaultColumns
+ */
+export function transformColumnModelsToFormData(
```
nickgros, line 286: This is new, and is used to mark if a column is a 'default column' so we know to not let the user modify the name.

packages/synapse-react-client/src/components/TableColumnSchemaEditor/TableColumnSchemaForm.tsx
```diff
@@ -68,53 +83,86 @@ export type SubmitHandle = {
...
+    useGetDefaultColumnModels(
+      viewScope?.viewEntityType!,
+      viewScope?.viewTypeMask,
+      {
+        enabled: hasDefaultColumnModels,
+        staleTime: Infinity, // The default column models will never change
+      },
+    )
+
+  const hasAnnotationColumnModels = isView && !isSubmissionView
```
nickgros, line 134: I think this is wrong, submission can have annotations

packages/synapse-react-client/src/components/TableColumnSchemaEditor/TableColumnSchemaForm.tsx
```diff
@@ -68,53 +83,86 @@ export type SubmitHandle = {
...
+  const {
+    data: annotationColumnModels,
```
nickgros, line 136: Annotation data is present so we can easily validate against sizes when we tackle SWC-6602

packages/synapse-react-client/src/mocks/msw/handlers/asyncJobHandlers.ts
```diff
@@ -0,0 +1,126 @@
...
+ * Generates MSW handlers for asynchronous jobs.
+ * @param requestPath the path to the service-specific asynchronous job request endpoint.
+ *   For example, `'/repo/v1/entity/:id/table/query/async/start'`
+ * @param responsePath a function that takes the token parameter ID and returns a path to the service-specific asynchronous job response endpoint.
+ *   For example, `param => /repo/v1/entity/:id/table/query/async/get/${param}`
+ *   The generic asynchronous job retrieval handler will be generated automatically.
+ * @param responseBody the response body, or a function that takes the request body and returns the response body
+ * @param backendOrigin the backend origin to use for the handlers.
+ */
+export function generateAsyncJobHandlers<
```
nickgros, line 34: Before this change, we weren't able to mock out multiple types of asynchronous jobs, like table query + fetching annotation columns for a given scope. Now we can.

packages/synapse-react-client/src/components/TableColumnSchemaEditor/TableColumnSchemaEditor.tsx
```diff
@@ -81,32 +50,35 @@ export default function TableColumnSchemaEditor(
...
+      const columnModels = transformFormDataToColumnModels(formData)
 
-      // Create the request to update the table
-      // This call will create new column models as appropriate
-      const request = await createTableUpdateTransactionRequest(
-        accessToken!,
-        entityId,
-        queryResultBundle!.responseBody!.columnModels!,
-        formData,
-      )
```
nickgros, line 103: This step is now included in the call to `mutate`. This is now more simple, and state (loading, error, etc) related to this action is all handled by one `useMutation` hook.

packages/synapse-react-client/src/components/TableColumnSchemaEditor/TableColumnSchemaEditorUtils.test.ts
```diff
@@ -288,173 +289,86 @@ describe('TableColumnSchemaEditorUtils', () => {
...
-      const columnModel1: ColumnModel = {
-        id: OLD_COLUMN_MODEL_ID1,
-        name: 'col1',
-        columnType: ColumnTypeEnum.STRING,
-      }
-      const columnModel2: ColumnModel = {
-        id: OLD_COLUMN_MODEL_ID2,
-        name: 'col2',
-        columnType: ColumnTypeEnum.STRING,
+  describe('getViewScopeForEntity', () => {
```
nickgros, line 292: these are new tests

packages/synapse-react-client/src/components/TableColumnSchemaEditor/TableColumnSchemaEditorUtils.ts
```diff
@@ -245,78 +258,95 @@ export function canHaveRestrictedValues(
...
-  }
+  })
+}
 
-  // delete columns that are not represented in the changes already (create or update)
-  for (const oldColumnModel of oldSchema) {
-    const oldColumnId: string | null = oldColumnModel.id
-    if (!columnIds.has(oldColumnId)) {
-      changes.push({ oldColumnId, newColumnId: null })
+export function getViewScopeForEntity(entity: Entity): ViewScope | undefined {
```
nickgros, line 320: This is also new, is used to ask for annotation columns

packages/synapse-react-client/src/mocks/msw/handlers/tableQueryHandlers.ts
```diff
@@ -54,85 +55,44 @@ export function getHandlersForTableQuery(
...
+  return generateAsyncJobHandlers<
+    ViewColumnModelRequest,
+    ViewColumnModelResponse
+  >(
+    '/repo/v1/column/view/scope/async/start',
+    tokenParam => `/repo/v1/column/view/scope/async/get/${tokenParam}`,
+    response,
+    backendOrigin,
+  )
+}
```
nickgros, line 79: Creating new mock services that use async jobs is now super easy too

packages/synapse-react-client/src/components/TableColumnSchemaEditor/TableColumnSchemaEditorUtils.ts
```diff
@@ -245,78 +258,95 @@ export function canHaveRestrictedValues(
...
+  columnModels: ColumnModel[],
+  defaultColumns: ColumnModel[] = [],
+): ColumnModelFormData[] {
+  return columnModels.map((cm): ColumnModelFormData => {
+    return {
+      ...cm,
+      isSelected: false,
+      // If the name matches a known default column model, then we consider it to be a default column model itself
+      isOriginallyDefaultColumn: defaultColumns.some(
+        dcm => dcm.name === cm.name,
```
jay-hodgson, line 308: should we also check that the default column type also matches?

packages/synapse-react-client/src/components/TableColumnSchemaEditor/TableColumnSchemaEditorUtils.ts
```diff
@@ -245,78 +258,95 @@ export function canHaveRestrictedValues(
...
+  if (isEntityView(entity)) {
+    return {
+      scope: entity.scopeIds,
+      viewTypeMask: entity.viewTypeMask,
+      viewEntityType: convertToEntityType(
+        entity.concreteType,
+      ) as ViewEntityType,
+    }
+  } else if (isDataset(entity) || isDatasetCollection(entity)) {
+    const mask = isDataset(entity)
```
jay-hodgson, line 330: Is this logic reversed?  If isDataset(entity), shouldn't it use ENTITY_VIEW_TYPE_MASK_DATASET?

packages/synapse-react-client/src/components/TableColumnSchemaEditor/TableColumnSchemaForm.tsx
jay-hodgson: Nice prep, thank you

packages/synapse-react-client/src/mocks/msw/handlers/tableQueryHandlers.ts
jay-hodgson: Would it be helpful (for project spinup) document common tasks and solutions like this (perhaps links in the README?)?

jay-hodgson (approved):

packages/synapse-react-client/src/mocks/msw/handlers/tableQueryHandlers.ts
nickgros: Totally agree. Wondering how much effort to put into that considering the following:
1. MSW just released a new major version that (I think) simplifies writing these handlers
2. What we have today could use a bit of cleaning up before we try to 'standardize' it
3. I'm also kind of hoping we can find a tool that makes it easier to generate these stubs using the OpenAPI spec when that work is done

packages/synapse-react-client/src/components/TableColumnSchemaEditor/TableColumnSchemaEditorUtils.ts
nickgros: Name alone is how SWC does it today: https://github.com/Sage-Bionetworks/SynapseWebClient/blob/6785747266c6e7177a94d03208eb95bcdc20e2e6/src/main/java/org/sagebionetworks/web/client/widget/table/v2/schema/ColumnModelsEditorWidget.java#L126-L128

Maybe a typical case to consider: if you deleted and re-added a default column `id` to be STRING instead of ENTITYID, should we still treat it as a default column?

packages/synapse-react-client/src/components/TableColumnSchemaEditor/TableColumnSchemaEditorUtils.ts
nickgros: No, the mask represents what the table contains. Datasets contain Files, and DatasetCollections contain Datasets.

jay-hodgson: Oh, right!

packages/synapse-react-client/src/components/TableColumnSchemaEditor/TableColumnSchemaEditorUtils.ts
jay-hodgson: Probably a bug in SWC, but it's an edge case and I'm not going to push for a fix.
If your column name `id` with type STRING was added to an entity view, I suspect that the backend would not properly populate the column (since it's not really a default column), but I might be wrong!

PR #541, opened 2023-11-03
PORTALS-2848

jay-hodgson:


packages/synapse-react-client/src/components/TimelinePlot/TimelinePlot.tsx
```diff
@@ -157,6 +159,9 @@ export const TimelinePlot = ({
     return <NoContentAvailable />
   }
 
+  const widthPx = dimensions.width ? dimensions.width / phaseData.length : 0
+  const gridTemplateColumns = phaseData.map(() => `${widthPx}px`).join(' ')
```
nickgros, line 163: Just curious, does this work in a way that `auto auto ...` wouldn't?

packages/synapse-react-client/src/utils/hooks/useRefDimensions.ts
```diff
@@ -0,0 +1,46 @@
...
+      >
+        Dimensions: {dimensions.width}w {dimensions.height}h
+      </div>
+    </div>
+  )
+ */
+
+const useRefDimensions = (ref: RefObject<Element>) => {
+  const [dimensions, setDimensions] = useState({ width: 1, height: 2 })
+  useEffect(() => {
```
nickgros, line 21: Not 100% sure but this might be one of the rare cases where you want to use [useLayoutEffect](https://react.dev/reference/react/useLayoutEffect)

nickgros (approved):

packages/synapse-react-client/src/components/TimelinePlot/TimelinePlot.tsx
jay-hodgson: You're right, I guess I was showing off :) The plot.ly component still needs the actual width, though

jay-hodgson: Updated to `auto auto ... `

packages/synapse-react-client/src/utils/hooks/useRefDimensions.ts
jay-hodgson: I didn't see any improvement in the layout behavior when using useLayoutEffect(), so I'd rather just keep using useEffect()

PR #543, opened 2023-11-03
update versions based on dependabot PRs

jay-hodgson:


packages/synapse-react-client/package.json
```diff
@@ -183,7 +183,6 @@
     "@types/react-window": "^1.8.5",
     "@types/shortid": "^0.0.29",
     "@types/spark-md5": "^3.0.2",
-    "@types/testing-library__jest-dom": "^5.14.7",
```
jay-hodgson, line 186: Package deprecated (jest-dom now includes types): https://www.npmjs.com/package/@types/testing-library__jest-dom

jay-hodgson:
As an extra step (besides verifying projects could build), I also pulled the SRC UMD build into synapse.org and it was loaded properly (using Chrome local override)

apps/SageAccountWeb/vite.config.ts
```diff
@@ -5,5 +5,6 @@ export default mergeConfig(vitestConfig, {
   test: {
     include: ['src/tests/**/*.test.[jt]s?(x)'],
     setupFiles: ['src/tests/setupTests.ts'],
+    environment: 'jsdom', // introduced due to random "ReferenceError: window is not defined" during tests
```
hallieswan, line 8: Wonder why this popped up now! Looks like vite's [default environment](https://vitest.dev/config/#environment) is node

hallieswan (approved):
Looks good to me!

PR #552, opened 2023-11-08
PORTALS-2849

jay-hodgson:


jay-hodgson:
More information can be found in Jira: https://sagebionetworks.jira.com/browse/PORTALS-2849?focusedCommentId=189242

package.json
```diff
@@ -43,7 +43,8 @@
       "semver": "^7.5.4",
       "@types/react": "18.0.27",
       "goober": "2.1.9",
-      "react-hot-toast": "2.2.0"
+      "react-hot-toast": "2.2.0",
+      "postcss": "8.4.31"
```
nickgros, line 47: I don't see why we should restrict minor/patch updates:
```suggestion
      "postcss": "^8.4.31"
```

nickgros (approved):

PR #555, opened 2023-11-08
PORTALS-2844

jay-hodgson:


jay-hodgson:
NF portal specific visualization change

PR #556, opened 2023-11-08
SWC-6548 - Export TableColumnSchemaForm and apply small fixes

nickgros:
- Release new version of synapse-react-client
 - Upgrade MUI
 - Add `getEditedColumnModels` imperative handle to TableColumnSchemaForm to match SWC interface
 - Fix animation/z-index of MUI Select causing issues in a GWT Bootstrap Modal
 - Fix issue where JSON Subcolumn name was not used if forceDisplayOriginalColumnName was true
 - Fix missing `key` in Breadcrumbs.tsx


packages/synapse-react-client/src/components/QueryVisualizationWrapper/QueryVisualizationWrapper.tsx
```diff
@@ -174,7 +174,7 @@ export function QueryVisualizationWrapper(
       const forceDisplayOriginalColumnName =
         localStorage.getItem('force-display-original-column-names') === 'true'
 
-      if (!columnName || forceDisplayOriginalColumnName) {
+      if (!columnName || (forceDisplayOriginalColumnName && !jsonPath)) {
```
nickgros, line 177: If we have a `jsonPath` don't use `columnName` even if `forceDisplayOriginalColumnName` is true

packages/synapse-react-client/src/components/TableColumnSchemaEditor/ColumnModelForm.tsx
```diff
@@ -200,11 +201,20 @@ export default function ColumnModelForm(props: ColumnModelFormProps) {
                 'aria-label': 'Column Type',
               }}
               sx={fieldSx}
+              MenuProps={{
+                TransitionComponent: Fade,
+              }}
```
nickgros, line 206: The animation + z-index of the items caused issues in an SWC modal.

packages/synapse-react-client/src/components/TableColumnSchemaEditor/TableColumnSchemaForm.tsx
```diff
@@ -174,6 +177,9 @@ const TableColumnSchemaForm = React.forwardRef<
         submit() {
           onSubmit(readFormData())
         },
+        getEditedColumnModels() {
+          return transformFormDataToColumnModels(readFormData())
+        },
```
nickgros, line 182: The alternative to introducing this handle would probably be something like completely re-implementing the table/view creation wizard.

packages/synapse-react-client/src/components/TableColumnSchemaEditor/TableColumnSchemaForm.tsx
```diff
@@ -225,7 +231,12 @@ const TableColumnSchemaForm = React.forwardRef<
 
   const addDefaultColumns = useCallback(() => {
     if (defaultColumnModels) {
-      addColumnSet(defaultColumnModels)
+      addColumnSet(
+        defaultColumnModels.map(cm => ({
+          ...cm,
+          id: undefined,
```
nickgros, line 237: For some reason, including the `id` broke SWC's RPC call to create and mutate the columns. Omit the ID to fix this.

packages/synapse-react-client/src/umd.index.ts
```diff
@@ -143,6 +144,7 @@ const SynapseComponents = {
   SubscriptionPage,
   OrientationBanner,
   AccessRequirementList,
+  TableColumnSchemaForm,
```
nickgros, line 147: Emit the form from the UMD bundle to be consumed in SWC

packages/synapse-react-client/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "synapse-react-client",
-  "version": "3.1.61",
+  "version": "3.1.62",
```
nickgros, line 3: Release a new version of synapse-react-client so this can be pulled in

packages/synapse-react-client/src/components/TableColumnSchemaEditor/TableColumnSchemaForm.tsx
```diff
@@ -90,8 +93,8 @@ type TableColumnSchemaFormProps = {
   /* If this is an entity view, the ViewScope can be used to determine the default column models and fetch annotation column models */
   viewScope?: ViewScope
   initialData?: ColumnModel[]
-  onSubmit: (formData: ColumnModelFormData[]) => void
-  isSubmitting: boolean
+  onSubmit?: (formData: ColumnModelFormData[]) => void
+  isSubmitting?: boolean
```
nickgros, line 97: SWC won't use these, but they will be used if/when we convert the "next layer" to React components.

jay-hodgson (approved):

PR #562, opened 2023-11-09
Added entity type when viewing trash can 

linglp:
Related to https://sagebionetworks.jira.com/browse/SWC-6367 
I added entity type on trash can page. 

More context: 
I tried making a synapse api call using synapse python client to endpoint: https://rest-docs.synapse.org/rest/GET/trashcan/view.html
By using my own synapse PAT, I could see that the response object is something like: 
```
[{'entityId': 'syn52920298', 'entityName': 'mockrdb_synapse_storage_manifest_table', 'entityType': 'table', 'deletedByPrincipalId': '3443707', 'deletedOn': '2023-11-09T17:58:42.000Z', 'originalParentId': 'syn52854552'}, {'entityId': 'syn52854759', 'entityName': 'mockrdb_synapse_storage_manifest_table', 'entityType': 'table', 'deletedByPrincipalId': '3443707', 'deletedOn': '2023-11-09T17:56:49.000Z', 'originalParentId': 'syn52854552'}]
```
This seems to suggest that entityType is string and it doesn't require further processing.


packages/synapse-react-client/src/components/trash/TrashCanList.tsx
```diff
@@ -38,6 +38,7 @@ function TrashCanListItem(props: TrashCanListItemProps) {
       </td>
       <td>{item.entityId}</td>
       <td>{item.entityName}</td>
+      <td>{item.entityType}</td>
```
nickgros, line 41: Use `entityTypeToFriendlyName` from EntityTypeUtils.ts

packages/synapse-types/src/TrashedEntity.ts
```diff
@@ -3,11 +3,15 @@
  *
  * https://docs.synapse.org/rest/org/sagebionetworks/repo/model/TrashedEntity.html
  */
+import { EntityType } from './EntityType'
+
```
nickgros, line 7: Please move this import to the top of the file, so the above comment is applied to the `TrashedEntity` type

nickgros (changes requested):

packages/synapse-react-client/src/components/trash/TrashCanList.tsx
linglp: Thanks Nick! i just updated that.

packages/synapse-types/src/TrashedEntity.ts
linglp: Thanks @nickgros ! I just moved that line to the top

nickgros (approved):

nickgros (changes requested):
@linglp this looks great, but it looks like there are type errors in TrashCanList.integration.test.tsx . Can you investigate

linglp:
> @linglp this looks great, but it looks like there are type errors in TrashCanList.integration.test.tsx . Can you investigate?

Thanks Nick. Is there a way to run all the integration tests locally? I modified the test by adding entityType. I want to double check if the tests are passing now.

nickgros (approved):
Thanks @linglp ! Nice work!

PR #563, opened 2023-11-09
PORTALS-2854: Species column type change

jay-hodgson:


packages/synapse-react-client/src/components/TimelinePlot/TimelinePlot.tsx
```diff
@@ -93,6 +92,19 @@ export const TimelinePlot = ({
...
+    if (species) {
+      const phasesForTargetSpecies =
+        hardcodedPhasesQueryResponseData.queryResult?.queryResults.rows.filter(
+          row => {
+            return row.values[phaseSpeciesIndex] == species
+          },
+        )
+      setPhaseData(phasesForTargetSpecies)
+    }
+  }, [species])
```
nickgros, line 106: Suggest instead doing `useMemo` and compute the value of `phaseData` instead of making it a state variable, since you never call `setPhaseData` anywhere else.

packages/synapse-react-client/src/components/TimelinePlot/TimelinePlotSpeciesSelector.tsx
```diff
@@ -52,9 +50,11 @@ export const TimelinePlotSpeciesSelector = ({
         <Select
           sx={{ marginLeft: '2px', marginBottom: '2px' }}
           value={species}
-          defaultValue={rows[0].values[0]}
+          defaultValue={defaultSpecies}
```
nickgros, line 53: Since this is a controlled component, this prop can probably just be omitted.

nickgros (commented):
Couple of suggestions but this looks good

PR #576, opened 2023-11-15
PORTALS-2852

jay-hodgson:


jay-hodgson:
Screen Shot 2023-11-15 at 12 46 36 PM

jay-hodgson:
Wrap description in a ShowMore component, and remove the hard coded height for FeaturedToolCard

packages/synapse-react-client/src/style/components/_featured-tool-card.scss
```diff
@@ -9,7 +9,7 @@
 
 .FeaturedToolCard {
   word-wrap: break-word;
-  height: 250px;
+  min-height: 250px;
```
nickgros, line 12: The cutoff for `ShowMore` is based on character count, not height, so I think you could end up in a situation where these cards have different heights.

Is it likely/possible for this to happen with varied text lengths? What if I change the default font size in my browser? And how much do we care if the cards are different heights?

This situation might be unlikely, and the "perfect" solution might be complex (might have to use grid or measure the card sizes?). I don't want perfect to get in the way of better here, so take with a grain of salt 🙂

nickgros (approved):

packages/synapse-react-client/src/style/components/_featured-tool-card.scss
jay-hodgson: This card already uses a css grid layout, so this height (or min-height) is unnecessary (all cards will still have the same height due to the layout).  Removing

PR #580, opened 2023-11-17
SWC-6608: define page size options that are aware of the data maxRows…

jay-hodgson:
…PerPage

### DIFF
--- a/packages/synapse-react-client/src/components/SynapseTable/TablePagination.test.tsx
+++ b/packages/synapse-react-client/src/components/SynapseTable/TablePagination.test.tsx
@@ -118,7 +118,7 @@ describe('TablePagination component', () => {
     expect(mockSetPageSize).toHaveBeenCalledWith('25')
   })
 
-  it('hides pagination when on the first page and query count is less than 10', async () => {
+  it('hides pagination when on the first page and query count is 1', async () => {
     render(
       <Wrapper
         goToPage={mockGoToPage}
@@ -131,7 +131,7 @@ describe('TablePagination component', () => {
     act(() => {
       setTableData({
         concreteType: 'org.sagebionetworks.repo.model.table.QueryResultBundle',
-        queryCount: 5,
+        queryCount: 1,
         maxRowsPerPage: 1000,
       })
     })
--- a/packages/synapse-react-client/src/components/SynapseTable/TablePagination.test.tsx
+++ b/packages/synapse-react-client/src/components/SynapseTable/TablePagination.test.tsx
@@ -1,5 +1,5 @@
 import React from 'react'
-import { render, screen, fireEvent, act } from '@testing-library/react'
+import { render, screen, act } from '@testing-library/react'
 import { TablePagination } from './TablePagination'
 import {
   PaginatedQueryContextType,
@@ -9,6 +9,7 @@ import {
 import { tableQueryDataAtom } from '../QueryWrapper/QueryWrapper'
 import { Provider, SetStateAction, useSetAtom } from 'jotai'
 import { QueryResultBundle } from '@sage-bionetworks/synapse-types'
+import userEvent from '@testing-library/user-event'
 
 const renderedTextConfirmation = 'TablePagination rendered!'
 
@@ -24,7 +25,7 @@ const Receiver = () => {
   return <>{renderedTextConfirmation}</>
 }
 
-function Wrapper(queryContext: Partial<PaginatedQueryContextType>) {
+const Wrapper: React.FC<Partial<PaginatedQueryContextType>> = queryContext => {
   return (
     <Provider>
       <QueryContextProvider queryContext={queryContext as QueryContextType}>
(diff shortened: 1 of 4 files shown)


jay-hodgson:
https://github.com/Sage-Bionetworks/synapse-web-monorepo/assets/1864447/9497e0ff-98ab-421e-a816-d66a8710cfa7

packages/synapse-react-client/src/components/SynapseTable/TablePagination.tsx
```diff
@@ -22,15 +34,16 @@ export const TablePagination = () => {
 
   // PORTALS-2259: Special case.  If we're on the first page,
   // and the total query count is less than the min page size, then do not show pagination UI.
-  if (currentPage == 1 && queryCount && queryCount < 10) {
+  // Also hide pagination if the query count is unavailable.
+  if ((currentPage == 1 && queryCount && queryCount < 10) || !queryCount) {
     return <></>
   }
```
nickgros, line 40: I think 10 here should be pageSizeOptionsBasedOnData[0], right?
```suggestion
  const smallestPageSizeOption = pageSizeOptionsBasedOnData[0]
  if ((currentPage == 1 && queryCount && queryCount < smallestPageSizeOption) || !queryCount) {
    return <></>
  }
```

packages/synapse-react-client/src/components/SynapseTable/TablePagination.test.tsx
```diff
@@ -0,0 +1,153 @@
...
+  })
+
+  it('renders pagination and page size elements', () => {
+    render(
+      Wrapper({
+        goToPage: mockGoToPage,
+        setPageSize: mockSetPageSize,
+        pageSize: 10,
+        currentPage: 1,
+      }),
```
nickgros, line 52: Use JSX, not function call
```suggestion
      <Wrapper 
        goToPage={mockGoToPage}
        setPageSize={mockSetPageSize}
        pageSize={10}
        currentPage={1}
      } />,
```

packages/synapse-react-client/src/components/SynapseTable/TablePagination.test.tsx
```diff
@@ -0,0 +1,153 @@
...
+    )
+    act(() => {
+      screen.findByText(renderedTextConfirmation)
+    })
```
nickgros, line 56: You will also need to add `async` to the function definition
```suggestion
    await screen.findByText(renderedTextConfirmation)
```

packages/synapse-react-client/src/components/SynapseTable/TablePagination.test.tsx
```diff
@@ -0,0 +1,153 @@
...
+      setTableData({
+        concreteType: 'org.sagebionetworks.repo.model.table.QueryResultBundle',
+        queryCount: 100,
+        maxRowsPerPage: 100,
+      })
+    })
+
+    // Simulate changing the page
+    act(() => {
+      fireEvent.click(screen.getByText('2')) // Click on page 2
```
nickgros, line 92: can you use `await userEvent.click` instead? You can remove the `act` if you do that

packages/synapse-react-client/src/components/SynapseTable/TablePagination.test.tsx
```diff
@@ -0,0 +1,153 @@
...
+        maxRowsPerPage: 1000,
+      })
+    })
+
+    // Simulate changing the page size
+    act(() => {
+      fireEvent.change(screen.getByRole('combobox'), {
+        target: { value: '25' },
+      }) // Change page size to 25
+    })
```
nickgros, line 124: If you can't do this one with `userEvent`, then you might need to wrap the following `expect` assertions in a `waitFor`

packages/synapse-react-client/src/components/SynapseTable/TablePagination.test.tsx
```diff
@@ -0,0 +1,153 @@
...
+
+    expect(mockSetPageSize).toHaveBeenCalledTimes(1)
+    expect(mockSetPageSize).toHaveBeenCalledWith('25')
+  })
+
+  it('hides pagination when on the first page and query count is less than 10', () => {
```
nickgros, line 130: Consider adding one case where `maxRowsPerPage` is 5 and there are > 10 results--verify that pagination controls are visible and the combobox has one option that is 5

packages/synapse-react-client/src/components/SynapseTable/TablePagination.test.tsx
jay-hodgson: Sure thing, this was code modified from a chatgpt response :D

packages/synapse-react-client/src/components/SynapseTable/TablePagination.test.tsx
jay-hodgson: Replaced with userEvent

packages/synapse-react-client/src/components/SynapseTable/TablePagination.test.tsx
jay-hodgson: Done

packages/synapse-react-client/src/components/SynapseTable/TablePagination.tsx
jay-hodgson: This is an interesting case!  I think we can hide the pagination as soon as all of the results are shown in a single page

nickgros: I think that makes this unusual sequence possible:

1. Receive 40 results, 25 results shown per page
2. Change the results shown per page to 40
3. Pagination is now hidden and you can't change it to 10 or back to 25

jay-hodgson: This is true, so it works like the Show More button (expand to all results, no way to contract).  It's kind of strange, but I think the UX is ok.  If you disagree, let's talk about the scenario with design.  (How important is it to hide pagination when the results are less than the page size?)

nickgros (approved):

nickgros (approved):

PR #585, opened 2023-11-21
PORTALS-2861, PORTALS-2863: Work around double click issue, and use xaxis tickvals and ticktext

jay-hodgson:


hallieswan (approved):
Looks good to me!

PR #586, opened 2023-11-22
SWC-6613

jay-hodgson:
Walk through bug with current SRC code.  Then load up the locally built SRC override (by bringing up the dev tools and refreshing the screen) to show that the bug is solved.  While the dev tools are open, the uncaught js exception from MUI (that I think is the root cause) is also caught by our handler:

https://github.com/Sage-Bionetworks/synapse-web-monorepo/assets/1864447/5d61c60a-ca5e-4cb5-a605-3e64cb0e340c

### DIFF
--- a/packages/synapse-react-client/src/components/Evaluation/CalendarWithIconFormGroup.tsx
+++ b/packages/synapse-react-client/src/components/Evaluation/CalendarWithIconFormGroup.tsx
@@ -27,16 +27,23 @@ export const CalendarWithIconFormGroup: React.FunctionComponent<
   CalendarWithIconFormGroupProps
 > = ({ value, setterCallback, label, disabled = false }) => {
   const [id] = useState(uniqueId('calendar-with-icon-form-group-'))
+
+  // SWC-6613: Force re-render when DateTimePicker dialog is closed to re-initialize (clean up MUI dialog)
+  const [dateTimePickerKey, setDateTimePickerKey] = useState(1)
   return (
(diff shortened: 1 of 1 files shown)


nickgros (approved):
I think this works.

This issue looks similar to other issues where the transition seems to be bugged in SWC. We fixed it in those places by changing the `TransitionComponent`. You could try adding this to the theme.

```js
MuiDateTimePicker: {
      defaultProps: {
        TransitionComponent: TRANSITION_COMPONENT_OVERRIDE,
      },
    },
```

Though since MUI doesn't provide an easy way to change all transitions, I think at this point I'll open a ticket to see if we can understand why these transitions don't work in SWC, then we can fix it and not have to worry about this issues reappearing in new components.

I was also just looking at making some changes to this component (including upgrading the major version of the MUI X date-picker library) since I'm seeing issues with using UTC vs. local time when using this in the table column schema editor, so I can make sure this issue doesn't regress as a part of that upgrade.

jay-hodgson:
> ```js
> MuiDateTimePicker: {
>       defaultProps: {
>         TransitionComponent: TRANSITION_COMPONENT_OVERRIDE,
>       },
>     },
> ```
Tried, but it resulted in an error because MuiDateTimePicker is not in the list of components that can be styled in the theme.  Maybe this works in a later version?

Object literal may only specify known properties, and 'MuiDateTimePicker' does not exist in type 'Components<Omit<Theme, "components">>'.ts(2322)

PR #589, opened 2023-11-27
SWC-6612 - Improve datetime input + display

nickgros:
General improvements for datetime handling that we will benefit from in SWC-6612
 - Replace CalendarWithIconFormGroup with a generic, reusable DateTimePicker component that comes from MUI
 - Upgrade @mui/x-date-pickers to v6
 - Update CreatedOnByUserDiv to use `formatDate`, fixing an issue where it would display format 'MM/DD/YYYY z' if UTC was displayed
 - Add msw handlers for EvaluationEditorPage story
 - Bundle react-transition-group. Dependency version mismatch caused animation/transition issues in SynapseWebClient
 - Fix y margin issue on Evaluation Editor Page
 - Improve datetime annotation rendering in AnnotationsTable


packages/synapse-react-client/package.json
```diff
@@ -38,7 +38,7 @@
     "@mui/system": "^5.14.17",
     "@mui/utils": "^5.14.17",
     "@mui/x-data-grid": "5.17.26",
-    "@mui/x-date-pickers": "^5.0.20",
+    "@mui/x-date-pickers": "^6.18.1",
```
nickgros, line 41: The `@mui/x-...` packages do not follow the same versioning as the stable MUI packages. So mixing v6 with @mui/core v5 here is ok

packages/synapse-react-client/src/components/Evaluation/CreatedOnByUserDiv.tsx
```diff
@@ -1,36 +1,18 @@
...
-      <span>
-        Created on{' '}
-        {date
-          .toLocaleDateString(
-            undefined,
-            utcTime ? dateFormatOptionUTC : dateFormatOptionLocal,
-          )
-          .replace(',', '')}{' '}
-        by{' '}
-      </span>
```
nickgros, line 33: This ended up showing text like "Created on 3/24/2022 UTC"--use the utility to fix this

packages/synapse-react-client/src/components/Evaluation/EvaluationEditorPage.stories.ts
```diff
@@ -15,6 +20,69 @@ export const Entity: Story = {
 }
 
 export const Evaluation: Story = {
+  parameters: {
```
nickgros, line 23: Add some mock data to make it easier to interact with this this component in Storybook.

packages/synapse-react-client/src/components/Evaluation/EvaluationRoundEditor.tsx
```diff
@@ -131,10 +133,12 @@ export const EvaluationRoundEditor: React.FunctionComponent<
   }, [error])
 
   const [startDate, setStartDate] = useState<string | Dayjs | null>(
-    dayjs(evaluationRoundInput.roundStart),
+    evaluationRoundInput.roundStart
+      ? dayjs(evaluationRoundInput.roundStart)
+      : null,
```
nickgros, line 138: We are now using the built-in MUI validation for the input datetime. Ensure that the starting value is `null` for a new evaluation round, so that the validation error isn't triggered before the user picks a datetime.

packages/synapse-react-client/src/components/Evaluation/EvaluationRoundEditor.tsx
```diff
@@ -257,28 +268,46 @@ export const EvaluationRoundEditor: React.FunctionComponent<
...
+                        : undefined,
+                    },
+                  }}
                 />
               </Col>
             </Row>
             <Row>
               <Col>
-                <h5>SUBMISSION LIMITS</h5>
+                <h5 style={{ marginTop: '20px' }}>SUBMISSION LIMITS</h5>
```
nickgros, line 310: Quick margin fix. Ideally would replace all of these Bootstrap components with MUI

packages/synapse-react-client/src/utils/functions/DateFormatter.test.ts
```diff
@@ -0,0 +1,76 @@
...
+    default: mockDayjsInstance,
+  }
+})
+
+const mockGetUseUtcTimeFromCookie = jest.mocked(getUseUtcTimeFromCookie)
+
+const mockDate = Symbol('mockDate') as unknown as Dayjs
+const mockDayjs = jest.mocked(dayjs)
+
+describe('DateFormatter', () => {
```
nickgros, line 35: Add a couple of tests for our `formatDate` function

packages/synapse-react-client/src/utils/functions/DateFormatter.ts
```diff
@@ -1,16 +1,24 @@
...
-  pattern: string = 'M/D/YYYY h:mm A',
-): string {
+export function formatDate(time: Dayjs, pattern?: string): string {
+  if (!pattern) {
+    // By default, only show timezone if using UTC time
+    pattern = 'M/D/YYYY h:mm A'
+    if (SynapseClient.getUseUtcTimeFromCookie()) {
+      pattern += ' z'
+    }
+  }
```
nickgros, line 18: Before, we would always append `' UTC'` if the UTC cookie was set.

Now, only show the timezone by default if there's no pattern override.

This will fix a few cases where we're just showing a date (`'01/01/2024'`). It usually doesn't make sense to append a timezone if you're showing a date without a time!


---

I think it would be good to change our default pattern to use [localized formats](https://day.js.org/docs/en/display/format#localized-formats), but that probably deserves some dedicated time to ensure we use it consistently and don't mix date formats throughout the app.

jay-hodgson:
Do these changes happen to fix https://sagebionetworks.jira.com/browse/PORTALS-2816 and https://sagebionetworks.jira.com/browse/PORTALS-2815 ?

packages/synapse-react-client/src/components/Evaluation/EvaluationRoundEditor.tsx
```diff
@@ -257,28 +268,46 @@ export const EvaluationRoundEditor: React.FunctionComponent<
...
+                  onChange={setStartDate}
+                  onError={error => setStartDateError(error)}
+                  disabled={disableStartInput}
+                  // Don't allow the value to be a past date only if the input is not disabled.
+                  // i.e. allow a past date if the round has already started
+                  disablePast={!disableStartInput}
+                  slotProps={{
+                    textField: {
+                      helperText: startDateError
+                        ? 'Start date cannot be in the past'
```
jay-hodgson, line 283: Noticed this hard coded error message.  Is this the only error that can happen with this start date selector?

packages/synapse-react-client/src/components/Evaluation/EvaluationRoundEditor.tsx
```diff
@@ -257,28 +268,46 @@ export const EvaluationRoundEditor: React.FunctionComponent<
...
+                  value={endDate}
+                  onChange={setEndDate}
+                  onError={error => setEndDateError(error)}
+                  disablePast={
+                    !dayjs(endDate).isSame(evaluationRoundInput.roundEnd)
+                  }
+                  slotProps={{
+                    textField: {
+                      helperText: endDateError
+                        ? 'End date cannot be in the past'
```
jay-hodgson, line 301: Same comment

packages/synapse-react-client/src/utils/functions/DateFormatter.ts
jay-hodgson: Using localized formats is a great idea - is there a Jira ticket to track?

jay-hodgson (approved):

packages/synapse-react-client/src/components/DateTimePicker/DateTimePicker.tsx
```diff
@@ -0,0 +1,83 @@
...
+  const tzDisplay = utcTime ? 'UTC' : dayjs().tz(dayjs.tz.guess()).format('z')
+
+  return (
+    <TextField
+      {...props}
+      InputProps={{
+        ...props.InputProps,
+        endAdornment: (
+          <>
+            <span>{tzDisplay}</span>
```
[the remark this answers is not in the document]
jay-hodgson: Defining the custom renderer to show the tz is brilliant

packages/synapse-react-client/src/utils/functions/DateFormatter.ts
nickgros: I'll add some more context to this old ticket and move it to dev triage! https://sagebionetworks.jira.com/browse/SWC-5410

packages/synapse-react-client/src/components/Evaluation/EvaluationRoundEditor.tsx
nickgros: Just realized that this also catches the invalid date case. Will replace with a message translator

nickgros:
@jay-hodgson 
> Do these changes happen to fix [sagebionetworks.jira.com/browse/PORTALS-2816](https://sagebionetworks.jira.com/browse/PORTALS-2816) and [sagebionetworks.jira.com/browse/PORTALS-2815](https://sagebionetworks.jira.com/browse/PORTALS-2815) ?

I do not think this fixes PORTALS-2816. 

I think this improves the situation for PORTALS-2815 a bit. That issue doesn't prescribe any specific improvement, so more could be done.

PR #590, opened 2023-11-28
SWC-6612

nickgros:
- Add a bit of spacing before the timezone in the DateTimePicker
 - Bump chromatic to v9

### DIFF
--- a/packages/synapse-react-client/package.json
+++ b/packages/synapse-react-client/package.json
@@ -191,7 +191,7 @@
     "babel-jest": "^29.7.0",
     "babel-loader": "^9.1.3",
     "buffer": "^6.0.3",
-    "chromatic": "^6.19.9",
+    "chromatic": "^9.1.0",
     "copyfiles": "^2.4.1",
     "core-js": "^3.33.2",
     "esbuild": "^0.18.11",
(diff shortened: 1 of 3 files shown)


packages/synapse-react-client/package.json
```diff
@@ -253,6 +253,7 @@
     "build:esbuild": "node esbuild.config.mjs",
     "build:docs": "pnpm storybook build",
     "build:gh-pages": "pnpm storybook build && touch storybook-static/.nojekyll",
+    "build-storybook": "pnpm storybook build",
```
nickgros, line 256: I removed this in a previous change because I assumed that chromatic would use the newer/preferred `storybook build` command, but it still relies on `build-storybook`

packages/synapse-react-client/src/components/DateTimePicker/DateTimePicker.tsx
```diff
@@ -30,7 +30,7 @@ function TextFieldWithTzShown(props: TextFieldProps) {
         ...props.InputProps,
         endAdornment: (
           <>
-            <span>{tzDisplay}</span>
+            <Box ml={0.5}>{tzDisplay}</Box>
```
nickgros, line 33: Add a bit of space to distinguish the input time from this time zone text

jay-hodgson (approved):

PR #593, opened 2023-12-01
Fix mutations, update files vsn

allaway:
Fix mutations sql (current one points to a table that does not exist), and bump to latest portal fileview snapshot (113).


allaway:
@jay-hodgson - the toolApplicationsSql is referencing a table that doesn't exist, so I also switched that to an appropriate reference. However, looking at it - I'm a little confused how it works. Since it is just selecting "applications", how can it filter for a particular resourceId? Does this need to be changed?

nickgros (approved):

PR #596, opened 2023-12-05
SWC-6612: Add validation functions for Table Column Schema Editor

nickgros:
- Add validation functions for Table Column Schema editor
- Simplify DateTimePicker

### DIFF
--- a/packages/synapse-react-client/package.json
+++ b/packages/synapse-react-client/package.json
@@ -113,7 +113,8 @@
     "ua-parser-js": "^1.0.35",
     "universal-cookie": "^4.0.4",
     "use-deep-compare-effect": "^1.8.1",
-    "xss": "^1.0.14"
+    "xss": "^1.0.14",
+    "zod": "^3.22.4"
   },
   "peerDependencies": {
     "react": "^18.2.0",
--- a/packages/synapse-react-client/src/components/DateTimePicker/DateTimePicker.tsx
+++ b/packages/synapse-react-client/src/components/DateTimePicker/DateTimePicker.tsx
@@ -1,4 +1,4 @@
-import React, { useId, useMemo } from 'react'
+import React, { useMemo } from 'react'
 import dayjs, { Dayjs } from 'dayjs'
 import { LocalizationProvider } from '@mui/x-date-pickers/LocalizationProvider'
 import { AdapterDayjs } from '@mui/x-date-pickers/AdapterDayjs'
@@ -11,7 +11,6 @@ import utc from 'dayjs/plugin/utc'
 import timezone from 'dayjs/plugin/timezone'
 import advancedFormat from 'dayjs/plugin/advancedFormat'
 import { useSynapseContext } from '../../utils'
-import { DateTimeField, DateTimeFieldProps } from '@mui/x-date-pickers'
 
 dayjs.extend(utc)
 dayjs.extend(timezone)
@@ -39,23 +38,8 @@ function TextFieldWithTzShown(props: TextFieldProps) {
   )
 }
 
-function DateTimeFieldWithTzShown(props: DateTimeFieldProps<string | Dayjs>) {
-  const id = useId()
-
-  return (
-    <DateTimeField<string | Dayjs>
-      id={id}
-      {...props}
-      slots={{
-        ...props.slots,
-        textField: TextFieldWithTzShown,
-      }}
-    />
-  )
-}
-
 export default function DateTimePicker(props: DateTimePickerProps) {
-  const { value, ...rest } = props
+  const { value, slots, ...rest } = props
   const { utcTime } = useSynapseContext()
 
   const valueAsDayjs = useMemo(() => {
@@ -74,8 +58,8 @@ export default function DateTimePicker(props: DateTimePickerProps) {
       <MuiDateTimePicker<string | Dayjs>
         value={valueAsDayjs}
         slots={{
-          ...props.slots,
-          field: DateTimeFieldWithTzShown,
+          ...slots,
+          textField: TextFieldWithTzShown,
         }}
         {...rest}
       />
--- a/packages/synapse-react-client/src/components/TableColumnSchemaEditor/Validators/BooleanSchema.ts
+++ b/packages/synapse-react-client/src/components/TableColumnSchemaEditor/Validators/BooleanSchema.ts
@@ -5,6 +5,4 @@ const stringToBoolean = z.enum(['true', 'false']).transform(v => v === 'true')
 /**
  * Validates and converts a value to a valid Synapse BOOLEAN value
  */
-export const booleanSchema = z
-  .union([stringToBoolean, z.boolean()])
-  .pipe(z.coerce.boolean())
+export const booleanSchema = z.union([stringToBoolean, z.boolean()])
--- a/packages/synapse-react-client/src/components/TableColumnSchemaEditor/Validators/BooleanSchema.ts
+++ b/packages/synapse-react-client/src/components/TableColumnSchemaEditor/Validators/BooleanSchema.ts
@@ -1,0 +1,10 @@
+import { z } from 'zod'
+
+const stringToBoolean = z.enum(['true', 'false']).transform(v => v === 'true')
+
+/**
+ * Validates and converts a value to a valid Synapse BOOLEAN value
+ */
+export const booleanSchema = z
+  .union([stringToBoolean, z.boolean()])
+  .pipe(z.coerce.boolean())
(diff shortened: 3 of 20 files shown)


packages/synapse-react-client/src/components/DateTimePicker/DateTimePicker.tsx
```diff
@@ -74,8 +58,8 @@ export default function DateTimePicker(props: DateTimePickerProps) {
       <MuiDateTimePicker<string | Dayjs>
         value={valueAsDayjs}
         slots={{
-          ...props.slots,
-          field: DateTimeFieldWithTzShown,
+          ...slots,
+          textField: TextFieldWithTzShown,
```
nickgros, line 62: We can just plug the custom TextField component in here.

The prior solution caused the `InputProps` passed to `TextField` to be incorrectly merged, which broke keyboard input when custom `InputProps` were specified.

packages/synapse-react-client/src/components/TableColumnSchemaEditor/Validators/BooleanSchema.ts
```diff
@@ -0,0 +1,8 @@
+import { z } from 'zod'
+
+const stringToBoolean = z.enum(['true', 'false']).transform(v => v === 'true')
+
+/**
+ * Validates and converts a value to a valid Synapse BOOLEAN value
+ */
+export const booleanSchema = z.union([stringToBoolean, z.boolean()])
```
nickgros, line 8: Accepts only `'true'`, `'false'`, `true`, `false` and when parsed, the returned value is necessarily `true` or `false`

packages/synapse-react-client/src/components/TableColumnSchemaEditor/Validators/ColumnModelValidator.test.ts
```diff
@@ -0,0 +1,664 @@
+import { ColumnTypeEnum } from '@sage-bionetworks/synapse-types'
+import { ColumnModelFormDataZodSchema } from './ColumnModelValidator'
+
+describe('ColumnModel validation', () => {
```
nickgros, line 4: So many tests to make sure this works reasonably. At a high level, check the following behavior
- Doesn't require fields that should be optional
- Respects max/min values for maximumSize, maximumListLength
- Converts data properly, such as
  - Drops empty strings
  - Handles data that may be numbers (returned from the API) or strings (entered by a form field)
  - Validates the contents `defaultValue` and `enumValues` fields based on the column type

packages/synapse-react-client/src/components/TableColumnSchemaEditor/Validators/ColumnModelValidator.ts
```diff
@@ -0,0 +1,145 @@
...
+      } else {
+        result.error.issues.forEach(issue => {
+          ctx.addIssue({ ...issue, path: ['defaultValue'] })
+        })
+        transformedValue = undefined
+      }
+      return { ...data, defaultValue: transformedValue }
+    }
+    return omit(data, 'defaultValue')
+  })
```
nickgros, line 110: Now it gets pretty complicated...the schema we use to validate `defaultValue` will depend on the column type. So...

1. Conditionally fetch a schema
2. `safeParse` the data with the schema
3. If the result has errors, pass those errors along
4. If the result doesn't have errors, transform the value and keep going!

packages/synapse-react-client/src/components/TableColumnSchemaEditor/Validators/ColumnModelValidator.ts
```diff
@@ -0,0 +1,145 @@
...
+        transformedValue = result.data
+      } else {
+        result.error.issues.forEach(issue => {
+          ctx.addIssue({ ...issue, path: ['enumValues'] })
+        })
+        transformedValue = undefined
+      }
+      return { ...data, enumValues: transformedValue }
+    }
+    return omit(data, 'enumValues')
```
nickgros, line 127: This is the exact same set of steps as the previous `transform`, but on a different field, and different data validator. Do you think this is worth extracting to a new function that takes (fieldName, validatorProvider)?

packages/synapse-react-client/src/components/TableColumnSchemaEditor/Validators/ColumnModelValidator.ts
```diff
@@ -0,0 +1,145 @@
...
+  formData: SetOptional<
+    ColumnModelFormData,
+    'isOriginallyDefaultColumn' | 'isSelected'
+  >[],
+): SafeParseReturnType<
+  SetOptional<
+    ColumnModelFormData,
+    'isOriginallyDefaultColumn' | 'isSelected'
+  >[],
+  SetOptional<ColumnModel, 'id'>[]
```
nickgros, line 142: Small note but pretty cool: this line denotes that the _specified_ return type of the schema's `parse`/`safeParse` functions is our `ColumnModel` object.

The _actual_ return type is inferred based on the Zod schema.

If the specified type and the actual types don't match, then TypeScript returns a compiler error.

So with just this line, we basically have a compile-time test that our schema actually returns a `ColumnModel` object.

packages/synapse-react-client/package.json
```diff
@@ -113,7 +113,8 @@
     "ua-parser-js": "^1.0.35",
     "universal-cookie": "^4.0.4",
     "use-deep-compare-effect": "^1.8.1",
-    "xss": "^1.0.14"
+    "xss": "^1.0.14",
+    "zod": "^3.22.4"
```
nickgros, line 117: Recommend skimming https://zod.dev/, especially recommend reading the [Basic Usage](https://zod.dev/?id=basic-usage) section

packages/synapse-react-client/src/components/TableColumnSchemaEditor/Validators/ColumnModelValidator.ts
jay-hodgson: > Do you think this is worth extracting to a new function that takes (fieldName, validatorProvider)?

I think this slight refactor is worth doing

packages/synapse-react-client/src/components/TableColumnSchemaEditor/Validators/ColumnModelValidator.ts
```diff
@@ -0,0 +1,145 @@
...
+      } else {
+        result.error.issues.forEach(issue => {
+          ctx.addIssue({ ...issue, path: ['enumValues'] })
+        })
+        transformedValue = undefined
+      }
+      return { ...data, enumValues: transformedValue }
+    }
+    return omit(data, 'enumValues')
+  })
```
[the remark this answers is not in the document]
jay-hodgson: Yes!  This will vastly improve the experience for data curators

packages/synapse-react-client/src/components/TableColumnSchemaEditor/Validators/DatetimeSchema.ts
```diff
@@ -0,0 +1,21 @@
...
+import dayjs from 'dayjs'
+import utc from 'dayjs/plugin/utc'
+
+dayjs.extend(utc)
+
+const dateTimeString = z.string().transform(data => {
+  // Note: can't use parseInt on a timestamp! "2000-01-01T00:00:00Z" will be parsed as 2000
+  const isInt = data.match(/^\d+$/)
+  if (isInt) {
+    return dayjs.utc(parseInt(data))
```
jay-hodgson, line 11: not calling toISOString() in this case?

packages/synapse-react-client/src/components/TableColumnSchemaEditor/Validators/EntityIdSchema.ts
```diff
@@ -0,0 +1,9 @@
+import { z } from 'zod'
+
+/**
+ * Validates and converts a value to a valid Synapse ENTITYID value
+ */
+export const entityIdSchema = z.union([
+  z.string().regex(/^(syn)?\d+$/),
```
jay-hodgson, line 7: Even though this regex is straightforward, you should probably add a comment that explains in english (most regex test services provide this copy if you don't want to write it yourself)

jay-hodgson (approved):

packages/synapse-react-client/src/components/TableColumnSchemaEditor/Validators/ColumnModelValidator.ts
nickgros: I spent some time trying to do this and it's very hard, if not impossible, to write a function that doesn't break Zod's ability to infer the output type of the parser. I will not refactor this at this time, but I will add a comment to document this.

packages/synapse-react-client/src/components/TableColumnSchemaEditor/Validators/DatetimeSchema.ts
nickgros: This was an oversight, so I will change it. It ultimately gets re-wrapped in `dayjs` and `toIsoString` is called again, so it doesn't change the behavior of the exported validator.

PR #599, opened 2023-12-06
SWC-6614

jay-hodgson:
https://github.com/Sage-Bionetworks/synapse-web-monorepo/assets/1864447/1b146dd1-5fc7-4b27-81fa-9bb6901eeeff


packages/synapse-react-client/src/components/EntityHeaderTable/DebouncedInput.tsx
```diff
@@ -0,0 +1,51 @@
...
+  React.InputHTMLAttributes<TextFieldProps>,
+  'type' | 'min' | 'max' | 'value' | 'placeholder' | 'className' | 'list'
+>) {
+  const [value, setValue] = useState(initialValue)
+  useDebouncedEffect(
+    () => {
+      onChange(value)
+    },
+    [value],
+    300,
```
jay-hodgson, line 26: Could probably reduce even more, since this does not result in a service call.

packages/synapse-react-client/src/components/EntityHeaderTable/EntityHeaderTable.tsx
```diff
@@ -0,0 +1,542 @@
...
+    EntityType.TABLE,
+    EntityType.LINK,
+    EntityType.ENTITY_VIEW,
+    EntityType.DOCKER_REPO,
+    EntityType.SUBMISSION_VIEW,
+    EntityType.DATASET,
+    EntityType.DATASET_COLLECTION,
+    EntityType.MATERIALIZED_VIEW,
+    EntityType.VIRTUAL_TABLE,
+  ]
```
nickgros, line 287: If this is everything, and it's only used for the entity finder, you can delete this. The default value is all types.

If it's used for something other than the EntityFinder, you can use `Object.values(EntityType)`, or consider adding a new entry `'ALL_TYPES'` to `EntityTypeGroup` in `EntityTypeUtils`.

packages/synapse-react-client/src/components/EntityHeaderTable/EntityHeaderTable.tsx
```diff
@@ -0,0 +1,542 @@
...
+
+export type EntityHeaderTableProps = {
+  references: ReferenceList
+  isEditable: boolean
+  onUpdate?: (updatedRefs: ReferenceList) => void // when the references are updated, EntityHeaderTable will call this function with the updated list
+  removeSelectedRowsButtonText?: string
+  onUpdateEntityIDsTextbox?: (value: string) => void // when the entity IDs text box is updated, this is called
+}
+
+const UNMANAGEABLE_SUBJECT_COUNT = 10
```
nickgros, line 58: Recommend a brief comment to explain what we'll use this constant for

packages/synapse-react-client/src/components/EntityHeaderTable/EntityHeaderTable.tsx
```diff
@@ -0,0 +1,542 @@
...
+        </div>
+      </Box>
+      {totalRowCount > 0 && (
+        <Box
+          sx={{
+            overflow: 'auto',
+            maxHeight: '250px',
+            paddingLeft: '2px',
+            th: {
+              backgroundColor: '#eee',
```
nickgros, line 331: Looks like you override this below, unless it helps with sticky styling somehow?

packages/synapse-react-client/src/components/EntityHeaderTable/EntityHeaderTable.tsx
```diff
@@ -0,0 +1,542 @@
...
+                    }
+                    return (
+                      <th
+                        key={header.id}
+                        colSpan={header.colSpan}
+                        style={{
+                          width: columnSize,
+                          position: 'sticky',
+                          top: '0px',
+                          background: '#fff',
```
nickgros, line 363: Think this might work to prevent hard-coding (in my dreams we can one day justify a dark theme)
```suggestion
                          background: 'background.default',
```

packages/synapse-react-client/src/components/EntityHeaderTable/EntityHeaderTable.tsx
```diff
@@ -0,0 +1,542 @@
...
+                                    wrap={false}
+                                    sx={{
+                                      color: 'primary.main',
+                                      backgroundColor: 'none',
+                                      float: 'right',
+                                      marginRight: '5px',
+                                    }}
+                                  />
+                                ),
+                              }[header.column.getIsSorted() as string] ?? null}
```
nickgros, line 406: Maybe declare this object in a different place just to simplify code

packages/synapse-react-client/src/components/EntityHeaderTable/EntityHeaderTable.tsx
```diff
@@ -0,0 +1,542 @@
...
+                            </div>
+                            {header.column.getCanFilter() &&
+                            totalRowCount > UNMANAGEABLE_SUBJECT_COUNT ? (
+                              <div>
+                                <Filter column={header.column} table={table} />
```
nickgros, line 411: Love how simple it is to render this control! I think if we decide we want to align more closely with the design, it seems like this would be really easy to pull out of the table, as long as we extract the right column object!

packages/synapse-react-client/src/components/EntityHeaderTable/EntityHeaderTable.tsx
```diff
@@ -0,0 +1,542 @@
...
+                        )}
+                        {header.column.getCanResize() && (
+                          <div
+                            className={`resizer ${
+                              header.column.getIsResizing() ? 'isResizing' : ''
+                            }`}
+                            onMouseDown={header.getResizeHandler()}
+                            onTouchStart={header.getResizeHandler()}
+                          />
+                        )}
```
nickgros, line 424: Does this work with the hard-coded percentages for column width?

packages/synapse-react-client/src/components/EntityHeaderTable/EntityHeaderTable.tsx
```diff
@@ -0,0 +1,542 @@
...
+          const newEntityIDsString =
+            newEntityIDs.trim().length > 0
+              ? newEntityIDs.concat(',')
+              : newEntityIDs
+          const newValue = newEntityIDsString.concat(
+            newEntityIDsArray.join(','),
+          )
+          updateNewEntityIDs(newValue)
+          setShowEntityFinder(false)
+        }}
```
nickgros, line 480: It's super weird to me to not just add these after confirming. I think we should make it a long-term goal to have design investigate flipping this and adding a "Paste list of IDs" to the Entity Finder, instead of using the entity finder to maybe populate a text field.

packages/synapse-react-client/src/components/EntityHeaderTable/DebouncedInput.tsx
```diff
@@ -0,0 +1,51 @@
+import React, { useState } from 'react'
+import { TextField, TextFieldProps, Autocomplete } from '@mui/material'
+import { useDebouncedEffect } from '../../utils/hooks'
+
+// A debounced input react component
+
+export function DebouncedInput({
+  value: initialValue,
```
nickgros, line 8: Rename prop to `initialValue` because the internal `value` will not update if this prop updates...if you want this to act as 'fully-controlled', and have the value update when the prop updates, I think you may need to debounce where you apply the filter, instead of where you fire the input's `onChange`.

packages/synapse-react-client/src/components/EntityHeaderTable/DebouncedInput.tsx
```diff
@@ -0,0 +1,51 @@
...
+  onChange,
+  options,
+  ...props
```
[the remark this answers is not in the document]
jay-hodgson: Good idea.  (Note that this code is based on a tanstack example)

packages/synapse-react-client/package.json
```diff
@@ -48,6 +48,7 @@
     "@rjsf/utils": "5.12.1",
     "@rjsf/validator-ajv8": "5.12.1",
     "@sage-bionetworks/react-base-table": "^1.13.4",
+    "@tanstack/match-sorter-utils": "^8.8.4",
```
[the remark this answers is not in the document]
jay-hodgson: Nice catch, will remove intermediate change

packages/synapse-react-client/src/components/EntityHeaderTable/EntityHeaderTable.tsx
```diff
@@ -0,0 +1,542 @@
...
+                </IconButton>
+              </Tooltip>
+            </Box>
+            <Button
+              variant="outlined"
+              onClick={addPastedValuesToArray}
+              disabled={isLoading || newEntityIDs.trim().length == 0}
+            >
+              <AddCircleTwoTone />
+              &nbsp;Add Entities
```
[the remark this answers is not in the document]
jay-hodgson: Is this available for MUI buttons?  Or is that just for TextFields?

packages/synapse-react-client/src/components/EntityHeaderTable/EntityHeaderTable.tsx
jay-hodgson: I don't think so

packages/synapse-react-client/src/components/EntityHeaderTable/EntityHeaderTable.tsx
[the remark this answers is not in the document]
nickgros: Sorry, the correct prop name is `startIcon`: https://mui.com/material-ui/api/button/

nickgros (approved):
Looks good! Lots of remaining changes you could make, nothing blocking approval

PR #600, opened 2023-12-08
PORTALS-2868

jay-hodgson:


packages/synapse-react-client/src/components/SynapseTable/SendToCavaticaConfirmationDialog.tsx
```diff
@@ -102,132 +111,191 @@ export default function SendToCavaticaConfirmationDialog(
...
-                  Read more about CAVATICA{' '}
-                  <Link href={cavaticaHelpURL} target="_blank">
-                    here
-                  </Link>
-                  .
-                </>
-              )}
+    <div>
+      <ConfirmationDialog
+        open={isShowingExportToCavaticaModal}
```
nickgros, line 116: This state variable is really confusing me...did you hijack `isShowingExportToCavaticaModal` to now mean showing disclaimer only?
Would it make more sense to have two variables?
```
const showDisclaimer = isShowingExportToCavaticaModal && !disclaimerAcknowledged
const showExport = isShowingExportToCavaticaModal && disclaimerAcknowledged
```

Then pass those into the `open` props?

PR #605, opened 2023-12-09
PORTALS-2858: use SelectColumns when generating the IN filters for se…

jay-hodgson:
…lected rows (CAVATICA export or query for to add to the Download Cart)


packages/synapse-react-client/src/utils/functions/QueryFilterUtils.ts
```diff
@@ -1,24 +1,24 @@
...
     throw new Error('rowSelectionPrimaryKey must be defined and have length 1')
   }
   // Apply a filter that should only return the selected rows
-  const primaryKeyColumnIndex = columnModels.findIndex(
-    cm => cm.name === primaryKeyColumnNames[0],
+  const primaryKeyColumnIndex = selectColumns.findIndex(
+    sc => sc.name === primaryKeyColumnNames[0],
   )
   const filter: ColumnSingleValueQueryFilter = {
     concreteType:
```
jay-hodgson, line 24: If columns were selected in a different order than the ColumnModels order in the Table schema, then row.values[primaryKeyColumnIndex] does not return the expected result!  Use the SelectColumns instead (since the row data is synced with the SelectColumns)

nickgros (approved):
Nice catch!

PR #609, opened 2023-12-12
SWC-6622 - Handle case where restricted values for INTEGER columns could not be edited

nickgros:
SWC-6622

### DIFF
--- a/packages/synapse-react-client/src/components/TableColumnSchemaEditor/TableColumnSchemaEditor.integration.test.tsx
+++ b/packages/synapse-react-client/src/components/TableColumnSchemaEditor/TableColumnSchemaEditor.integration.test.tsx
@@ -1,5 +1,5 @@
 import React from 'react'
-import { render, screen, waitFor } from '@testing-library/react'
+import { render, screen, waitFor, within } from '@testing-library/react'
 import TableColumnSchemaEditor, {
   TableColumnSchemaEditorProps,
 } from './TableColumnSchemaEditor'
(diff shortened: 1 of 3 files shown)


packages/synapse-react-client/src/components/TableColumnSchemaEditor/TableColumnSchemaEditor.integration.test.tsx
```diff
@@ -381,4 +381,135 @@ describe('TableColumnSchemaEditor', () => {
       expect(updateTableSpy).toHaveBeenCalled()
     })
   })
+
+  it('Can update restricted values for STRING type', async () => {
```
nickgros, line 385: This already worked, just adding a test for this scenario

packages/synapse-react-client/src/components/TableColumnSchemaEditor/TableColumnSchemaEditor.integration.test.tsx
```diff
@@ -381,4 +381,135 @@ describe('TableColumnSchemaEditor', () => {
...
+    })
+    expect(restrictedValuesInput).toHaveValue('bar, true, 3, null')
+
+    await user.click(saveButton)
+    await waitFor(() => {
+      expect(updateTableSpy).toHaveBeenCalled()
+    })
+  })
+
+  it('Can update restricted values for INTEGER type', async () => {
```
nickgros, line 447: Test the change

packages/synapse-react-client/src/components/TableColumnSchemaEditor/TableColumnSchemaEditorUtils.ts
```diff
@@ -263,6 +263,19 @@ export function canHaveRestrictedValues(
...
+  enumValues: string[],
+  columnType: ColumnType | ColumnTypeEnum,
+): (string | number)[] {
+  // SWC-6622 - Special case: if these are INTEGER, convert to numbers before inserting data into form components
+  // The editor validates that these are integers, so inserting the strings returned by the API causes an
+  // unexpected validation error
+  if (columnType === ColumnTypeEnum.INTEGER) {
+    return enumValues.map(value => parseInt(value))
+  }
+  return enumValues
```
nickgros, line 276: The only column types that can have enumValues are STRING, INTEGER, ENTITYID, so this is the only special case

jay-hodgson (approved):

PR #610, opened 2023-12-13
SWC-6627 - Improvements for custom RadioGroup

nickgros:
- Allow passing a ReactNode label
 - Allow disabling the group
 - Make `id` optional, and use `useId` to generate a unique ID when one is not provided

In SWC-6627 we will be reimplementing the ability to add evaluations to a SubmissionView. The existing UI uses a radio group. We will want to disable the radio group inputs while the data changes, e.g. loading new pages

### DIFF
--- a/packages/synapse-react-client/src/components/widgets/RadioGroup.tsx
+++ b/packages/synapse-react-client/src/components/widgets/RadioGroup.tsx
@@ -17,7 +17,7 @@ export function RadioGroup<T extends string | boolean | number = string>(
     : `radiogroup`
 
   return (
-    <div id={props.id} className={className} role="radiogroup">
+    <div className={className} role="radiogroup">
       {props.options.map((option, index) => (
         <RadioOption<T>
           key={index.toString()}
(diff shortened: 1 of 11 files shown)


packages/synapse-react-client/src/components/widgets/RadioGroup.tsx
```diff
@@ -17,34 +17,34 @@ export function RadioGroup<T extends string | boolean | number = string>(
     : `radiogroup`
 
   return (
-    <div className={className} role="radiogroup">
+    <div id={props.id} className={className} role="radiogroup">
```
hallieswan, line 20: Should an `id` always be specified in the `div`? If so, should an id be generated with `useId` when `props.id` is undefined?

nickgros: I put it in the div since we weren't using the `id` prop otherwise. I don't think it would cause any problems to not have an ID here--I'd probably try to get rid of it entirely instead

hallieswan (approved):
The changes look good to me! The type check [failed](https://github.com/Sage-Bionetworks/synapse-web-monorepo/actions/runs/7199772423/job/19612215250?pr=610), so may need to clean up some types before merging

packages/synapse-react-client/src/components/widgets/RadioGroup.tsx
nickgros: I decided to completely remove the prop

PR #620, opened 2023-12-22
SWC-6617: add AR type column to AR table in ACT dashboard

hallieswan:
I mapped the AR concrete type to a display value, which is shown in the AR table. If there is an established mapping or a better approach, happy to use that instead! 

- org.sagebionetworks.repo.model.SelfSignAccessRequirement -> Click-Wrap
- org.sagebionetworks.repo.model.ManagedACTAccessRequirement -> Managed
- org.sagebionetworks.repo.model.ACTAccessRequirement -> Basic
- org.sagebionetworks.repo.model.LockAccessRequirement -> Lock
- org.sagebionetworks.repo.model.TermsOfUseAccessRequirement -> Terms of Use

SWC-6617

### DIFF
--- a/packages/synapse-react-client/src/components/AccessRequirementList/AccessRequirementList.tsx
+++ b/packages/synapse-react-client/src/components/AccessRequirementList/AccessRequirementList.tsx
@@ -4,11 +4,15 @@ import useGetInfoFromIds, {
   UseGetInfoFromIdsProps,
 } from '../../utils/hooks/useGetInfoFromIds'
 import {
+  ACT_ACCESS_REQUIREMENT_CONCRETE_TYPE_VALUE,
   AccessRequirement,
   EntityHeader,
+  MANAGED_ACT_ACCESS_REQUIREMENT_CONCRETE_TYPE_VALUE,
   ManagedACTAccessRequirement,
   Renewal,
   Request,
+  SELF_SIGN_ACCESS_REQUIREMENT_CONCRETE_TYPE_VALUE,
+  TERMS_OF_USE_ACCESS_REQUIREMENT_CONCRETE_TYPE_VALUE,
 } from '@sage-bionetworks/synapse-types'
 import IconSvg from '../IconSvg/IconSvg'
 import StandaloneLoginForm from '../Authentication/StandaloneLoginForm'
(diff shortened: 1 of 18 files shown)


packages/synapse-react-client/src/components/AccessRequirementList/AccessRequirementList.tsx
```diff
@@ -58,10 +62,10 @@ export type AccessRequirementListProps = {
...
   AccessRequirement['concreteType']
 >([
-  'org.sagebionetworks.repo.model.SelfSignAccessRequirement',
-  'org.sagebionetworks.repo.model.TermsOfUseAccessRequirement',
-  'org.sagebionetworks.repo.model.ManagedACTAccessRequirement',
-  'org.sagebionetworks.repo.model.ACTAccessRequirement',
+  SELF_SIGN_ACCESS_REQUIREMENT_CONCRETE_TYPE_VALUE,
+  TERMS_OF_USE_ACCESS_REQUIREMENT_CONCRETE_TYPE_VALUE,
+  MANAGED_ACT_ACCESS_REQUIREMENT_CONCRETE_TYPE_VALUE,
+  ACT_ACCESS_REQUIREMENT_CONCRETE_TYPE_VALUE,
```
hallieswan, line 68: Replace the AR concrete type strings with the newly defined constants

packages/synapse-react-client/src/components/dataaccess/AccessRequirementTable.integration.test.tsx
```diff
@@ -145,21 +162,28 @@ describe('Access Requirement Table tests', () => {
     screen.getByRole('columnheader', {
       name: /^Access Requirement Name/,
     })
+    screen.getByRole('columnheader', { name: 'Type' })
```
hallieswan, line 165: Confirm that newly added "Type" column is present

packages/synapse-react-client/src/components/dataaccess/AccessRequirementTable.integration.test.tsx
```diff
@@ -145,21 +162,28 @@ describe('Access Requirement Table tests', () => {
...
     screen.getByRole('columnheader', { name: 'Related to Projects' })
     screen.getByRole('columnheader', { name: 'Reviewer' })
     screen.getByRole('columnheader', { name: 'Last Modified' })
     screen.getByRole('columnheader', { name: /^Created On/ })
 
     // check the first row of data
-    screen.getByRole('cell', { name: mockSearchResults.results[0].id })
-    screen.getByRole('cell', { name: mockSearchResults.results[0].name })
+    const row = screen.getAllByRole('row')[1]
```
hallieswan, line 172: Get the first row, then check the cells within that row, so that cells in other rows aren't inadvertently matched when there are multiple cells with the same values (e.g. AR Type).

nickgros (approved):
This looks great, thanks for all the cleanup!

PR #625, opened 2024-01-02
SWC-6644: avoid password managers by not including "username" in the placeholder text

jay-hodgson:


jay-hodgson:
Here were some workarounds that I found online, none of which worked:
Input: props => {
  return <components.Input
  {...props}
  id={'user-search'}
  autoComplete='prevent autocomplete random string'
  data-lpignore="true"
  className="search"
/> 
}

nickgros (commented):
LGTM

PR #626, opened 2024-01-03
PORTALS-2884

jay-hodgson:


packages/synapse-react-client/src/components/SynapseTable/SendToCavaticaConfirmationDialog.tsx
```diff
@@ -44,7 +44,7 @@ export default function SendToCavaticaConfirmationDialog(
     fileNameColumnName,
     fileVersionColumnName,
     cavaticaHelpURL,
-    cavaticaConnectAccountURL = 'https://help.eliteportal.org/help/limited-data-commons#LimitedDataCommons-GainingAccess',
```
nickgros, line 47: Should this prop now be required? Link's `href` can now be undefined which I think is not desirable

jay-hodgson: Good point

jay-hodgson: If link is undefined, then we should not show the portal-branded "Click here for instructions" link

nickgros (approved):

PR #629, opened 2024-01-03
PORTALS-2887

jay-hodgson:


apps/portals/src/configurations/elportal/synapseConfigs/data.ts
```diff
@@ -18,7 +18,6 @@ const data: SynapseConfig = {
     cavaticaConnectAccountURL: cavaticaConnectAccountURL,
     visibleColumnCount: 10,
     isRowSelectionVisible: true,
-    additionalFiltersLocalStorageKey: 'cohort-builder-files-perspective',
```
jay-hodgson, line 21: This is a bug fix.  The V1 Files tab in the ELITE portal is not part of the cohort builder, and should not look at this special key to apply custom filters.

apps/portals/src/configurations/elportal/synapseConfigs/cohortbuilder.ts
```diff
@@ -97,7 +97,8 @@ export const individualsView: SynapseConfig = {
         },
       ],
     },
-    additionalFiltersLocalStorageKey: 'cohort-builder-individuals-perspective',
+    additionalFiltersSessionStorageKey:
```
jay-hodgson, line 100: Renamed many instances of LocalStorage to SessionStorage to be accurate.

packages/synapse-react-client/src/components/QueryWrapperPlotNav/QueryWrapperPlotNav.tsx
```diff
@@ -262,12 +262,12 @@ const QueryWrapperPlotNav: React.FunctionComponent<QueryWrapperPlotNavProps> = (
     initQueryJson,
     showLastUpdatedOn,
     unitDescription,
-    additionalFiltersLocalStorageKey,
+    additionalFiltersSessionStorageKey,
   } = props
 
   const entityId = parseEntityIdFromSqlStatement(sql)
   const additionalFilters = getAdditionalFilters(
-    additionalFiltersLocalStorageKey ?? entityId,
```
jay-hodgson, line 270: Instead of falling back to looking for an entry using the entity ID, just don't look for a filter unless the additional filter key is supplied.

packages/synapse-react-client/src/utils/functions/SqlFunctions.ts
```diff
@@ -38,26 +38,20 @@ export const getIgnoredQueryFilterSearchParamKey = (
...
+      )
+    : undefined
   let additionalFilters: QueryFilter[] = []
-  if (localStorageQueryFiltersString) {
+  if (sessionStorageQueryFiltersString) {
     additionalFilters = JSON.parse(
-      localStorageQueryFiltersString,
+      sessionStorageQueryFiltersString,
     ) as QueryFilter[]
-    // delay clearing out the value so re-rendering during React Strict mode does not destroy the filter
```
jay-hodgson, line 53: This is the final significant change.  Since it's in session storage, leave it!  If the same window is refreshed (or the user browses somewhere and returns to this page) the stored filter will be re-applied.  Until the user removes the filter, or overwrites it with another selection.

nickgros (approved):
Without a truly persistent solution for cohort builder queries, sessionStorage seems to be a much better fit than localStorage. Looks good!

PR #630, opened 2024-01-04
PORTALS-2879: update details page to also recognize entity ID lists

jay-hodgson:


apps/portals/src/portal-components/DetailsPage/DetailsPage.tsx
```diff
@@ -229,7 +229,8 @@ const SynapseObject: React.FC<{
     return <></>
   } else if (
     columnType === ColumnTypeEnum.STRING_LIST ||
-    columnType === ColumnTypeEnum.INTEGER_LIST
+    columnType === ColumnTypeEnum.INTEGER_LIST ||
+    columnType === ColumnTypeEnum.ENTITYID_LIST
```
jay-hodgson, line 233: This is the change that fixes the issue.  The option to hide UI instead of reporting No Content is only useful because of the way that we split these list values into individual components.  I think ideally they would all be fed into a single CardContainerLogic (for this specific case)

nickgros (approved):
I like the quick fix, I could see the `HIDDEN` placeholder being useful for other scenarios as well

PR #632, opened 2024-01-04
SWC-6627 - EntityHeaderTable style updates + small fixes

nickgros:
- Move filter inputs above table
 - Change sort icon to always be visible if table can be sorted
 - Handle case where filters would not be cleared if items were removed to cause the filter controls to be hidden
 - Update styling to get closer to design
 - Fix prop signature for DebouncedInput to accept `label`

|Before|After|
|---|---|
| image | image |

### DIFF
--- a/packages/synapse-react-client/src/components/EntityHeaderTable/DebouncedInput.tsx
+++ b/packages/synapse-react-client/src/components/EntityHeaderTable/DebouncedInput.tsx
@@ -16,20 +16,21 @@ export function DebouncedInput({
   options: string[]
   delay?: number
 } & Pick<
-  React.InputHTMLAttributes<TextFieldProps>,
-  'type' | 'min' | 'max' | 'value' | 'placeholder' | 'className' | 'list'
+  TextFieldProps,
+  'type' | 'value' | 'placeholder' | 'className' | 'label'
 >) {
   const [value, setValue] = useState(initialValue)
   useDebouncedEffect(
     () => {
       onChange(value)
     },
-    [value],
+    [onChange, value],
     delay,
   )
 
   return (
     <Autocomplete
+      freeSolo
       disablePortal
       isOptionEqualToValue={(option, value) =>
         value.length == 0 || option === value
--- a/packages/synapse-react-client/src/components/EntityHeaderTable/EntityHeaderTable.integration.test.tsx
+++ b/packages/synapse-react-client/src/components/EntityHeaderTable/EntityHeaderTable.integration.test.tsx
@@ -11,6 +11,7 @@ import { EntityFinderModal } from '../EntityFinder/EntityFinderModal'
 import mockFileEntityData, {
   MOCK_FILE_ENTITY_ID,
   MOCK_FILE_NAME,
+  mockFileEntities,
 } from '../../mocks/entity/mockFileEntity'
 import mockDatasetData, {
   MOCK_DATASET_ENTITY_ID,
@@ -156,5 +157,52 @@ describe('EntityHeaderTable tests', () => {
     await userEvent.click(button)
 
     expect(mockOnUpdate).toHaveBeenCalledWith([])
+  })
+
+  it('removes filters when filters are hidden', async () => {
+    const mockOnUpdate = jest.fn()
+    // 10 FileEntities, and 1 Dataset
+    const refs: ReferenceList = [
+      ...mockFileEntities.slice(0, 10).map(entity => ({ targetId: entity.id })),
+      { targetId: mockDatasetData.id },
+    ]
+    const removeEntitiesButtonText = 'Remove Test Entities'
+    renderTable({
+      references: refs,
+      isEditable: true,
+      onUpdate: mockOnUpdate,
+      removeSelectedRowsButtonText: removeEntitiesButtonText,
+    })
+
+    // Should have 12 rows: 11 items + table header
+    await waitFor(() => expect(screen.getAllByRole('row')).toHaveLength(12))
+
+    // Filter to just show the file data
+    const typeFilterInput = await screen.findByLabelText('Filter by Type', {
+      exact: false,
+    })
+    await userEvent.type(typeFilterInput, 'File')
+
+    // One row should be hidden, 10 shown (+ table header)
+    await waitFor(() => expect(screen.getAllByRole('row')).toHaveLength(11))
+
+    const checkBoxes = await screen.findAllByRole('checkbox')
+    // Click the 'Select All' checkbox
+    await userEvent.click(checkBoxes[0])
+    const button = await screen.findByRole('button', {
+      name: removeEntitiesButtonText,
+    })
+    await userEvent.click(button)
+
+    expect(mockOnUpdate).toHaveBeenCalledWith([
+      { targetId: mockDatasetData.id },
+    ])
+
+    await waitFor(() => {
+      // The filter field was removed
+      expect(typeFilterInput).not.toBeInTheDocument()
+      // The filter should have been cleared, and the remaining row (+ header) should be shown
+      expect(screen.getAllByRole('row')).toHaveLength(2)
+    })
   })
 })
--- a/packages/synapse-react-client/src/components/EntityHeaderTable/EntityHeaderTable.tsx
+++ b/packages/synapse-react-client/src/components/EntityHeaderTable/EntityHeaderTable.tsx
@@ -1,4 +1,4 @@
-import React, { useState, useMemo, MouseEventHandler, useCallback } from 'react'
+import React, { useState, useMemo, useCallback, useEffect } from 'react'
 import {
   useReactTable,
   ColumnFiltersState,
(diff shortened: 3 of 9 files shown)


packages/synapse-react-client/src/components/EntityHeaderTable/DebouncedInput.tsx
```diff
@@ -16,20 +16,21 @@ export function DebouncedInput({
   options: string[]
   delay?: number
 } & Pick<
-  React.InputHTMLAttributes<TextFieldProps>,
-  'type' | 'min' | 'max' | 'value' | 'placeholder' | 'className' | 'list'
```
nickgros, line 20: I wanted to add `label`. Keeping `'min' | 'max' | 'list'` caused type errors. I am not sure if those options worked, but if we need them options we can revisit--we don't seem to be using them now

packages/synapse-react-client/src/components/EntityHeaderTable/DebouncedInput.tsx
```diff
@@ -16,20 +16,21 @@ export function DebouncedInput({
...
       onChange(value)
     },
-    [value],
+    [onChange, value],
     delay,
   )
 
   return (
     <Autocomplete
+      freeSolo
```
nickgros, line 33: This allows typing freely, which is desired. This does not seem to actually change how the user can interact with the text input, but it silences a warning that was logged during tests.

packages/synapse-react-client/src/components/EntityHeaderTable/EntityHeaderTable.integration.test.tsx
```diff
@@ -157,4 +158,51 @@ describe('EntityHeaderTable tests', () => {
...
+
+    expect(mockOnUpdate).toHaveBeenCalledWith([
+      { targetId: mockDatasetData.id },
+    ])
+
+    await waitFor(() => {
+      // The filter field was removed
+      expect(typeFilterInput).not.toBeInTheDocument()
+      // The filter should have been cleared, and the remaining row (+ header) should be shown
+      expect(screen.getAllByRole('row')).toHaveLength(2)
```
nickgros, line 205: Before changes, the filter controls were removed, but the filter was not, so the remaining Dataset would not be visible, and the filter could not be updated to show it!

packages/synapse-react-client/src/components/EntityHeaderTable/EntityHeaderTable.tsx
```diff
@@ -192,7 +195,9 @@ export const EntityHeaderTable = (props: EntityHeaderTableProps) => {
       },
       {
         accessorFn: (row: EntityHeaderOrDummy) =>
-          row.isDummy ? '-' : getEntityTypeFromHeader(row),
+          row.isDummy
+            ? '-'
+            : entityTypeToFriendlyName(getEntityTypeFromHeader(row)),
```
nickgros, line 200: use friendly name for type

packages/synapse-react-client/src/components/EntityHeaderTable/EntityHeaderTable.tsx
```diff
@@ -281,54 +281,94 @@ export const EntityHeaderTable = (props: EntityHeaderTableProps) => {
...
+        {isEditable && refsInState.length > 0 && (
+          <Button
+            variant="contained"
+            disabled={!isSelection}
+            onClick={onRemove}
+          >
+            {removeSelectedRowsButtonText}
+          </Button>
+        )}
+      </Box>
```
nickgros, line 341: Moved filter controls up above table

packages/synapse-react-client/src/components/EntityHeaderTable/EntityHeaderTable.tsx
```diff
@@ -281,54 +281,94 @@ export const EntityHeaderTable = (props: EntityHeaderTableProps) => {
...
+                {header.column.getCanFilter() && showFilterControls ? (
+                  <Box sx={{ flexGrow: 1 }}>
+                    <Filter column={header.column} table={table} />
+                  </Box>
+                ) : null}
+              </React.Fragment>
+            )
+          }),
+        )}
       </Box>
```
nickgros, line 356: Moved filter controls so they are rendered above the table

packages/synapse-react-client/src/components/EntityHeaderTable/EntityHeaderTable.tsx
```diff
@@ -281,54 +281,94 @@ export const EntityHeaderTable = (props: EntityHeaderTableProps) => {
...
             th: {
-              backgroundColor: '#eee',
+              height: '38px',
+              backgroundColor: theme.palette.grey[200],
               zIndex: 100,
             },
-          }}
+            ['tr:nth-of-type(2n)']: {
+              backgroundColor: theme.palette.grey[100],
+            },
```
nickgros, line 370: Styling rows to match design

packages/synapse-react-client/src/components/EntityHeaderTable/EntityHeaderTable.tsx
```diff
@@ -357,58 +397,42 @@ export const EntityHeaderTable = (props: EntityHeaderTableProps) => {
...
+                                  fontSize={'inherit'}
+                                  sx={{
+                                    color: header.column.getIsSorted()
+                                      ? 'primary.main'
+                                      : 'grey.700',
+                                    backgroundColor: 'none',
+                                  }}
+                                />
+                              </IconButton>
+                            )}
```
nickgros, line 434: If `getCanSort()` is true, always show the sort icon, changing the color if `getIsSorted` is true.

packages/synapse-react-client/src/components/EntityHeaderTable/EntityHeaderTable.tsx
```diff
@@ -459,7 +483,9 @@ export const EntityHeaderTable = (props: EntityHeaderTableProps) => {
         confirmButtonCopy={`Add ${pluralObjectName}`}
         onConfirm={items => {
           if (hideTextFieldToPasteValue) {
-            onUpdate([...refsInState, ...items])
+            const newRefs = [...refsInState, ...items]
+            setRefsInState(newRefs)
+            onUpdate(newRefs)
```
nickgros, line 488: We have a weird local/parent copied state issue here, so make this small change to try to keep them in sync

packages/synapse-react-client/src/components/EntityHeaderTable/Filter.tsx
```diff
@@ -28,10 +28,9 @@ export function Filter({
       options={sortedUniqueValues}
       initialValue={columnFilterValue}
       onChange={value => column.setFilterValue(value)}
-      placeholder={`Filter by ${column.id}... (${
+      label={`Filter by ${column.columnDef.header} (${
```
nickgros, line 31: Update label text to match column header

packages/synapse-react-client/src/components/EntityHeaderTable/Filter.tsx
```diff
@@ -28,10 +28,9 @@ export function Filter({
...
         column.getFacetedUniqueValues().size
       })`}
-      list={column.id + 'list'}
```
nickgros, line 34: Not sure what this did

packages/synapse-react-client/src/mocks/entity/mockFileEntity.ts
```diff
@@ -234,6 +236,17 @@ const mockFileEntityHeader: EntityHeader = {
...
+const generatedFileEntityData: MockEntityData<FileEntity>[] = times(50).map(
+  i =>
+    generateBaseEntity(
+      {
+        concreteType: 'org.sagebionetworks.repo.model.FileEntity',
+        parentId: mockProject.id,
+      },
+      30000 + i + 1,
+    ) as MockEntityData<FileEntity>,
+)
```
nickgros, line 248: Generate 50 file entities with faker

packages/synapse-react-client/src/mocks/faker/generateFakeEntity.ts
```diff
@@ -1,26 +1,54 @@
...
 
-export function generateProject(overrides?: Partial<Project>): Project {
   return {
-    ...generateBaseEntity(overrides),
+    id: entity.id!,
+    entity: entity,
+    name: entity.name,
+    entityHeader: header,
+  }
+}
```
nickgros, line 44: Updated to return the bundle of mock data, including entity headers, that we pass along to mock service worker, so that these are easier to use in those tests and stories.

packages/synapse-react-client/src/components/EntityHeaderTable/EntityHeaderTable.integration.test.tsx
jay-hodgson: Oops, nice catch

jay-hodgson (approved):

packages/synapse-react-client/src/components/EntityHeaderTable/DebouncedInput.tsx
nickgros: 😂

PR #635, opened 2024-01-04
SWC-6627 - Add EvaluationFinder

nickgros:
- Add EvaluationFinder, which will be used to choose evaluations for a SubmissionView
 - Update MarkdownPopover to not propagate click and prevent default behavior, fixing a case where opening a popover would toggle the checkbox in the EvaluationFinder
 - Update Checkbox to allow passing any ReactNode as a label
 - Update and reuse logic for `getNextPageParam` for infinite queries where the response is a PaginatedResults object
 - Change SynapseClient.getEvaluations signature to allow passing any parameters

https://github.com/Sage-Bionetworks/synapse-web-monorepo/assets/17580037/6de1bd04-a55f-47b6-abfb-9e085061481c

### DIFF
--- a/packages/synapse-react-client/src/components/EvaluationFinder/EvaluationFinder.test.tsx
+++ b/packages/synapse-react-client/src/components/EvaluationFinder/EvaluationFinder.test.tsx
@@ -38,7 +38,7 @@ describe('EvaluationFinder', () => {
     renderComponent({ onChange, selectedIds })
 
     // The first page is visible
-    const evaluationsToSelect = await screen.findAllByRole('checkbox')
+    let evaluationsToSelect = await screen.findAllByRole('checkbox')
     expect(evaluationsToSelect).toHaveLength(2)
 
     // Verify that the selection is correct
@@ -53,7 +53,8 @@ describe('EvaluationFinder', () => {
     await userEvent.click(nextPageButton)
 
     // Page 2 will be shown
-    await waitFor(() => expect(screen.getAllByRole('checkbox')).toHaveLength(1))
+    evaluationsToSelect = await screen.findAllByRole('checkbox')
+    expect(evaluationsToSelect).toHaveLength(1)
     expect(nextPageButton).toBeDisabled()
 
     // Go back to page 1
@@ -62,6 +63,12 @@ describe('EvaluationFinder', () => {
     })
     await userEvent.click(previousPageButton)
     await waitFor(() => expect(screen.getAllByRole('checkbox')).toHaveLength(2))
+
+    evaluationsToSelect = await screen.findAllByRole('checkbox')
+    expect(evaluationsToSelect).toHaveLength(2)
+
+    // Click the first evaluation and verify that the onChange callback is called
+    await userEvent.click(evaluationsToSelect[0])
+    expect(onChange).toHaveBeenCalledWith(expect.arrayContaining(['1', '2']))
   })
-  it('shows selection and calls onChange', () => {})
 })
--- a/packages/synapse-react-client/src/components/EvaluationFinder/EvaluationFinder.test.tsx
+++ b/packages/synapse-react-client/src/components/EvaluationFinder/EvaluationFinder.test.tsx
@@ -1,0 +1,67 @@
+import React from 'react'
+import { render, screen, waitFor } from '@testing-library/react'
+import EvaluationFinder, { EvaluationFinderProps } from './EvaluationFinder'
+import { Evaluation, PaginatedResults } from '@sage-bionetworks/synapse-types'
+import SynapseClient from '../../synapse-client'
+import userEvent from '@testing-library/user-event'
+import { createWrapper } from '../../testutils/TestingLibraryUtils'
+
+function renderComponent(props: EvaluationFinderProps) {
+  return render(<EvaluationFinder {...props} />, { wrapper: createWrapper() })
+}
+
+const onChange = jest.fn()
+const selectedIds = ['2']
+
+const mockPage1: PaginatedResults<Evaluation> = {
+  results: [
+    { id: '1', name: 'Eval 1', submissionInstructionsMessage: '_Markdown_' },
+    { id: '2', name: 'Eval 2' },
+  ],
+  totalNumberOfResults: 3,
+}
+const mockPage2: PaginatedResults<Evaluation> = {
+  results: [{ id: '3', name: 'Eval 3' }],
+  totalNumberOfResults: 3,
+}
+
+jest.spyOn(SynapseClient, 'getEvaluations').mockImplementation(params => {
+  if (params == null || params.offset == null || params.offset === 0) {
+    return Promise.resolve(mockPage1)
+  } else {
+    return Promise.resolve(mockPage2)
+  }
+})
+
+describe('EvaluationFinder', () => {
+  it('fetches and shows pages of evaluations', async () => {
+    renderComponent({ onChange, selectedIds })
+
+    // The first page is visible
+    const evaluationsToSelect = await screen.findAllByRole('checkbox')
+    expect(evaluationsToSelect).toHaveLength(2)
+
+    // Verify that the selection is correct
+    expect(evaluationsToSelect[0]).not.toBeChecked()
+    expect(evaluationsToSelect[1]).toBeChecked()
+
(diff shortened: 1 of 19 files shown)


packages/synapse-react-client/src/components/EvaluationFinder/EvaluationFinder.test.tsx
```diff
@@ -0,0 +1,67 @@
...
+    expect(nextPageButton).toBeDisabled()
+
+    // Go back to page 1
+    const previousPageButton = await screen.findByRole('button', {
+      name: 'Previous',
+    })
+    await userEvent.click(previousPageButton)
+    await waitFor(() => expect(screen.getAllByRole('checkbox')).toHaveLength(2))
+  })
+  it('shows selection and calls onChange', () => {})
```
nickgros, line 66: Todo

packages/synapse-react-client/src/components/Markdown/MarkdownPopover.tsx
```diff
@@ -108,6 +105,14 @@ export const MarkdownPopover: React.FunctionComponent<MarkdownPopoverProps> = ({
...
         role="button"
         className={'PopoverContainer'}
+        onClick={e => {
+          // Prevent the default action of the click event -- for example, if this is in a `label` tag for a checkbox,
+          // prevent the click from toggling the checkbox value
+          e.preventDefault()
+          // Prevent the click from propagating to the parent container
+          e.stopPropagation()
+          setOpenMarkdownPopoverId(currentId => (currentId == id ? null : id))
+        }}
```
nickgros, line 115: Clicking the Help popover that appears in a checkbox label should not toggle the checkbox.

packages/synapse-react-client/src/components/styled/LightTooltip.tsx
```diff
@@ -26,8 +26,8 @@ export const LightTooltip: StyledComponent<TooltipProps> = styled(
     },
     color: theme.palette.background.paper,
   },
-  [`& .${linkClasses.root}`]: {
-    color: theme.palette.primary.main,
+  [`& .${tooltipClasses.tooltip} .${linkClasses.root}`]: {
+    color: `${theme.palette.primary.main}`,
   },
```
nickgros, line 31:
```suggestion
  [`& .${tooltipClasses.tooltip} .${linkClasses.root}`]: {
    color: theme.palette.primary.main,
  },
```

packages/synapse-react-client/src/components/styled/LightTooltip.tsx
```diff
@@ -26,8 +26,8 @@ export const LightTooltip: StyledComponent<TooltipProps> = styled(
...
 }))
```
nickgros, line 32: Increase the specificity of this rule to ensure it actually applies. It currently has the same specificity as the rule that makes Links in Tooltips white, since our tooltips have a dark background by default. This is inappropriate for the `LightTooltip`, so we use the `main` color

packages/synapse-react-client/src/components/widgets/Checkbox.tsx
```diff
@@ -1,17 +1,27 @@
...
+    'data-testid'?: string
+    'aria-label'?: string
+  } & (
+    | {
+        label?: React.ReactNode
+        /* If the label is not a string, then we require an aria-label */
+        'aria-label': string
+      }
+    | { label: string }
+  )
```
nickgros, line 23: Add ability to add an arbitrary React component as a label, but if it's not a string, require an `aria-label`

packages/synapse-react-client/src/synapse-client/SynapseClient.ts
```diff
@@ -870,15 +869,40 @@ export const getGroupHeadersBatch = (
...
+  }
+
+  // If evaluation IDs were explicitly specified, fetch all pages
+  if (params.evaluationIds) {
+    const results = await getAllOfPaginatedService(fn)
+    return {
+      totalNumberOfResults: results.length,
+      results,
+    }
+  }
```
nickgros, line 902: When you edit a SubmissionView, the evaluation IDs that are already in the view scope can just be fetched all at once without having to worry about pagination.

packages/synapse-react-client/src/components/EvaluationFinder/EvaluationFinder.tsx
```diff
@@ -0,0 +1,102 @@
...
+        activeOnly,
+      },
+      { keepPreviousData: true },
+    )
+
+  if (isLoading) {
+    return <LinearProgress />
+  }
+  if (!data?.pages) {
+    return <></>
```
jay-hodgson, line 39: No content UI instead?

packages/synapse-react-client/src/components/EvaluationFinder/EvaluationFinder.tsx
```diff
@@ -0,0 +1,102 @@
...
+  }
+
+  const canGoToNextPage =
+    (data.pages.length - 1 > currentPage || hasNextPage) && !isFetchingNextPage
```
jay-hodgson, line 43: This suggests that there is a case when `hasNextPage `is `false `but `data.pages.length - 1 > currentPage` is `true`?

jay-hodgson: To be more clear, why is this not enough?  `hasNextPage && !isFetchingNextPage`

packages/synapse-react-client/src/components/EvaluationFinder/EvaluationFinder.tsx
```diff
@@ -0,0 +1,102 @@
...
+          </Button>
+        )}
+        <Button
+          variant={'outlined'}
+          disabled={!canGoToNextPage}
+          onClick={() => {
+            if (data.pages[currentPage + 1]) {
+              setCurrentPage(page => page + 1)
+            } else {
+              fetchNextPage().then(() => setCurrentPage(page => page + 1))
```
jay-hodgson, line 93: What if there's an error fetching the next page?

jay-hodgson (approved):

packages/synapse-react-client/src/components/EvaluationFinder/EvaluationFinder.tsx
nickgros: Mismatch between the infinite query API and how it's used in this component. `hasNextPage` tells us if there's another page on the server. This breaks down with the following sequence:
1. There are two pages of data
2. Component is rendered, page 1 loads
3. Go to page 2. `hasNextPage` is now false
4. Go back to page 1. `hasNextPage` is still false, but we need a way to go to page 2. 

I think I could do a better job of organizing + using better variable names to make this more clear.

packages/synapse-react-client/src/components/EvaluationFinder/EvaluationFinder.tsx
nickgros: IMO we should propagate an error to an error boundary, which makes this case improbable. If we do that, it seems reasonable to show UI that indicates that an unknown error occurred

packages/synapse-react-client/src/components/EvaluationFinder/EvaluationFinder.tsx
nickgros: The custom hook will return the error, so we shouldn't have to handle it here. I propose that we have the custom hook pass the error up to the nearest error boundary.

packages/synapse-react-client/src/components/EvaluationFinder/EvaluationFinder.tsx
```diff
@@ -23,24 +24,35 @@ export type EvaluationFinderProps = Pick<
...
   }
   if (!data?.pages) {
-    return <></>
+    // This should never happen since errors are propagated to the error boundary
+    return (
+      <Alert severity="error">
+        An unexpected error occurred and evaluations could not be loaded
+      </Alert>
+    )
   }
```
nickgros, line 51: If `isLoading` is true and `data?.pages` is falsy, then it must be true that either 1) the query is disabled, which should not happen here or 2) an error occurred. But if an error occurs, we're relying on the error boundary for UI. So this UI is now just fallback for a case that shouldn't happen

PR #634, opened 2024-01-04
PORTALS-2881: update e2e tests

hallieswan:

PR #636, opened 2024-01-04
SWC-6529: publish @sage-bionetworks/synapse-types package

hallieswan:
See passing npmpublish dryrun for `@sage-bionetworks/synapse-types` [here](https://github.com/Sage-Bionetworks/synapse-web-monorepo/actions/runs/7415916429/job/20179937481)

Note: failing e2e tests are fixed in #634

### DIFF
--- a/.github/workflows/npmpublish.yml
+++ b/.github/workflows/npmpublish.yml
@@ -10,16 +10,26 @@ jobs:
   tag-and-publish:
     strategy:
       matrix:
-        publishablePackage: [synapse-react-client, markdown-it-synapse]
+        publishablePackage:
+          [
+            synapse-react-client,
+            markdown-it-synapse,
+            '@sage-bionetworks/synapse-types',
+          ]
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v3.1.0
       - uses: ./.github/actions/pnpm-setup-action
+      - name: Get package path
+        id: package-path
+        run: |
+          packagePath=$(pnpm nx show project ${{ matrix.publishablePackage }} --json | jq -r '.sourceRoot')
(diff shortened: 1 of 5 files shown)


.github/workflows/npmpublish.yml
```diff
@@ -10,16 +10,26 @@ jobs:
...
+          ]
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v3.1.0
       - uses: ./.github/actions/pnpm-setup-action
+      - name: Get package path
+        id: package-path
+        run: |
+          packagePath=$(pnpm nx show project ${{ matrix.publishablePackage }} --json | jq -r '.sourceRoot')
+          echo "packagePath=${packagePath}" >> "$GITHUB_OUTPUT"
```
hallieswan, line 27: Handle case where package name (`@sage-bionetworks/synapse-types`) may differ from package path (`packages/synapse-types`)

packages/synapse-types/LICENSE
```diff
@@ -0,0 +1,201 @@
...
+      To apply the Apache License to your work, attach the following
+      boilerplate notice, with the fields enclosed by brackets "{}"
+      replaced with your own identifying information. (Don't include
+      the brackets!)  The text should be enclosed in the appropriate
+      comment syntax for the file format. We also recommend that a
+      file or class name and description of purpose be included on the
+      same "printed page" as the copyright notice for easier
+      identification within third-party archives.
+
+   Copyright 2023 Sage Bionetworks
```
hallieswan, line 189: Happy to update if there's more appropriate license or copyright date to use here.

nickgros: Apache 2 seems right, since we use that for other libs

nickgros (approved):
Looks good, let's see if it works 🙂

PR #638, opened 2024-01-05
PORTALS-2878: followup

jay-hodgson:

PR #641, opened 2024-01-06
Attempt to fix flaky AccessSubmissionDashboard tests

nickgros:
- Tweak waitFor assertions to check the props passed to a mock component 
 - Bump 'history' package to v4, v5 is not compatible with React Router v5
 - Remove test skips

Locally, I ran multiple iterations of the AccessSubmissionDashboard tests. I experienced irregular test failures with 'history' v5 that went away when I modified the waitFor assertions. Consistency may also have been helped by changing the version of history to v4. The documentation for the history package notes that v4 is compatible with React Router v5. If/when we upgrade to react-router v6, we should bump history back to v5.


packages/synapse-react-client/src/components/dataaccess/AccessSubmissionDashboard.integration.test.tsx
```diff
@@ -43,7 +43,6 @@ function renderComponent(modifyHistory?: (history: MemoryHistory) => void) {
     modifyHistory(history)
   }
   const renderResult = render(
-    // @ts-expect-error - seems to be an obscure type mismatch
```
nickgros, line 46: not that obscure! we were just using incompatible dependency versions

packages/synapse-react-client/src/components/dataaccess/AccessSubmissionDashboard.integration.test.tsx
```diff
@@ -161,10 +160,6 @@ describe('AccessSubmissionDashboard tests', () => {
     )
   })
 
-  //  TODO:  Error in Travis build.  Commenting out to release SRC.
-  //  Example: https://app.travis-ci.com/github/Sage-Bionetworks/Synapse-React-Client/builds/258312595
-  //  Expected: "999" (MOCK_USER_ID)
-  //  Received: null
   it('Updates the passed props and URLSearchParams when updating reviewerId', async () => {
```
nickgros, line 163: This test failed in #597, #635 , #638, and possibly more

packages/synapse-react-client/src/components/dataaccess/AccessSubmissionDashboard.integration.test.tsx
```diff
@@ -103,7 +102,7 @@ describe('AccessSubmissionDashboard tests', () => {
     )
   })
 
-  it.skip('Updates the passed props and URLSearchParams when updating arName', async () => {
```
nickgros, line 106: These tests were already skipped, but they seem less flaky when using the correct version of `history`

apps/portals/package.json
```diff
@@ -10,6 +10,7 @@
     "@mui/material": "^5.14.17",
     "@react-hookz/web": "^23.1.0",
     "@sage-bionetworks/react-base-table": "^1.13.4",
+    "@sage-bionetworks/synapse-types": "workspace:*",
```
nickgros, line 13: This package is not exclusively types and does have some JS code, such as enums, so `dependencies` is probably more appropriate than `devDependencies`. This should not impact builds.

apps/portals/src/configurations/elportal/synapseConfigs/handleFilesToParticipants.tsx
```diff
@@ -6,14 +6,14 @@ import {
   SynapseClient,
   SynapseConstants,
   SynapseUtilityFunctions,
+  CustomControlCallbackData,
 } from 'synapse-react-client'
-import { CustomControlCallbackData } from 'synapse-react-client/src/components/SynapseTable/TopLevelControls/TopLevelControls'
```
nickgros, line 10: This is a type import so it was removed from compilation, so it happened to work in this case, but we should not import from direct paths

apps/portals/src/configurations/elportal/synapseConfigs/handleFilesToParticipants.tsx
```diff
@@ -6,14 +6,14 @@ import {
...
 
 export const handleFilesToParticipants = async (
   event: CustomControlCallbackData,
 ) => {
   // add filter for files perspective, to show participants associated to the selected files only.
-  let token = undefined
+  let token: string | undefined = undefined
```
nickgros, line 16: This needed a type annotation otherwise I got a compilation error

packages/synapse-react-client/src/components/dataaccess/AccessRequirementDashboard.integration.test.tsx
```diff
@@ -60,18 +59,20 @@ describe('AccessRequirementDashboard tests', () => {
...
+      expect(mockAccessRequirementTable).toHaveBeenLastCalledWith(
+        expect.objectContaining({
+          nameContains: '',
+          relatedProjectId: undefined,
+          reviewerId: undefined,
+          onCreateNewAccessRequirementClicked: mockOnCreateNewAR,
+        }),
+        expect.anything(),
+      ),
     )
```
nickgros, line 72: Wait for the last call to this mocked component to have the expected props. Part of the flakiness may be that we aren't giving React enough time to get through its render loop

nickgros:
Note that I ran the test job on the latest commit 5 times and got 5 consecutive successful runs: https://github.com/Sage-Bionetworks/synapse-web-monorepo/actions/runs/7432473054/job/20265252842?pr=641

apps/synapse-oauth-signin/package.json
```diff
@@ -10,7 +10,7 @@
     "@mui/material": "^5.14.17",
     "@react-hookz/web": "^23.1.0",
     "@sage-bionetworks/react-base-table": "^1.13.4",
-    "history": "^5.3.0",
+    "history": "^4.10.1",
```
jay-hodgson, line 13: I wish we could add a comment in json to describe the situation.  What do you think about opening a ticket to upgrade react-router to v6, and bump history back to v5?

jay-hodgson (approved):

apps/synapse-oauth-signin/package.json
nickgros: I added a comment to PORTALS-2208

PR #644, opened 2024-01-08
SWC-6627 - Add SubmissionViewScopeEditor

nickgros:
https://github.com/Sage-Bionetworks/synapse-web-monorepo/assets/17580037/3522f494-8357-4a37-a1cf-1a3d609db4f5

### DIFF
--- a/packages/synapse-react-client/src/components/SubmissionViewScopeEditor/SubmissionViewScopeEditor.tsx
+++ b/packages/synapse-react-client/src/components/SubmissionViewScopeEditor/SubmissionViewScopeEditor.tsx
@@ -2,7 +2,7 @@ import React from 'react'
 import { ACCESS_TYPE } from '@sage-bionetworks/synapse-types'
 import EvaluationFinder from '../EvaluationFinder/EvaluationFinder'
 import { useGetEvaluation } from '../../synapse-queries/evaluation/useEvaluation'
-import { Box, IconButton, Typography } from '@mui/material'
+import { Box, IconButton, Skeleton, Typography } from '@mui/material'
 import IconSvg from '../IconSvg'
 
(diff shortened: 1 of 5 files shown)


packages/synapse-react-client/src/components/SubmissionViewScopeEditor/SubmissionViewScopeEditor.tsx
```diff
@@ -0,0 +1,96 @@
...
+  evaluationIds: string[]
+  onChange: (evaluationIds: string[]) => void
+}
+
+/**
+ * Component used to select evaluations to define a SubmissionView scope
+ * @param props
+ * @constructor
+ */
+export default function SubmissionViewScopeEditor(
```
nickgros, line 58: Slight improvement here where we show the selection at the same time as the full list of evaluations. This way you don't have to open a new modal to add each individual evaluation

packages/synapse-react-client/src/synapse-queries/evaluation/useEvaluation.ts
```diff
@@ -17,14 +35,26 @@ export function useGetEvaluationsInfinite(
...
-      ),
+      )
+
+      // Update the cache with each individual evaluation
+      responsePage.results.forEach(evaluation => {
+        queryClient.setQueryData(
+          keyFactory.getEvaluationByIdQueryKey(evaluation.id!),
+          evaluation,
+        )
+      })
```
nickgros, line 54: This can prevent a lot of unnecessary requests where many evaluations in your "Selected" list may already be in the cache because it was in the list of all evaluations when you were making your selection

hallieswan (approved):
Looks good to me! Really clear components and tests

PR #647, opened 2024-01-09
SWC-6627 - Add EntityViewMaskEditor and EntityViewScopeEditor

nickgros:
## Mask 

image

image

## Scope

https://github.com/Sage-Bionetworks/synapse-web-monorepo/assets/17580037/ce53dc63-ed51-4463-bf89-ad18173b603e

### DIFF
--- a/packages/synapse-react-client/src/components/EntityViewScopeEditor/EntityViewMaskEditor.test.tsx
+++ b/packages/synapse-react-client/src/components/EntityViewScopeEditor/EntityViewMaskEditor.test.tsx
@@ -52,7 +52,7 @@ describe('EntityViewMaskEditor tests', () => {
     )
   })
 
-  it('Disables input if an unsupported mask value is passed', async () => {
+  it('Disables input if an unsupported mask value is passed', () => {
     const value = ENTITY_VIEW_TYPE_MASK_DOCKER
     const onChange = jest.fn()
     renderComponent({ value, onChange })
--- a/packages/synapse-react-client/src/components/EntityViewScopeEditor/EntityViewMaskEditor.test.tsx
+++ b/packages/synapse-react-client/src/components/EntityViewScopeEditor/EntityViewMaskEditor.test.tsx
@@ -53,13 +53,18 @@ describe('EntityViewMaskEditor tests', () => {
   })
 
   it('Disables input if an unsupported mask value is passed', () => {
-    const value = ENTITY_VIEW_TYPE_MASK_DOCKER
+    const value = ENTITY_VIEW_TYPE_MASK_FILE | ENTITY_VIEW_TYPE_MASK_DOCKER
     const onChange = jest.fn()
     renderComponent({ value, onChange })
 
     screen
       .getAllByRole('checkbox')
       .forEach(checkbox => expect(checkbox).toBeDisabled())
+
+    expect(screen.getByLabelText('Files')).toBeChecked()
+    expect(screen.getByLabelText('Folders')).not.toBeChecked()
+    expect(screen.getByLabelText('Tables')).not.toBeChecked()
+    expect(screen.getByLabelText('Datasets')).not.toBeChecked()
 
     const alert = screen.getByRole('alert')
     within(alert).getByText(
--- a/packages/synapse-react-client/src/components/EntityViewScopeEditor/EntityViewMaskEditor.test.tsx
+++ b/packages/synapse-react-client/src/components/EntityViewScopeEditor/EntityViewMaskEditor.test.tsx
@@ -1,0 +1,85 @@
+import React from 'react'
+import EntityViewMaskEditor, {
+  EntityViewMaskEditorProps,
+  isMaskSupportedInUI,
+} from './EntityViewMaskEditor'
+import { render, screen, within } from '@testing-library/react'
+import { createWrapper } from '../../testutils/TestingLibraryUtils'
+import {
+  ENTITY_VIEW_TYPE_MASK_DOCKER,
+  ENTITY_VIEW_TYPE_MASK_FILE,
+  ENTITY_VIEW_TYPE_MASK_FOLDER,
+  ENTITY_VIEW_TYPE_MASK_TABLE,
+} from '@sage-bionetworks/synapse-types'
+import userEvent from '@testing-library/user-event'
+
+describe('EntityViewMaskEditor tests', () => {
+  function renderComponent(props: EntityViewMaskEditorProps) {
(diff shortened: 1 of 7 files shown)


packages/synapse-react-client/src/components/EntityViewScopeEditor/EntityViewMaskEditor.tsx
```diff
@@ -0,0 +1,64 @@
...
+]
+
+export function isMaskSupportedInUI(value: number) {
+  // Union of all the supported UI mask values
+  const allMasksSupportedInEditor = maskLabelPairs.reduce(
+    (acc, [, mask]) => acc | mask,
+    0,
+  )
+  // Flip the bits and & with the value -- if any bits remain, then an unsupported mask is in use
+  return (~allMasksSupportedInEditor & value) === 0
```
nickgros, line 30:
```suggestion
  
  const allMasksSupportedInEditor = maskLabelPairs.forEach(
    ([, mask]) => {
    	// Remove each supported bit
    	if ((value & mask) !== 0) {
    		value = value - mask
    	}
    }
  )
  // If value is not zero, then some unsupported bit remains
  return value === 0
```

packages/synapse-react-client/src/components/EntityViewScopeEditor/EntityViewScopeEditor.tsx
```diff
@@ -0,0 +1,58 @@
...
+export type EntityViewScopeEditorProps = {
+  scopeIds: string[]
+  onChange: (scopeIds: string[]) => void
+  isProjectView?: boolean
+}
+
+export default function EntityViewScopeEditor(
+  props: EntityViewScopeEditorProps,
+) {
+  const { scopeIds, isProjectView = false } = props
```
hallieswan, line 21: Could `onChange` be destructured here too? Just curious if there are situations where it's best to access a value directly from the `props` object

packages/synapse-react-client/src/components/EntityViewScopeEditor/EntityViewScopeEditor.tsx
```diff
@@ -0,0 +1,58 @@
...
+        </Typography>
+      )}
+      <EntityHeaderTable
+        references={references}
+        isEditable={true}
+        onUpdate={newReferences => {
+          props.onChange(newReferences.map(ref => ref.targetId))
+        }}
+        removeSelectedRowsButtonText={'Remove Selected Items from View Scope'}
+        objectNameCopy={'container'}
```
hallieswan, line 44: Not sure what the design specifies, but would it make sense to toggle `objectNameCopy` to 'projects' when `isProjectView` is true?

packages/synapse-react-client/src/components/EntityViewScopeEditor/EntityViewScopeEditor.test.tsx
```diff
@@ -0,0 +1,146 @@
...
+        },
+        expect.anything(),
+      ),
+    )
+  })
+
+  it('Restricts the selectable types based on isProjectView', async () => {
+    const scopeIds = ['syn123']
+    const onChange = jest.fn()
+    const isProjectView = true // !
```
hallieswan, line 111: Should there be more info added to this comment?

hallieswan: Ah, or is it referring the reader to the ! comment below?

packages/synapse-react-client/src/components/EntityViewScopeEditor/EntityViewMaskEditor.test.tsx
```diff
@@ -0,0 +1,85 @@
...
+
+    screen
+      .getAllByRole('checkbox')
+      .forEach(checkbox => expect(checkbox).toBeDisabled())
+
+    const alert = screen.getByRole('alert')
+    within(alert).getByText(
+      'A custom mask is in use. Changes cannot be made in the UI.',
+    )
+  })
```
hallieswan, line 68: If a supported and an unsupported mask were passed (e.g. `ENTITY_VIEW_TYPE_MASK_FILE | ENTITY_VIEW_TYPE_MASK_DOCKER`), would the supported mask be checked and disabled?

hallieswan (approved):
Looks good to me! Just a couple small questions

packages/synapse-react-client/src/components/EntityViewScopeEditor/EntityViewMaskEditor.tsx
```diff
@@ -0,0 +1,64 @@
...
+    <>
+      <Typography variant={'body1'} mt={2.5} mb={1.25} sx={{ fontWeight: 700 }}>
+        Include in View
+      </Typography>
+      <FormGroup sx={{ gap: 1 }}>
+        {maskLabelPairs.map(([label, mask]) => (
+          <Checkbox
+            key={label}
+            label={label}
+            checked={(value & mask) > 0}
```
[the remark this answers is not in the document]
nickgros: It wouldn't hurt, I'll add it

packages/synapse-react-client/src/components/EntityViewScopeEditor/EntityViewScopeEditor.tsx
nickgros: Just an inconsistency that I will fix. The only times I think it matters are
1. If you destructure some props and spread the rest e.g. ` const { foo, bar, ...rest } = props` since the destructured props affect the value of `rest`
2. Where Playwright somehow tracks which values you destructure in the function signature (which still doesn't make a lot of sense to me 😅)

packages/synapse-react-client/src/components/EntityViewScopeEditor/EntityViewScopeEditor.tsx
nickgros: Great idea!

packages/synapse-react-client/src/components/EntityViewScopeEditor/EntityViewScopeEditor.test.tsx
nickgros: Just shorthand to call out the parameter under test but it would not hurt to call it out more explicitly

packages/synapse-react-client/src/components/EntityViewScopeEditor/EntityViewMaskEditor.test.tsx
nickgros: Yes

PR #648, opened 2024-01-10
SWC-6640

jay-hodgson:


jay-hodgson:
Screen Shot 2024-01-10 at 2 51 00 PM

packages/synapse-react-client/src/utils/hooks/useDetectSSOCode.ts
```diff
@@ -54,13 +55,35 @@ export default function useDetectSSOCode(
   // state is used during OAuth based Synapse account creation (it's the username)
   const state = searchParams?.get('state')
 
+  const { accessToken } = useSynapseContext()
```
jay-hodgson, line 58: This is the tricky bit. If logged into Synapse and you're logging into ORCiD, then assume this is to bind the ORCiD to your account.  If not logged in and logging into ORCiD, then assume user is attempting to login via ORCiD to an account that is already bound to an ORCiD account

nickgros: Clever!

packages/synapse-react-client/src/utils/hooks/useDetectSSOCode.ts
```diff
@@ -80,13 +103,13 @@ export default function useDetectSSOCode(
...
           }
-          console.error('Error with Google account association: ', err)
+          console.error('Error with account login: ', err)
           if (onError) {
             onError(err.reason)
           }
         }
 
-        if (state) {
+        if (OAUTH2_PROVIDERS.GOOGLE == provider && state) {
```
nickgros, line 112: Got it, this is because we don't support account creation with ORCiD.

nickgros (approved):

packages/synapse-react-client/src/utils/hooks/useDetectSSOCode.ts
jay-hodgson: Exactly

PR #649, opened 2024-01-11
SWC-6638: separate certification and validation badges

hallieswan:
See corresponding changes in SWC: https://github.com/Sage-Bionetworks/SynapseWebClient/pull/5253

SRC_certified_tooltip
SRC_certified_validated


packages/synapse-react-client/src/components/AccountLevelBadge/AccountLevelBadge.test.tsx
```diff
@@ -1,79 +1,50 @@
...
-const mockVerified: UserBundle = {
-  userId: '345424',
-  isCertified: true,
-  isVerified: true,
-}
-
-describe('basic functionality', () => {
-  const props = {
-    userId: '1234',
-  }
```
hallieswan, line 38: Moved into AccountLevelBadges.test.tsx

packages/synapse-react-client/src/components/AccountLevelBadge/AccountLevelBadge.tsx
```diff
@@ -1,67 +1,66 @@
...
+}
 
 export const AccountLevelBadge: React.FunctionComponent<
   AccountLevelBadgeProps
-> = ({ userId }: AccountLevelBadgeProps) => {
-  const {
-    data: userBundle,
-    isLoading,
-    error,
-  } = useGetUserBundle(userId, CERTIFICATION_AND_VERIFICATION_BUNDLE_MASK)
```
hallieswan, line 28: Moved the service call into AccountLevelBadges. This component now just creates a particular type of badge.

packages/synapse-react-client/src/components/AccountLevelBadges/AccountLevelBadges/AccountLevelBadges.stories.ts
hallieswan, line 1: Replicates the original AccountLevelBadge stories

packages/synapse-react-client/src/components/AccountLevelBadges/AccountLevelBadges.tsx
```diff
@@ -0,0 +1,48 @@
...
+  const {
+    data: userBundle,
+    isLoading,
+    error,
+  } = useGetUserBundle(userId, CERTIFICATION_AND_VERIFICATION_BUNDLE_MASK)
+
+  const hasBadges = userBundle?.isCertified || userBundle?.isVerified
+
+  if (isLoading) {
+    return <></>
```
nickgros, line 28: Not sure if you have the container/label hidden by default in https://github.com/Sage-Bionetworks/SynapseWebClient/pull/5253

If you do, consider showing one or two Skeleton badges during loading.

You might be able to easily get a badge of the correct size if you do 
```
<Skeleton>
  <AccountLevelBadge badgeType="certified" />
</Skeleton>
```

packages/synapse-react-client/src/components/AccountLevelBadge/AccountLevelBadge.tsx
```diff
@@ -1,67 +1,66 @@
...
-    icon = <Certified />
-  }
-  if (userBundle?.isVerified) {
-    accountLevelString = accountLevelVerifiedLabel
-    icon = <Validated />
-  }
+> = ({ badgeType }: AccountLevelBadgeProps) => {
+  const badgeConfig = accountLevelBadgeConfig[badgeType]
   return (
     <div className={'AccountLevelBadge cardContainer'}>
```
nickgros, line 43: Kind of picky and maybe out-of-scope for this ticket, so feel free to ignore--consider checking if this looks closer to the design if we drop the `.cardContainer` styling and use a MUI `Card` component.

I think that the `.cardContainer` class styles and our design system have drifted, but I'd be worried to change `.cardContainer` in this PR since it is used in many places.

If that's an easy change, maybe we follow up with a ticket to remove the style and replace it with the component.

nickgros (approved):
Looks great! Couple of small suggestions

packages/synapse-react-client/src/components/AccountLevelBadges/AccountLevelBadges.tsx
nickgros: Looking at your other PR more closely, since it's visible by default, I think the fragment is fine--it's kind of a judgement call since a lot of users won't have anything shown after loading.

hallieswan: Oh good point! Since most users won't have a badge shown after loading, I'll keep it as is for now. If most users end up with badges in the future, then we can add the Skeleton!

packages/synapse-react-client/src/components/AccountLevelBadge/AccountLevelBadge.tsx
hallieswan: Good call! I updated this component to use the MUI `Card` and opened [this ticket](https://sagebionetworks.jira.com/browse/SWC-6652) to track the work elsewhere.

packages/synapse-react-client/src/components/AccountLevelBadge/AccountLevelBadge.tsx
```diff
@@ -40,28 +40,35 @@ export const AccountLevelBadge: React.FunctionComponent<
 > = ({ badgeType }: AccountLevelBadgeProps) => {
   const badgeConfig = accountLevelBadgeConfig[badgeType]
   return (
-    <div className={'AccountLevelBadge cardContainer'}>
+    <Card
```
hallieswan, line 43: Use MUI `Card` to more closely align with design. Move styles from `_account-level-badges.scss` into the component.

hallieswan: SRC_updated_card

packages/synapse-react-client/src/style/components/_all.scss
```diff
@@ -8,15 +8,14 @@
   './error', './goals', './feed-cards', './upset-plot', './resources',
   'cards-rotate', './expandable_content', './bar-loader', './featured-data-tabs',
   './programs', './user-card-list-groups', './evaluation_queue/all',
-  './personal-access-tokens', './account-level-badge', './terms-and-conditions',
```
hallieswan, line 11: Only change here was removing `./account-level-badge`

packages/synapse-react-client/src/components/AccountLevelBadge/AccountLevelBadge.tsx
nickgros: Awesome! Thanks for converting the whole component

nickgros (approved):

PR #650, opened 2024-01-11
SWC-6627 - Add WizardChoiceButton, TableTypeSelection, ViewTypeSelection, SqlDefinedTableEditor

nickgros:
- Add WizardChoiceButton
 - Add TableTypeSelection, ViewTypeSelection
 - Add SqlDefinedTableEditor
 - DefaultValueField: fallback to `undefined` in case `value` is null, which is not a valid input value
 - EntityHeaderTable: Number of skeleton/loading rows should match number of refs passed

## Screenshots

|Feature|Screenshot|
|---|---|
|TableTypeSelection (WizardChoiceButton)| image |
|ViewTypeSelection (WizardChoiceButton)| image |
|SqlDefinedTableEditor| image |

### DIFF
--- a/packages/synapse-react-client/src/components/CreateTableViewWizard/TableTypeSelection.tsx
+++ b/packages/synapse-react-client/src/components/CreateTableViewWizard/TableTypeSelection.tsx
@@ -6,7 +6,7 @@ import { Link } from '@mui/material'
 
 const TABLE_DESCRIPTION = (
   <>
-    <WizardChoiceButtonDescription my={1.25}>
+    <WizardChoiceButtonDescription>
       Synapse tables are used to organize web-accessible, sharable, and
       queryable data. Tables may be queried and edited with the Synapse UI, as
       well as with the Synapse programmatic clients.
@@ -27,7 +27,7 @@ const TABLE_DESCRIPTION = (
 
 const VIEW_DESCRIPTION = (
   <>
-    <WizardChoiceButtonDescription my={1.25}>
+    <WizardChoiceButtonDescription>
       Views allow you to see and query groups of other objects in Synapse, such
       as files, tables, projects, or submissions and any associated annotations
       about those items.
@@ -45,7 +45,7 @@ const VIEW_DESCRIPTION = (
 )
 
 export type TableTypeSelectionProps = {
-  /* Callback including the chosen type, and optional viewTypeMask if included in the choice (e.g. Project Views) */
+  /* Callback invoked if the user selects "Table" */
   onTableSelected: () => void
   /* Callback invoked if the user selects the "View" group, which will show more options */
   onViewSelected: () => void
--- a/packages/synapse-react-client/src/components/CreateTableViewWizard/TableTypeSelection.tsx
+++ b/packages/synapse-react-client/src/components/CreateTableViewWizard/TableTypeSelection.tsx
@@ -1,0 +1,76 @@
+import React from 'react'
+import WizardChoiceButtonGroup from '../WizardChoiceButton/WizardChoiceButtonGroup'
+import WizardChoiceButton from '../WizardChoiceButton/WizardChoiceButton'
+import WizardChoiceButtonDescription from '../WizardChoiceButton/WizardChoiceButtonDescription'
+import { Link } from '@mui/material'
+
+const TABLE_DESCRIPTION = (
+  <>
+    <WizardChoiceButtonDescription my={1.25}>
+      Synapse tables are used to organize web-accessible, sharable, and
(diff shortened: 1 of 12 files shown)


packages/synapse-react-client/src/components/EntityHeaderTable/EntityHeaderTable.tsx
```diff
@@ -305,11 +305,13 @@ export const EntityHeaderTable = (props: EntityHeaderTableProps) => {
...
       }
     },
-    [showFilterControls],
+    [table, showFilterControls],
   )
 
   if (isLoading) {
-    return <SkeletonTable numCols={3} numRows={10} />
+    return (
+      <SkeletonTable numCols={3} numRows={Math.min(10, refsInState.length)} />
```
nickgros, line 313: If we're fetching < 10 entity headers, the skeleton table should render as many rows.

packages/synapse-react-client/src/components/TableColumnSchemaEditor/ColumnModelFormFields/DefaultValueField.tsx
```diff
@@ -163,7 +163,7 @@ export default function DefaultValueField<T>(props: DefaultValueFieldProps<T>) {
     <TextField
       {...TextFieldProps}
       type={'text'}
-      value={value}
+      value={value || undefined}
```
nickgros, line 166: Fix a case where warnings were logged because `value` may have been `null`, which is not a valid input value. Now if `value` is `null`, the prop becomes `undefined`, which is valid

packages/synapse-react-client/src/components/WizardChoiceButton/WizardChoiceButtonGroup.tsx
```diff
@@ -0,0 +1,30 @@
+import React from 'react'
+import { Box, BoxProps } from '@mui/material'
+
+export default function WizardChoiceButtonGroup(props: BoxProps) {
```
nickgros, line 4: Would love to use `styled('div', ...)` instead but there's no way to set the `role` prop with the `styled` API

packages/synapse-react-client/src/components/CreateTableViewWizard/TableTypeSelection.tsx
```diff
@@ -0,0 +1,76 @@
+import React from 'react'
+import WizardChoiceButtonGroup from '../WizardChoiceButton/WizardChoiceButtonGroup'
+import WizardChoiceButton from '../WizardChoiceButton/WizardChoiceButton'
+import WizardChoiceButtonDescription from '../WizardChoiceButton/WizardChoiceButtonDescription'
+import { Link } from '@mui/material'
+
+const TABLE_DESCRIPTION = (
+  <>
+    <WizardChoiceButtonDescription my={1.25}>
```
hallieswan, line 9: As far as I can tell, `my={1.25}` is applied to each `WizardChoiceButtonDescription` instance in TableTypeSelection and ViewTypeSelection. How did you decide what to include in the default styling for `WizardChoiceButtonDescription` vs styling to apply directly to each instance of the component? Mostly just curious to understand how you're thinking about designing new components!

packages/synapse-react-client/src/components/CreateTableViewWizard/TableTypeSelection.tsx
```diff
@@ -0,0 +1,76 @@
...
+      }}
+    >
+      More about Views
+    </Link>
+  </>
+)
+
+export type TableTypeSelectionProps = {
+  /* Callback including the chosen type, and optional viewTypeMask if included in the choice (e.g. Project Views) */
+  onTableSelected: () => void
```
hallieswan, line 49: Could this comment be misplaced? It seems to match `onTypeSelected` more than this function

packages/synapse-react-client/src/components/CreateTableViewWizard/TableTypeSelection.tsx
nickgros: I think I was trying to be flexible because I wasn't sure how it was going to look +/- docs links. But since I ended up not needing that flexibility, it would make more sense to make this the default margin. Nice catch!

packages/synapse-react-client/src/components/CreateTableViewWizard/TableTypeSelection.tsx
nickgros: Yes, thanks for catching this. I originally had a different props signature and I changed it but forgot to change the comment.

PR #654, opened 2024-01-17
PORTALS-2156

jay-hodgson:


packages/synapse-react-client/src/components/DownloadCart/ShowDownloadV2.tsx
```diff
@@ -56,13 +57,12 @@ export function ShowDownloadV2({ to, className = '' }: ShowDownloadV2Props) {
...
   return (
-    <a
+    <MuiLink
       className={`Download-Link v2 ${className}`}
-      href={to}
       rel="noopener noreferrer"
     >
-      {content}
-    </a>
+      <Link to={to}>{content}</Link>
```
nickgros, line 64: Can we use MUIs `component` prop instead of nesting links ([see MUI docs](https://mui.com/material-ui/guides/routing/#link))?

PR #657, opened 2024-01-19
SWC-6627 - Update TableColumnSchemaEditor to be used in SWC

nickgros:
- TableColumnSchemaEditor is now in a dialog
 - Update `TableColumnSchemaUtils.createTableUpdateTransactionRequest` to fix bugs where columns were not properly updated or created
 - Add many tests to verify column update behavior

### DIFF
--- a/packages/synapse-react-client/src/components/TableColumnSchemaEditor/TableColumnSchemaEditor.integration.test.tsx
+++ b/packages/synapse-react-client/src/components/TableColumnSchemaEditor/TableColumnSchemaEditor.integration.test.tsx
@@ -8,13 +8,14 @@ import { server } from '../../mocks/msw/server'
 import mockTableEntityData from '../../mocks/entity/mockTableEntity'
 import {
   getAnnotationColumnHandlers,
+  getCreateColumnModelBatchHandler,
   getDefaultColumnHandlers,
+  getTableTransactionHandlers,
 } from '../../mocks/msw/handlers/tableQueryHandlers'
 import { syn17328596 as mockTableQueryData } from '../../mocks/query/syn17328596'
 import userEvent from '@testing-library/user-event'
 import { SynapseClient } from '../../index'
 import { rest } from 'msw'
-import { uniqueId } from 'lodash-es'
 import {
   ColumnModel,
   ColumnTypeEnum,
@@ -28,8 +29,13 @@ import { mockFileViewEntity } from '../../mocks/entity/mockFileView'
 import { ImportTableColumnsButtonProps } from './ImportTableColumnsButton'
 import { SetOptional } from 'type-fest'
 import { ENTITY_BUNDLE_V2 } from '../../utils/APIConstants'
-import { addColumnModelToForm } from './TableColumnSchemaEditorTestUtils'
+import {
+  addColumnModelToForm,
+  modifyColumnModelInForm,
+} from './TableColumnSchemaEditorTestUtils'
 import { MOCK_ANNOTATION_COLUMNS } from '../../mocks/mockAnnotationColumns'
+import defaultFileViewColumnModels from '../../mocks/query/defaultFileViewColumnModels'
+import { MOCK_ACCESS_TOKEN } from '../../mocks/MockSynapseContext'
 
 const mockedImportedColumns: SetOptional<ColumnModel, 'id'>[] = [
   {
@@ -77,10 +83,18 @@ async function setUp(props: TableColumnSchemaEditorProps) {
   const user = userEvent.setup()
   const component = renderComponent(props)
   const saveButton = await screen.findByRole('button', { name: 'Save' })
-  return { user, component, saveButton }
+  const cancelButton = await screen.findByRole('button', { name: 'Cancel' })
+  const moveDownButton = await screen.findByRole('button', {
+    name: 'Move Down',
+  })
+  return { user, component, saveButton, cancelButton, moveDownButton }
 }
 
 describe('TableColumnSchemaEditor', () => {
+  const mockOnColumnsUpdated = jest.fn()
+  const mockOnCancel = jest.fn()
+  const updateTableSpy = jest.spyOn(SynapseClient, 'updateTable')
+
   beforeAll(() => {
     server.listen()
   })
@@ -91,23 +105,20 @@ describe('TableColumnSchemaEditor', () => {
       /* Each test in this suite should register its own table query handler */
       ...getDefaultColumnHandlers(),
       ...getAnnotationColumnHandlers(MOCK_ANNOTATION_COLUMNS),
-      rest.post(
-        `${getEndpoint(
-          BackendDestinationEnum.REPO_ENDPOINT,
-        )}/repo/v1/column/batch`,
-        async (req, res, ctx) => {
-          const response: { list: ColumnModel[] } = await req.json()
-          response.list.forEach(cm => {
-            if (cm.id == null) {
-              cm.id = uniqueId()
-            }
-          })
-          return res(
-            ctx.status(201),
-            ctx.json<{ list: ColumnModel[] }>(response),
-          )
-        },
-      ),
+      getCreateColumnModelBatchHandler(),
+      ...getTableTransactionHandlers({
+        concreteType:
+          'org.sagebionetworks.repo.model.table.TableUpdateTransactionResponse',
+        results: [
+          {
+            concreteType:
+              'org.sagebionetworks.repo.model.table.TableSchemaChangeResponse',
+            schema: [
+              // We don't currently use this schema, so it's ok if it's inaccurate
+            ],
+          },
+        ],
+      }),
     )
   })
   afterAll(() => server.close())
@@ -124,6 +135,9 @@ describe('TableColumnSchemaEditor', () => {
     )
     await setUp({
       entityId: mockTableEntityData.id,
+      open: true,
+      onColumnsUpdated: mockOnColumnsUpdated,
+      onCancel: mockOnCancel,
     })
 
     const numColumns = mockTableQueryData.columnModels!.length
@@ -135,8 +149,8 @@ describe('TableColumnSchemaEditor', () => {
       expect(nameFields[index]).toHaveValue(cm.name)
     })
   })
+
   it('User can update the form and submit the new data', async () => {
-    const updateTableSpy = jest.spyOn(SynapseClient, 'updateTable')
     server.use(
       getEntityBundleHandler({
         entity: mockTableEntityData.entity,
@@ -149,6 +163,9 @@ describe('TableColumnSchemaEditor', () => {
 
     const { user, saveButton } = await setUp({
       entityId: mockTableEntityData.id,
+      open: true,
+      onColumnsUpdated: mockOnColumnsUpdated,
+      onCancel: mockOnCancel,
     })
 
     // Add a column
@@ -156,7 +173,32 @@ describe('TableColumnSchemaEditor', () => {
     await user.click(saveButton)
 
     await waitFor(() => {
-      expect(updateTableSpy).toHaveBeenCalled()
+      const expectedExistingColumnIds = mockTableQueryData.columnModels!.map(
+        cm => cm.id,
+      )
+      expect(updateTableSpy).toHaveBeenCalledWith(
+        {
+          changes: [
+            {
+              // The new column was added, no columns were changed, moved, or removed
+              changes: [{ newColumnId: expect.any(String), oldColumnId: null }],
+              concreteType:
+                'org.sagebionetworks.repo.model.table.TableSchemaChangeRequest',
+              entityId: mockTableEntityData.id,
+              // The new column is at the end
+              orderedColumnIds: [
+                ...expectedExistingColumnIds,
+                expect.any(String),
+              ],
+            },
+          ],
+          concreteType:
+            'org.sagebionetworks.repo.model.table.TableUpdateTransactionRequest',
+          entityId: mockTableEntityData.id,
+        },
+        MOCK_ACCESS_TOKEN,
+      )
+      expect(mockOnColumnsUpdated).toHaveBeenCalled()
     })
   })
 
@@ -172,8 +214,11 @@ describe('TableColumnSchemaEditor', () => {
         },
       }),
     )
-    await setUp({
+    const { user, saveButton } = await setUp({
       entityId: mockFileViewEntity.id,
+      open: true,
+      onColumnsUpdated: mockOnColumnsUpdated,
+      onCancel: mockOnCancel,
     })
 
     const addDefaultColumnsButton = await screen.findByRole('button', {
@@ -185,7 +230,7 @@ describe('TableColumnSchemaEditor', () => {
     // Verify we have no columns -- there will only be one checkbox on the screen for select all/none
     expect(screen.getAllByRole('checkbox')).toHaveLength(1)
 
-    await userEvent.click(addDefaultColumnsButton)
+    await user.click(addDefaultColumnsButton)
 
     await waitFor(() => {
       // The columns should have been added.
@@ -194,6 +239,35 @@ describe('TableColumnSchemaEditor', () => {
 
     // The default columns aren't editable, so we won't have a name text field.
     expect(screen.queryByLabelText('Name')).not.toBeInTheDocument()
+
+    await user.click(saveButton)
+
+    await waitFor(() => {
+      expect(updateTableSpy).toHaveBeenCalledWith(
+        {
+          changes: [
+            {
+              // The default columns were added; they are all new
+              changes: expect.arrayContaining(
+                defaultFileViewColumnModels.map(cm => ({
+                  newColumnId: cm.id,
+                  oldColumnId: null,
+                })),
+              ),
+              concreteType:
+                'org.sagebionetworks.repo.model.table.TableSchemaChangeRequest',
+              entityId: mockFileViewEntity.id,
+              orderedColumnIds: defaultFileViewColumnModels.map(cm => cm.id),
+            },
+          ],
+          concreteType:
+            'org.sagebionetworks.repo.model.table.TableUpdateTransactionRequest',
+          entityId: mockFileViewEntity.id,
+        },
+        MOCK_ACCESS_TOKEN,
+      )
+      expect(mockOnColumnsUpdated).toHaveBeenCalled()
+    })
   })
   it('User can add annotations columns to a view', async () => {
     // Must use a file view to get annotations
@@ -207,8 +281,11 @@ describe('TableColumnSchemaEditor', () => {
         },
       }),
     )
-    await setUp({
+    const { user, saveButton } = await setUp({
       entityId: mockFileViewEntity.id,
+      open: true,
+      onColumnsUpdated: mockOnColumnsUpdated,
+      onCancel: mockOnCancel,
     })
 
     const addAnnotationColumnsButton = await screen.findByRole('button', {
@@ -220,7 +297,7 @@ describe('TableColumnSchemaEditor', () => {
     // Verify we have no columns -- there will only be one checkbox on the screen for select all/none
     expect(screen.getAllByRole('checkbox')).toHaveLength(1)
 
-    await userEvent.click(addAnnotationColumnsButton)
+    await user.click(addAnnotationColumnsButton)
 
     await waitFor(() => {
       // The column(s) should have been added.
@@ -229,6 +306,37 @@ describe('TableColumnSchemaEditor', () => {
 
     // The annotation column(s) are editable, so we can find a name text field.
     await screen.findAllByLabelText('Name')
+
+    await user.click(saveButton)
+
+    await waitFor(() => {
+      expect(updateTableSpy).toHaveBeenCalledWith(
+        {
+          changes: [
+            {
+              // The annotation columns were added; they are all new
+              changes: expect.arrayContaining(
+                MOCK_ANNOTATION_COLUMNS.results.map(cm => ({
+                  newColumnId: cm.id,
+                  oldColumnId: null,
+                })),
+              ),
+              concreteType:
+                'org.sagebionetworks.repo.model.table.TableSchemaChangeRequest',
+              entityId: mockFileViewEntity.id,
+              orderedColumnIds: MOCK_ANNOTATION_COLUMNS.results.map(
+                cm => cm.id,
+              ),
+            },
+          ],
+          concreteType:
+            'org.sagebionetworks.repo.model.table.TableUpdateTransactionRequest',
+          entityId: mockFileViewEntity.id,
+        },
+        MOCK_ACCESS_TOKEN,
+      )
+      expect(mockOnColumnsUpdated).toHaveBeenCalled()
+    })
   })
 
   it('Existing default columns are not editable, except facetType', async () => {
@@ -251,6 +359,9 @@ describe('TableColumnSchemaEditor', () => {
     )
     const { user } = await setUp({
       entityId: mockFileViewEntity.id,
+      open: true,
+      onColumnsUpdated: mockOnColumnsUpdated,
+      onCancel: mockOnCancel,
     })
 
     // Verify we have one column, which has a checkbox, in addition to the "Select All" checkbox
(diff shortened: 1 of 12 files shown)


packages/synapse-react-client/src/components/TableColumnSchemaEditor/TableColumnSchemaEditor.tsx
```diff
@@ -6,23 +6,30 @@ import {
...
 import { SetOptional } from 'type-fest'
+import { SynapseSpinner } from '../LoadingScreen/LoadingScreen'
+import { DialogBase } from '../DialogBase'
+import { noop } from 'lodash-es'
 
 export type TableColumnSchemaEditorProps = {
   entityId: string
+  open: boolean
+  onColumnsUpdated?: () => void
+  onCancel?: () => void
```
nickgros, line 21: Add `open` prop because this is now a modal, add callback props

packages/synapse-react-client/src/utils/functions/TableColumnSchemaUtils.ts
```diff
@@ -4,40 +4,46 @@ import {
...
     if (copy.id != null) {
-      // any changes to the existing column model?
+      // If there are any changes to the existing column model, remove the ID before we create the new one
       const oldColumnModel: ColumnModel | undefined =
         oldColumnModelId2Model.get(copy.id)
-      if (oldColumnModel != undefined && !isEqual(oldColumnModel, copy)) {
+      if (
+        oldColumnModel != undefined &&
+        !isEqualTreatUndefinedAsMissing(oldColumnModel, copy)
+      ) {
```
nickgros, line 46: This was one bug. These columns are equivalent, but our code did not treat them as equal until changing the `isEqual` method

```js
{
  "id": "123",
  "name": "foo",
  "columnType": "STRING"
}
```

```js
{
  "id": "123",
  "name": "foo",
  "columnType": "STRING",
  "maximumLength": undefined
}
```

packages/synapse-react-client/src/utils/functions/TableColumnSchemaUtils.ts
```diff
@@ -47,23 +53,33 @@ export async function createTableUpdateTransactionRequest(
...
+      // The column has changed
+      oldColumnId != null &&
+      oldColumnId !== newColumnId
+    ) {
       changes.push({ oldColumnId, newColumnId })
+    } else if (
+      // the column is new and is defined by the user
+      oldColumnId == null ||
+      // the column is new but already has an ID (e.g. we are adding a default column to a view)
+      !oldColumnModelId2Model.has(oldColumnId)
```
nickgros, line 73: We were missing this case before. The editor would fail to add a change for default view columns. The check `oldColumnId !== newColumnId` assumed that all of the `oldColumnId` values would be in the `oldSchema`, but that was not true for default view columns.

packages/synapse-react-client/src/components/TableColumnSchemaEditor/TableColumnSchemaEditor.integration.test.tsx
```diff
@@ -77,39 +84,48 @@ async function setUp(props: TableColumnSchemaEditorProps) {
...
-              cm.id = uniqueId()
-            }
-          })
-          return res(
-            ctx.status(201),
-            ctx.json<{ list: ColumnModel[] }>(response),
-          )
-        },
-      ),
+      getCreateColumnModelBatchHandler(),
```
nickgros, line 112: This is the same handler, just pulled from the shared mocks

packages/synapse-react-client/src/components/TableColumnSchemaEditor/TableColumnSchemaEditor.integration.test.tsx
```diff
@@ -77,39 +84,48 @@ async function setUp(props: TableColumnSchemaEditorProps) {
...
+        results: [
+          {
+            concreteType:
+              'org.sagebionetworks.repo.model.table.TableSchemaChangeResponse',
+            schema: [
+              // We don't currently use this schema, so it's ok if it's inaccurate
+            ],
+          },
+        ],
+      }),
```
nickgros, line 125: Add a handler for the final update service.

packages/synapse-react-client/src/components/TableColumnSchemaEditor/TableColumnSchemaEditor.integration.test.tsx
```diff
@@ -124,6 +140,9 @@ describe('TableColumnSchemaEditor', () => {
     )
     await setUp({
       entityId: mockTableEntityData.id,
+      open: true,
+      onColumnsUpdated: mockOnColumnsUpdated,
+      onCancel: mockOnCancel,
```
nickgros, line 145: In each test, add the new props

packages/synapse-react-client/src/components/TableColumnSchemaEditor/TableColumnSchemaEditor.integration.test.tsx
```diff
@@ -149,14 +168,35 @@ describe('TableColumnSchemaEditor', () => {
...
+            maximumSize: 50,
+          },
+        ],
+      )
+
+      expect(updateTableSpy).toHaveBeenCalledWith(
+        await createTableUpdateTransactionRequestSpy.mock.results[0].value,
+        MOCK_ACCESS_TOKEN,
+      )
+      expect(mockOnColumnsUpdated).toHaveBeenCalled()
```
nickgros, line 199: For each test with a successful update call, verify 3 things
1. We passed the expected data to the update fn (skipping the react-query layer, we just verify a call to `createTableUpdateTransactionRequest`)
2. Verify that `SynapseClient.updateTable` is called with the result of `createTableUpdateTransactionRequest`
3. Verify that our `onColumnUpdated` callback is invoked

#1 is very important that we get right, else we can have bugs that can result in unexpected data loss. For example, we need to capture the nuance between 
* Updating column 123 to now have the schema defined by column 456. Data should be preserved.
* Deleting column 123 and adding column 456 does not preserve data

This PR fixes bugs where the user tries to do an update but the implementation erroneously did a delete + create

packages/synapse-react-client/src/components/TableColumnSchemaEditor/TableColumnSchemaEditor.integration.test.tsx
```diff
@@ -194,6 +237,26 @@ describe('TableColumnSchemaEditor', () => {
 
     // The default columns aren't editable, so we won't have a name text field.
     expect(screen.queryByLabelText('Name')).not.toBeInTheDocument()
+
+    await user.click(saveButton)
```
nickgros, line 241: For many of these existing tests, follow through with clicking "Save" and verify the calls are correct

packages/synapse-react-client/src/components/TableColumnSchemaEditor/TableColumnSchemaEditor.integration.test.tsx
```diff
@@ -480,8 +634,263 @@ describe('TableColumnSchemaEditor', () => {
...
+        ],
+      )
+      expect(updateTableSpy).toHaveBeenCalledWith(
+        await createTableUpdateTransactionRequestSpy.mock.results[0].value,
+        MOCK_ACCESS_TOKEN,
+      )
+      expect(mockOnColumnsUpdated).toHaveBeenCalled()
+    })
+  })
+  it('Maintains the column ID passed to the update function when a column is modified', async () => {
```
nickgros, line 656: Add a test scenario for updating an existing column

packages/synapse-react-client/src/components/TableColumnSchemaEditor/TableColumnSchemaEditor.integration.test.tsx
```diff
@@ -480,8 +634,263 @@ describe('TableColumnSchemaEditor', () => {
...
+      )
+      expect(updateTableSpy).toHaveBeenCalledWith(
+        await createTableUpdateTransactionRequestSpy.mock.results[0].value,
+        MOCK_ACCESS_TOKEN,
+      )
+      expect(mockOnColumnsUpdated).toHaveBeenCalled()
+    })
+  })
+
+  it('Can reorder and edit the same column', async () => {
```
nickgros, line 758: Add a test scenario for reordering an edited column

packages/synapse-react-client/src/components/TableColumnSchemaEditor/TableColumnSchemaEditor.integration.test.tsx
```diff
@@ -480,8 +634,263 @@ describe('TableColumnSchemaEditor', () => {
...
+      )
+      expect(updateTableSpy).toHaveBeenCalledWith(
+        await createTableUpdateTransactionRequestSpy.mock.results[0].value,
+        MOCK_ACCESS_TOKEN,
+      )
+      expect(mockOnColumnsUpdated).toHaveBeenCalled()
+    })
+  })
+
+  it('Handles failure on the update call', async () => {
```
nickgros, line 823: Add a test scenario for a failed update

packages/synapse-react-client/src/components/TableColumnSchemaEditor/TableColumnSchemaEditor.integration.test.tsx
```diff
@@ -480,8 +634,263 @@ describe('TableColumnSchemaEditor', () => {
...
+    })
+
+    const errorAlert = await screen.findByRole('alert')
+    within(errorAlert).getByText(errorMessage)
+
+    // `onColumnsUpdated` should NOT have been called because the update failed
+    expect(mockOnColumnsUpdated).not.toHaveBeenCalled()
+  })
+
+  it('Clicking cancel calls onCancel', async () => {
```
nickgros, line 871: Test the new `onCancel` prop

packages/synapse-react-client/src/components/TableColumnSchemaEditor/TableColumnSchemaForm.tsx
```diff
@@ -238,6 +238,12 @@ function TableColumnSchemaFormInternal(
           return !currentFormData.find(fd => fd.name === cm.name)
         },
       )
+      // Delete the ID column so TableColumnSchemaUtils.createTableUpdateTransactionRequest recognizes these as new columns
+      // createTableUpdateTransactionRequest uses the existing column ID to track column updates in the table, so any new columns should have no ID
+      // The user can also modify these columns before submitting them, so the ID may not be accurate when we end up submitting them
+      columnsToAdd.forEach(cm => {
+        delete cm.id
+      })
```
nickgros, line 246: This is inside of the function called when _existing_ columns are added to a table. Examples include 
* Columns imported from another table
* Columns for a view that correspond to annotations on objects in the view

The update utility expects that columns that are _new to the table_ do not include IDs, so we remove them here. This is how the Java implementation works. 

To me, this feels a bit dangerous. I personally didn't quite grasp how this worked until I tried to fix a few bugs here. I think it might feel less 'dangerous' if we were to track our own set of changes instead of relying on presence/absence of IDs, but that is a pretty large change since it would also affect the editor implementation.

packages/synapse-react-client/src/mocks/msw/handlers/asyncJobHandlers.ts
```diff
@@ -75,10 +81,13 @@ export function generateAsyncJobHandlers<
         const responseObject: TResponseBody =
           typeof response === 'function' ? response(request) : response
 
+        const jobState: AsynchJobState =
+          serviceSpecificEndpointResponseStatus < 400 ? 'COMPLETE' : 'FAILED'
+
         return res(
-          ctx.status(201),
+          ctx.status(200),
```
nickgros, line 88: This looks like a bug but it's not, should probably add a comment
```suggestion
          // This endpoint returns a successful status code regardless of the job status
          ctx.status(200),
```

packages/synapse-react-client/src/mocks/msw/handlers/asyncJobHandlers.ts
```diff
@@ -119,7 +128,10 @@ export function generateAsyncJobHandlers<
         const responseObject: TResponseBody =
           typeof response === 'function' ? response(request) : response
 
-        return res(ctx.status(201), ctx.json(responseObject))
+        return res(
+          ctx.status(serviceSpecificEndpointResponseStatus),
```
nickgros, line 132: We call this endpoint to get the status code and figure out the type of error.

Our SynapseClient code calls this endpoint because if the job failed, the other service doesn't give us the information we need to determine if we should retry or not. This endpoint gives us an HTTP status code, which we can use to decide to retry.

packages/synapse-react-client/src/utils/functions/TableColumnSchemaUtils.test.ts
```diff
@@ -54,6 +54,56 @@ describe('TableColumnSchemaUtils', () => {
       expect(tableUpdateRequest.orderedColumnIds[1]).toBe(OLD_COLUMN_MODEL_ID1)
       expect(tableUpdateRequest.changes).toHaveLength(0)
     })
+
+    test('reorder with a change', async () => {
```
nickgros, line 58: Add a test case for reordering + changing the same column

packages/synapse-react-client/src/utils/functions/TableColumnSchemaUtils.test.ts
```diff
@@ -93,6 +143,42 @@ describe('TableColumnSchemaUtils', () => {
         newColumnId: NEW_COLUMN_MODEL_ID,
       })
     })
+    test('add new column that has an existing ID', async () => {
```
nickgros, line 146: Remove this test case or decide if it's possible to throw an error, because this is an invalid scenario

packages/synapse-react-client/src/utils/functions/TableColumnSchemaUtils.test.ts
```diff
@@ -93,6 +144,38 @@ describe('TableColumnSchemaUtils', () => {
         newColumnId: NEW_COLUMN_MODEL_ID,
       })
     })
+
+    test('adding a new column that has an existing ID causes a thrown error', async () => {
```
nickgros, line 148: Add a test case to handle one of the current bugs -- no new columns should have IDs

packages/synapse-react-client/src/components/CreateTableViewWizard/CreateTableViewWizard.integration.test.tsx
```diff
@@ -49,6 +50,24 @@ const createEntitySpy = jest.mocked(SynapseClient.createEntity)
...
+
+async function getLatestCreatedColumnModelIdsFromSpy(
+  spy: typeof createColumnModelsSpy,
+) {
+  return (
+    await (spy.mock.results[0].value as ReturnType<
+      typeof SynapseClient.createColumnModels
+    >)
+  ).list.map(cm => cm.id)
+}
```
nickgros, line 69: Get the IDs of the new column models returned by the most recent call to `createColumnModels`. We'll verify that these IDs get passed to another function.

packages/synapse-react-client/src/components/CreateTableViewWizard/CreateTableViewWizard.tsx
```diff
@@ -137,7 +138,12 @@ export default function CreateTableViewWizard(
       entityType !== EntityType.VIRTUAL_TABLE
     ) {
       try {
-        createdColumnModels = await createColumnModels(columnModels)
+        createdColumnModels = await createColumnModels(
+          columnModels.map(cm => omitBy(cm, isUndefined)) as SetOptional<
+            ColumnModel,
+            'id'
+          >[],
+        )
```
nickgros, line 146: Remove undefined fields from the object before we pass it to the mutation

packages/synapse-react-client/src/utils/functions/TableColumnSchemaUtils.ts
```diff
@@ -4,40 +4,67 @@ import {
...
+  for (const columnModel of oldSchema) {
+    oldColumnModelIds.add(columnModel.id)
+  }
+  for (const columnModel of proposedSchema) {
+    if (columnModel.id != null && !oldColumnModelIds.has(columnModel.id)) {
+      throw new Error(
+        `Proposed schema contains a new column model with ID ${columnModel.id} that is not in the old schema.`,
+      )
+    }
+  }
```
nickgros, line 26: At the layers above this, we remove IDs from new columns, but add this validation check to prevent regression in the future

packages/synapse-react-client/src/utils/functions/TableColumnSchemaUtils.ts
```diff
@@ -4,40 +4,67 @@ import {
...
 /**
- * Creates the column models for a change request and returns
- * the change request to be passed to the Synapse backend.
- * @param accessToken
- * @param tableId
- * @param oldSchema
- * @param proposedSchema
+ * Creates the column models for a change request and returns the change request to be passed to the Synapse backend.
+ *
+ * Note that to update a column model and preserve table data, the ID of the
```
hallieswan, line 32: Was this comment cutoff early?

packages/synapse-react-client/src/components/TableColumnSchemaEditor/TableColumnSchemaEditor.tsx
```diff
@@ -42,15 +49,26 @@ function _TableColumnSchemaEditor(props: TableColumnSchemaEditorProps) {
   const entity = bundle?.entity
   const originalColumnModels = bundle?.tableBundle?.columnModels
 
-  const { mutate, isLoading: isMutating, error } = useUpdateTableColumns()
+  const {
+    mutate,
+    isLoading: isMutating,
+    error,
+  } = useUpdateTableColumns({
```
nickgros, line 56: Consider trying to consolidate omitting undefined properties in some shared method called by both this component and the CreateTableViewWizard

packages/synapse-react-client/src/utils/functions/TableColumnSchemaUtils.ts
nickgros: Yes

nickgros, line 32:
```suggestion
 * Note that to update a column model and preserve table data, the ID of the column in the oldSchema should be the same as the column ID in the proposedSchema
```

packages/synapse-react-client/src/utils/functions/TableColumnSchemaUtils.ts
```diff
@@ -4,40 +4,67 @@ import {
...
+ * @param accessToken the Synapse access token
+ * @param tableId the ID of the table to change
+ * @param oldSchema the current schema of the table
+ * @param proposedSchema the proposed schema of the table. For TableEntities, the IDs should be preserved for columns
+ *                       that should maintain their data, even if the column model has changed
  */
 export async function createTableUpdateTransactionRequest(
```
nickgros, line 39: @nickgros to create design ticket to address that users cannot discern between a modified column and a new column in the UI

packages/synapse-react-client/src/utils/functions/TableColumnSchemaUtils.ts
```diff
@@ -47,23 +74,34 @@ export async function createTableUpdateTransactionRequest(
...
+      // The column has changed
+      columnIdInProposedSchema != null &&
+      columnIdInProposedSchema !== newColumnId
+    ) {
+      changes.push({ oldColumnId: columnIdInProposedSchema, newColumnId })
+    } else if (
+      // the column is new and is defined by the user
+      columnIdInProposedSchema == null ||
+      // the column is new but already has an ID (e.g. we are adding a default column to a view)
+      !oldColumnModelId2Model.has(columnIdInProposedSchema)
```
nickgros, line 95: This should never happen now; IDs should be removed from _any_ new column before this method is called, including default columns

packages/synapse-react-client/src/utils/functions/TableColumnSchemaUtils.ts
```diff
@@ -47,23 +74,34 @@ export async function createTableUpdateTransactionRequest(
...
+      columnIdsThatWereChangedOrAdded.add(columnIdInProposedSchema)
+    columnIdsThatWereChangedOrAdded.add(newColumnId)
+    if (
+      // The column has changed
+      columnIdInProposedSchema != null &&
+      columnIdInProposedSchema !== newColumnId
+    ) {
+      changes.push({ oldColumnId: columnIdInProposedSchema, newColumnId })
+    } else if (
+      // the column is new and is defined by the user
```
nickgros, line 92:
```suggestion
      // the column is new to the table
```

packages/synapse-react-client/src/utils/functions/TableColumnSchemaUtils.ts
```diff
@@ -47,23 +74,34 @@ export async function createTableUpdateTransactionRequest(
     .list
 
   // now that all columns have been created, figure out the column changes (create, update, and no-op)
-  // keep track of column ids to figure out what columns were deleted.
   const changes: ColumnChange[] = []
-  const columnIds: Set<string | null> = new Set()
+  const columnIdsThatWereChangedOrAdded: Set<string> = new Set()
```
nickgros, line 78: Rename to `idsThatShouldNotBeDeletedFromSchema`,

Consider adding comments about _why_ we add each value to this set when we do so.

packages/synapse-react-client/src/utils/functions/TableColumnSchemaUtils.ts
```diff
@@ -47,23 +74,34 @@ export async function createTableUpdateTransactionRequest(
...
-    const oldColumnId: string | null = proposedSchema[i].id ?? null
-    const newColumnId: string | null = createdColumnModels[i].id ?? null
-    columnIds.add(oldColumnId)
-    columnIds.add(newColumnId)
-    if (oldColumnId !== newColumnId) {
-      changes.push({ oldColumnId, newColumnId })
+    const columnIdInProposedSchema: string | null = proposedSchema[i].id ?? null
+    const newColumnId: string = createdColumnModels[i].id
+    if (columnIdInProposedSchema)
+      columnIdsThatWereChangedOrAdded.add(columnIdInProposedSchema)
```
nickgros, line 83:
```suggestion
      // We add the columnIdInProposedSchema because the column was modified. It should not be deleted.
      columnIdsThatWereChangedOrAdded.add(columnIdInProposedSchema)
```

packages/synapse-react-client/src/components/TableColumnSchemaEditor/TableColumnSchemaEditor.tsx
```diff
@@ -75,48 +93,57 @@ function _TableColumnSchemaEditor(props: TableColumnSchemaEditorProps) {
...
-export default function TableColumnSchemaEditor(
-  props: TableColumnSchemaEditorProps,
-) {
-  // Wrap in a Jotai provider to ensure the Jotai atomic state is unique to this component tree
-  // i.e. other instances of TableColumnSchemaEditor will not share state with this instance
-  return (
-    <Provider>
-      <_TableColumnSchemaEditor {...props} />
-    </Provider>
+    <DialogBase
```
nickgros, line 96: See if `ConfirmationDialog` works for this case

hallieswan (approved):
Looks good to me, thanks for adding all of the new tests!

packages/synapse-react-client/src/utils/functions/TableColumnSchemaUtils.ts
nickgros: SWC-6666

packages/synapse-react-client/src/components/TableColumnSchemaEditor/TableColumnSchemaEditor.tsx
nickgros: Could do this in `SynapseClient.createColumnModels` but hesitant to put business logic in SynapseClient

hallieswan (approved):
The ConfirmationDialog refactor is a great improvement!

PR #658, opened 2024-01-19
SWC-6434: add userinfo format choice to OAuth client editor

hallieswan:
https://github.com/Sage-Bionetworks/synapse-web-monorepo/assets/26949006/ce954a21-c3e9-4b99-a447-a49478e3fc5a

### DIFF
--- a/packages/synapse-react-client/src/components/OAuthClientManagement/CreateOAuthClient.integration.test.tsx
+++ b/packages/synapse-react-client/src/components/OAuthClientManagement/CreateOAuthClient.integration.test.tsx
@@ -1,9 +1,17 @@
-import { render, screen, waitFor } from '@testing-library/react'
+import { OIDCSigningAlgorithm } from '@sage-bionetworks/synapse-types'
+import { act, render, screen, waitFor } from '@testing-library/react'
 import userEvent from '@testing-library/user-event'
 import React from 'react'
 import { createWrapper } from '../../testutils/TestingLibraryUtils'
+import { MOCK_ACCESS_TOKEN } from '../../mocks/MockSynapseContext'
 import { rest, server } from '../../mocks/msw/server'
-import { CreateOAuthModal, CreateOAuthModalProps } from './CreateOAuthClient'
+import {
+  CreateOAuthModal,
+  CreateOAuthModalProps,
+  defaultUserInfoSignedResponseAlgorithm,
+  warningBody,
+  warningHeader,
+} from './CreateOAuthClient'
 import { mockClientList1 } from '../../mocks/oauth/MockClient'
 import * as ToastMessage from '../ToastMessage/ToastMessage'
 import {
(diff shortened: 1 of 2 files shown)


packages/synapse-react-client/src/components/OAuthClientManagement/CreateOAuthClient.integration.test.tsx
```diff
@@ -35,9 +43,10 @@ const defaultProps: CreateOAuthModalProps = {
   setIsShowingConfirmModal: jest.fn(),
 }
 
-jest
-  .spyOn(SynapseClient, 'isOAuthClientReverificationRequired')
-  .mockReturnValue(Promise.resolve({ reverificationRequired: true }))
```
hallieswan, line 40: I think tests were flaky because the call to `isOAuthClientReverificationRequired` is wrapped in `useDebouncedEffect`, so depending on the test timing, the service might be called and cause the warning modal to appear when reverification was not expected (Creates a client on save, Updates a client) or not be called when the warning modal was expected to appear (show warning modal when changing redirect uri).

packages/synapse-react-client/src/components/OAuthClientManagement/CreateOAuthClient.integration.test.tsx
```diff
@@ -47,18 +56,68 @@ function renderComponent(props: CreateOAuthModalProps = defaultProps) {
 
 function setUp(props: CreateOAuthModalProps = defaultProps) {
   const user = userEvent.setup()
+
+  // useFakeTimers to ensure that isOAuthClientReverificationRequired is called
+  // at least once per test, since the call is wrapped in useDebouncedEffect
+  jest.useFakeTimers()
```
hallieswan, line 62: Use fake timers, so that `isOAuthClientReverificationRequired` is called at least once per test, then switch back to using real timers for the rest of the test. This seems to resolve the flakiness in local test runs.

packages/synapse-react-client/src/components/OAuthClientManagement/CreateOAuthClient.integration.test.tsx
```diff
@@ -47,18 +56,68 @@ function renderComponent(props: CreateOAuthModalProps = defaultProps) {
...
+      ),
+      rest.put(
+        `${getEndpoint(
+          BackendDestinationEnum.REPO_ENDPOINT,
+        )}/auth/v1/oauth2/client/${mockClient.client_id!}`,
+        async (req, res, ctx) => {
+          return res(ctx.status(200), ctx.json(req.body))
+        },
+      ),
+      mockVerificationPrecheckService(false),
```
hallieswan, line 118: Most tests expect that `isOAuthClientReverificationRequired` will return `false`, so set that as the default value.

packages/synapse-react-client/src/components/OAuthClientManagement/CreateOAuthClient.tsx
```diff
@@ -25,6 +29,10 @@ import SynapseClient from '../../synapse-client'
 import { useDebouncedEffect } from '../../utils/hooks'
 import { ConfirmationDialog } from '../ConfirmationDialog/ConfirmationDialog'
 
+export const defaultUserInfoSignedResponseAlgorithm = 'JSON'
+export const warningHeader = 'Are you absolutely sure?'
+export const warningBody =
+  'Editing this detail will render your client invalid and will require you to resubmit verification. This action cannot be undone.'
```
hallieswan, line 35: Pull these into constants that can be used in tests

packages/synapse-react-client/src/components/OAuthClientManagement/CreateOAuthClient.tsx
```diff
@@ -98,13 +110,17 @@ export const CreateOAuthModal: React.FunctionComponent<
...
+      client?.userinfo_signed_response_alg ??
+        defaultUserInfoSignedResponseAlgorithm,
+    )
     setTosUri(client?.tos_uri ?? '')
   }, [isShowingModal, client])
 
   useDebouncedEffect(
     () => {
-      if (accessToken) {
+      if (accessToken && oAuthClient.client_id) {
```
hallieswan, line 123: Prevent a 400 when calling `isOAuthClientReverificationRequired` when `client_id` is undefined

nickgros (approved):
Great improvement to tests

PR #661, opened 2024-01-22
PORTALS-2902: move Portal preferences to local storage instead of a cookie

hallieswan:


apps/SageAccountWeb/src/components/CookiesNotification.tsx
hallieswan, line 1: Move this component into SRC, so it can be reused here and in Portals

apps/portals/src/portal-components/elportal/ELBetaLaunchBanner.tsx
hallieswan, line 1: https://github.com/Sage-Bionetworks/synapse-web-monorepo/assets/26949006/d1441174-629c-4d5d-95c7-332be8dfcabd

packages/synapse-react-client/src/components/CookiesNotification/CookiesNotification.stories.tsx
hallieswan, line 1: PORTALS-2902_CookiesNotification

jay-hodgson (approved):

PR #664, opened 2024-01-22
PORTALS-2898: bump playwright version to name blob report in playwright config

hallieswan:


apps/portals/e2e/explore.spec.ts
```diff
@@ -341,7 +342,7 @@ test.describe('Explore', () => {
     const exploreTabTypes = exploreTabType.split('_')
 
     for (const exploreTab of exploreTabs) {
-      test(exploreTab, async ({ page }) => {
+      test(tagTestTitle(exploreTab), async ({ page }) => {
```
nickgros, line 345: Is there a reason we use the title tag instead of using [Custom Annotations](https://playwright.dev/docs/test-annotations#custom-annotations)?

If I had to guess, is it that the title tag allows us to run e.g. `@adknowledgeportal` tests from the command line, but custom annotations don't let us do that?

.github/workflows/end-to-end-test-portals.yml
```diff
@@ -48,10 +48,8 @@ jobs:
         run: |
           cd "${{ env.PORTALS_SUBDIR }}"
           pnpm playwright install --with-deps
-      # Name report in playwright.config once feature is released:
-      # https://github.com/microsoft/playwright/issues/27284
       - name: Run e2e tests
-        run: PWTEST_BLOB_REPORT_NAME="${{ matrix.PORTAL }}" pnpm nx run portals:e2e
```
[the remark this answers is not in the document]
nickgros: Thanks for filing the issue in the playwright repo

nickgros (approved):

apps/portals/e2e/explore.spec.ts
hallieswan: The title tag would allow us to run particular tests from the command line, if we add tests that only run against a particular portal in the future! Currently, we run all tests against all portals. 

My primary reason was how easy it would be to distinguish test runs by Portal in the HTML report. 

Custom annotations are shown on the test page, but not on the summary page - so you'd have to click on each test run to find the results for a particular Portal:

https://github.com/Sage-Bionetworks/synapse-web-monorepo/assets/26949006/763f03f1-32bb-4197-9c00-6445da227721

Tags are shown on the summary page and can be used to filter down to test runs for a particular Portal, so it's quick to find the test run for a particular Portal:

https://github.com/Sage-Bionetworks/synapse-web-monorepo/assets/26949006/8dae7b47-121e-4e6f-b3b3-d2cbe17e58c5

nickgros: Gotcha, tags != annotations. Thanks!

PR #668, opened 2024-01-24
PORTALS-2880: increase size of data icons in generic card

hallieswan:
staging:
PORTALS-2880_staging

feature:
PORTALS-2880_feature

### DIFF
--- a/packages/synapse-react-client/src/assets/mui_components/Other.tsx
+++ b/packages/synapse-react-client/src/assets/mui_components/Other.tsx
@@ -2,9 +2,9 @@ import React from 'react'
 import SvgIcon, { SvgIconProps } from '@mui/material/SvgIcon'
 
 const Other = (props: SvgIconProps) => {
-  const { fill, style } = props
+  const { fill, style, ...rest } = props
   return (
-    <SvgIcon style={style}>
+    <SvgIcon style={style} {...rest}>
       <circle cx="4.71456" cy="12.7146" r="1.71456" fill={fill} />
       <circle cx="11.5728" cy="12.7146" r="1.71456" fill={fill} />
       <circle cx="18.431" cy="12.7146" r="1.71456" fill={fill} />
--- a/packages/synapse-react-client/src/components/GenericCard/GenericCard.tsx
+++ b/packages/synapse-react-client/src/components/GenericCard/GenericCard.tsx
@@ -27,7 +27,7 @@ import {
   TargetEnum,
 } from '../CardContainerLogic'
 import HeaderCard from '../HeaderCard'
-import IconList, { IconListProps } from '../IconList'
+import IconList from '../IconList'
 import IconSvg, { type2SvgIconName } from '../IconSvg/IconSvg'
 import { CardFooter, Icon } from '../row_renderers/utils'
 import { FileHandleLink } from '../widgets/FileHandleLink'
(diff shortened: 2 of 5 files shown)


packages/synapse-react-client/src/components/IconList.tsx
```diff
@@ -25,7 +32,13 @@ const IconList: React.FunctionComponent<IconListProps> = props => {
         noMatch = true
         return
       } else {
-        return <IconSvg key={el} {...iconConfig} />
+        return (
+          <IconSvg
+            style={{ fontSize: iconFontSize }}
```
nickgros, line 37: I think this is fine. Wondering if it's possible to use `sx` instead of `style` but guessing that might not be easy because
* You might have to deeply merge props
* Not all of the IconSvg items use MUI's `SvgIcon`, which is the component that actually uses `sx`

nickgros (approved):

packages/synapse-react-client/src/assets/mui_components/Other.tsx
```diff
@@ -2,9 +2,9 @@ import React from 'react'
 import SvgIcon, { SvgIconProps } from '@mui/material/SvgIcon'
 
 const Other = (props: SvgIconProps) => {
-  const { fill, style } = props
+  const { fill, style, ...rest } = props
```
hallieswan, line 5: Ensure that Other icon respects all passed props, e.g. fontSize

packages/synapse-react-client/src/components/IconList.tsx
```diff
@@ -1,20 +1,42 @@
...
-    [index: string]: IconSvgProps // if the icon option has the "label" set, it will show tooltip in IconSvg
-  }
+  iconConfigs: IconConfigs
   iconNames: string[]
+  iconFontSize?: string
   useTheme?: boolean
   useBackground?: boolean
 }
 
+const mergeIconFontSizeIntoIconConfigs = (
```
hallieswan, line 16: Set fontSize within sx of each icon's config object

packages/synapse-react-client/src/components/IconList.tsx
```diff
@@ -1,20 +1,42 @@
...
+  [index: string]: IconSvgProps // if the icon option has the "label" set, it will show tooltip in IconSvg
+}
+
 export type IconListProps = {
-  iconConfigs: {
-    [index: string]: IconSvgProps // if the icon option has the "label" set, it will show tooltip in IconSvg
-  }
+  iconConfigs: IconConfigs
   iconNames: string[]
+  iconFontSize?: string
```
nickgros, line 11: I was thinking we could just have `iconSx?: SxProps` instead of `iconFontSize`, but as I look at this, could you just pass this to `iconConfigs`?
```js
{
  sx: {
    fontSize: "40px"
  }
}

File: packages/synapse-react-client/src/components/GenericCard/GenericCard.tsx [hunk 5]
```diff
@@ -544,6 +544,11 @@ class _GenericCard extends React.Component<GenericCardPropsInternal> {
                     <IconList
                       iconConfigs={columnIconOptions.columns.dataType}
                       iconNames={JSON.parse(dataTypeIconNames)}
+                      commonIconProps={
+                        {
+                          sx: { fontSize: '40px' },
+                        } as IconListProps['commonIconProps']
```
nickgros, line 550: Do we need this cast here?

nickgros (approved):

packages/synapse-react-client/src/components/GenericCard/GenericCard.tsx
hallieswan: We don't! I'll remove it, thanks!

PR #669, opened 2024-01-25
PORTALS-2896: expand the existing test for visibility and update exploreConfig

jinjunoh:


apps/portals/e2e/explore.spec.ts
```diff
@@ -108,6 +113,70 @@ const getAvailableFacetsDiv = (page: Page) => {
...
     .filter({ hasText: 'Available Facets' })
 }
+const toggleChartVisibilityAndVerifyCount = async (page: Page) => {
+  const chartsPanel = page.locator('.FacetNav')
+
+  const allChartCells = chartsPanel.locator('.FacetNavPanel')
+  const allChartsCount = await allChartCells.count()
+
+  const chartCells = chartsPanel.locator('.FacetNavPanel:visible')
+  const initialCount = await chartCells.count()
```
nickgros, line 123: Consider using `getByRole` if possible. I think we explicitly have the `"graphics-document"` role on these visualizations. I am not sure if `getByRole` works with hidden nodes

apps/portals/e2e/explore.spec.ts
```diff
@@ -108,6 +113,70 @@ const getAvailableFacetsDiv = (page: Page) => {
...
+  await test.step('confirm chat visibility and count', async () => {
```
nickgros, line 124: small typo
```suggestion
  await test.step('confirm chart visibility and count', async () => {
```

apps/portals/e2e/explore.spec.ts
```diff
@@ -108,6 +113,70 @@ const getAvailableFacetsDiv = (page: Page) => {
...
+    let foundChartIndex = -1
+    for (let i = 0; i < allChartsCount; i++) {
+      const chart = allChartCells.nth(i)
+      if (await chart.isVisible()) {
+        foundChartIndex = i
+        break
+      }
+    }
+    const chartCell = allChartCells.nth(foundChartIndex)
+    const chartTitle = chartCell.locator('.FacetNavPanel__title')
```
nickgros, line 134: It's preferable to use roles/labels vs CSS classes to find UI elements on a page, especially in E2E tests. For some of these deeply-nested selectors, they should basically be treated as implementation details where they can be changed/removed at any time, which should minimally impact these tests.

What do you think of this approach instead?
1. Add a prop `role="toolbar"` to [the div](https://github.com/Sage-Bionetworks/synapse-web-monorepo/blob/8d098093961c76e042310e27ab3a930e7b1cb98c/packages/synapse-react-client/src/components/widgets/facet-nav/FacetNavPanel.tsx#L362) that contains these buttons (currently has the class `FacetNavPanel__title__tools`)
2. Replace this query with something like this (note: I am not sure if you can mix `getByRole` with `locator`, but hopefully something like this works)
```suggestion
    const chartToolbar = await chartCell.getByRole('toolbar')
```

apps/portals/e2e/explore.spec.ts
```diff
@@ -108,6 +113,70 @@ const getAvailableFacetsDiv = (page: Page) => {
...
+) => {
+  const showHideButton = page.getByRole('button', {
+    name: 'Show / Hide Visualizations',
+  })
+  const ageDistChart = page.locator('text=Age Distribution')
+  const isAgeDistributionChartPresent = (await ageDistChart.count()) > 0
+
+  const chartsPanel = page.locator('.FacetNav')
+  const chartCells = chartsPanel.locator('.FacetNavPanel:visible')
+  const initialCount = await chartCells.count()
```
nickgros, line 158: Consider extracting duplicate logic into another function

apps/portals/e2e/explore.spec.ts
nickgros: I want to acknowledge that some of the tests in this file are doing this, so it's understandable that you just followed by example

apps/portals/e2e/explore.spec.ts
```diff
@@ -108,6 +113,70 @@ const getAvailableFacetsDiv = (page: Page) => {
...
+}
+
+const toggleShowHideVisualizationsAndVerify = async (
+  page: Page,
+  shouldBeVisible: boolean,
+) => {
+  const showHideButton = page.getByRole('button', {
+    name: 'Show / Hide Visualizations',
+  })
+  const ageDistChart = page.locator('text=Age Distribution')
```
nickgros, line 153: why `.locator('text=...` instead of `getByText`?

apps/portals/e2e/explore.spec.ts
```diff
@@ -108,6 +113,70 @@ const getAvailableFacetsDiv = (page: Page) => {
...
+  const isAgeDistributionChartPresent = (await ageDistChart.count()) > 0
+
+  const chartsPanel = page.locator('.FacetNav')
+  const chartCells = chartsPanel.locator('.FacetNavPanel:visible')
+  const initialCount = await chartCells.count()
+
+  await test.step('toggle visualization and count', async () => {
+    for (let i = 0; i < 2; i++) {
```
nickgros, line 161: Why is this for loop here? Should this be removed, or am I missing where `i` is used?

nickgros (changes requested):

apps/portals/e2e/explore.spec.ts
jinjunoh: I wanted to simulate toggling the Show / Hide Visualization twice to look at scenarios when all the visuals are hidden and when it's toggled back to show all the visuals. I also wanted to check whether or not the Show / Hide Visualization button preserve the number of visuals from before toggling. Would it be better to get rid of the loop and repeat the code twice? or making a helper function to extract duplicate logic?

nickgros: That makes sense. The loop is confusing to me, so I like the idea to add a helper function, that you call twice

apps/portals/e2e/explore.spec.ts
```diff
@@ -109,6 +114,109 @@ const getAvailableFacetsDiv = (page: Page) => {
...
+  const initialCount = visibleChartCount
+
+  await test.step('confirm chart visibility and count', async () => {
+    const chartCell = allChartCells.nth(chartToTestIndex)
+    const chartToolbar = await chartCell.getByRole('toolbar')
+    await chartToolbar
+      .getByRole('button', { name: 'Hide graph under Show More' })
+      .click()
+
+    const newCount = await countVisibleElements(allChartCells, allChartsCount)
```
nickgros, line 155: Does this work? If so, maybe get rid of this function all together and just do this
```suggestion
    const newCount = await allChartCells.filter({ has: page.locator('visible') }).count()
```

apps/portals/e2e/explore.spec.ts
```diff
@@ -109,6 +114,109 @@ const getAvailableFacetsDiv = (page: Page) => {
...
+  let firstVisibleindex = -1
+  for (let i = 0; i < allChartsCount; i++) {
+    const chart = allChartCells.nth(i)
+    if (await chart.isVisible()) {
+      visibleChartCount++
+    }
+    if ((await chart.isVisible()) && firstVisibleindex == -1) {
+      firstVisibleindex = i
+    }
+  }
```
nickgros, line 144: Prefer readability here and reuse your function even if it means looping twice; we don't show enough charts to make it worth optimizing to loop only once.

```suggestion
  const visibleChartCount = await countVisibleElements(allChartCells)
  let firstVisibleindex = -1
  for (let i = 0; i < allChartsCount; i++) {
    const chart = allChartCells.nth(i)
    if (await chart.isVisible()) {
      firstVisibleindex = i
      break;
    }
  }
```

apps/portals/e2e/explore.spec.ts
```diff
@@ -109,6 +114,109 @@ const getAvailableFacetsDiv = (page: Page) => {
...
+const countVisibleElements = async (allCells: Locator, count: number) => {
+  let visibleCount = 0
+  for (let i = 0; i < count; i++) {
+    const cell = allCells.nth(i)
+    if (await cell.isVisible()) {
+      visibleCount++
+    }
+  }
+  return visibleCount
+}
```
nickgros, line 126: Seems like we don't need the 2nd parameter, right? See below comments first, we may not need this function

```suggestion
const countVisibleElements = async (allCells: Locator) => {
  const count = await allCells.count()
  let visibleCount = 0
  for (let i = 0; i < count; i++) {
    const cell = allCells.nth(i)
    if (await cell.isVisible()) {
      visibleCount++
    }
  }
  return visibleCount
}
```

apps/portals/e2e/explore.spec.ts
nickgros: If `filter` ends up working (comment below), do this:

```suggestion
  const visibleChartCount = await allChartCells.filter({ has: page.locator('visible') }).count()
  const chartToTest = await allChartCells.filter({ has: page.locator('visible') }).first()
  let firstVisibleindex = -1
  for (let i = 0; i < allChartsCount; i++) {
    const chart = allChartCells.nth(i)
    if (await chart.isVisible()) {
      firstVisibleindex = i
      break;
    }
  }
```

apps/portals/e2e/explore.spec.ts
```diff
@@ -109,6 +114,109 @@ const getAvailableFacetsDiv = (page: Page) => {
...
+    await chartToolbar
+      .getByRole('button', { name: 'Hide graph under Show More' })
+      .click()
+
+    const newCount = await countVisibleElements(allChartCells, allChartsCount)
+    expect(newCount).toBe(initialCount - 1)
+  })
+}
+
+const toggleAndVerifyVisbility = async (
```
nickgros, line 160: small typo
```suggestion
const toggleAndVerifyVisibility = async (
```

nickgros (approved):
Couple of suggestions to consider, but do not affect the actual behavior of the tests, looks good!

PR #673, opened 2024-01-25
PORTALS-2910: custom command must support row selection and all

jay-hodgson:


apps/portals/src/configurations/crc-researcher/synapseConfigs/hidden.ts
```diff
@@ -26,7 +26,6 @@ export const hidden: SynapseConfig = {
         onClick: (event) => {
           handleParticipantWorkflowChange(event, 'Uncategorized')
         },
-        isRowSelectionSupported: true,
```
jay-hodgson, line 29: Removed isRowSelectionSupported from props.  Now all custom commands are assumed to support either case (when row selections are available or unavailable)

apps/portals/src/configurations/elportal/synapseConfigs/cohortbuilder.ts
```diff
@@ -6,14 +6,8 @@ import {
...
-import {
-  handleParticipantsToFiles,
-  handleSelectedParticipantsToFiles,
-} from './handleParticipantsToFiles'
-import {
-  handleFilesToParticipants,
-  handleSelectedFilesToParticipants,
-} from './handleFilesToParticipants'
+import { handleSelectedParticipantsToFiles } from './handleParticipantsToFiles'
+import { handleSelectedFilesToParticipants } from './handleFilesToParticipants'
```
jay-hodgson, line 10: only one handler for each case, which supports both cases.

apps/portals/src/configurations/elportal/synapseConfigs/handleFilesToParticipants.tsx
```diff
@@ -9,10 +9,7 @@ import {
   CustomControlCallbackData,
 } from 'synapse-react-client'
 
-export const handleFilesToParticipants = async (
-  event: CustomControlCallbackData,
-) => {
-  // add filter for files perspective, to show participants associated to the selected files only.
+const getAllIDs = async (event: CustomControlCallbackData) => {
```
jay-hodgson, line 12: Cohort Builder consists of 2 views - participant centric or files centric.
When no rows are selected, this queries for a single field (all pages) and passes the appropriate IDs to the other view.

hallieswan (approved):
Looks good to me!

PR #676, opened 2024-01-29
SWC-6660

hallieswan:
component | change | main | feature
:---:|:---:|:---:|:---:
title bar - file in external S3 bucket | storage location | SWC-6660_externalS3_file_main | SWC-6660_externalS3_file_feature
title bar - folder in external S3 bucket | not changed | SWC-6660_externalS3_folder_main | SWC-6660_externalS3_folder_feature
entity modal metadata - file in external S3 bucket | storage | SWC-6660_entityModalMetadata_main | SWC-6660_entityModalMetadata_feature

### DIFF
--- a/packages/synapse-react-client/src/components/entity/metadata/MetadataTable.tsx
+++ b/packages/synapse-react-client/src/components/entity/metadata/MetadataTable.tsx
@@ -12,6 +12,7 @@ import {
 } from '../../../utils/functions/FileHandleUtils'
 import useGetEntityBundle from '../../../synapse-queries/entity/useEntityBundle'
 import { UserBadge } from '../../UserCard/UserBadge'
+import { useGetUploadDestinationForStorageLocation } from '../../../synapse-queries/file/useUploadDestination'
 
 export type MetadataTableProps = {
   readonly entityId: string
@@ -29,10 +30,19 @@ export const MetadataTable = ({
   const dataFileHandle = entityBundle
     ? getDataFileHandle(entityBundle)
     : undefined
+  const parentId = entityBundle?.entity?.parentId
+  const storageLocationId = dataFileHandle?.storageLocationId
+  const { data: storageLocationUploadDestination } =
+    useGetUploadDestinationForStorageLocation(parentId!, storageLocationId!, {
+      enabled: parentId !== undefined && storageLocationId !== undefined,
+    })
 
   let fileLocationName = undefined
   if (dataFileHandle) {
-    fileLocationName = getStorageLocationName(dataFileHandle)
+    fileLocationName = getStorageLocationName(
+      dataFileHandle,
+      storageLocationUploadDestination,
+    )
   }
 
   return entityBundle ? (
--- a/packages/synapse-react-client/src/components/entity/metadata/MetadataTable.tsx
+++ b/packages/synapse-react-client/src/components/entity/metadata/MetadataTable.tsx
@@ -36,7 +36,7 @@ export const MetadataTable = ({
     useGetUploadDestinationForStorageLocation(parentId!, storageLocationId!, {
       enabled:
         parentId !== undefined &&
-        !(storageLocationId === undefined || storageLocationId === null),
+        storageLocationId != null,
     })
 
   let fileLocationName = undefined
--- a/packages/synapse-react-client/src/components/entity/metadata/MetadataTable.tsx
+++ b/packages/synapse-react-client/src/components/entity/metadata/MetadataTable.tsx
@@ -34,7 +34,9 @@ export const MetadataTable = ({
   const storageLocationId = dataFileHandle?.storageLocationId
   const { data: storageLocationUploadDestination } =
     useGetUploadDestinationForStorageLocation(parentId!, storageLocationId!, {
-      enabled: parentId !== undefined && storageLocationId !== undefined,
+      enabled:
+        parentId !== undefined &&
+        !(storageLocationId === undefined || storageLocationId === null),
     })
 
   let fileLocationName = undefined
--- a/packages/synapse-react-client/src/components/entity/page/title_bar/TitleBarProperties.integration.test.tsx
+++ b/packages/synapse-react-client/src/components/entity/page/title_bar/TitleBarProperties.integration.test.tsx
@@ -5,6 +5,8 @@ import {
   EntityType,
   ExternalFileHandle,
   ExternalObjectStoreFileHandle,
+  S3FileHandle,
+  S3_FILE_HANDLE_CONCRETE_TYPE_VALUE,
   VersionableEntity,
 } from '@sage-bionetworks/synapse-types'
 import { render, screen } from '@testing-library/react'
@@ -32,6 +34,10 @@ import TitleBarProperties, {
   TitleBarPropertiesProps,
 } from './TitleBarProperties'
 import * as UseGetEntityPropertiesModule from './useGetEntityTitleBarProperties'
+import {
+  MOCK_EXTERNAL_S3_STORAGE_LOCATION_ID,
+  mockExternalS3UploadDestination,
+} from '../../../../mocks/mock_upload_destination'
 
 const HAS_ACCESS_V2_DATA_TEST_ID = 'mock-has-access-v2'
 
@@ -216,6 +222,27 @@ describe('TitleBarProperties', () => {
 
       await screen.findByText(`Storage Location`)
       await screen.findByText('Synapse Storage')
+    })
+    it('File handle storage location bucket and baseKey', async () => {
+      const bucketName = mockExternalS3UploadDestination.bucket
+      useEntityBundleOverride({
+        ...mockFileEntity.bundle,
+        fileHandles: [
+          {
+            ...mockFileHandle,
+            concreteType: S3_FILE_HANDLE_CONCRETE_TYPE_VALUE,
+            bucketName: bucketName,
+            storageLocationId: MOCK_EXTERNAL_S3_STORAGE_LOCATION_ID,
+          } as S3FileHandle,
+        ],
+      })
+      renderComponent()
+      await expandPropertiesIfPossible()
+
+      await screen.findByText(`Storage Location`)
+      await screen.findByText(
+        `s3://${bucketName}/${mockExternalS3UploadDestination.baseKey}`,
+      )
     })
     it('File handle endpoint, bucket, key', async () => {
       const endpointUrl = 'https://my-endpoint.fake'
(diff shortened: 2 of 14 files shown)


packages/synapse-react-client/src/mocks/entity/mockProject.ts
```diff
@@ -20,6 +21,19 @@ export const mockProjectIds = times(20).map(i => i + 10001)
 const MOCK_PROJECT_ID = `syn10000`
 const MOCK_PROJECT_NAME = 'A Mock Project'
 
+export const mockDoiAssociation: DoiAssociation = {
```
hallieswan, line 24: Move into an exported mock, which can be used to resolve lint errors in TitleBar integration tests

packages/synapse-react-client/src/components/entity/page/title_bar/TitleBarProperties.integration.test.tsx
```diff
@@ -226,7 +258,7 @@ describe('TitleBarProperties', () => {
             endpointUrl,
             bucket,
             fileKey,
-          },
```
hallieswan, line 229: The remaining changes in this file are related to linting errors found after colocating test file with component

packages/synapse-react-client/src/utils/functions/FileHandleUtils.test.ts
```diff
@@ -4,41 +4,55 @@ import {
...
+  mockExternalUploadDestination,
+  mockS3UploadDestination,
+} from '../../mocks/mock_upload_destination'
 
 describe('File Handle Utils', () => {
   describe('getDataFileHandle', () => {
     it('Returns the data file handle ID if it exists', () => {
       const dataFileHandleId = '123'
       const fileHandle = {
+        ...mockFileHandle,
```
hallieswan, line 48: Changes in this describe block are related to lint errors raised after colocating test with component

packages/synapse-react-client/src/utils/functions/FileHandleUtils.ts
```diff
@@ -72,12 +80,22 @@ export function getFileHandleStorageInfo(
...
+  const baseKeyText = baseKey ? `/${baseKey}` : ''
+
   switch (fileHandle.concreteType) {
     case PROXY_FILE_HANDLE_CONCRETE_TYPE_VALUE:
       return (fileHandle as ProxyFileHandle).filePath
     case EXTERNAL_OBJECT_STORE_FILE_HANDLE_CONCRETE_TYPE_VALUE:
-      return (fileHandle as ExternalObjectStoreFileHandle).bucket
+      return `${
+        (fileHandle as ExternalObjectStoreFileHandle).bucket
+      }${baseKeyText}`
```
hallieswan, line 98: @nickgros Do you know whether the value of bucket in `fileHandle` always match the value in `storageLocationUploadDestination`? Would it be appropriate to call [`getUploadDestinationString`](https://github.com/Sage-Bionetworks/synapse-web-monorepo/blob/a4d978ba1de54da55ece084c75d077fefbb3562c/packages/synapse-react-client/src/utils/functions/FileHandleUtils.ts#L123) on the `storageLocationUploadDestination` instead?

packages/synapse-react-client/src/components/entity/page/title_bar/useGetEntityTitleBarProperties.tsx
```diff
@@ -42,18 +45,42 @@ export function useGetEntityTitleBarProperties(
...
+  // is where the file handle was actually uploaded. We must use a service that
+  // takes the storage location id in the input to get the storage location for a
+  // file entity.
+  const { data: storageLocationUploadDestination } =
+    useGetUploadDestinationForStorageLocation(parentId!, storageLocationId!, {
+      enabled:
+        !isContainer &&
+        parentId !== undefined &&
+        storageLocationId !== undefined,
+    })
```
nickgros, line 83: Thank you for this clarifying comment! What do you think about encapsulating both of these calls into a single `useGetUploadDestination` hook that calls the correct hook based on whether the entity is a container? Or maybe it makes more sense to keep them separate like this, since they really are fetching different things? I have no strong preference

nickgros (approved):
This looks great! Thank you for the clean-up!

packages/synapse-react-client/src/utils/functions/FileHandleUtils.ts
nickgros: I am not 100% sure but I would expect it does always match the value (the backend probably JOINs this data when we request it). These implementations are looking very similar, so I think it would be a good idea to reuse them.

nickgros: Are there cases where a storage location may not have an upload destination?

hallieswan: > Are there cases where a storage location may not have some related upload destination?

If a `storageLocationId` exists, then the [service](https://rest-docs.synapse.org/rest/GET/entity/id/uploadDestination/storageLocationId.html) will always return an upload destination. However, `storageLocationId` is optional in the [FileHandle](https://rest-docs.synapse.org/rest/org/sagebionetworks/repo/model/file/FileHandle.html). Based on [this comment](https://github.com/Sage-Bionetworks/synapse-web-monorepo/blob/a4d978ba1de54da55ece084c75d077fefbb3562c/packages/synapse-react-client/src/utils/functions/FileHandleUtils.ts#L84), when `storageLocationId` is `null`, we assume the file handle is  in Synapse Storage. So even if use `getUploadDestinationString` to deduplicate some logic, we would still need to handle this case. Are there other cases you're considering that we'd need to handle?

packages/synapse-react-client/src/components/entity/page/title_bar/useGetEntityTitleBarProperties.tsx
hallieswan: Oh that's a good idea! I'll make that change

hallieswan: Actually, on second thought, I might keep them separate, since they are fetching different things -- for example, in `MetadataTable`, we get the storage location for file handles, but don't use the default upload destination for container entities, so we only need to call `useGetUploadDestinationForStorageLocation`. Keeping the hooks separate more clearly delineates which service will be called. What do you think?

packages/synapse-react-client/src/utils/functions/FileHandleUtils.ts
```diff
@@ -82,38 +79,32 @@ export function getFileHandleStorageInfo(
...
-      ) {
-        return SYNAPSE_STORAGE
-      } else {
-        return `s3://${(fileHandle as S3FileHandle).bucketName}${baseKeyText}`
-      }
     case GOOGLE_CLOUD_FILE_HANDLE_CONCRETE_TYPE_VALUE:
-      return `gs://${
-        (fileHandle as GoogleCloudFileHandle).bucketName
-      }${baseKeyText}`
+      if (storageLocationUploadDestination) {
```
hallieswan, line 99: Reuse `getUploadDestinationString` when we could get an upload destination using the storage location id. Otherwise, default to the bucket value stored in the file handle.

packages/synapse-types/src/File/FileHandle.ts
```diff
@@ -27,7 +27,7 @@ export interface FileHandle {
   /** The short, user visible name for this file. */
   fileName: string
   /** The optional storage location descriptor */
-  storageLocationId: number
+  storageLocationId: number | null
```
hallieswan, line 30: Update type to handle case where storage location id is null when storage location field was not back-filled, as per comment in FileHandleUtils

packages/synapse-react-client/src/components/entity/metadata/MetadataTable.tsx
```diff
@@ -34,7 +34,9 @@ export const MetadataTable = ({
   const storageLocationId = dataFileHandle?.storageLocationId
   const { data: storageLocationUploadDestination } =
     useGetUploadDestinationForStorageLocation(parentId!, storageLocationId!, {
-      enabled: parentId !== undefined && storageLocationId !== undefined,
+      enabled:
+        parentId !== undefined &&
+        !(storageLocationId === undefined || storageLocationId === null),
```
nickgros, line 39: Can be simplified
```suggestion
        storageLocationId != null,
```

packages/synapse-react-client/src/components/entity/page/title_bar/useGetEntityTitleBarProperties.tsx
```diff
@@ -79,7 +79,7 @@ export function useGetEntityTitleBarProperties(
       enabled:
         !isContainer &&
         parentId !== undefined &&
-        storageLocationId !== undefined,
+        !(storageLocationId === undefined || storageLocationId === null),
```
nickgros, line 82: Ditto
```suggestion
        storageLocationId != null,
```

packages/synapse-types/src/File/FileHandle.ts
nickgros: It's definitely sometimes undefined, not sure if it is ever `null`, so you may consider changing it to this. But if this works, I'm ok with the type imprecision; we may be moving to generated models soon, so I don't think it's that important that this is perfect for nullish values
```suggestion
  storageLocationId?: number
```

packages/synapse-react-client/src/utils/functions/FileHandleUtils.ts
nickgros: Thanks for this consolidation

packages/synapse-react-client/src/utils/functions/FileHandleUtils.ts
nickgros: I was thinking of [ExternalFileHandles](https://rest-docs.synapse.org/rest/org/sagebionetworks/repo/model/file/ExternalFileHandle.html) but those are unaffected by this change

nickgros (approved):

packages/synapse-types/src/File/FileHandle.ts
hallieswan: For now, I'll include `null` in the type since the [previous implementation](https://github.com/Sage-Bionetworks/synapse-web-monorepo/blob/045b8597f9a12ee96b76da55e768eddd64740a5d/packages/synapse-react-client/src/utils/functions/FileHandleUtils.ts#L87) checked for null storage location ids and I maintained the null check in the [PR changes](https://github.com/hallieswan/synapse-web-monorepo/blob/0e0ad995e3b5361123882d311b314c07d4e0ee84/packages/synapse-react-client/src/utils/functions/FileHandleUtils.ts#L88), but happy to change in the future!

PR #681, opened 2024-01-31
PORTALS-2768: only show icon, no text in SynapseTable access column

hallieswan:
### Request Access - Table

https://github.com/Sage-Bionetworks/synapse-web-monorepo/assets/26949006/5b32f799-2c96-4de8-a0ee-4ec54ffd02b5

https://github.com/Sage-Bionetworks/synapse-web-monorepo/assets/26949006/767506a3-b97d-4477-b9c8-5363a2ebafa5

### View Terms - Table

https://github.com/Sage-Bionetworks/synapse-web-monorepo/assets/26949006/caa0e87f-c298-48ec-8cd1-69e496a23e44

https://github.com/Sage-Bionetworks/synapse-web-monorepo/assets/26949006/8ae58c45-3d7d-4146-b8fc-f82d26fa3c6d

### Request Access - Options

https://github.com/Sage-Bionetworks/synapse-web-monorepo/assets/26949006/c569093e-ba77-41e2-93a2-410e10f0dc86

https://github.com/Sage-Bionetworks/synapse-web-monorepo/assets/26949006/0f8e0fd5-cd7f-4f21-aa38-da97c752bb01

### Title Bar

main: 
main_titleBar

feature: 
feature_titleBar

### DIFF
--- a/packages/synapse-react-client/src/components/HasAccess/HasAccessV2.integration.test.tsx
+++ b/packages/synapse-react-client/src/components/HasAccess/HasAccessV2.integration.test.tsx
@@ -21,6 +21,7 @@ import { MOCK_CONTEXT_VALUE } from '../../mocks/MockSynapseContext'
 import {
   mockOpenRestrictionInformation,
   mockUnmetControlledDataRestrictionInformationACT,
+  mockUnmetControlledDataRestrictionInformationRestricted,
 } from '../../mocks/mock_has_access_data'
 import { rest, server } from '../../mocks/msw/server'
 
@@ -37,6 +38,22 @@ const renderComponent = (
 }
 const props: HasAccessProps = {
   entityId,
+}
+
+const expectIcon = async (iconName: string, tooltipText: string) => {
+  const icon = await screen.findByRole('img')
+  expect(icon).toHaveAccessibleName(tooltipText)
+  expect(icon.getAttribute('data-svg')).toEqual(iconName)
+}
+const expectSignInButton = async () => {
+  const button = await screen.findByRole('button')
+  expect(button).toHaveTextContent('')
+  expect(button).toHaveClass(SRC_SIGN_IN_CLASS)
+}
+const expectArButton = async (linkText: string) => {
+  const button = await screen.findByRole('button')
+  expect(button).toHaveTextContent(linkText)
+  expect(button).not.toHaveClass(SRC_SIGN_IN_CLASS)
 }
 
 const onGetRestrictionInformation = jest.fn()
@@ -92,9 +109,7 @@ describe('HasAccess tests', () => {
 
     renderComponent(props)
 
-    expect((await screen.findByRole('img')).getAttribute('data-svg')).toEqual(
-      'accessOpen',
-    )
+    await expectIcon('accessOpen', '')
     expect(screen.queryByRole('button')).not.toBeInTheDocument()
 
     const expectedRestrictionInformationRequest: RestrictionInformationRequest =
@@ -122,9 +137,7 @@ describe('HasAccess tests', () => {
 
     renderComponent(props)
 
-    expect((await screen.findByRole('img')).getAttribute('data-svg')).toEqual(
-      'accessOpen',
-    )
+    await expectIcon('accessOpen', '')
     expect(screen.queryByRole('button')).not.toBeInTheDocument()
 
     const expectedRestrictionInformationRequest: RestrictionInformationRequest =
@@ -153,10 +166,10 @@ describe('HasAccess tests', () => {
 
       renderComponent(props)
 
-      const icon = await screen.findByRole('img', {
-        name: 'You do not have download access for this item.',
-      })
-      expect(icon.getAttribute('data-svg')).toEqual('accessClosed')
+      await expectIcon(
+        'accessClosed',
+        'You do not have download access for this item.',
+      )
       expect(screen.queryByRole('button')).not.toBeInTheDocument()
 
       const expectedRestrictionInformationRequest: RestrictionInformationRequest =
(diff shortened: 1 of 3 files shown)


packages/synapse-react-client/src/components/HasAccess/HasAccessV2.tsx
```diff
@@ -18,18 +18,22 @@ import AccessRequirementList, {
...
 
 export type HasAccessProps = {
   onHide?: () => void
   entityId: string
   entityVersionNumber?: string
   className?: string
+  showButtonText?: boolean
 }
 
+const buttonSx = { p: '0px', minWidth: 'unset' }
```
hallieswan, line 32: Matches styling between button-wrapped icon and icon alone

packages/synapse-react-client/src/components/HasAccess/HasAccessV2.integration.test.tsx
```diff
@@ -260,12 +298,12 @@ describe('HasAccess tests', () => {
...
-        name: 'You must request access to this restricted file.',
-      })
-      expect(icon.getAttribute('data-svg')).toEqual('accessClosed')
+      await expectIcon(
+        'accessClosed',
+        'You must request access to this restricted file.',
+      )
       // User will be prompted to sign in in the AR modal
-      expect(screen.queryByRole('button')).not.toBeInTheDocument()
+      await expectArButton('Request Access')
```
hallieswan, line 306: Previously used a button to open sign in modal and a link to open the AR modal, but now using a button in both cases so must check for the type of button rather than asserting that there are no buttons in the document

packages/synapse-react-client/src/components/HasAccess/HasAccessV2.tsx
```diff
@@ -18,18 +18,22 @@ import AccessRequirementList, {
...
+
 export enum FileHandleDownloadTypeEnum {
   Accessible = 'Accessible',
+  AccessibleWithTerms = 'AccessibleWithTerms',
```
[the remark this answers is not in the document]
nickgros: Ah, because the only difference before was the presence of the `View Terms` text, but now that that's gone, we need to add a completely new state. Makes sense!

packages/synapse-react-client/src/components/HasAccess/HasAccessV2.tsx
```diff
@@ -90,6 +100,13 @@ function AccessIcon(props: { downloadType: FileHandleDownloadTypeEnum }) {
   return <></>
 }
 
+function isFileHandleDownloadTypeUndefined(
+  fileHandleDownloadType: FileHandleDownloadTypeEnum | undefined,
+) {
+  // note, this can't be "if (!downloadType)" since DownloadTypeEnum has a 0 value (which is falsy)
+  return fileHandleDownloadType === undefined
```
nickgros, line 107: Is this really true? `FileHandleDownloadTypeEnum` is a string enum

nickgros: I see now that this comment was just copied.

nickgros: Here's the original change where the comment should have been removed: https://github.com/Sage-Bionetworks/synapse-web-monorepo/commit/fd9c270d53f6c3f2042809ba8b14e24679ee9c90

Can you go ahead and remove this comment?

packages/synapse-react-client/src/components/HasAccess/HasAccessV2.tsx
```diff
@@ -302,15 +329,14 @@ export function HasAccessV2(props: HasAccessProps) {
...
     ) : (
-      <AccessIcon downloadType={fileHandleDownloadType} />
+      <AccessIcon downloadType={fileHandleDownloadType!} />
     )
 
   return (
     <span style={{ whiteSpace: 'nowrap' }}>
-      {iconContainer}
-      {accessRequirementsJsx}
+      {accessRequirementsJsx ? accessRequirementsJsx : iconContainer}
```
nickgros, line 339: So it looks like we are rendering the `iconContainer` when `accessRequirementsJsx` is null. Would it make more sense/be possible to this? 
1. Define `iconContainer` first
2. Return `iconContainer` instead of `null` in the `accessRequirementsJsx` logic

Then we could remove this ternary and it might be a bit more straightforward.

packages/synapse-react-client/src/components/HasAccess/HasAccessV2.integration.test.tsx
```diff
@@ -97,9 +109,7 @@ describe('HasAccess tests', () => {
 
     renderComponent(props)
 
-    expect((await screen.findByRole('img')).getAttribute('data-svg')).toEqual(
-      'accessOpen',
-    )
+    await expectIcon('accessOpen', '')
```
nickgros, line 112: Big thanks for this cleanup

nickgros (approved):
Had a few small suggestions. Looks great!

packages/synapse-react-client/src/components/HasAccess/HasAccessV2.tsx
hallieswan: Oh will do! Thanks for tracking down the original change.

packages/synapse-react-client/src/components/HasAccess/HasAccessV2.tsx
hallieswan: Great idea, I'll make that improvement!

packages/synapse-react-client/src/components/HasAccess/HasAccessV2.tsx
```diff
@@ -100,13 +100,6 @@ function AccessIcon(props: { downloadType: FileHandleDownloadTypeEnum }) {
   return <></>
 }
 
-function isFileHandleDownloadTypeUndefined(
-  fileHandleDownloadType: FileHandleDownloadTypeEnum | undefined,
-) {
-  // note, this can't be "if (!downloadType)" since DownloadTypeEnum has a 0 value (which is falsy)
-  return fileHandleDownloadType === undefined
-}
-
```
hallieswan: Comment is no longer accurate, so a separate method is not needed -- can just check with `!fileHandleDownloadType` as needed

packages/synapse-react-client/src/components/HasAccess/HasAccessV2.tsx
```diff
@@ -260,7 +285,7 @@ export function HasAccessV2(props: HasAccessProps) {
       linkText = 'Request Access'
     } else if (RestrictionLevel.OPEN === restrictionLevel) {
       // they need to sign in
-      return null
+      return iconContainer
```
hallieswan, line 288: Return `iconContainer` here, so ternary can be removed below

nickgros (approved):

PR #684, opened 2024-02-02
SWC-6678 - Automatically add default columns for views

nickgros:
https://github.com/Sage-Bionetworks/synapse-web-monorepo/assets/17580037/c6cf0b19-46b6-4b6a-8397-8ad7fab0039f

### DIFF
--- a/packages/synapse-react-client/src/components/EvaluationFinder/EvaluationFinder.test.tsx
+++ b/packages/synapse-react-client/src/components/EvaluationFinder/EvaluationFinder.test.tsx
@@ -46,7 +46,7 @@ describe('EvaluationFinder', () => {
     expect(evaluationsToSelect[1]).toBeChecked()
 
     const nextPageButton = await screen.findByRole('button', {
-      name: 'Next',
+      name: 'Next Page',
     })
 
     await waitFor(() => expect(nextPageButton).not.toBeDisabled())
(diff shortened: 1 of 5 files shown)


packages/synapse-react-client/src/components/EvaluationFinder/EvaluationFinder.test.tsx
```diff
@@ -46,7 +46,7 @@ describe('EvaluationFinder', () => {
     expect(evaluationsToSelect[1]).toBeChecked()
 
     const nextPageButton = await screen.findByRole('button', {
-      name: 'Next',
+      name: 'Next Page',
```
nickgros, line 49: When shown in the Submission View wizard, the modal also has a "Next" button. Adding "Page" here so we don't have two buttons with the same name

packages/synapse-react-client/src/components/EvaluationFinder/EvaluationFinder.tsx
```diff
@@ -111,7 +111,7 @@ export default function EvaluationFinder(props: EvaluationFinderProps) {
             }
           }}
         >
-          Next
+          Next Page
```
nickgros, line 114: When shown in the Submission View wizard, the modal also has a "Next" button. Adding "Page" here so we don't have two buttons with the same text on screen at the same time

jay-hodgson (approved):

PR #689, opened 2024-02-06
[WIP] SWC-6662 - openapi client gen with openapi-generator typescript-fetch

nickgros:


packages/synapse-client/.gitignore
```diff
@@ -0,0 +1 @@
+src/generated/
```
nickgros, line 1: Don't commit the generated models to git

packages/synapse-client/README.md
```diff
@@ -0,0 +1 @@
+pnpm openapi-generator-cli config-help -g typescript-fetch
```
nickgros, line 1: TODO

packages/synapse-client/out/modelGeneric.mustache
```diff
@@ -0,0 +1,174 @@
...
+    {{#required}}
+    if (!('{{name}}' in value)) return false;
+    {{! START CUSTOM TEMPLATING FOR synapse-client }}
+    {{! if the property is a required enum, check that the value is a valid enum value }}
+    {{#isEnum}}
+    if (!Object.values({{classname}}{{enumName}}).includes(value["{{name}}"] as {{classname}}{{enumName}})) {
+      return false
+    }
+    {{/isEnum}}
+    {{! END CUSTOM TEMPLATING FOR synapse-client }}
```
nickgros, line 40: In the `instanceOf` method, if a model's property is a required enum, but the property's value does not match a known enum value, return `false`. This is especially useful because it will use our `concreteType` property, which is now always defined as an enum following PLFM-8257

packages/synapse-client/out/modelGeneric.mustache
```diff
@@ -0,0 +1,174 @@
+import { mapValues } from '../runtime{{importFileExtension}}';
```
nickgros, line 1: Most of this code is copied/generated from the original template source.

Look for `START CUSTOM TEMPLATING FOR synapse-client`

packages/synapse-client/out/modelGeneric.mustache
```diff
@@ -0,0 +1,174 @@
...
+{{/discriminator}}
+{{>modelGenericInterfaces}}
+
+/**
+ * Check if a given object implements the {{classname}} interface.
+ */
+{{! START CUSTOM TEMPLATING FOR synapse-client }}
+{{! Change the return type from `boolean` to a type predicate so this method can be used for type narrowing }}
+export function instanceOf{{classname}}(value: object): value is {{classname}} {
+{{! END CUSTOM TEMPLATING FOR synapse-client }}
```
nickgros, line 29: The return type of this function was originally `boolean` but we can make it a [type predicate](https://www.typescriptlang.org/docs/handbook/2/narrowing.html#using-type-predicates) so this function can be used as a type guard, e.g.

```
const obj = {
  // ... other properties
  "concreteType":  "org.sagebionetworks.repo.model.FileEntity"
}

if (instanceOfFileEntity(obj)) {
  // in this block, obj is known to be FileEntity and can be treated as such without type casting
}

```

packages/synapse-client/src/SynapseClient.ts
```diff
@@ -0,0 +1,164 @@
...
+  UserProfileServicesApi,
+  VerificationServicesApi,
+  WikiPageServices2Api,
+  WikiPageServicesApi,
+} from './generated'
+
+/**
+ * Creates one class that encapsulates all sets of Synapse API services.
+ */
+export class SynapseClient extends BaseAPI {
```
nickgros, line 50: The generator creates one API client class per Java controller. I created this convenience wrapper over every service.

packages/synapse-client/src/generatedSpecTest.test.ts
```diff
@@ -0,0 +1,70 @@
+import { describe, expect, expectTypeOf, test } from 'vitest'
+import {
+  EntityFromJSONTyped,
+  FileEntity,
+  instanceOfFileEntity,
+} from './generated'
+
+describe('Tests for the generated specification', () => {
+  test('instanceOfXxx works for a model with a concreteType', () => {
```
nickgros, line 9: Test the template change that checks that required enum properties have a valid value in `instanceOf...`

packages/synapse-client/src/generatedSpecTest.test.ts
```diff
@@ -0,0 +1,70 @@
...
+    expect(instanceOfFileEntity(fileEntity)).toBe(true)
+    expect(instanceOfFileEntity(withDifferentConcreteType)).toBe(false)
+
+    // Verify that EntityFromJSONTyped inspects the concreteType
+    expect(EntityFromJSONTyped(fileEntity, false)).toEqual(fileEntity)
+    expect(() =>
+      EntityFromJSONTyped(withDifferentConcreteType, false),
+    ).toThrow()
+  })
+  test('instanceOfXxx asserts that an object is of the checked type', () => {
```
nickgros, line 43: Test the template change that allows us to use `instanceOf...` as a type predicate

packages/synapse-client/src/spec/openapispecification.json
nickgros, line 1: Synapse OpenAPI definition is checked in to ensure it is consistent

packages/synapse-client/src/util/ShortenCanonicalName.ts
```diff
@@ -0,0 +1,123 @@
...
+ * Output:
+ * ```
+ *  {
+ *    'org.sagebionetworks.foo.Bar': 'FooBar',
+ *    'org.sagebionetworks.foo.Baz': 'Baz',
+ *    'org.sagebionetworks.bar.Bar': 'BarBar',
+ *  }
+ * ```
+ */
+export function shortenCanonicalNames(
```
nickgros, line 47: This function is used to generate the names of the models so we don't have to use model names like `OrgSagebionetworksRepoModelFileEntity`

packages/synapse-react-client/src/components/OAuthClientManagement/CreateOAuthClient.tsx
nickgros, line 1: Integrated into synapse-react-client as an example -- needed to pick models/services that had a low code surface, because our hand-written models are incompatible with the generated models. The OAuth client services seemed to be ok for that

packages/synapse-react-client/src/components/OAuthClientManagement/CreateOAuthClient.tsx
```diff
@@ -106,31 +101,37 @@ export const CreateOAuthModal: React.FunctionComponent<
   useEffect(() => {
     setClientName(client?.client_name ?? '')
     setRedirectUris(
-      client?.redirect_uris.map(str => ({ uri: str })) ?? [{ uri: '' }],
+      client?.redirect_uris?.map((str: string) => ({ uri: str })) ?? [
+        { uri: '' },
+      ],
```
nickgros, line 106: Note that redirect_uris is NOT required in the spec, so we have to do additional null handling or assertions to work with the new model. This will be common, and is IMO the biggest downside and reason to not use the generated code right now.

packages/synapse-react-client/src/components/OAuthClientManagement/CreateOAuthClient.tsx
```diff
@@ -106,31 +101,37 @@ export const CreateOAuthModal: React.FunctionComponent<
...
     setPolicyUri(client?.policy_uri ?? '')
     setClientUri(client?.client_uri ?? '')
     setSectorUri(client?.sector_identifier_uri ?? '')
     setUserinfoSignedResponseAlg(
-      client?.userinfo_signed_response_alg ??
-        defaultUserInfoSignedResponseAlgorithm,
+      (client?.userinfo_signed_response_alg ??
+        defaultUserInfoSignedResponseAlgorithm) as
+        | 'JSON'
+        | OIDCSigningAlgorithm,
```
nickgros, line 115: Bug where the model for `userinfo_signed_response_alg` has the wrong type. The bug is in the spec, see PLFM-8319

packages/synapse-react-client/src/components/OAuthClientManagement/CreateOAuthClient.tsx
```diff
@@ -106,31 +101,37 @@ export const CreateOAuthModal: React.FunctionComponent<
...
-        ).then(precheckResult => {
-          setWarnTrigger(precheckResult.reverificationRequired)
-        })
+        synapseClient.openIDConnectServicesClient
+          .putAuthV1Oauth2ClientIdVerificationPrecheck({
+            id: oAuthClient.client_id!,
+            oAuthClient: oAuthClient,
+          })
+          .then(precheckResult => {
+            setWarnTrigger(precheckResult.reverificationRequired!)
```
nickgros, line 130: `reverificationRequired` is another property that is never null, but because the spec doesn't indicate it as required, we have to add the `!` non-null assertion

packages/synapse-react-client/src/components/OAuthClientManagement/CreateOAuthClient.tsx
```diff
@@ -106,31 +101,37 @@ export const CreateOAuthModal: React.FunctionComponent<
...
+      if (oAuthClient.client_id) {
         // SWC-6365: use the pre-check service to determine if we need to show a warning on edit
-        SynapseClient.isOAuthClientReverificationRequired(
-          oAuthClient,
-          accessToken,
-        ).then(precheckResult => {
-          setWarnTrigger(precheckResult.reverificationRequired)
-        })
+        synapseClient.openIDConnectServicesClient
+          .putAuthV1Oauth2ClientIdVerificationPrecheck({
```
nickgros, line 125: These method names are not great. Should we create our own wrapper? Push for the ability to incrementally curate these in the backend?

packages/synapse-react-client/src/mocks/MockSynapseContext.tsx
```diff
@@ -12,6 +14,12 @@ export const MOCK_CONTEXT_VALUE: SynapseContextType = {
   downloadCartPageUrl: '/DownloadCart',
   withErrorBoundary: false,
   keyFactory: new KeyFactory(MOCK_ACCESS_TOKEN),
+  synapseClient: new SynapseClient(
+    new Configuration({
+      fetchApi: fetchWithExponentialTimeout,
```
nickgros, line 19: Super fortunate that our existing abstraction just plugs in here. We pass in our `fetchWithExponentialTimeout` and we maintain the same behavior that we already have with automatic retries and all errors converted to our custom `SynapseClientError` class.

packages/synapse-react-client/src/utils/context/SynapseContext.tsx
```diff
@@ -50,6 +54,16 @@ export function SynapseContextProvider(props: SynapseContextProviderProps) {
     [providedContext.accessToken],
   )
 
+  const synapseApiClient = useMemo(() => {
+    const configuration = new Configuration({
+      fetchApi: fetchWithExponentialTimeout,
```
nickgros, line 59: Super fortunate that our existing abstraction just plugs in here. We pass in our `fetchWithExponentialTimeout` and we maintain the same behavior that we already have with automatic retries and all errors converted to our custom `SynapseClientError` class.

packages/synapse-react-client/src/components/OAuthClientManagement/CreateOAuthClient.tsx
```diff
@@ -106,31 +101,37 @@ export const CreateOAuthModal: React.FunctionComponent<
...
+        | 'JSON'
+        | OIDCSigningAlgorithm,
     )
     setTosUri(client?.tos_uri ?? '')
   }, [isShowingModal, client])
 
   useDebouncedEffect(
     () => {
-      if (accessToken && oAuthClient.client_id) {
+      if (oAuthClient.client_id) {
```
nickgros, line 122: I don't think you can ever see this component without being logged in, so I'm not worried about the `accessToken` checks

nickgros:
Current approach will increase our IIFE bundle size from 3MB to 5MB - this is already very large, so need to rethink the approach

packages/synapse-client/out/modelGeneric.mustache
nickgros: type predicate was added in upstream

packages/synapse-client/out/modelGeneric.mustache
```diff
@@ -0,0 +1,168 @@
...
+    {{#required}}
+    if (!('{{name}}' in value) || value['{{name}}'] === undefined) return false;
+    {{! START CUSTOM TEMPLATING FOR synapse-client }}
+    {{! if the property is a required enum, check that the value is a valid enum value }}
+    {{#isEnum}}
+      if (!Object.values({{classname}}{{enumName}}).includes(value["{{name}}"] as {{classname}}{{enumName}})) {
+      return false
+      }
+    {{/isEnum}}
+    {{! END CUSTOM TEMPLATING FOR synapse-client }}
```
nickgros, line 34: In the `instanceOf` method, if a model's property is a required enum, but the property's value does not match a known enum value, return `false`. This is especially useful because it will use our `concreteType` property, which is now always defined as an enum following PLFM-8257

packages/synapse-client/out/modelGeneric.mustache
```diff
@@ -0,0 +1,168 @@
...
+    {{#required}}
+    if (!('{{name}}' in value) || value['{{name}}'] === undefined) return false;
+    {{! START CUSTOM TEMPLATING FOR synapse-client }}
+    {{! if the property is a required enum, check that the value is a valid enum value }}
+    {{#isEnum}}
+      if (!Object.values({{classname}}{{enumName}}).includes(value["{{name}}"] as {{classname}}{{enumName}})) {
+        return false
+      }
+    {{/isEnum}}
+    {{! END CUSTOM TEMPLATING FOR synapse-client }}
```
nickgros, line 34: In the `instanceOf` method, if a model's property is a required enum, but the property's value does not match a known enum value, return `false`. This is especially useful because it will use our `concreteType` property, which is now always defined as an enum following PLFM-8257

Here is an example of generated code that includes this logic. See that the generated code checks that the `concreteType` matches one of the allowed enumeration values, which in this case is only `'org.sagebionetworks.repo.model.FileEntity'`.

```typescript
/**
 * Check if a given object implements the FileEntity interface.
 */
export function instanceOfFileEntity(value: object): value is FileEntity {
  if (!('concreteType' in value) || value['concreteType'] === undefined)
    return false
  if (
    !Object.values(FileEntityConcreteTypeEnum).includes(
      value['concreteType'] as FileEntityConcreteTypeEnum,
    )
  ) {
    return false
  }
  return true
}
```

nickgros:
Done in #1213

PR #690, opened 2024-02-06
Add error lint rule "react/no-object-type-as-default-prop"

nickgros:
This ESLint rule prevents us from specifying new objects as default props in components, which are re-created on each new render cycle, which can lead to performance issues. Note that default primitive values (which are effectively compared by value by React) are unaffected.

I also added the `quiet` option to the root `lint` script, which will cause the command to only output errors instead of errors and warnings, which is more useful in CI.

### DIFF
--- a/.eslintrc.cjs
+++ b/.eslintrc.cjs
@@ -36,6 +36,7 @@ module.exports = {
         additionalHooks: "useDebouncedEffect",
       },
     ],
+    "react/no-object-type-as-default-prop": "error",
     "no-extra-semi": "off",
     "prefer-const": "warn",
     "jest/expect-expect": "off",
--- a/package.json
+++ b/package.json
@@ -9,7 +9,7 @@
   "scripts": {
     "prepare": "husky install",
     "build": "nx run-many --target=build",
-    "lint": "nx run-many --target=lint",
+    "lint": "nx run-many --target=lint --quiet",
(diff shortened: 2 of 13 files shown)


.eslintrc.cjs
```diff
@@ -36,6 +36,7 @@ module.exports = {
         additionalHooks: "useDebouncedEffect",
       },
     ],
+    "react/no-object-type-as-default-prop": "error",
```
nickgros, line 39: [See doc](https://github.com/jsx-eslint/eslint-plugin-react/blob/master/docs/rules/no-object-type-as-default-prop.md)

Primitive default values for props are OK but object/array props are not unless you define them outside of the definition of the function component.

A few times I have made this mistake, which leads to excessive re-renders and is hard to track down. I see no downsides to adding this rule.

package.json
```diff
@@ -9,7 +9,7 @@
   "scripts": {
     "prepare": "husky install",
     "build": "nx run-many --target=build",
-    "lint": "nx run-many --target=lint",
+    "lint": "nx run-many --target=lint --quiet",
```
nickgros, line 12: Added `--quiet`, which will omit warnings and only show lint errors. This will make reading the CI output a bit more useful.

jay-hodgson (approved):

PR #692, opened 2024-02-06
PORTALS-2916 - Facet plot colors contrast within each plot

nickgros:
- Add and use `getContrastColorPalette` which traverses through the color palettes to return a sequence of contrasting colors

image


packages/synapse-react-client/src/components/ColorGradient/ColorGradient.tsx
```diff
@@ -12,6 +12,8 @@ export function rgba2rgb(background: number[], color: number[]) {
   ]
 }
 
+const VARIANTS_PER_COLOR = 5
+
 export function getColorPalette(
```
nickgros, line 17: Made a few modifications to this original function, but its behavior is unchanged

packages/synapse-react-client/src/components/ColorGradient/ColorGradient.tsx
```diff
@@ -41,4 +48,29 @@ export function getColorPalette(
   return { colorPalette, textColors }
 }
 
-export default getColorPalette
+export function getContrastColorPalette(
```
nickgros, line 51: Added a new function, since we may not want the highly-contrasting colors in all cases (e.g. TimelinePlot)

packages/synapse-react-client/src/components/widgets/facet-nav/FacetNavPanel.tsx
```diff
@@ -80,8 +80,9 @@ export async function extractPlotDataArray(
   plotType: PlotType,
   accessToken?: string,
 ) {
-  const { colorPalette } = getColorPalette(
-    index,
+  const colorPalette = getContrastColorPalette(
+    index % 2 === 0 ? 'even' : 'odd',
+    Math.floor(index / 2),
```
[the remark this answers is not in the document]
jay-hodgson: Nice catch

jay-hodgson (approved):

PR #694, opened 2024-02-07
PORTALS-2919: add redirect from DataAccess/Instructions to /Data Access

jay-hodgson:


jay-hodgson:
Also spent time verifying all of the other client redirects for ADKP.  Found one that was no longer functional (went to /Explore/Explore/target), so I fixed it while I was here

PR #697, opened 2024-02-07
PORTALS-2670 - Use Synapse theme/config for PersonalAccessToken, OAuth2 Client pages

nickgros:
- Store SourceAppConfig in context to apply a particular SourceApp + Theme to an entire subtree
- Use synapse.org theme for OAuthClientManagementPage and PersonalAccessTokensPage
- Clean up related effects between `useSourceApp` and `AppInitializer` and put them in the new `SourceAppProvider

### DIFF
--- a/apps/SageAccountWeb/src/App.tsx
+++ b/apps/SageAccountWeb/src/App.tsx
@@ -17,9 +17,9 @@ import {
   ApplicationSessionManager,
   CookiesNotification,
   SynapseClient,
-  SynapseToastContainer,
-  SynapseContextType,
   SynapseContextConsumer,
+  SynapseContextType,
+  SynapseToastContainer,
 } from 'synapse-react-client'
 import { getSearchParam } from './URLUtils'
 import './App.scss'
@@ -30,6 +30,7 @@ import TwoFactorAuthEnrollmentPage from './components/TwoFactorAuth/TwoFactorAut
 import TwoFactorAuthBackupCodesPage from './components/TwoFactorAuth/TwoFactorAuthBackupCodesPage'
 import { PersonalAccessTokensPage } from './components/PersonalAccessTokensPage'
 import { OAuthClientManagementPage } from './components/OAuthClientManagementPage'
+import { SourceAppProvider } from './components/useSourceApp'
 
 const isCodeSearchParam = getSearchParam('code') !== undefined
 const isProviderSearchParam = getSearchParam('provider') !== undefined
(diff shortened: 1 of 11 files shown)


apps/SageAccountWeb/src/App.tsx
```diff
@@ -51,131 +52,143 @@ function App() {
         <StyledEngineProvider injectFirst>
           <QueryClientProvider client={queryClient}>
             <ApplicationSessionManager>
-              <AppInitializer>
-                <CookiesNotification />
-                <Switch>
-                  <Route exact path="/">
+              <SourceAppProvider>
```
nickgros, line 55: Diff is bad, the only change here is that I added this `SourceAppProvider` to wrap `AppInitializer`

apps/SageAccountWeb/src/AppInitializer.tsx
```diff
@@ -16,56 +11,13 @@ import { AppContextProvider } from './AppContext'
...
 
 function AppInitializer(props: { children?: React.ReactNode }) {
-  const [appId, setAppId] = useState<string>()
   const [redirectURL, setRedirectURL] = useState<string>()
-  const [theme, setTheme] = useState<Theme>(
-    createTheme(
-      SynapseTheme.defaultMuiThemeOptions,
-      sageAccountWebThemeOverrides,
-    ),
-  )
```
nickgros, line 26: Theme, appId now managed by `SourceAppProvider`

apps/SageAccountWeb/src/AppInitializer.tsx
```diff
@@ -16,56 +11,13 @@ import { AppContextProvider } from './AppContext'
...
-      localStorage.setItem('sourceAppId', searchParamAppId)
-      setAppId(searchParamAppId)
-    } else if (localStorageAppId) {
-      setAppId(localStorageAppId)
-    } else {
-      // fallback to Sage Bionetworks
-      localStorage.setItem('sourceAppId', 'SAGE')
-      setAppId('SAGE')
-    }
-  }, [])
```
nickgros, line 46: This hook updated localStorage if the URL had the `appId` query param, and also set local state here. I removed the local state and moved the effect logic to `SourceAppProvider`.

apps/SageAccountWeb/src/AppInitializer.tsx
```diff
@@ -16,56 +11,13 @@ import { AppContextProvider } from './AppContext'
...
-
-  const sourceApp = useSourceApp(appId)
-
-  useEffect(() => {
-    if (sourceApp?.friendlyName) {
-      currentSourceAppNameState.set(sourceApp.friendlyName)
-    }
-  }, [currentSourceAppNameState, sourceApp?.friendlyName])
```
nickgros, line 54: This effect set a localStorage value that we would use to inform you you could use your Sage account thaat you used to log into a different app. Moved to `SourceAppProvider`

apps/SageAccountWeb/src/AppInitializer.tsx
```diff
@@ -16,56 +11,13 @@ import { AppContextProvider } from './AppContext'
...
-        createTheme(
-          SynapseTheme.defaultMuiThemeOptions,
-          sageAccountWebThemeOverrides,
-          {
-            palette: sourceApp.palette,
-          },
-        ),
-      )
-    }
-  }, [sourceApp?.appId])
```
nickgros, line 68: Moved this logic to `SourceAppProvider`, though it didn't need to be an effect

apps/SageAccountWeb/src/components/ProfileAvatar.tsx
```diff
@@ -36,18 +36,18 @@ export const ProfileAvatar = (props: ProfileAvatarProps) => {
...
       SynapseClient.getFileHandleByIdURL(
-        userProfile?.profilePicureFileHandleId as string,
+        userProfile.profilePicureFileHandleId,
         accessToken,
       ).then(picUrl => {
         setProfilePicUrl(picUrl)
       })
     }
-  }, [userProfile])
+  }, [accessToken, userProfile])
```
nickgros, line 45: Bug where if you logged out and back in, your profile picture would not show up

apps/SageAccountWeb/src/components/SageResourcesPage.tsx
```diff
@@ -5,15 +5,17 @@ import { useHistory } from 'react-router-dom'
...
+} from './useSourceApp'
 
-export type SageResourcesPageProps = {}
-
-export const SageResourcesPage = (props: SageResourcesPageProps) => {
+export function SageResourcesPageInternal() {
   const history = useHistory()
   const theme = useTheme()
   const sourceAppConfigs = useSourceAppConfigs()
-  const sageSourceAppConfig = useSourceApp('SAGE')
```
nickgros, line 16: This is how we got a custom app config before, but it just updated branding components and did not update the theme, which is why I added the context object.

apps/SageAccountWeb/src/components/useSourceApp.tsx
```diff
@@ -1,25 +1,135 @@
...
+const SOURCE_APP_ID_LOCALSTORAGE_KEY = 'sourceAppId'
+export const DEFAULT_SOURCE_APP_ID = 'SAGE'
+export const SYNAPSE_SOURCE_APP_ID = 'synapse.org'
+
+/**
+ * This must be exported to use the context in class components.
+ */
+export const SourceAppContext = React.createContext<SourceAppContextType>(
+  STATIC_SOURCE_APP_CONFIG,
+)
```
nickgros, line 26: Use context so we an apply a SourceApp/theme to an entire subtree

apps/SageAccountWeb/src/components/useSourceApp.tsx
```diff
@@ -1,25 +1,135 @@
...
+    SOURCE_APP_ID_LOCALSTORAGE_KEY,
+  )
+
+  useEffect(() => {
+    if (idFromLocalStorage && sourceApp?.friendlyName) {
+      currentSourceAppNameState.set(sourceApp.friendlyName)
+    }
+  }, [idFromLocalStorage, currentSourceAppNameState, sourceApp?.friendlyName])
+}
+
```
nickgros, line 66: Moved + commented the effect that set the `lastLoginInfoState` value, which the Login component will use to tell you that you might already have a Sage account that you used for a different app

apps/SageAccountWeb/src/components/useSourceApp.tsx
```diff
@@ -1,25 +1,135 @@
...
+    </SourceAppContext.Provider>
+  )
+}
+
+export function useSourceApp(): SourceAppContextType {
+  const context = useContext(SourceAppContext)
+  if (context === undefined) {
+    console.error('useSourceApp must be used within a SourceAppContextProvider')
+    return STATIC_SOURCE_APP_CONFIG
   }
```
nickgros, line 133: This should never happen

packages/synapse-react-client/src/components/OAuthClientManagement/OAuthManagement.tsx
```diff
@@ -57,6 +58,7 @@ export const OAuthManagement: React.FunctionComponent = () => {
           setIsEdit(false)
         }}
         sx={{ float: 'right' }}
+        startIcon={<AddCircleTwoTone />}
```
nickgros, line 61: Small changes to match design -- add an icon to the button, button text should not be allcaps

apps/SageAccountWeb/src/components/useSourceApp.tsx
```diff
@@ -1,25 +1,135 @@
...
+/**
+ * Synchronizes the current state of the `currentSourceAppName` localStorage value.
+ * The value may be used at a later date in the Login component to inform a user if they have a OneSage account
+ * that was used for a different application.
+ */
+export function useSynchronizeCurrentSourceAppNameForLogin(
+  sourceApp?: SourceAppConfig,
+) {
+  const { currentSourceAppNameState } = useLastLoginInfoState()
+  const { value: idFromLocalStorage } = useLocalStorageValue(
```
hallieswan, line 56: Could consider using the same name for the value as the previous function (e.g. `localStorageAppId`), but totally up to you!

apps/SageAccountWeb/src/components/useSourceApp.tsx
```diff
@@ -1,25 +1,135 @@
...
+    }
+  }, [idFromLocalStorage, currentSourceAppNameState, sourceApp?.friendlyName])
+}
+
+/**
+ * Provides context necessary for most components in SRC.
+ *
+ * The SynapseContextProvider must be wrapped in a react-query QueryClientProvider.
+ * @param props sourceAppId: An optional source app ID to use. If not provided, the source app ID will be read from localStorage, or default to 'SAGE'.
+ */
```
hallieswan, line 72: Maybe a copy/paste issue -- should this comment be updated for `SourceAppProvider`? e.g. something like -- 

```suggestion
/**
 * Provides source app context necessary for most components in SageAccountWeb.
 *
 * The SourceAppProvider must be wrapped in a react-query QueryClientProvider.
 * @param props sourceAppId: An optional source app ID to use. If not provided, the source app ID will be read from localStorage, or default to 'SAGE'.
 */
```

apps/SageAccountWeb/src/components/useSourceApp.tsx
hallieswan: Looks like `idFromLocalStorage` is also used in `SourceAppProvider` so could consider changing `useConfigureSourceAppFromQueryParams` to match the other functions instead. Up to you!

apps/SageAccountWeb/src/components/useSourceApp.tsx
```diff
@@ -1,25 +1,135 @@
...
+      {/* The theme is determined by the sourceApp, so render the ThemeProvider here*/}
+      <ThemeProvider theme={theme}>{children}</ThemeProvider>
+    </SourceAppContext.Provider>
+  )
+}
+
+export function useSourceApp(): SourceAppContextType {
+  const context = useContext(SourceAppContext)
+  if (context === undefined) {
+    console.error('useSourceApp must be used within a SourceAppContextProvider')
```
hallieswan, line 131:
```suggestion
    console.error('useSourceApp must be used within a SourceAppProvider')
```

hallieswan (approved):
Looks good to me! Just a couple small suggestions

apps/SageAccountWeb/src/components/useSourceApp.tsx
nickgros: Yes, every time I create a new context I copy this boilerplate but forget to change the comments 😬

Thanks for catching

PR #700, opened 2024-02-07
PORTALS-2858: in progress

jay-hodgson:


packages/synapse-react-client/src/synapse-queries/entity/useExportToCavatica.ts
```diff
@@ -23,7 +23,9 @@ export function useExportToCavatica(
   return async () => {
     try {
       // add drs_uri to select
-      const selectColumnsList = selectColumns?.map(col => col.name).join(',')
+      const selectColumnsList = selectColumns
+        ?.map(col => `"${col.name}"`)
```
jay-hodgson, line 27: Some column names have spaces, so surround with double quotes

jay-hodgson:
Verified sending files to the download cart from a materialized view works when pointing to the staging stack.  And exporting files to CAVATICA works properly (import from CAVATICA works!)

nickgros (approved):

PR #701, opened 2024-02-07
SWC-6685 - Turn on `liveValidate` for entities with present annotations + schema

nickgros:
SWC-6685

In SchemaDrivenAnnotationEditor

- turn on liveValidate if annotations and schema are present 
- Adjust some spacing values
- Separate out behavior to hide form actions into separate prop


packages/synapse-react-client/src/components/SchemaDrivenAnnotationEditor/SchemaDrivenAnnotationEditor.tsx
```diff
@@ -60,16 +61,18 @@ export type SchemaDrivenAnnotationEditorProps = {
   schemaId?: string
   /** May be used to directly provide a JSON Schema to use for the form */
   validationSchema?: JSONSchema7
-  /** Optionally supply a ref to the form to handle submission externally with `formRef.current.submit()`. If provided, the editor will not render its own submit UI. */
+  /** Optionally supply a ref to the form to handle submission externally with `formRef.current.submit()`. */
   formRef?: React.RefObject<RJSF>
-  /** Provide live input validation. This can cause major performance degradation. */
+  /** Provide live input validation. This can cause performance degradation. By default, liveValidate will be true if an entity with a schema and existing annotations is being edited */
   liveValidate?: boolean
```
nickgros, line 67: The original issue pointed out how confusing it is to see that an entity has invalid annotations, but you can't actually see which ones are invalid until you click 'Save'!

We can make this more user-friendly by turning on `liveValidate`. Since we started using this library, the performance of `liveValidate` has gotten significantly better due to improvements in RJSF.

If this does become a performance bottleneck, we can revisit and consider other strategies.

packages/synapse-react-client/src/components/SchemaDrivenAnnotationEditor/SchemaDrivenAnnotationEditor.tsx
```diff
@@ -60,16 +61,18 @@ export type SchemaDrivenAnnotationEditorProps = {
...
   /** Invoked after a successful form submission */
   onSuccess?: () => void
   /** If defined and formRef is not supplied, shows a 'Cancel' button and runs this effect on click */
   onCancel?: () => void
   /** Passes new form data upon each change to the form */
   onChange?: (annotations: Record<string, unknown>) => void
+  /** If true, the editor will not render its own submit UI. */
+  hideActions?: boolean
```
nickgros, line 75: This behavior was based on supplying a `formRef`, but really it should be a separate prop

packages/synapse-react-client/src/components/SchemaDrivenAnnotationEditor/AnnotationEditorUtils.ts
```diff
@@ -133,3 +134,17 @@ export function convertToArray<T>(value: T): Array<unknown> {
...
+ * @param existingAnnotations
+ * @param validationSchema
+ */
+export function shouldLiveValidate(
+  existingAnnotations: Record<string, unknown> | undefined,
+  validationSchema: JSONSchema7 | undefined,
+): boolean {
+  const hasExistingAnnotations =
+    existingAnnotations && Object.keys(existingAnnotations).length > 0
+  return Boolean(hasExistingAnnotations && validationSchema)
```
nickgros, line 149: If the entity doesn't start with annotations, or if it doesn't have a custom schema, we should not turn on liveValidate (i.e. validate only at the end).

packages/synapse-react-client/src/components/SchemaDrivenAnnotationEditor/SchemaDrivenAnnotationEditor.tsx
```diff
@@ -259,127 +266,125 @@ export function SchemaDrivenAnnotationEditor(
               </Box>
             </Alert>
           )}
```
nickgros, line 268: Bad diff below, I will call out the changes

packages/synapse-react-client/src/components/SchemaDrivenAnnotationEditor/SchemaDrivenAnnotationEditor.tsx
```diff
@@ -259,127 +266,125 @@ export function SchemaDrivenAnnotationEditor(
...
-          {formData != undefined && (
```
nickgros, line 262: Hiding the form until we had `formData` was causing issues in Storybook where stories where we rendered the form with a schema but no entity failed to appear.

I think this condition was part of a workaround for another issue that has since been fixed.

packages/synapse-react-client/src/components/SchemaDrivenAnnotationEditor/SchemaDrivenAnnotationEditor.tsx
```diff
@@ -259,127 +266,125 @@ export function SchemaDrivenAnnotationEditor(
...
-          )}
+              }
+              setShowSubmissionError(false)
+              setFormData(formData)
+              submitChangedEntity()
+            }}
+            onError={(errors: RJSFValidationError[]) => {
+              // invoked when submit is clicked and there are client-side validation errors
+              setValidationError(errors)
+              if ((validationError || liveValidate) && entityId) {
```
nickgros, line 348: If `liveValidate` is off, and the annotations are invalid against the schema, clicking this once triggers validation, then clicking it again opens a confirmation modal to save the invalid annotations.

With this change, if `liveValidate` is on, then the user can already see the errors, so we should just 'skip' the first click and go straight to the confirmation modal.

packages/synapse-react-client/src/components/SchemaDrivenAnnotationEditor/template/FieldDescriptionTable.tsx
```diff
@@ -41,7 +41,7 @@ export default function FieldDescriptionTable(
 
   return (
     <Collapse className="field-description" in={true}>
-      <table className="FieldDescriptionTable">
+      <Box component="table" className="FieldDescriptionTable" mb={1}>
```
nickgros, line 44: Tweaking some spacing

packages/synapse-react-client/src/components/SchemaDrivenAnnotationEditor/widget/SelectWidget.tsx
```diff
@@ -9,15 +9,15 @@ const filter = createFilterOptions<EnumOptionsType>()
...
  */
 export function findValueOption(
   value: string | null | undefined,
   options: EnumOptionsType[],
-): EnumOptionsType | undefined {
+): EnumOptionsType | string | null {
   if (value == null) {
-    return undefined
+    return null
   }
```
nickgros, line 21: This was causing issues where MUI `Autocomplete` components were switching from controlled <--> uncontrolled, which triggered console warnings

jay-hodgson: How widespread is this kind of issue?

jay-hodgson (approved):

packages/synapse-react-client/src/components/SchemaDrivenAnnotationEditor/widget/SelectWidget.tsx
nickgros: This particular bit of code was only exposed in a few places where we were using RJSF + the QuerySortSelector component used for portals' CardContainer, so not very widespread.

The general "switching between controlled/uncontrolled" warning issue could be happening in many more places, though.

PR #704, opened 2024-02-09
PORTALS-2916 - Temporarily simplify the colors used in facet plots until the palette is improved

nickgros:
Temporary improvement for PORTALS-2916

Also fix an issue where facets would appear full-width instead of in a grid


packages/synapse-react-client/src/components/widgets/facet-nav/PlotsContainer.tsx
```diff
@@ -257,7 +257,7 @@ export default function PlotsContainer(props: PlotsContainerProps) {
               showPlotVisualization ? '' : 'hidden'
             } ${showMoreButtonState === 'LESS' ? 'less' : ''}`}
           >
-            <div className="PlotNav__row" role="list">
+            <div className="PlotsContainer__row" role="list">
```
nickgros, line 260: This class was erroneously renamed, causing it to not get the appropriate styles from CSS

jinjunoh (approved):

jay-hodgson (approved):

PR #706, opened 2024-02-09
Fix issue where pagination did not work for the AccessApprovalsTable

nickgros:
Fix a case where `fetchNextPage` was never called, and fixed tests that made it seem like this worked.

I don't think this significantly impacted production, I think few users (if any) have enough approvals to hit the page limit (which I think is 50).

### DIFF
--- a/packages/synapse-react-client/src/components/dataaccess/AccessApprovalsTable.tsx
+++ b/packages/synapse-react-client/src/components/dataaccess/AccessApprovalsTable.tsx
@@ -140,13 +140,13 @@ export const AccessApprovalsTable: React.FunctionComponent<
           No Results
         </Typography>
       )}
-      {!hasNextPage ? (
-        ''
-      ) : (
+      {hasNextPage && (
         <Button
           variant="outlined"
           color="primary"
-          onClick={() => fetchNextPage}
+          onClick={() => {
+            fetchNextPage()
+          }}
         >
           Show More
         </Button>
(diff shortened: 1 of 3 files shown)


packages/synapse-react-client/src/mocks/MockAccessApprovals.ts
```diff
@@ -5,28 +5,25 @@ import {
...
+      state: ApprovalState.APPROVED,
+      submitterId: MOCK_USER_ID.toString(),
+    },
+  ],
   nextPageToken: '50a0',
 } satisfies AccessApprovalSearchResponse
 
 export const mockAccessApprovalSearchResult2: AccessApprovalSearchResult = {
-  accessRequirementId: '9602629',
+  accessRequirementId: '9602630',
```
nickgros, line 26: The bug was masked by a bad test because the test checked for this ID, which was the same for search results on both pages!

packages/synapse-react-client/src/components/dataaccess/AccessApprovalsTable.tsx
```diff
@@ -140,13 +140,13 @@ export const AccessApprovalsTable: React.FunctionComponent<
...
-        ''
-      ) : (
+      {hasNextPage && (
         <Button
           variant="outlined"
           color="primary"
-          onClick={() => fetchNextPage}
+          onClick={() => {
+            fetchNextPage()
+          }}
```
nickgros, line 149: `fetchNextPage` was never called, so future pages of data could never be fetched

packages/synapse-react-client/src/mocks/MockAccessApprovals.ts
```diff
@@ -5,28 +5,25 @@ import {
 } from '@sage-bionetworks/synapse-types'
 import { MOCK_USER_ID } from './user/mock_user_profile'
 
-const mockSearchResults = []
-for (let i = 0; i < 50; i++) {
```
nickgros, line 9: We didn't really need 50 results

jay-hodgson (approved):

PR #709, opened 2024-02-13
SWC-6615

jay-hodgson:


packages/synapse-react-client/src/components/dataaccess/AccessRequirementTable.tsx
```diff
@@ -73,16 +76,27 @@ export function AccessRequirementTable(props: AccessRequirementTableProps) {
   })
 
   const searchRequest: Omit<AccessRequirementSearchRequest, 'nextPageToken'> =
-    useMemo(
-      () => ({
+    useMemo(() => {
+      // SWC-6615: If the input string is a single integer, assume it's the AR ID.  Otherwise use as the nameContains field.
```
jay-hodgson, line 80: Here's the core of the PR.  If the input string looks like a single integer, then treat it as an AR ID.  Otherwise send it as the nameContains value (as it did before).

nickgros (approved):

PR #711, opened 2024-02-14
SWC-6545 pre-work - Upgrade Storybook + expose ability to enable StrictMode

nickgros:
- Upgrade Storybook to latest version (7.6.15)
- Expose ability to enable React StrictMode in Storybook
- Replace deprecated @storybook/addon-styling with @storybook/addon-themes
- Replace deprecated storybook-addon-designs with @storybook/addon-designs

I was trying to upgrade react-query to v5 (SWC-6545) and found that our Challenge Portal components relied on features that were removed from react-query (e.g. onSuccess/onSettled callbacks for useQuery). The features were removed because they were often used improperly, leading to bugs. The Challenge Portal components indeed used these callbacks improperly, so I was working to fix them and additionally remove unnecessary `useEffect` hooks.

To attempt to identify new issues with the effect hooks, I enabled StrictMode. However, StrictMode did not work properly in Storybook, with a fix in v7.6.13 (https://github.com/storybookjs/storybook/issues/22472), so this upgrade is necessary to enable StrictMode.


packages/synapse-react-client/.storybook/main.ts
```diff
@@ -10,15 +10,16 @@ const config: StorybookConfig = {
...
       options: {},
     },
-    'storybook-addon-designs',
+    '@storybook/addon-designs',
   ],
 
   framework: {
     name: '@storybook/react-vite',
     options: {
+      strictMode: true,
```
nickgros, line 22:
```suggestion
      strictMode: false,
```

packages/synapse-react-client/.storybook/main.ts
```diff
@@ -10,15 +10,16 @@ const config: StorybookConfig = {
...
       options: {},
     },
-    'storybook-addon-designs',
+    '@storybook/addon-designs',
   ],
 
   framework: {
     name: '@storybook/react-vite',
     options: {
+      strictMode: false,
```
nickgros, line 22: I considered enabling StrictMode by default, but wasn't sure if that might cause problems with stories. Explicitly setting it here makes it easy to go in and change it locally when desired. In the future, we can decide to enable it globally.

jay-hodgson (approved):

PR #714, opened 2024-02-16
SWC-6545, PORTALS-2936 - Type improvements

nickgros:
- Add various types for request objects that are used in Challenge Portal components.

This is part of an incremental set of changes that are designed to 
- Fix bugs and improve maintainability of the Challenge Portal components
- Allow us to upgrade react-query to v5

### DIFF
--- a/packages/synapse-types/src/MembershipInvitation.ts
+++ b/packages/synapse-types/src/MembershipInvitation.ts
@@ -2,10 +2,16 @@
 export type MembershipInvitation = {
   id: string //The id of the MembershipInvitation.
   teamId: string //The id of the Team which the user is invited to join.
-  inviteeId: string //The principal ID of the user being invited to join the Team.
-  inviteeEmail: string //The email address of the user being invited to join the Team.
-  message: string //The invitation message (optional).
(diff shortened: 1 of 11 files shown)


packages/synapse-types/src/MembershipInvitation.ts
```diff
@@ -2,10 +2,16 @@
...
+  inviteeId?: string //The principal ID of the user being invited to join the Team.
+  inviteeEmail?: string //The email address of the user being invited to join the Team.
+  message?: string //The invitation message (optional).
+  expiresOn?: string //The date this invitation expires (optional).
   createdOn: string //The date this MembershipInvitation was created.
   createdBy: string //The ID of the user that created this MembershipInvitation.
 }
+
+// The fields that should be included in the request to create a new MembershipInvitation
+export type CreateMembershipInvitationRequest = Pick<
```
jay-hodgson, line 14: That's convenient!  I imagine that if we have auto-generated models from the Synapse OpenAPI spec, these are types that we'll want to continue to define.

jay-hodgson (approved):

PR #720, opened 2024-02-19
PORTALS-2936 - Update AccessRequirementList to accept subjectId and `subjectType` props, deprecate `entityId` and `teamId` props

nickgros:
- Update AccessRequirementList to only render dialog components when `renderAsModal` is true. This supports PORTALS-2936 because, in a later change, we can create a new dialog that shows the contents of the AccessRequirementList, but with custom actions that are relevant to registering for a challenge.

- Update AccessRequirementList to accept subjectId and `subjectType` props, deprecate `entityId` and `teamId` props. The component(s) implicitly relied on an `entityId` in a few cases. This is a more appropriate signature since access requirements can be used to restrict a variety of different objects.


packages/synapse-react-client/src/components/AccessRequirementList/AccessRequirementList.tsx
```diff
@@ -44,20 +28,48 @@ import {
...
-import { useSynapseContext } from '../../utils/context/SynapseContext'
+import { useSynapseContext } from '../../utils'
 import { useCanShowManagedACTWikiInWizard } from './AccessRequirementListUtils'
-import { DialogActions } from '@mui/material'
+import { noop } from 'lodash-es'
+import { DialogBase } from '../DialogBase'
+import UserOrTeamBadge from '../UserOrTeamBadge'
+import { EntityLink } from '../EntityLink'
 
 export type AccessRequirementListProps = {
```
nickgros, line 38: Summary of changes to the props:

- Caller must specify one of (subjectId AND subjectType), entityId, teamId
- entityId, teamId are deprecated
- Made `onHide` optional, since it only make sense when `renderAsModal` is true

packages/synapse-react-client/src/components/AccessRequirementList/AccessRequirementList.tsx
```diff
@@ -127,15 +139,30 @@ export default function AccessRequirementList(
...
+      : props.teamId
+
+  const subjectType =
+    'subjectType' in props
+      ? props.subjectType
+      : isShowingRequirementsForEntity
+      ? RestrictableObjectType.ENTITY
+      : RestrictableObjectType.TEAM
+
+  let { dialogTitle = 'Data Access Request' } = props
```
nickgros, line 165: Instantiate `dialogTitle` with `let` so we can update it as needed.

packages/synapse-react-client/src/components/AccessRequirementList/AccessRequirementList.tsx
```diff
@@ -148,30 +175,27 @@ export default function AccessRequirementList(
     Request | Renewal | undefined
   >()
 
-  const entityHeaderProps: UseGetInfoFromIdsProps<EntityHeader> = {
-    ids: [entityId],
-    type: 'ENTITY_HEADER',
-  }
   const canShowManagedACTWikiInWizard = useCanShowManagedACTWikiInWizard()
 
-  const entityInformation = useGetInfoFromIds<EntityHeader>(entityHeaderProps)
```
nickgros, line 157: Use react-query instead of this old custom cache

packages/synapse-react-client/src/components/AccessRequirementList/AccessRequirementList.tsx
```diff
@@ -243,64 +267,53 @@ export default function AccessRequirementList(
...
-        >
-          {entityInformation[0]?.name}
-        </Link>
-      </>
-    )
-  }, [entityId, entityInformation, numberOfFilesAffected, requestObjectName])
+    if (subjectType === RestrictableObjectType.ENTITY) {
+      return <EntityLink entity={subjectId} />
+    } else if (subjectType === RestrictableObjectType.TEAM) {
+      return <UserOrTeamBadge principalId={subjectId} />
```
nickgros, line 273: Use existing renderers for the restricted object

packages/synapse-react-client/src/components/AccessRequirementList/AccessRequirementList.tsx
```diff
@@ -243,64 +267,53 @@ export default function AccessRequirementList(
...
+    } else {
+      console.warn(
+        `Unhandled case for AccessRequirementList requestDetails: ${subjectType}: ${subjectId}`,
+      )
+      return <></>
```
nickgros, line 278: At the moment, the only remaining case in [RestrictableObjectType](https://rest-docs.synapse.org/rest/org/sagebionetworks/repo/model/RestrictableObjectType.html) is EVALUATION--I've never heard of an AR on an evaluation (since it's usually applied to the participant team. Do we need to address this ASAP?

packages/synapse-react-client/src/components/AccessRequirementList/AccessRequirementList.tsx
```diff
@@ -243,64 +267,53 @@ export default function AccessRequirementList(
...
+    }
+  }, [numberOfFilesAffected, requestObjectName, subjectId, subjectType])
 
   const content = (
     <>
```
nickgros, line 283: Remove `DialogTitle` and `DialogContent` from `content`. These will now be handled by our `DialogBase` component, and the wrapper components will not appear when `renderAsModal` is false.

packages/synapse-react-client/src/components/AccessRequirementList/ManagedACTAccessRequirementRequestFlow/ManagedACTAccessRequirementItem.tsx
```diff
@@ -17,7 +16,6 @@ import {
 import { RequirementItemStatus } from '../AccessApprovalCheckMark'
 
 export type ManagedACTAccessRequirementItemProps = {
-  entityId: string
```
nickgros, line 20: Unused

packages/synapse-react-client/src/components/AccessRequirementList/AccessRequirementList.tsx
jay-hodgson: Please ignore the EVALUATION subject type for now (no known use case)

packages/synapse-react-client/src/components/AccessRequirementList/AccessRequirementListItem.tsx
```diff
@@ -39,15 +42,15 @@ export function AccessRequirementListItem(
...
+          subjectId={subjectId}
+          subjectType={subjectType}
         />
       )
     case MANAGED_ACT_ACCESS_REQUIREMENT_CONCRETE_TYPE_VALUE:
       return (
         <ManagedACTAccessRequirementItem
           accessRequirement={accessRequirement}
           onHide={onHide}
-          entityId={entityId}
```
jay-hodgson, line 50: Am I missing something?  I thought the subjectId (and type) would be passed into the ManagedACTAccessRequirementItem as well

packages/synapse-react-client/src/components/AccessRequirementList/ManagedACTAccessRequirementRequestFlow/ManagedACTAccessRequirementItem.tsx
jay-hodgson: Oh I see!

packages/synapse-react-client/src/components/AccessRequirementList/AccessRequirementListItem.tsx
nickgros: The prop was unused, so I removed it. The relationship between the AR and the subject is handled in the `onRequestAccess` callback.

packages/synapse-react-client/src/components/AccessRequirementList/ManagedACTAccessRequirementRequestFlow/ManagedACTAccessRequirementItem.tsx
jay-hodgson: We already have the AR, so we only need it in the other case because we will send the user to the Synapse.org page with this legacy AR.  Got it

jay-hodgson (approved):

PR #728, opened 2024-02-20
PORTALS-2952

jay-hodgson:


apps/portals/src/Navbar.tsx
```diff
@@ -365,7 +365,7 @@ function Navbar() {
                     <React.Fragment key={`${topLevelTo}-${topLevelIndex}`}>
                       {el.routes &&
                         el.routes.map((route, index) => {
-                          const { path: to, link } = route
+                          const { path: to, link, target } = route
```
jay-hodgson, line 368: Respect target in dropdown nav anchor (if set)

nickgros (approved):

PR #755, opened 2024-02-28
SWC-6698: reduce debounce delay, but reset the debounce when hovering…

jay-hodgson:
… over facet enum values


jay-hodgson:
Also, the checkbox label does not seem to be aligned with the checkbox.
This is on ADKP:
Screen Shot 2024-02-28 at 10 59 45 AM

Local output, with line height:
Screen Shot 2024-02-28 at 11 50 29 AM

nickgros (approved):

nickgros:
> Also, the checkbox label does not seem to be aligned with the checkbox.

@jay-hodgson is this an artifact of the font change?

PR #756, opened 2024-02-28
SWC-6669: create EntityViewScopeEditorModal for Editing EntityView scope/mask

jinjunoh:


packages/synapse-react-client/src/components/EntityHeaderTable/EntityHeaderTable.tsx
```diff
@@ -229,7 +231,10 @@ export const EntityHeaderTable = (props: EntityHeaderTableProps) => {
     const newDataEntityIds = new Set()
     newData.map(entityHeader => newDataEntityIds.add(entityHeader.id))
     const missingRefs = refsInState.filter(
-      ref => !newDataEntityIds.has(ref.targetId),
+      ref =>
+        !newDataEntityIds.has(
+          ref.targetId.startsWith('syn') ? ref.targetId : `syn${ref.targetId}`,
```
nickgros, line 236: Can you extract this to a utility function called something like `normalizeSynPrefix`? You could put it in `EntityTypeUtils.ts` unless you can think of a better place

packages/synapse-react-client/src/components/EntityHeaderTable/useEntityHeaderTableState.ts
```diff
@@ -15,6 +15,10 @@ export function useEntityHeaderTableState(
   const [newEntityIDs, _setNewEntityIDs] = useState<string>('')
   const [parseErrors, setParseErrors] = useState<string[]>([])
 
+  useEffect(() => {
+    _setRefsInState(cloneDeep(references))
+  }, [references])
```
nickgros, line 20: Just want a comment here for clarity
```suggestion
  /* Reset state if the `references` argument changes */
  useEffect(() => {
    _setRefsInState(cloneDeep(references))
  }, [references])
```

packages/synapse-react-client/src/components/EntityHeaderTable/useEntityHeaderTableState.ts
```diff
@@ -25,6 +29,7 @@ export function useEntityHeaderTableState(
 
   const setRefsInState = useCallback(
     (refs: ReferenceList) => {
+      console.log('setRefsInState called')
```
nickgros, line 32:
```suggestion
```

packages/synapse-react-client/src/components/EntityViewScopeEditor/EntityViewScopeEditorModal.test.tsx
```diff
@@ -0,0 +1,235 @@
...
+} from '@sage-bionetworks/synapse-types'
+import { MOCK_ACCESS_TOKEN } from '../../mocks/MockSynapseContext'
+import { SynapseClientError } from '../../utils'
+import { mockProjectViewEntity } from '../../mocks/entity/mockProjectView'
+
+const mockTableEntityInstance = mockTableEntityData.entity
+
+jest.mock('../EntityViewScopeEditor/EntityViewScopeEditor', () => ({
+  __esModule: true,
+  default: jest.fn(() => <div data-testid={'EntityViewScopeEditorrMocked'} />),
```
nickgros, line 28:
```suggestion
  default: jest.fn(() => <div data-testid={'EntityViewScopeEditorMocked'} />),
```

packages/synapse-react-client/src/components/EntityViewScopeEditor/EntityViewScopeEditorModal.test.tsx
```diff
@@ -0,0 +1,235 @@
...
+    jest.spyOn(SynapseClient, 'getEntity').mockResolvedValue({
+      ...mockTableEntity,
+      scopeIds: ['syn123'],
+      viewTypeMask: ENTITY_VIEW_TYPE_MASK_FILE | ENTITY_VIEW_TYPE_MASK_FOLDER,
+      concreteType: ENTITY_VIEW_CONCRETE_TYPE_VALUE,
+    } as Entity)
+  })
+
+  it('displays the correct scope editor and mask editor after initial render', async () => {
+    jest.clearAllMocks()
```
nickgros, line 54: Please move the `clearAllMocks` call to the `beforeEach` arrow fn

packages/synapse-react-client/src/components/EntityViewScopeEditor/EntityViewScopeEditorModal.test.tsx
```diff
@@ -0,0 +1,235 @@
...
+    })
+
+    const dialog = screen.getByRole('dialog')
+    expect(
+      within(dialog).getByRole('heading', { name: 'Edit Scope' }),
+    ).toBeVisible()
+    expect(within(dialog).getByRole('button', { name: 'Cancel' })).toBeVisible()
+    expect(within(dialog).getByRole('button', { name: 'Save' })).toBeVisible()
+
+    const filesButton = within(dialog).getByRole('checkbox', { name: 'Files' })
```
nickgros, line 69: Prefer renaming these `filesCheckbox` or something like that; these are not buttons

packages/synapse-react-client/src/components/EntityViewScopeEditor/EntityViewScopeEditorModal.test.tsx
```diff
@@ -0,0 +1,235 @@
...
+
+    await waitFor(() => {
+      expect(filesButton).toBeChecked()
+      expect(foldersButton).toBeChecked()
+      expect(tablesButton).not.toBeChecked()
+      expect(datasetsButton).not.toBeChecked()
+    })
+  })
+
+  it('display error for invalid entity', async () => {
```
nickgros, line 96: Rename test for clarity; there's no validity issue with the entity here
```suggestion
  it('displays error response returned by update call', async () => {
```

packages/synapse-react-client/src/components/EntityViewScopeEditor/EntityViewScopeEditorModal.test.tsx
```diff
@@ -0,0 +1,235 @@
...
+    )
+
+    const saveButton = within(dialog).getByRole('button', { name: 'Save' })
+    await waitFor(() => {
+      expect(
+        within(dialog).getByRole('button', { name: 'Save' }),
+      ).not.toBeDisabled()
+    })
+    await userEvent.click(saveButton)
+    expect(mockOnUpdate).not.toHaveBeenCalled()
```
nickgros, line 121: I would expect that you would waitFor an alert to appear here, verify that the error message is shown, then I would expect mockOnUpdate to have been called (since that's what threw the error)--without that, I am not sure what you are testing here

packages/synapse-react-client/src/components/EntityViewScopeEditor/EntityViewScopeEditorModal.test.tsx
```diff
@@ -0,0 +1,235 @@
...
+      viewTypeMask: ENTITY_VIEW_TYPE_MASK_PROJECT,
+      concreteType: ENTITY_VIEW_CONCRETE_TYPE_VALUE,
+    } as Entity)
+
+    renderComponent({
+      entityId: mockProjectViewEntity.id,
+      open: true,
+      onCancel: mockOnCancel,
+      onUpdate: mockOnUpdate,
+    })
```
nickgros, line 139: Not that it matters, since you mocked `getEntity`, to always return it, but I would expect to see `mockProjectViewEntity` instead of `mockTableEntity`--it is kind of weird to pass one ID as props but have your mock service return a completely different entity. It might confuse someone working on this test in the future.

packages/synapse-react-client/src/components/EntityViewScopeEditor/EntityViewScopeEditorModal.test.tsx
```diff
@@ -0,0 +1,235 @@
...
+    const dialog = screen.getByRole('dialog')
+    expect(
+      within(dialog).getByRole('heading', { name: 'Edit Scope' }),
+    ).toBeVisible()
+
+    expect(within(dialog).getByRole('button', { name: 'Cancel' })).toBeVisible()
+    expect(within(dialog).getByRole('button', { name: 'Save' })).toBeVisible()
```
nickgros, line 146: You never change the visibility of these buttons, so I don't think it's necessary to check `toBeVisible()`

Also, to reduce the amount of duplicate code/simplify reading these tests, you can return the buttons in `renderComponent` so you have them available right away. For example see `setUp` in TableColumnSchemaEditor.integration.test.tsx

nickgros: You can do the same thing with the checkboxes if you use `queryByRole` in the `renderComponent` function. You can then do `expect(filesCheckbox).toBeInTheDocument()` or `expect(filesCheckbox).not.toBeInTheDocument()` depending on the test

nickgros (changes requested):

packages/synapse-react-client/src/utils/functions/EntityTypeUtils.ts
```diff
@@ -306,6 +307,15 @@ export function getVersionDisplay(entity: Entity): string {
   }
 }
 
+export function normalizeSynPrefix(
+  ref: Reference,
+  newDataEntityIds: Set<unknown>,
+) {
+  return !newDataEntityIds.has(
+    ref.targetId.startsWith('syn') ? ref.targetId : `syn${ref.targetId}`,
```
nickgros, line 315: Sorry for being unclear, can you just keep this line in the util? The `Set.has` part should stay in the component, since that part is component-specific

nickgros (commented):
Test changes look great! Can you just fix the util?

packages/synapse-react-client/src/utils/functions/EntityTypeUtils.ts
nickgros: I want to reuse this in PORTALS-2971 !

nickgros (approved):

PR #760, opened 2024-02-29
PORTALS-2940 - Allow accepting a team invitation from ChallengeTeamWizard

nickgros:
PORTALS-2940

- Allow joining a challenge team to which you have been invited
- Completely refactor ChallegeTeamWizard
- Rename JoinRequestForm to MembershipRequestForm
- Add OpenMembershipInvitation component to display an invitation to a team

image

image


packages/synapse-react-client/src/components/ChallengeTeamWizard/ChallengeTeamTable.tsx
```diff
@@ -17,26 +22,24 @@ export type ChallengeTeamTableProps = {
...
-const getTeamRow = (team: Team): ChallengeTeamRow => {
-  return {
-    id: team.id,
-    name: team.name,
-    created: formatDate(dayjs(team.createdOn), 'MM/DD/YY'),
-    description: team.description,
-  }
+type RowData = Team & {
+  hasInvitation: boolean
 }
```
nickgros, line 27: The row data included the fetched teams, so just reuse the `Team` type. Additionally, we will fetch open invitations.

packages/synapse-react-client/src/components/ChallengeTeamWizard/ChallengeTeamTable.tsx
```diff
@@ -46,14 +49,17 @@ export default function ChallengeTeamTable(props: ChallengeTeamTableProps) {
...
     const teams =
       teamsList?.list.filter(
         team =>
           team.canPublicJoin ||
           // `canRequestMembership` is true by default if the property is not present
-          team.canRequestMembership !== false,
+          team.canRequestMembership !== false ||
+          openInvitationsToJoinTeams?.some(
+            invitation => invitation.teamId === team.id,
+          ),
```
nickgros, line 62: If there is an open invitation for the registered team, then that team can also be shown

packages/synapse-react-client/src/components/ChallengeTeamWizard/ChallengeTeamTable.tsx
```diff
@@ -102,24 +118,39 @@ export default function ChallengeTeamTable(props: ChallengeTeamTableProps) {
         disableColumnMenu: true,
         renderCell: ({ row }) => (
           <Link
```
nickgros, line 120: Use MUI link instead of react-router link

packages/synapse-react-client/src/components/ChallengeTeamWizard/ChallengeTeamTable.tsx
```diff
@@ -102,24 +118,39 @@ export default function ChallengeTeamTable(props: ChallengeTeamTableProps) {
...
+        hideable: false,
+        disableColumnMenu: true,
+        align: 'center',
+        renderCell: ({ value }) =>
+          value ? (
+            <Tooltip title={'You have been invited to join this team'}>
+              <EmailTwoTone fontSize={'small'} sx={{ color: 'grey.700' }} />
+            </Tooltip>
+          ) : null,
+      },
```
nickgros, line 145: Add a new column to indicate an invitation: 
image

packages/synapse-react-client/src/components/ChallengeTeamWizard/ChallengeTeamTable.tsx
```diff
@@ -102,24 +118,39 @@ export default function ChallengeTeamTable(props: ChallengeTeamTableProps) {
...
       {
         field: 'created',
         headerName: 'Created On',
         width: 100,
         filterable: false,
         hideable: false,
         disableColumnMenu: true,
+        renderCell: ({ value }) => formatDate(dayjs(value), 'MM/DD/YY'),
```
nickgros, line 153: Moved this transformation to the column definition instead of when creating the row objects

packages/synapse-react-client/src/components/ChallengeTeamWizard/ChallengeTeamTable.tsx
```diff
@@ -149,6 +180,7 @@ export default function ChallengeTeamTable(props: ChallengeTeamTableProps) {
         rowCount={5}
         hideFooter
         density="compact"
+        rowBuffer={5} // Buffer ensures all rows required in test are loaded.
```
nickgros, line 183: MUI DataGrid has virtualization enabled by default, which causes problems in JSDOM when it doesn't load all of the rows you expect to see. This prop ensures enough rows are loaded for our tests to work.

packages/synapse-react-client/src/components/ChallengeTeamWizard/ChallengeTeamWizard.integration.test.tsx
```diff
@@ -0,0 +1,259 @@
...
+  act(() => {
+    mockAllIsIntersecting(true)
+  })
+
+  const dialog = screen.getByRole('dialog')
+
+  return { user, component, dialog }
+}
+
+describe('ChallengeTeamWizard tests', () => {
```
nickgros, line 52: Completely rewrite this test file so we have a test for every successful flow

packages/synapse-react-client/src/components/ChallengeTeamWizard/ChallengeTeamWizard.stories.tsx
```diff
@@ -1,16 +1,37 @@
...
 export const Demo: Story = {
   args: {
-    projectId: 'syn51565244',
+    projectId: mockProject.id,
     isShowingModal: true,
   },
+  parameters: {
+    stack: 'mock',
+  },
+}
```
nickgros, line 28: Add a mock story so we can click through this without registering for a challenge in prod

packages/synapse-react-client/src/components/ChallengeTeamWizard/ChallengeTeamWizard.test.tsx
nickgros, line 1: Renamed to *.integration.test.tsx and rewrote

packages/synapse-react-client/src/components/ChallengeTeamWizard/ChallengeTeamWizard.tsx
```diff
@@ -1,76 +1,56 @@
...
   JOIN_REQUEST_SENT = 'JOIN_REQUEST_SENT',
   CREATE_NEW_TEAM = 'CREATE_NEW_TEAM',
   REGISTRATION_SUCCESSFUL = 'REGISTRATION_SUCCESSFUL',
 }
-type StepKey = keyof typeof StepsEnum
-type StepList = {
-  [key in StepKey]: Step
-}
 
-const steps: StepList = {
```
nickgros, line 41: Removed this entire `StepList` object, and removed usage of the `StepperDialog` component in this file.

My opinion is that the `StepList` was confusing because we would override some of these properties on-the-fly. The `StepperDialog` put a complicated interface on top of `DialogBase`, which exposed more complexity than it abstracted. I think it makes more sense to just use `DialogBase`.

packages/synapse-react-client/src/components/ChallengeTeamWizard/ChallengeTeamWizard.tsx
```diff
@@ -79,366 +59,368 @@ export type ChallengeTeamWizardProps = {
...
-  const { accessToken, keyFactory } = useSynapseContext()
-  const [loading, setLoading] = useState<boolean>(true)
-  const [step, setStep] = useState<Step>(steps.SELECT_YOUR_CHALLENGE_TEAM)
-  const [errorMessage, setErrorMessage] = useState<string>()
+/**
+ * The ChallengeTeamWizard is used to guide a user through the process of joining or creating a team for a challenge.
+ *
+ * A required precondition is that the user is NOT on any registered submission team for the challenge.
+ */
+function ChallengeTeamWizard(props: ChallengeTeamWizardProps) {
```
nickgros, line 67: Completely refactored this component to store as little state as possible to get through the wizard steps by delegating state/interactions to child components when possible.

There is still complexity here (most of it is essential), but I tried to detangle it as best as I could.

packages/synapse-react-client/src/components/ChallengeTeamWizard/ChallengeTeamWizard.tsx
```diff
@@ -79,366 +59,368 @@ export type ChallengeTeamWizardProps = {
...
 
-  useEffect(() => {
-    if (teamMembershipStatus) {
-      setMembershipStatus(prev => ({
-        ...prev,
-        [teamMembershipStatus.teamId]: teamMembershipStatus,
-      }))
+  // Determine modal content based on step.id
+  const { actions = <></>, content = <></> } = useMemo(() => {
+    if (!isLoggedIn) {
```
nickgros, line 175: Instead of using an effect to `setErrorMessage` when not logged in, or when already on a submission team, just return the error in an `Alert`.

packages/synapse-react-client/src/components/ChallengeTeamWizard/ChallengeTeamWizard.tsx
```diff
@@ -79,366 +59,368 @@ export type ChallengeTeamWizardProps = {
...
-        canJoin,
-        membershipApprovalRequired,
-        isMember,
-        hasOpenInvitation,
-        hasOpenRequest,
-      } = membershipStatus[selectedTeamId]
 
-      if (isMember) {
-        // User cannot join this team, disable Next button
+    switch (step) {
```
nickgros, line 198: Most of the logic to drive rendering child components is now contained in this `switch`, which I think is easier to follow than a list of objects with properties that are overridden on-the-fly with effects based on local state.

packages/synapse-react-client/src/components/ChallengeTeamWizard/ChallengeTeamWizard.tsx
```diff
@@ -79,366 +59,368 @@ export type ChallengeTeamWizardProps = {
...
+            <>
+              {backButton}
+              <AcceptMembershipInvitationButton
+                teamId={selectedTeamId!}
+                onSuccess={() => {
+                  setStep(ChallengeTeamWizardStep.REGISTRATION_SUCCESSFUL)
+                }}
+              />
+            </>
+          ),
```
nickgros, line 292: This is a new step to show a membership invitation and prompt the user to accept it

packages/synapse-react-client/src/components/ChallengeTeamWizard/JoinRequestForm.tsx
nickgros, line 1: Renamed to `MembershipRequestForm` since it is used to create a `MembershipRequest`

packages/synapse-react-client/src/components/ChallengeTeamWizard/MembershipRequestForm.tsx
```diff
@@ -0,0 +1,104 @@
...
+
+type JoinRequestFormProps = {
+  teamId: string
+  onRequestSubmitted: () => void
+}
+
+export type MembershipRequestFormHandle = {
+  // Allow the parent component to trigger a submit of the form, so this may be embedded in an arbitrary modal.
+  submit: () => void
+}
```
nickgros, line 19: Changed to expose a `submit` button to submit the membership request, so that way we capture the logic required to make the service call in this component instead of requiring the parent to configure it.

packages/synapse-react-client/src/components/ChallengeTeamWizard/OpenMembershipInvitation.tsx
```diff
@@ -0,0 +1,112 @@
...
+/**
+ * For the current, authenticated user, displays the details of an open MembershipInvitation for the provided team ID.
+ *
+ * This component requires
+ *   1. The current user is logged-in
+ *   2. The current user has an open invitation to join the teamId passed in via props
+ *
+ * To prompt the user to accept the invitation, use the {@link AcceptMembershipInvitationButton} component.
+ */
+export function OpenMembershipInvitation(props: AcceptInvitationProps) {
```
nickgros, line 61: image

packages/synapse-react-client/src/components/ChallengeTeamWizard/OpenMembershipInvitation.tsx
```diff
@@ -0,0 +1,112 @@
...
+import { Alert, Button, ButtonProps, Stack, Typography } from '@mui/material'
+import UserOrTeamBadge from '../UserOrTeamBadge'
+import { noop } from 'lodash-es'
+
+export type AcceptInvitationProps = {
+  teamId: string
+  onSuccess?: () => void
+}
+
+export function AcceptMembershipInvitationButton(
```
nickgros, line 19: Button to accept joining a team--since this doesn't actually need to share data with the `OpenMembershipInvitation` component, it can be rendered separately, as long as the same `teamId` prop is passed to both components. A bit more simple than the `submit/useImperativeHandle` pattern, which isn't necessary for this case.

packages/synapse-react-client/src/components/ChallengeTeamWizard/ChallengeTeamTable.tsx
```diff
@@ -89,6 +104,7 @@ export default function ChallengeTeamTable(props: ChallengeTeamTableProps) {
                 onSelectTeam(teamId)
               }}
               label=""
+              aria-label={`Select ${row.name}`}
```
nickgros, line 107: Add `aria-label` so it is easier to manipulate selections in tests

packages/synapse-react-client/src/components/widgets/RadioGroup.tsx
```diff
@@ -36,28 +36,29 @@ export type RadioOptionProps<T extends string | boolean | number = string> = {
   value: T
   style?: React.CSSProperties
   onChange: (value: T) => void
-  disabled?: boolean
-  checked: boolean
-}
+} & Omit<
+  React.HTMLProps<HTMLInputElement>,
+  'label' | 'value' | 'style' | 'onChange'
+>
```
nickgros, line 42: Just changed the prop signature to allow passing any valid prop to the `input` component. This let me pass an `aria-label` prop in `ChallengeTeamTable`

packages/synapse-react-client/src/mocks/challenge/mockChallenge.ts
```diff
@@ -95,3 +101,22 @@ export const mockTeamList = (teams: ChallengeTeam[]): ListWrapper<Team> => {
     list,
   }
 }
+
+export const mockRegisteredChallengeTeams: ChallengeTeam[] = [
```
nickgros, line 105: Moved mock declaration to the `mockChallenge` file instead of the `challengeHandlers`

jay-hodgson (approved):

PR #765, opened 2024-03-01
PORTALS-2971 - Normalize 'syn' prefix on entity IDs before requesting headers

nickgros:


packages/synapse-react-client/src/components/EntityHeaderTable/EntityHeaderTable.tsx
```diff
@@ -230,9 +230,11 @@ export const EntityHeaderTable = (props: EntityHeaderTableProps) => {
     //create dummy entries for values that were not returned by the getEntityHeaders call!
     const newData = results ? results?.results : []
     const newDataEntityIds = new Set()
-    newData.map(entityHeader => newDataEntityIds.add(entityHeader.id))
+    newData.forEach(entityHeader =>
```
nickgros, line 233: `forEach` is more appropriate than `map` here, this isn't a transformation

packages/synapse-react-client/src/synapse-queries/entity/useGetEntityHeaders.ts
```diff
@@ -21,6 +21,11 @@ export function useGetEntityHeaders(
 ) {
   const { accessToken, keyFactory } = useSynapseContext()
   const queryClient = useQueryClient()
+  references = references.map(ref => ({
+    ...ref,
+    targetId: normalizeSynPrefix(ref.targetId),
+  }))
```
nickgros, line 27: transform the list of references so the `targetId`s all start with `syn`

packages/synapse-react-client/src/synapse-queries/entity/useGetEntityHeaders.ts
```diff
@@ -80,15 +85,19 @@ export function useGetEntityHeaders(
...
+  entityId: string = '',
   versionNumber?: number,
   options?: Partial<
     UseQueryOptions<EntityHeader | undefined, SynapseClientError>
   >,
 ) {
   const { accessToken, keyFactory } = useSynapseContext()
+  if (entityId) {
+    entityId = normalizeSynPrefix(entityId)
+  }
```
nickgros, line 97: If there's a truthy (not the empty string) entityId, add `syn`.

Don't do anything with empty string, because it may be that the caller of the hook has not yet given us an ID (it could be the result of a different request). If that's the case, we should disable the query until then.

packages/synapse-react-client/src/utils/functions/EntityTypeUtils.ts
```diff
@@ -307,8 +306,16 @@ export function getVersionDisplay(entity: Entity): string {
...
 }
 
-export function normalizeSynPrefix(ref: Reference) {
-  return ref.targetId.startsWith('syn') ? ref.targetId : `syn${ref.targetId}`
+/**
+ * Given an entityId, returns the entity ID with the `syn` prefix.
+ * If the entity already has the `syn` prefix, the entityId will not be changed
+ * @param entityId
+ */
+export function normalizeSynPrefix(entityId: string) {
```
nickgros, line 314: Use string instead of `Reference`

hallieswan (approved):
Looks good to me!

PR #766, opened 2024-03-01
SWC-6105 - Get annotations for specified version in AnnotationsTable.tsx

nickgros:
SWC-6105
- Get annotations for specified version in AnnotationsTable.tsx
- Simplify entity query keys

### DIFF
--- a/packages/synapse-react-client/src/components/SchemaDrivenAnnotationEditor/SchemaDrivenAnnotationEditor.tsx
+++ b/packages/synapse-react-client/src/components/SchemaDrivenAnnotationEditor/SchemaDrivenAnnotationEditor.tsx
@@ -150,6 +150,7 @@ export function SchemaDrivenAnnotationEditor(
 
   const { data } = useGetJson(
     entityId!,
+    undefined,
     // Derived annotations will be precomputed and displayed as placeholders in the form
     false,
     {
--- a/packages/synapse-react-client/src/components/entity/metadata/AnnotationsTable.tsx
+++ b/packages/synapse-react-client/src/components/entity/metadata/AnnotationsTable.tsx
@@ -35,19 +35,18 @@ function getDisplayedAnnotation(value: string | number | boolean): string {
 }
 
 export function AnnotationsTable(props: AnnotationsTableProps) {
-  const { entityId } = props
+  const { entityId, versionNumber } = props
   const [isManuallyRefetching, setIsManuallyRefetching] = useState(false)
   /**
    * Currently, schema/validation features are only shown in experimental mode.
    */
   const { isInExperimentalMode } = useSynapseContext()
 
-  // TODO: Support versioned annotations, see PLFM-7290
   const {
     data: entityData,
     isLoading,
     refetch: refetchEntityData,
-  } = useGetJson(entityId, true)
+  } = useGetJson(entityId, versionNumber, true)
   const entityMetadata = entityData?.entityMetadata
   const annotations = entityData?.annotations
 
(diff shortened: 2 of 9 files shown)


packages/synapse-react-client/src/components/SchemaDrivenAnnotationEditor/SchemaDrivenAnnotationEditor.tsx
```diff
@@ -150,6 +150,7 @@ export function SchemaDrivenAnnotationEditor(
 
   const { data } = useGetJson(
     entityId!,
+    undefined,
```
nickgros, line 153: For the annotation editor, always get the annotations for the latest version

packages/synapse-react-client/src/components/entity/metadata/AnnotationsTable.tsx
```diff
@@ -35,19 +35,18 @@ function getDisplayedAnnotation(value: string | number | boolean): string {
...
    */
   const { isInExperimentalMode } = useSynapseContext()
 
-  // TODO: Support versioned annotations, see PLFM-7290
   const {
     data: entityData,
     isLoading,
     refetch: refetchEntityData,
-  } = useGetJson(entityId, true)
+  } = useGetJson(entityId, versionNumber, true)
```
nickgros, line 49: For the AnnotationsTable (read-only), use the prop to get the current version of the annotations

packages/synapse-react-client/src/synapse-queries/KeyFactory.ts
```diff
@@ -44,17 +44,18 @@ const entityQueryKeyObjects = {
...
+  entity: (id: string, versionNumber?: string | number) => {
+    const key: Record<string, unknown> = {
+      ...entityQueryKeyObjects.all,
+      id: id,
+    }
+    if (versionNumber) {
+      key['version'] = versionNumber
+    }
+    return key
+  },
```
nickgros, line 56: Change the query key so our query keys look like 
```
[ { objectType: 'entity', id: 'syn123' } ]
[ { objectType: 'entity', id: 'syn123', version: 2 } ]

[ { objectType: 'entity', id: 'syn123' }, 'activity' ]
[ { objectType: 'entity', id: 'syn123', version: 2 }, 'activity' ]
```

I verified that the invalidations will work as expected. For example, calling `queryClient.invalidateQueries` with the query key  `[ { objectType: 'entity', id: 'syn123' } ]` will invalidate all queries related to syn123, including those that include a version number.

packages/synapse-react-client/src/synapse-queries/KeyFactory.ts
```diff
@@ -44,17 +44,18 @@ const entityQueryKeyObjects = {
...
   // List of Entity Versions
-  versions: (id: string) => ({
-    ...entityQueryKeyObjects.entity(id),
-    scope: 'versions',
```
nickgros, line 56: I removed the `scope` property from the key and moved the value to be the next array item. This is effectively the same, I think it's a bit easier to reason with when the entity id/version are kept in a separate object from this label

packages/synapse-react-client/src/synapse-queries/entity/useEntity.ts
```diff
@@ -301,7 +304,7 @@ export function useUpdateViaJson(
       await invalidateAllQueriesForEntity(queryClient, keyFactory, entityId)
       queryClient.setQueryData(
         // This annotation data will never include derived annotations, which are calculated by the backend asynchronously
-        keyFactory.getEntityJsonQueryKey(entityId, false),
+        keyFactory.getEntityJsonQueryKey(entityId, undefined, false),
```
hallieswan, line 307: Just to double check, but based on the updates to the query keys, this should invalidate all queries for the entity, including those with the version number?

hallieswan (approved):
Looks good to me!

packages/synapse-react-client/src/synapse-queries/entity/useEntity.ts
nickgros: The invalidation will invalidate all queries for the entity. This `setQueryData` will only apply to the "versionless"/"current" annotations data

PR #767, opened 2024-03-01
SWC-6707: increase login page and standalone login form to fix text w…

nickgros:
…rapping issue


nickgros (commented):
LGTM, we will fix to replace the SVG with a more maintainable solution in a later PR

PR #771, opened 2024-03-04
PORTALS-2961, PORTALS-2962: add GENIE release cards

hallieswan:
PORTALS-2961 - large release card: 

https://github.com/Sage-Bionetworks/synapse-web-monorepo/assets/26949006/e4f79e46-35dd-4feb-913c-4f85d205aff5

PORTALS-2962 - medium release card: 

https://github.com/Sage-Bionetworks/synapse-web-monorepo/assets/26949006/447ca00b-851c-47e4-9dc7-eca6eb0123ce


packages/synapse-react-client/src/components/CardContainer/CardContainer.tsx
```diff
@@ -174,16 +188,24 @@ export function CardContainer(props: CardContainerProps) {
...
-      {title && <h2 className="SRC-card-overview-title">{title}</h2>}
-      {!title && unitDescription && (
-        <TotalQueryResults frontText={'Displaying'} />
-      )}
-      {/* ReactCSSTransitionGroup adds css fade in property for cards that come into view */}
-      {cards}
+    <>
+      <Box
+        role="list"
+        sx={isReleaseCardMediumList ? releaseCardMediumListSx : defaultListSx}
```
hallieswan, line 198: Add special styling for ReleaseCardMedium, so that text will be aligned across cards using a grid for the list of cards and a subgrid for each card's rows.

hallieswan: Otherwise, the only other change here is to move the `showViewMoreButton` out of the list, so that the grid styling doesn't apply to the button.

packages/synapse-react-client/src/components/ReleaseCard/ReleaseCardLarge.tsx
```diff
@@ -0,0 +1,115 @@
...
+  value,
+  label,
+}: ReleaseCardStat) => {
+  return (
+    <Stack
+      direction="column"
+      alignItems="center"
+      padding="20px 40px 20px 40px"
+      gap="10px"
+      width="200px" // ensure that stats are aligned when multiple cards are stacked vertically
```
hallieswan, line 21: Using a set width ensures that the statistics are aligned across a list of large release cards. Could consider using grid/subgrid instead, but this would be a bit more involved, so I was planning to hold off on until we need to display statistics with widely varying widths so using a set 200px width wouldn't work well.

packages/synapse-react-client/src/components/ReleaseCard/ReleaseCardMedium.tsx
```diff
@@ -0,0 +1,100 @@
...
+
+  return (
+    <Box
+      className={`ReleaseCard ${
+        includePortalCardClass ? 'SRC-portalCard' : ''
+      }`}
+      display="grid"
+      gridTemplateColumns={`repeat(${nCols}, 1fr)`} // columns are controlled by this component, not the parent grid
+      columnGap="10px"
+      gridTemplateRows="subgrid" // rows are controlled by parent grid, so text will be aligned across cards
```
hallieswan, line 58: Use a grid for columns in each Card, but use the grid of the list of Cards to determine each Card's rows in a subgrid. This ensures that row heights are aligned across Cards, so even if text wraps to the next row in one Card.

packages/synapse-react-client/src/components/ReleaseCard/ReleaseCardTypes.ts
```diff
@@ -0,0 +1,50 @@
...
+  // defining SQL for the data (table, view...) driving the Explore page
+  // which is used in generating the button's query string
+  exploreDataSql?: string
+  // column name of the selected facet in the data (table, view...) driving the Explore page
+  // which is used in generating the button's query string
+  exploreDataFacetColumnName?: string
+  // column name in the data (table, view...) driving the page where the button is located
+  // which is used to to populate the value for the selected facet in the button's query string
+  sourceDataFacetValueColumnName?: string
+}
```
hallieswan, line 28: In the GENIE portal, ["current release" cards](https://www.figma.com/file/BI4y33EHA95onN8DourTNZ/Two-Projects?type=design&node-id=275-22889&mode=design&t=76oHvfvp9FWFtDSR-4) link to different pages within the Portal depending on release type. For GENIE releases, the explore button redirects to `/explore/files/genie`. For other types of releases, the explore button redirects to `/explore/files/genie-bpc` with the current release type facet selected (e.g.'crc'). 

I used this config to specify how the button's link should be built using the table that drives the Release Cards, e.g. https://www.synapse.org/#!Synapse:syn53701326/tables/, but happy to adjust if there's a better/established way to dynamically construct these types of links! 

*The Figma only shows one current release, but there may be multiple current releases shown on the same page, as shown in [the ticket](https://sagebionetworks.jira.com/browse/PORTALS-2961).

packages/synapse-react-client/src/components/ReleaseCard/ReleaseCardTypes.ts
```diff
@@ -0,0 +1,50 @@
...
+  label: string
+  value: string
+}
+
+export type StatConfig = {
+  // specifies which column should be used to populate the statistic value
+  columnName: string
+  // the label that will be displayed with the statistic in the card
+  label: string
+}
```
hallieswan, line 11: In the GENIE portal, cards include specimen and patient counts. However, if these cards are reused for other Portal data releases, we may want to show different statistics. Use this config to allow different statistics to be defined for each set of release cards.

packages/synapse-react-client/src/components/ReleaseCard/ReleaseCardTypes.ts
```diff
@@ -0,0 +1,50 @@
...
+  dataGuidePath?: string
+}
+
+export type ReleaseCardMediumConfig = {
+  cardSize: 'medium'
+  // path to request access to the release data
+  requestAccessPath: string
+  // statistics to display in the release card
+  statsConfig: StatConfig[]
+}
```
hallieswan, line 48: I tried to make these cards generic so that either size could be used in other portals to show current, previous, or a full set of releases. I think the GENIE portal would filter the source table by some column (e.g. `isCurrentRelease`) to display the correct set of releases -- see stories for an example.

packages/synapse-react-client/src/components/ReleaseCard/ReleaseCardUtils.ts
```diff
@@ -0,0 +1,146 @@
...
+const formatNumber = (num: number) => {
+  if (num < 1_000_000) return num.toLocaleString()
+  const numberFormatter = new Intl.NumberFormat([], {
+    minimumFractionDigits: 1,
+    maximumFractionDigits: 2,
+    notation: 'compact',
+    localeMatcher: 'best fit',
+  })
+  return numberFormatter.format(num)
+}
```
hallieswan, line 22: When a statistic is a very large number, use [Intl.NumberFormat](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/NumberFormat) to display a shortened version of the number, e.g. 1.234T instead of 123,456,789,012,345.

packages/synapse-react-client/src/components/ReleaseCard/ReleaseCardUtils.ts
```diff
@@ -0,0 +1,146 @@
...
+  statsConfig: StatConfig[],
+) => {
+  return statsConfig.map(statConfig => {
+    const { columnName, label } = statConfig
+    const statString = getValueFromData(schema, data, columnName)
+    const cardStat: ReleaseCardStat = {
+      label: label,
+      value: statString
+        ? formatNumber(Number(statString))
+        : MISSING_STAT_DEFAULT_VALUE,
```
hallieswan, line 52: I'm currently assuming that all statistics will be numbers.

packages/synapse-react-client/src/mocks/entity/mockReleaseCardsTable.ts
```diff
@@ -0,0 +1,59 @@
+import { EntityHeader, TableEntity } from '@sage-bionetworks/synapse-types'
+import { MOCK_USER_ID } from '../user/mock_user_profile'
+import { MockEntityData } from './MockEntityData'
+import mockProject from './mockProject'
+
+export const MOCK_RELEASE_CARDS_TABLE_ID = 'syn40001'
+const MOCK_RELEASE_CARDS_TABLE_NAME = 'releases table'
+const parentId = mockProject.id
+
+export const mockReleaseCardsTable = {
```
hallieswan, line 10: Create a mock table to drive the release cards in Storybook until a public entity is created for the GENIE portal.

packages/synapse-react-client/src/components/ReleaseCard/ReleaseCardTypes.ts
```diff
@@ -0,0 +1,50 @@
...
+export type ReleaseCardLargeConfig = {
+  cardSize: 'large'
+  // whether the release version should be prefixed with "Release "
+  prependRelease: boolean
+  // statistics to display in the release card
+  statsConfig: StatConfig[]
+  // button to explore data in the release
+  buttonToExplorePageConfig?: ButtonToExplorePageConfig
+  // path to release's data guide
+  dataGuidePath?: string
```
hallieswan, line 39: Currently assuming that all cards will point to the same release data guide page.

packages/synapse-react-client/src/components/ReleaseCard/ReleaseCardTypes.ts
```diff
@@ -0,0 +1,50 @@
...
+  // column name in the data (table, view...) driving the page where the button is located
+  // which is used to to populate the value for the selected facet in the button's query string
+  sourceDataFacetValueColumnName?: string
+}
+
+export type ReleaseCardLargeConfig = {
+  cardSize: 'large'
+  // whether the release version should be prefixed with "Release "
+  prependRelease: boolean
+  // statistics to display in the release card
```
hallieswan, line 34: Cards can be shown [without "Release"](https://www.figma.com/file/BI4y33EHA95onN8DourTNZ/Two-Projects?type=design&node-id=195-13613&mode=design&t=76oHvfvp9FWFtDSR-4) or [with "Release"](https://www.figma.com/file/BI4y33EHA95onN8DourTNZ/Two-Projects?type=design&node-id=277-24184&mode=design&t=76oHvfvp9FWFtDSR-4) before the data release version name.

packages/synapse-react-client/src/components/ReleaseCard/ReleaseCardTypes.ts
```diff
@@ -0,0 +1,50 @@
...
+  // button to explore data in the release
+  buttonToExplorePageConfig?: ButtonToExplorePageConfig
+  // path to release's data guide
+  dataGuidePath?: string
+}
+
+export type ReleaseCardMediumConfig = {
+  cardSize: 'medium'
+  // path to request access to the release data
+  requestAccessPath: string
```
hallieswan, line 45: In the [Figma](https://www.figma.com/file/BI4y33EHA95onN8DourTNZ/Two-Projects?type=design&node-id=263-15875&mode=design&t=76oHvfvp9FWFtDSR-4), previous release cards all point to the same "Request Access" page. If we don't expect that medium release cards will always point to a "Request Access" page, I could change this field name to be more generic (e.g. "cardLinkPath"?).

packages/synapse-react-client/src/components/CardContainerLogic/CardContainerLogic.stories.ts
```diff
@@ -101,3 +112,76 @@ export const PublicationCard: Story = {
...
+        label: 'Explore Current Data Release',
+        sourcePathColumnName: 'releaseExplorePath',
+        exploreDataSql: currentReleaseCardSql, // generally this would refer to a different table, not the source table
+        exploreDataFacetColumnName: 'releaseType',
+        sourceDataFacetValueColumnName: 'releaseType',
+      },
+      dataGuidePath: 'data guide',
+    },
+  },
+  parameters: {
```
nickgros, line 142: So the storybook user doesn't have to explicitly switch to the mock stack
```suggestion
  parameters: {
    stack: 'mock',
```

packages/synapse-react-client/src/components/CardContainerLogic/CardContainerLogic.stories.ts
```diff
@@ -101,3 +112,76 @@ export const PublicationCard: Story = {
...
+    sql: `SELECT * FROM ${MOCK_RELEASE_CARDS_TABLE_ID} WHERE isCurrentRelease = false`,
+    type: RELEASE_CARD,
+    initialLimit: 5,
+    releaseCardConfig: {
+      cardSize: 'medium',
+      requestAccessPath: 'data access',
+      statsConfig: statsConfig,
+    },
+  },
+  parameters: {
```
nickgros, line 171:
```suggestion
  parameters: {
    stack: 'mock',
```

packages/synapse-react-client/src/components/ReleaseCard/ReleaseCardMedium.tsx
nickgros: Can't wait to use this more and more in these portals components!

Based on my inspection, this looks ok for users who use a browser that does not support subgrid (actually imperceptible in the demo because all cards have same # of stats and each sub-row has the same height).

nickgros (approved):
This looks great! Had a few comments but none blocking merge.

packages/synapse-react-client/src/components/ReleaseCard/ReleaseCardUtils.ts
```diff
@@ -0,0 +1,146 @@
...
+    exploreDataFacetColumnName &&
+    sourceDataFacetValueColumnName
+  ) {
+    const facetValue = getValueFromData(
+      schema,
+      data,
+      sourceDataFacetValueColumnName,
+    )
+    // column not found in source table or cell did not have value in source table
+    if (facetValue === null) return null
```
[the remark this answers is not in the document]
hallieswan: For the GENIE data releases, the source table will have a null value in the facet value column, so that the button will redirect to `/explore/files/genie` without selected facets. Would you include that context in the warning, i.e. that the facet value may be null by design?

packages/synapse-react-client/src/components/ReleaseCard/ReleaseCardUtils.ts
```diff
@@ -0,0 +1,146 @@
...
+          concreteType:
+            'org.sagebionetworks.repo.model.table.FacetColumnValuesRequest',
+          columnName: exploreDataFacetColumnName,
+          facetValues: [facetValue],
+        },
+      ],
+    }
+    return `?QueryWrapper0=${encodeURIComponent(JSON.stringify(query))}`
+  }
+  return null
```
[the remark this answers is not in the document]
hallieswan: The parameters used to build the query string (`exploreDataSql`, `exploreDataFacetColumnName`, `sourceDataFacetValueColumnName`) would be intentionally undefined in the release card config when all cards should redirect to an explore page without selecting a particular facet. Would you also include this context in the warning here? Just wondering how to include information that would be useful if there is a mistake in the config, but not cause confusion when the config is working as intended.

packages/synapse-react-client/src/components/QueryWrapper/generateEncodedPathAndQueryForSelectedFacetURL.ts
```diff
@@ -0,0 +1,34 @@
+import { Query } from '@sage-bionetworks/synapse-types'
+
+const generateEncodedQueryForSelectedFacetURL = (
```
hallieswan, line 3: Extract common logic from ReleaseCardUtils and onPointClick

packages/synapse-react-client/src/components/ReleaseCard/ReleaseCardUtils.ts
```diff
@@ -82,65 +82,44 @@ export const formatReleaseCardData = (
...
+    schema,
+    data,
+    buttonToExploreDataConfig.sourcePathColumnName,
+  )
+  if (!path) {
+    console.warn(
+      `Column not found in source table or cell did not have value in source table for ${buttonToExploreDataConfig.sourcePathColumnName}`,
+    )
+    return null
+  }
```
hallieswan, line 101: Output a warning if the user specified a `buttonToExploreDataConfig`, but a value for the path was not found in the source table. Assume that this is more likely to be an issue with the config or source table than by design. 

If we find that users only specify an explore button path for a subset of release cards by design, could change to a log message instead.

packages/synapse-react-client/src/components/ReleaseCard/ReleaseCardUtils.ts
```diff
@@ -82,65 +82,44 @@ export const formatReleaseCardData = (
...
-) => {
-  if (!buttonToExploreDataConfig) return null
-  const path = getValueFromData(
-    schema,
-    data,
-    buttonToExploreDataConfig.sourcePathColumnName,
-  )
-  // column not found in source table or cell did not have value in source table
-  if (!path) return null
+  if (!hasSelectedFacet) console.log(`No selected facet for ${path}`)
```
hallieswan, line 115: In the GENIE portal, release cards will either point directly to an explore page and or also include a selected facet. The source table will have `null` values in the selected facet column for the cards that should point directly to the explore page. Since this is an expected case, log a message about the missing facet rather than a warning.

apps/portals/src/configurations/cancercomplexity/synapseConfigs/onPointClick.ts
```diff
@@ -50,12 +30,14 @@ export const onPointClick = ({
   if (typeUpperCase === 'Grants' || typeUpperCase === 'Projects') {
     facet = 'consortium'
   }
-  const encodedQuery = generateEncodedQueryForURL(
-    typeUpperCase,
+
+  const url = generateEncodedPathAndQueryForSelectedFacetURL(
```
hallieswan, line 34: https://github.com/Sage-Bionetworks/synapse-web-monorepo/assets/26949006/6f9c4a25-2f54-46b0-b177-2e6ac4961c15

packages/synapse-react-client/src/components/ReleaseCard/ReleaseCardUtils.ts
nickgros: Oh sorry, I misinterpreted the last version of the code. It's much more clear now with your explanation and the changes to the logical flow here. I don't think you need to log anything, since no facets is definitely a valid case!

nickgros (approved):

PR #775, opened 2024-03-04
PORTALS-2978: Set default portal page title at build time rather than…

jay-hodgson:
… relying on an onmount js effect

### DIFF
--- a/apps/portals/index.html
+++ b/apps/portals/index.html
@@ -37,7 +37,7 @@
       user's mobile device or desktop. See https://developers.google.com/web/fundamentals/web-app-manifest/
     -->
     <link rel="manifest" href="./manifest.json" />
-    <title> </title>
+    <title>%VITE_PORTAL_NAME%</title>
     <!-- This gets populated on app load -->
     <!-- links below are SRC dependencies -->
     <link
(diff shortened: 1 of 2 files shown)


apps/portals/index.html
```diff
@@ -37,7 +37,7 @@
       user's mobile device or desktop. See https://developers.google.com/web/fundamentals/web-app-manifest/
     -->
     <link rel="manifest" href="./manifest.json" />
-    <title> </title>
+    <title>%VITE_PORTAL_NAME%</title>
```
jay-hodgson, line 40: Set default page title at build time

apps/portals/src/AppInitializer.tsx
```diff
@@ -13,27 +13,13 @@ import { useLogInDialogContext } from './LogInDialogContext'
...
 
-/** On mount, update the document title and meta description using data from the portal config */
-function useSetDocumentMetadataFromConfig() {
-  useEffect(() => {
-    if (document.title !== import.meta.env.VITE_PORTAL_NAME) {
-      document.title = import.meta.env.VITE_PORTAL_NAME
-    }
-    document
-      .querySelector('meta[name="description"]')!
-      .setAttribute('content', import.meta.env.VITE_PORTAL_DESCRIPTION)
```
jay-hodgson, line 24: Default description already set at build time, this is unnecessary

nickgros (approved):

PR #776, opened 2024-03-05
Portals 2875

jinjunoh:


apps/portals/src/configurations/nf/synapseConfigs/datasets.ts
```diff
@@ -8,6 +7,14 @@ import { DetailsPageProps } from 'types/portal-util-types'
 export const newDatasetsSql = `${datasetsSql} order by ROW_ID desc limit 3`
 const rgbIndex = 8
 
+const columnAliases = {
+  pmid: 'PubMed Id',
+  id: 'File ID',
+  nf1Genotype: 'NF1 Genotype',
+  nf2Genotype: 'NF2 Genotype',
+  studyId: 'Study Name',
+}
```
jinjunoh, line 16: columnAliases pulled from ./commonProps had overlapping attribute studyId: 'On Synapse'.

jinjunoh: image

packages/synapse-react-client/src/components/EntityLink.tsx
```diff
@@ -15,7 +15,7 @@ type EntityLinkProps = {
   entity: string | EntityHeader | Entity
   versionNumber?: number
   /** Whether the component should link to the entity page in Synapse. Default true */
-  link?: boolean
+  link?: boolean | string
```
nickgros, line 18: Please update the comment to indicate that the link itself can be overridden by passing a string

packages/synapse-react-client/src/components/EntityLink.tsx
```diff
@@ -64,7 +64,21 @@ export const EntityLink = (props: EntityLinkProps) => {
...
+          rel="noopener noreferrer"
+          href={link}
+        >
+          {showIcon && (
+            <EntityTypeIcon type={type} style={{ marginRight: '6px' }} />
+          )}
+          {entity[displayTextField as keyof typeof entity]}
+        </Link>
+      )
+    }
```
nickgros, line 81: Here you duplicated the `<Link>` return, so we have to independently re-implement all of the other features like `showIcon`, `displayTextField`, etc.

Can you refactor so that we just have one returned `Link` component, where the `href` prop is dynamically configured?

nickgros (changes requested):

packages/synapse-react-client/src/components/EntityLink.tsx
```diff
@@ -64,16 +64,19 @@ export const EntityLink = (props: EntityLinkProps) => {
...
           rel="noopener noreferrer"
-          href={`${PRODUCTION_ENDPOINT_CONFIG.PORTAL}#!Synapse:${entity.id!}${
-            versionNumber ? `.${versionNumber}` : ''
-          }`}
+          href={
+            typeof link === 'string'
+              ? link
+              : `${PRODUCTION_ENDPOINT_CONFIG.PORTAL}#!Synapse:${entity.id!}${
+                  versionNumber ? `.${versionNumber}` : ''
+                }`
```
nickgros, line 78: This error was here before your change, but can you change `PRODUCTION_ENDPOINT_CONFIG.PORTAL` to `getEndpoint(BackendDestinationEnum.PORTAL_ENDPOINT)`?

This ensures that the link keeps you on the stack that you were already on, e.g. if you click this in staging, it should keep you in staging. Right now, it always takes you to production, which is wrong!

packages/synapse-react-client/src/components/GenericCard/SynapseCardLabel.tsx
```diff
@@ -99,6 +100,12 @@ export const SynapseCardLabel: React.FC<SynapseCardLabelProps> = props => {
     return <Linkify text={str} className={newClassName} />
   }
 
+  if ('resolveEntityName' in labelLink && labelLink.resolveEntityName && str) {
+    const { baseURL, URLColumnName } = labelLink
+    const href = `/${baseURL}?${URLColumnName}=${str}`
+    return <EntityLink entity={str} link={href} />
```
nickgros, line 106: Let's hide the icon by default since we aren't linking to Synapse

```suggestion
    return <EntityLink entity={str} link={href} showIcon={false} />
```

apps/portals/src/configurations/nf/synapseConfigs/datasets.ts
```diff
@@ -26,6 +32,15 @@ export const datasetCardConfiguration: CardConfiguration = {
       'yearProcessed',
     ],
   },
+  labelLinkConfig: [
+    {
+      isMarkdown: false,
+      matchColumnName: 'studyId',
+      URLColumnName: 'studyId',
+      baseURL: 'Explore/Datasets/DetailsPage',
```
nickgros, line 40: Shouldn't this link to the Studies page? Can you check a few of the "Datasets" details pages and verify that these links work?
```suggestion
      baseURL: 'Explore/Studies/DetailsPage',
```

nickgros (changes requested):

PR #783, opened 2024-03-06
Enable deploy app

xschildw:
This PR enables actual deployment of an app to an S3 bucket + update couple deps.


jay-hodgson:
Let's try it!  build-and-deploy-app used by synapse-oauth-signin and sage-account-web (staging buckets).
build-and-deploy-portal is used by deploy-adknowledge-portal (test for one of the portals) and deploy-challenge-portal (also points to staging bucket only).

jay-hodgson (approved):

PR #791, opened 2024-03-07
PORTALS-2974 - Reset table query offset after changing facets/filters

nickgros:


packages/synapse-react-client/src/utils/functions/queryUtils.ts
```diff
@@ -246,7 +246,7 @@ export function removeEmptyQueryParams(q: Query) {
   if (query.limit == null) {
     delete query.limit
   }
-  if (query.offset == null) {
+  if (query.offset == null || query.offset == 0) {
```
nickgros, line 249: If query offset is 0, delete it (improves likelihood of cache hit)

jay-hodgson (approved):

PR #803, opened 2024-03-11
SWC-6724 - Fix issue where DATE columns would incorrectly parse a string that was a unix timestamp

nickgros:
See SWC-6724

### DIFF
--- a/packages/synapse-react-client/src/components/DateTimePicker/DateTimePicker.tsx
+++ b/packages/synapse-react-client/src/components/DateTimePicker/DateTimePicker.tsx
@@ -16,7 +16,7 @@ dayjs.extend(utc)
 dayjs.extend(timezone)
 dayjs.extend(advancedFormat)
 
-export type DateTimePickerProps = MuiDateTimePickerProps<string | Dayjs>
+export type DateTimePickerProps = MuiDateTimePickerProps<Dayjs>
 
 function TextFieldWithTzShown(props: TextFieldProps) {
   const { utcTime } = useSynapseContext()
@@ -55,7 +55,7 @@ export default function DateTimePicker(props: DateTimePickerProps) {
 
   return (
     <LocalizationProvider dateAdapter={AdapterDayjs}>
-      <MuiDateTimePicker<string | Dayjs>
+      <MuiDateTimePicker<Dayjs>
         value={valueAsDayjs}
         slots={{
           ...slots,
--- a/packages/synapse-react-client/src/components/Evaluation/EvaluationRoundEditor.tsx
+++ b/packages/synapse-react-client/src/components/Evaluation/EvaluationRoundEditor.tsx
@@ -147,12 +147,12 @@ export const EvaluationRoundEditor: React.FunctionComponent<
     }
   }, [error])
 
-  const [startDate, setStartDate] = useState<string | Dayjs | null>(
+  const [startDate, setStartDate] = useState<Dayjs | null>(
     evaluationRoundInput.roundStart
       ? dayjs(evaluationRoundInput.roundStart)
       : null,
   )
-  const [endDate, setEndDate] = useState<string | Dayjs | null>(
+  const [endDate, setEndDate] = useState<Dayjs | null>(
     evaluationRoundInput.roundEnd ? dayjs(evaluationRoundInput.roundEnd) : null,
   )
 
--- a/packages/synapse-react-client/src/components/TableColumnSchemaEditor/ColumnModelFormFields/DefaultValueField.test.tsx
+++ b/packages/synapse-react-client/src/components/TableColumnSchemaEditor/ColumnModelFormFields/DefaultValueField.test.tsx
@@ -1,11 +1,22 @@
 import React from 'react'
 import DefaultValueField, { DefaultValueFieldProps } from './DefaultValueField'
 import { createWrapper } from '../../../testutils/TestingLibraryUtils'
-import { render, screen, within } from '@testing-library/react'
+import { act, render, screen, within } from '@testing-library/react'
 import { ColumnTypeEnum } from '@sage-bionetworks/synapse-types'
 import userEvent from '@testing-library/user-event'
 import { formatDate } from '../../../utils/functions/DateFormatter'
-import dayjs from 'dayjs'
+import dayjs, { isDayjs } from 'dayjs'
+import DateTimePicker from '../../DateTimePicker/DateTimePicker'
+
+const MOCK_DATETIME_PICKER_TEST_ID = 'MockDateTimePicker'
+
+// Mock the DateTimePicker because how the date renders can vary based on time zone and localization, making it hard to test
+jest.mock('../../DateTimePicker/DateTimePicker', () => ({
+  __esModule: true,
+  default: jest.fn(() => <div data-testid={MOCK_DATETIME_PICKER_TEST_ID} />),
+}))
+
+const mockDateTimePicker = jest.mocked(DateTimePicker)
 
 function renderComponent<T>(props: DefaultValueFieldProps<T>) {
   return render(<DefaultValueField<T> {...props} />, {
@@ -14,6 +25,9 @@ function renderComponent<T>(props: DefaultValueFieldProps<T>) {
 }
 
 describe('DefaultValueField', () => {
+  beforeEach(() => {
+    jest.clearAllMocks()
+  })
   describe('STRING columnType', () => {
     it('shows an initial value', () => {
       const onChange = jest.fn()
@@ -91,16 +105,24 @@ describe('DefaultValueField', () => {
         isOriginallyDefaultColumn: false,
       },
       onChange,
-      value: '2021-01-15T00:00:00.000Z',
+      value: '1710388800000',
     })
 
-    // The actual value shown can vary based on localization
-    // So we will just test for the presence of a value and successful onChange call
-    const datePicker = screen.getByRole('textbox')
-    expect(datePicker.getAttribute('value')).not.toBeNull()
+    await screen.findByTestId(MOCK_DATETIME_PICKER_TEST_ID)
+    const mockDateTimePickerPassedProps = mockDateTimePicker.mock.lastCall![0]
+    // Verify that we pass the correct Dayjs value to the mock
+    expect(isDayjs(mockDateTimePickerPassedProps.value)).toBe(true)
+    expect(mockDateTimePickerPassedProps.value!.valueOf()).toBe(1710388800000)
 
-    await userEvent.type(datePicker, '2')
-    expect(onChange).toHaveBeenCalled()
+    // Simulate changing the date
+    act(() => {
+      mockDateTimePickerPassedProps.onChange!(dayjs(1710399900000), {
+        validationError: null,
+      })
+    })
+
+    // Verify that `onChange` gets a millisecond timestamp encoded in a string
+    expect(onChange).toHaveBeenLastCalledWith('1710399900000')
   })
 
   it('handles a STRING_LIST column type', async () => {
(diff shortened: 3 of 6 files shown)


packages/synapse-react-client/src/components/DateTimePicker/DateTimePicker.tsx
```diff
@@ -16,7 +16,7 @@ dayjs.extend(utc)
 dayjs.extend(timezone)
 dayjs.extend(advancedFormat)
 
-export type DateTimePickerProps = MuiDateTimePickerProps<string | Dayjs>
+export type DateTimePickerProps = MuiDateTimePickerProps<Dayjs>
```
nickgros, line 19: Change our DateTimePicker to only use Dayjs. The point is that how we manage datetimes can vary across the different services, so we should force whoever is using the component to deal with the complexity of parsing and returning the correct format.

packages/synapse-react-client/src/components/Evaluation/EvaluationRoundEditor.tsx
```diff
@@ -147,12 +147,12 @@ export const EvaluationRoundEditor: React.FunctionComponent<
...
-  const [startDate, setStartDate] = useState<string | Dayjs | null>(
+  const [startDate, setStartDate] = useState<Dayjs | null>(
     evaluationRoundInput.roundStart
       ? dayjs(evaluationRoundInput.roundStart)
       : null,
   )
-  const [endDate, setEndDate] = useState<string | Dayjs | null>(
+  const [endDate, setEndDate] = useState<Dayjs | null>(
     evaluationRoundInput.roundEnd ? dayjs(evaluationRoundInput.roundEnd) : null,
   )
```
nickgros, line 157: This is never a string, so it was a safe/easy change

packages/synapse-react-client/src/components/TableColumnSchemaEditor/ColumnModelFormFields/DefaultValueField.test.tsx
```diff
@@ -91,16 +105,24 @@ describe('DefaultValueField', () => {
         isOriginallyDefaultColumn: false,
       },
       onChange,
-      value: '2021-01-15T00:00:00.000Z',
+      value: '1710388800000',
```
nickgros, line 108: Bad test - the backend passes a string containing an integer unix timestamp with millisecond precision

packages/synapse-react-client/src/components/TableColumnSchemaEditor/ColumnModelFormFields/DefaultValueField.test.tsx
```diff
@@ -91,16 +105,24 @@ describe('DefaultValueField', () => {
...
+    act(() => {
+      mockDateTimePickerPassedProps.onChange!(dayjs(1710399900000), {
+        validationError: null,
+      })
+    })
 
-    await userEvent.type(datePicker, '2')
-    expect(onChange).toHaveBeenCalled()
+    // Verify that `onChange` gets a millisecond timestamp encoded in a string
+    expect(onChange).toHaveBeenLastCalledWith('1710399900000')
```
nickgros, line 125: ensure that the changed data matches the format of the input data

packages/synapse-react-client/src/components/TableColumnSchemaEditor/ColumnModelFormFields/DefaultValueField.test.tsx
```diff
@@ -168,20 +187,27 @@ describe('DefaultValueField', () => {
...
 
     await userEvent.click(screen.getByRole('button', { name: 'OK' }))
 
-    // Saving will convert these to ISO strings
-    expect(onChange).toHaveBeenCalledWith([
-      dayjs(1678852800000).toISOString(),
-      dayjs(1710475200000).toISOString(),
-    ])
+    // Saving should invoke onChange, which should return the original type, millisecond timestamps
+    expect(onChange).toHaveBeenCalledWith([DATE_ONE, DATE_TWO])
```
nickgros, line 211: Slight change to the DATE_LIST case. The data passed to `onChange` now matches the original format--integer timestamps

packages/synapse-react-client/src/components/TableColumnSchemaEditor/ColumnModelFormFields/DefaultValueField.tsx
```diff
@@ -57,22 +61,24 @@ function DefaultValueDateField(props: DefaultValueFieldProps<string>) {
   const { onChange, value = null, disabled, TextFieldProps } = props
 
   const valueAsDayjs = useMemo(() => {
-    if (value != null) {
-      return dayjs(value)
+    if (value) {
+      // The backend service returns a string that contains an integer unix timestamp with millisecond precision
+      // Passing the raw string to dayjs without a format option will parse the value incorrectly
+      // 'x' indicates parsing the value as a ms Unix Timestamp - https://day.js.org/docs/en/parse/string-format
+      return dayjs(value, 'x')
```
nickgros, line 68: Fixes SWC-6724

packages/synapse-react-client/src/components/TableColumnSchemaEditor/ColumnModelFormFields/DefaultValueField.tsx
```diff
@@ -102,7 +111,7 @@ function DefaultValueListField(
         // if it's a DATE_LIST, then the values are Unix timestamps in milliseconds
         // The `MultiValueField` component expects ISO strings to match the JSON Schema 'date-time' format
         return _value.map(v => {
-          return dayjs(v).toISOString()
+          return dayjs(v, 'x').toISOString()
```
nickgros, line 114: No bug here since we parse the integers at a higher level, but this makes it a little safer

packages/synapse-react-client/src/components/TableColumnSchemaEditor/ColumnModelFormFields/DefaultValueField.tsx
```diff
@@ -115,8 +124,15 @@ function DefaultValueListField(
     <MultiValueField
       value={value}
       onChange={arr => {
-        // Convert null to undefined
-        onChange(arr || undefined)
+        if (arr == null) {
+          // Convert null to undefined
+          onChange(undefined)
+        } else if (columnModel.columnType === ColumnTypeEnum.DATE_LIST) {
+          onChange(arr.map(val => dayjs(val).valueOf()))
```
nickgros, line 131: Change behavior to match what is passed in - a list of integer timestamps

packages/synapse-react-client/src/components/TableColumnSchemaEditor/ColumnModelFormFields/DefaultValueField.tsx
```diff
@@ -115,8 +124,15 @@ function DefaultValueListField(
...
+        } else {
+          // Convert null to undefined
```
hallieswan, line 133: Looks like this comment may not apply here, since this conversion is handled above

packages/synapse-react-client/src/components/TableColumnSchemaEditor/ColumnModelFormFields/DefaultValueField.tsx
hallieswan: Just curious, why does this `dayjs` call not need the 'x' argument to indicate parsing the value as a ms Unix Timestamp?

packages/synapse-react-client/src/components/TableColumnSchemaEditor/ColumnModelFormFields/DefaultValueField.tsx
nickgros, line 133:
```suggestion
```

packages/synapse-react-client/src/components/TableColumnSchemaEditor/ColumnModelFormFields/MultiValueField.stories.tsx
```diff
@@ -0,0 +1,36 @@
...
+}
+export const IntegerList: Story = {
+  args: {
+    value: [-5, 20, 0],
+    columnType: ColumnTypeEnum.INTEGER_LIST,
+  },
+}
+export const DateList: Story = {
+  args: {
+    value: ['2023-10-05T14:48:00.000Z', '2024-03-11T07:24:39.000Z'],
```
hallieswan, line 33: Based on the processing in DefaultValueListField, I was expecting these to be Unix timestamps with millisecond precision. Does MultiValueField handle Date lists with other datetime formats as well?

nickgros: `MultiValueField` consumes ISO strings (internally it uses React JSON Schema Form, and JSON Schema date-time uses ISO strings). `DefaultValueListField` converts integers to ISO strings to pass into `MultiValueField`, and then transforms ISO strings back to integers.

Since this story just touches MultiValueField, ISO timestamps are appropriate. Managing all of these different levels of translation can get confusing, and it makes it pretty easy to write bugs 🙃

packages/synapse-react-client/src/components/TableColumnSchemaEditor/ColumnModelFormFields/DefaultValueField.tsx
nickgros: In this case, these are ISO timestamps. I'll add a comment to clarify

packages/synapse-react-client/src/components/TableColumnSchemaEditor/ColumnModelFormFields/MultiValueField.stories.tsx
hallieswan: Oh that is confusing! So that's also why the 'x' argument is not needed [here](https://github.com/Sage-Bionetworks/synapse-web-monorepo/pull/803#discussion_r1520132015), since `MultiValueField` returns an ISO string `onChange`, which can be directly converted to a Day.js object with the default `dayjs` parse? If that's right, would it be worth adding a comment with that context?

packages/synapse-react-client/src/components/TableColumnSchemaEditor/ColumnModelFormFields/DefaultValueField.tsx
```diff
@@ -115,8 +124,14 @@ function DefaultValueListField(
     <MultiValueField
       value={value}
       onChange={arr => {
-        // Convert null to undefined
-        onChange(arr || undefined)
+        if (arr == null) {
+          // Convert null to undefined
+          onChange(undefined)
+        } else if (columnModel.columnType === ColumnTypeEnum.DATE_LIST) {
+          onChange(arr.map(val => dayjs(val).valueOf()))
```
nickgros, line 131:
```suggestion
          // Transform the ISO Timestamps to integer unix timestamps to match the passed value
          onChange(arr.map(val => dayjs(val).valueOf()))
```

packages/synapse-react-client/src/components/TableColumnSchemaEditor/ColumnModelFormFields/MultiValueField.stories.tsx
hallieswan: Oh just saw your comment on that point -- thanks for helping me understand, Nick!

nickgros: Just did!

hallieswan (approved):
Looks good to me!

PR #808, opened 2024-03-12
PORTALS-3002: Upgrade Vite, TypeScript, tsup, esbuild

nickgros:
The goal was to just upgrade Vite, but we should also upgrade:
- esbuild - Vite uses esbuild, and we try to keep configs aligned, so we should try to keep them in sync
- tsup - Uses esbuild, same reasons
- TypeScript - the latest Vite changes required tsconfig changes. Upgrading TypeScript wasn't strictly required, but it ensured I wouldn't use deprecated/removed features as workarounds to new problems that we would have to re-address when upgrading TypeScript separately.

## Summary of changes:

Across entire project
- Upgrade TypeScript to v5.4.2
- Upgrade Vite to v5
- Upgrade Vite plugins to latest versions (except vite-plugin-node-polyfills)
- Upgrade tsup to v8
- Upgrade esbuild to v0.20
- Change tsconfig moduleResolution to "bundler" (Vite 5 compatibility) -

vite-config
- Replace @esbuild-plugins/node-globals-polyfill, @esbuild-plugins/node-modules-polyfill,rollup-plugin-polyfill-node with vite-plugin-node-polyfills. Pin to v0.17.0 because of davidmyersdev/vite-plugin-node-polyfills#81
- Update vite-plugin-svgr config to handle breaking changes in v4

synapse-react-client
- Expose ThemeTypes module augmentation via package.json `exports`
- Update storybook-vite config to use same SVGR config

apps
- Append `?url` to SVG imports used as URLs so they are handled by Vite
- Add 'vite/client' types for SVG URL import compatibility with TypeScript


apps/SageAccountWeb/src/components/CertificationQuiz.tsx
```diff
@@ -2,7 +2,7 @@ import React from 'react'
 import { Container } from 'react-bootstrap'
 import { Link } from 'react-router-dom'
 import { CertificationQuiz as Quiz } from 'synapse-react-client'
-import ReturnArrow from '../assets/ReturnArrow.svg'
+import ReturnArrow from '../assets/ReturnArrow.svg?url'
```
nickgros, line 5: Upgrading `vite-plugin-svgr` resulted in a breaking change -- the end result is that we cannot import SVGs as strings using the default export AND as React components using `import { ReactComponent as ...} from ...`.

However, Vite natively (no plugin required!) handles importing resource URLs if we add `?url`. The plugin was configured to still support using `import { ReactComponent as ... } from ...`.

apps/SageAccountWeb/src/components/CropImage.ts
```diff
@@ -1,4 +1,4 @@
-import { Area } from 'react-easy-crop/types'
+import { Area } from 'react-easy-crop'
```
nickgros, line 1: `/types` folder is not exported by `react-easy-crop`

apps/SageAccountWeb/tsconfig.json
```diff
@@ -8,7 +8,8 @@
     "resolveJsonModule": true,
     "isolatedModules": true,
     "noFallthroughCasesInSwitch": true,
-    "useUnknownInCatchVariables": false
+    "useUnknownInCatchVariables": false,
+    "types": ["vite/client"]
```
nickgros, line 12: Type support for `*.svg?url` imports

apps/portals/package.json
```diff
@@ -11,6 +11,7 @@
     "@react-hookz/web": "^23.1.0",
     "@sage-bionetworks/react-base-table": "^1.13.4",
     "@sage-bionetworks/synapse-types": "workspace:*",
+    "@tanstack/react-query": "5.22.2",
```
nickgros, line 14: Just relocated automatically by pnpm to be in lexicographical order--not changed

apps/portals/src/configurations/elportal/synapseConfigs/cohortbuilder.ts
```diff
@@ -10,8 +10,8 @@ import { handleSelectedFilesToParticipants } from './handleFilesToParticipants'
 import {
   QueryWrapperSynapsePlotProps,
   QueryWrapperSynapsePlotRowClickEvent,
-} from 'synapse-react-client/src/components/QueryWrapperPlotNav/QueryWrapperSynapsePlot'
```
nickgros, line 13: /src/components/... is not exported by synapse-react-client, so rewrite this import (synapse-react-client was also modified to export these types)

apps/portals/tsconfig.json
```diff
@@ -8,7 +8,9 @@
     "resolveJsonModule": true,
     "isolatedModules": true,
     "noFallthroughCasesInSwitch": true,
-    "noImplicitAny": false
+    "noImplicitAny": false,
+    "moduleResolution": "bundler",
```
nickgros, line 12: `"bundler"` is one of the values for `moduleResolution` permitted by Vite 5. It makes the most sense for all of the packages in this repo (the alternative is `'node16'`/`'nodenext'`)

packages/synapse-react-client/.storybook/main.ts
```diff
@@ -21,7 +21,7 @@ const config: StorybookConfig = {
     options: {
       strictMode: false,
       builder: {
-        viteConfigPath: './.storybook/storybook-vite.config.ts',
+        viteConfigPath: './.storybook/storybook-vite.config.mts',
```
nickgros, line 24: Moved Storybook Vite config to `.mts` so the Storybook manager treats it as an esm source

packages/synapse-react-client/.storybook/storybook-vite.config.mts
```diff
@@ -0,0 +1,6 @@
+import { defineConfig } from 'vite'
+import { vitestConfig } from 'vite-config'
+
+export default defineConfig(
+  vitestConfig,
```
nickgros, line 5: Remove custom storybook config and just use the same config we use for apps!

I think we had a custom config before because we adopted the Storybook + Vite integration when it was pretty nascent, and our app config didn't work in Storybook. That integration has since matured, and our shared config seems to work just fine now.

packages/synapse-react-client/jest.config.cjs
```diff
@@ -3,6 +3,7 @@ const esModules = [
   '@react-hook',
   'lodash-es',
   'nanoid',
+  'mui-one-time-password-input'
```
nickgros, line 6: Upgraded `mui-one-time-password-input` to v2, which only emits ESM. Need to account for that in the Jest config to allow our tests to run.

apps/SageAccountWeb/src/types.d.ts
```diff
@@ -1,2 +1,2 @@
-// Import from 'synapse-react-client/src' is fine here because this is type definitions only
-import 'synapse-react-client/src/theme/ThemeTypes'
+// Import MUI type augmentations from 'synapse-react-client' so we can use custom property values defined for our MUI theme
+import 'synapse-react-client/ThemeTypes'
```
nickgros, line 2: This was simplified in synapse-react-client so we don't have to follow the path

packages/synapse-react-client/package.json
```diff
@@ -14,10 +14,8 @@
       "import": "./dist/*",
       "require": "./dist/*"
     },
-    "./package.json": {
-      "import": "./package.json",
-      "require": "./package.json"
-    }
+    "./package.json": "./package.json",
+    "./ThemeTypes": "./src/theme/ThemeTypes.d.ts"
```
nickgros, line 18: Because we changed our TSConfig `moduleResolution` to now be `bundler`, the apps in the monorepo now use this `exports` block to locate scripts to import into those individual projects.

For this reason, apps can no longer import by doing `synapse-react-client/src/...`. We were trying to prevent this from happening before by convention, but now our TypeScript config enforces it.

For our MUI type augmentations to work (such as our custom Typography variant strings), the apps have to import this `ThemesType` declaration file. This entry allows consumers to import `synapse-react-client/ThemeTypes`.

Note that this isn't portable outside of the monorepo, because the `src` folder isn't uploaded to pnpm, but this is not a new problem.

packages/synapse-react-client/package.json
```diff
@@ -79,7 +77,7 @@
     "markdown-it-synapse-heading": "^1.0.1",
     "markdown-it-synapse-math": "^3.0.5",
     "markdown-it-synapse-table": "^1.0.6",
-    "mui-one-time-password-input": "1.1.0",
+    "mui-one-time-password-input": "^2.0.2",
```
nickgros, line 80: `mui-one-time-password-input` started causing tests to fail. I am not totally sure why, might be CJS/ESM/TypeScript compatibility issues. Upgrading to 2.0.2 and treating it as an ESM-only package seemed to solve all of the related problems.

packages/synapse-react-client/package.json
```diff
@@ -227,18 +225,18 @@
...
     "type-fest": "^3.13.0",
-    "typescript": "5.1.6",
+    "typescript": "5.4.2",
     "util": "^0.12.5",
-    "vite": "^4.5.2",
-    "vite-plugin-svgr": "^3.2.0",
+    "vite": "^5.1.6",
+    "vite-plugin-svgr": "^4.2.0",
     "vite-plugin-turbosnap": "^1.0.2",
+    "vite-config": "workspace:*",
```
nickgros, line 239: We are now using the shared vite-config for Storybook

packages/synapse-react-client/src/components/AccessRequirementList/AccessApprovalCheckMark.tsx
```diff
@@ -6,8 +6,7 @@ import {
 } from '@mui/icons-material'
 import Skeleton from '@mui/material/Skeleton'
 import ConditionalWrapper from '../utils/ConditionalWrapper'
-import { Avatar, AvatarProps, styled, SxProps, useTheme } from '@mui/material'
-import { StyledComponent } from '@emotion/styled/dist/emotion-styled.cjs'
```
nickgros, line 10: `dist/emotion-styled.cjs` is not exported by `@emotion/styled`, so just remove this type

packages/synapse-react-client/src/components/ChallengeSubmission/AvailableEvaluationQueueList.test.tsx
```diff
@@ -1,6 +1,5 @@
 import { render, screen } from '@testing-library/react'
 import userEvent from '@testing-library/user-event'
-import { UserEvent } from '@testing-library/user-event/dist/types/setup/setup'
```
nickgros, line 3: Use of non-exported type

packages/synapse-react-client/src/components/QueryWrapperPlotNav/index.ts
```diff
@@ -1,4 +1,16 @@
 import QueryWrapperPlotNav from './QueryWrapperPlotNav'
 import type { QueryWrapperPlotNavProps } from './QueryWrapperPlotNav'
-export { QueryWrapperPlotNav, QueryWrapperPlotNavProps }
+import QueryWrapperSynapsePlot from './QueryWrapperSynapsePlot'
+import type {
+  QueryWrapperSynapsePlotProps,
+  QueryWrapperSynapsePlotRowClickEvent,
+} from './QueryWrapperSynapsePlot'
```
nickgros, line 7: Re-export these types, which are used in portals

packages/synapse-react-client/src/synapse-client/SynapseClient.test.ts
```diff
@@ -67,6 +67,7 @@ describe('SynapseClient tests', () => {
 
   describe('fetch tests', () => {
     it('fetch error results in a nice network unavailable message', async () => {
+      doGetSpy.mockRestore()
```
nickgros, line 70: Fix a new test failure

packages/synapse-react-client/src/utils/hooks/useDetectSSOCode.test.ts
```diff
@@ -46,16 +47,30 @@ const mockBindOAuthProviderToAccount = jest.spyOn(
...
+}
+
 describe('useDetectSSOCode tests', () => {
+  function renderHook(
+    options: UseDetectSSOCodeOptions,
+    contextOverrides?: Partial<SynapseContextType>,
+  ) {
+    return _renderHook(() => useDetectSSOCode(options), {
+      wrapper: createWrapper(contextOverrides),
+    })
```
nickgros, line 62: This hook was recently changed and it now uses `useSynapseContext`. Instead of mocking `useSynapseContext` in just a few key tests, always use the real thing and contextually pass a fake context object depending on the test.

packages/synapse-react-client/src/utils/hooks/useDetectSSOCode.test.ts
```diff
@@ -95,18 +111,18 @@ describe('useDetectSSOCode tests', () => {
...
       '',
       `/?code=${authorizationCode}&provider=${OAUTH2_PROVIDERS.ORCID}`,
     )
     mockBindOAuthProviderToAccount.mockResolvedValue(successfulLoginResponse)
 
-    const hookReturn = renderHook(() => useDetectSSOCode({ onSignInComplete }))
+    const hookReturn = renderHook(
+      { onSignInComplete },
+      // User is logged in
+      AUTHENTICATED_CONTEXT,
```
nickgros, line 124: Supplying this context is not necessary but helps add context to the test

packages/vite-config/package.json
```diff
@@ -34,15 +34,13 @@
     }
   },
   "devDependencies": {
-    "@esbuild-plugins/node-globals-polyfill": "^0.2.3",
-    "@esbuild-plugins/node-modules-polyfill": "^0.2.2",
```
nickgros, line 38: These are not maintained anymore. Note we still use them to create the UMD bundle for synapse-react-client

packages/vite-config/package.json
```diff
@@ -34,15 +34,13 @@
...
-    "@vitejs/plugin-react": "^4.0.3",
-    "rollup-plugin-polyfill-node": "^0.12.0",
+    "@vitejs/plugin-react": "^4.2.1",
     "svgo": "^3.0.2",
-    "typescript": "5.1.6",
-    "vite": "^4.5.2",
-    "vite-plugin-svgr": "^3.2.0",
+    "typescript": "5.4.2",
+    "vite": "^5.1.6",
+    "vite-plugin-node-polyfills": "0.17.0",
```
nickgros, line 42: This plugin allows us to replace both the esbuild and rollup Node polyfill plugins.

Using v0.17 because there is a bug in later versions (currently v0.21)

packages/vite-config/src/vite-config.ts
```diff
@@ -17,18 +15,19 @@ export const config: UserConfig = {
...
-      plugins: [rollupNodePolyFill()],
-    },
   },
   plugins: [
     react(),
     svgr({
       svgrOptions: {
         plugins: ['@svgr/plugin-jsx'],
         ref: true,
+        exportType: 'named',
```
nickgros, line 25: `named` allows to use named imports as a React component, e.g. `import { ReactComponent as Foo } from 'foo.svg'`

packages/synapse-react-client/src/components/widgets/query-filter/EnumFacetFilter/EnumFacetFilter.integration.test.tsx
```diff
@@ -26,25 +19,13 @@ import { createWrapper } from '../../../../testutils/TestingLibraryUtils'
 import QueryWrapper from '../../../QueryWrapper'
 import { mockTableEntity } from '../../../../mocks/entity/mockTableEntity'
 import { DEBOUNCE_DELAY_MS } from '../../../../utils/hooks/useImmutableTableQuery/useImmutableTableQuery'
-
-jest.mock('../../../../synapse-client', () => {
```
nickgros, line 30: Tests started failing due to this mock breaking. Improved the test surface by using the data provided by mock-service-worker instead of mocking SynapseClient

packages/synapse-react-client/src/mocks/msw/handlers/entityHandlers.ts
```diff
@@ -30,6 +30,22 @@ import { SynapseApiResponse } from '../handlers'
...
+    normalizedPassedIds.includes(normalizeSynPrefix(entity.id)),
+  )
+}
+function getMatchingMockEntity(entityId: string): MockEntityData | undefined {
+  const matches = getMatchingMockEntities([entityId])
+  if (matches.length == 0) {
+    return undefined
+  }
+  return matches[0]
+}
```
nickgros, line 48: Update mock-service-worker functionality to be able to look up mock entities without a `syn` prefix.

packages/synapse-react-client/src/mocks/msw/handlers/entityHandlers.ts
```diff
@@ -231,11 +237,18 @@ export const getEntityHandlers = (backendOrigin: string) => [
...
-        )
-        .map(entity => entity.entityHeader)
+      const entityData: EntityHeader[] = referenceList.references
+        .map(ref => {
+          const entityHeader = getMatchingMockEntity(ref.targetId)?.entityHeader
+          if (entityHeader) {
+            return {
+              ...entityHeader,
+              // ID should match what is passed in (i.e. may or may not have syn prefix)
+              id: ref.targetId,
```
nickgros, line 247: Match behavior of the backend - if the client asks for `syn123`, the returned ID is `syn123`. If the client asks for `123`, the returned ID is `123`.

apps/SageAccountWeb/src/svg.d.ts
```diff
@@ -4,7 +4,4 @@ declare module '*.svg' {
   export const ReactComponent: React.FunctionComponent<
     React.SVGProps<SVGSVGElement> & { title?: string }
   >
-
-  const dataUrl: string
-  export default dataUrl
```
nickgros, line 9: This type of svg import no longer works (you need to import from `'path/to/foo.svg?url'`, so remove it from our custom type definitions

packages/synapse-react-client/src/mocks/msw/handlers/entityHandlers.ts
hallieswan: What do you think about logging a warning if multiple entities with the same id are found? Just wondering if it could help us catch when mock entities are added with ids that have already been used in existing mock entities

hallieswan (approved):
Looks good to me!

PR #811, opened 2024-03-13
SWC-6681: Split definingSQL form from entity description and validate

jinjunoh:
- https://rest-docs.synapse.org/rest/POST/validateDefiningSql.html
- create custom useValidateDefiningSql to handle queries and error response
- split TABLE_SQL and TABLE_NAME to two different steps


packages/synapse-react-client/src/components/CreateTableViewWizard/CreateTableViewWizard.tsx
```diff
@@ -124,6 +126,28 @@ export default function CreateTableViewWizard(
...
+    onSuccess: (isSqlValid: ValidateDefiningSqlResponse) => {
+      if (isSqlValid && isSqlValid.isValid) {
+        setStep('TABLE_NAME')
+      }
+    },
+  })
+
+  const showSqlValidationError: string | null = useMemo(() => {
+    if (sqlValidationError?.message) {
+      return sqlValidationError.message
```
nickgros, line 143: Is this necessary? I think we set up the generics so `useValidateDefiningSql`'s error will always be `SynapseClientError` if it is defined

packages/synapse-react-client/src/components/CreateTableViewWizard/CreateTableViewWizard.tsx
```diff
@@ -124,6 +126,28 @@ export default function CreateTableViewWizard(
...
+    error: sqlValidationError,
+  } = useValidateDefiningSql(sql, entityType!, {
+    onSuccess: (isSqlValid: ValidateDefiningSqlResponse) => {
+      if (isSqlValid && isSqlValid.isValid) {
+        setStep('TABLE_NAME')
+      }
+    },
+  })
+
+  const showSqlValidationError: string | null = useMemo(() => {
```
nickgros, line 141: Since it is a string, suggest renaming
```suggestion
  const sqlValidationError: string | null = useMemo(() => {
```

packages/synapse-react-client/src/components/CreateTableViewWizard/CreateTableViewWizard.tsx
```diff
@@ -206,6 +230,15 @@ export default function CreateTableViewWizard(
     sql,
     viewTypeMask,
   ])
+
+  const onTableSql = useCallback(async () => {
+    try {
+      await validateSql()
+    } catch (e) {
+      return
+    }
```
nickgros, line 239: It's not a good idea to swallow errors, though in this case we know that the error is shown to the user after it is returned by `useMutation`.

To remove this bad pattern, let's use `mutate` instead of `mutateAsync`. `mutate` will not throw on an error, so there is no need to `try/catch`.

I think then `onTableSql` could be deleted (since it would just call another fn) and in `onNextButtonClicked` you can just call `validateSql`

packages/synapse-react-client/src/synapse-queries/table/useDefiningSql.ts
```diff
@@ -0,0 +1,18 @@
...
+export function useValidateDefiningSql(
+  sql: string,
+  entityType: EntityType,
+  options?: UseMutationOptions<ValidateDefiningSqlResponse, SynapseClientError>,
+) {
+  return useMutation<ValidateDefiningSqlResponse, SynapseClientError>({
+    ...options,
+    mutationFn: () => validateDefiningSql(sql, entityType),
+  })
+}
```
nickgros, line 18: This doesn't look right, the `sql` and `entityType` args should be passed to the `mutationFn`, not the hook

packages/synapse-react-client/src/components/CreateTableViewWizard/CreateTableViewWizard.tsx
nickgros: May or may not want to keep this fn after updating the mutation hook per [this comment](https://github.com/Sage-Bionetworks/synapse-web-monorepo/pull/811/files#r1523625082) as you will still need to call `validateSql({sql, entityType})`

PR #820, opened 2024-03-14
SWC-6571:  DropdownSelect component to make dropdown menu visible on …

jinjunoh:
…hover


jinjunoh:
On Production 
Screenshot 2024-03-15 at 11 36 33 AM

On Feature 
Screenshot 2024-03-15 at 11 33 51 AM

packages/synapse-react-client/src/components/EntityFinder/tree/EntityTree.tsx
```diff
@@ -413,38 +413,22 @@ export function EntityTree(props: EntityTreeProps) {
...
-                  >
-                    {scopeOption}
-                  </Dropdown.Item>
-                )
-              })}
-            </Dropdown.Menu>
-          </Dropdown>
+                ),
+            )}
+            selectedIndex={Object.values(FinderScope).indexOf(scope)}
```
hallieswan, line 425: When `projectId` is undefined, `options` is a subset of `FinderValues`. However, `selectedIndex` is calculated using the full list of `FinderScope` values. Will the expected option be displayed? 

For example, when `initialScope` is `FinderScope.ALL_PROJECTS` and `projectId` is undefined, we would expect that 'All Projects' would be displayed.

packages/synapse-react-client/src/components/EntityFinder/tree/EntityTree.tsx
```diff
@@ -413,38 +413,22 @@ export function EntityTree(props: EntityTreeProps) {
...
+            setSelectedIndex={index => {
+              const selectedScope = Object.values(FinderScope)[index]
+              setScope(selectedScope)
+              setCurrentContainer(getScopeOptionDefaultContainer(selectedScope))
+            }}
```
hallieswan, line 430: Should this [still check](https://github.com/Sage-Bionetworks/synapse-web-monorepo/blob/506a12d9afd9ef45b99f94b2d423c50dcec24db1/packages/synapse-react-client/src/components/EntityFinder/tree/EntityTree.tsx#L434) whether the scope has changed (e.g. `scope !== selectedScope`) before updating the state with `setScope` and `setCurrentContainer`? I'm wondering if this check was added to prevent unnecessary re-renders, but @nickgros may know more.

hallieswan (changes requested):
Looking good! A couple questions to consider

packages/synapse-react-client/src/components/DropdownSelect/DropdownSelect.tsx
```diff
@@ -71,8 +71,14 @@ export default function DropdownSelect(props: DropdownSelectProps) {
...
     <React.Fragment>
-      <ButtonGroup variant={variant} ref={anchorRef} aria-label="split button">
+      <ButtonGroup
+        variant={variant}
+        ref={anchorRef}
+        aria-label="split button"
+        size="small"
+      >
         <Button
+          className="optionsButton"
```
nickgros, line 81: This component is used in other places so the `size` and `className` props should be passed as their own props

packages/synapse-react-client/src/components/EntityFinder/tree/EntityTree.tsx
nickgros: I believe the purpose of the check is to avoid resetting `currentContainer` when `scope` doesn't change.

packages/synapse-react-client/src/style/components/entity_finder/_tree-view.scss
```diff
@@ -76,4 +76,8 @@
       font-size: large;
     }
   }
+
+  .optionsButton {
```
nickgros, line 80: You might be able to use the selector `.Header .MuiButton-root` so you don't have to worry about passing a `className` prop to the DropdownSelect

packages/synapse-react-client/src/components/DropdownSelect/DropdownSelect.tsx
nickgros: I suggest not worrying about passing a className to `Button` and just have `DropdownSelectProps` extend `ButtonGroup` props, then forward extra props to the `ButtonGroup` component

nickgros (approved):
Looks great!

PR #839, opened 2024-03-20
SWC-6734:  Provide way to send back progress during upload, as well as ability to cancel the upload in progress

jay-hodgson:


packages/synapse-react-client/src/synapse-client/SynapseClient.ts
```diff
@@ -1788,8 +1788,14 @@ export const signOut = async () => {
...
 
+// Progress sent back during upload, will report when a part of a file is finished processing
+export type ProgressCallback = {
+  value: number
+  total: number
+}
+
 /**
- * Upload file.  Note that this currently only supports Synapse storage (Sage s3 bucket)
+ * Upload file to Synapse, creating a FileHandle
```
jay-hodgson, line 1798: Verified that I could use this code to upload to a custom storage location ID.
Screen Shot 2024-03-20 at 10 30 49 AM

packages/synapse-react-client/src/synapse-client/SynapseClient.ts
```diff
@@ -2019,6 +2053,16 @@ export const startMultipartUpload = (
         .split('')
         .map(bit => bit === '1')
       status.clientSidePartsState = clientSidePartsState
+      let progress = 0
```
jay-hodgson, line 2056: progress is reported by number of file parts that we have processed, so the fidelity has been reduced (used to report on individual bytes in Synapse.org).  But for any file (large with multiple parts, or smaller than a single part) where the user has a reasonably good connection, the UX will be ok.

nickgros (approved):
These are great improvements, so I am happy to approve this, with a few suggested changes. 

I do have concerns about existing code quality and test quality for such a critical feature. We do capture file upload in our SWC e2e tests. Is that good enough?

packages/synapse-react-client/src/synapse-client/SynapseClient.ts
```diff
@@ -1985,17 +2003,31 @@ const uploadFilePart = async (
...
   // The parent would still need to figure out progress (for the total file set).
   await fetch(presignedUrl, {
     method: 'PUT',
     headers: {
       'Content-Type': contentType,
     },
     body: file,
+    signal,
   })
+  clearInterval(timer)
```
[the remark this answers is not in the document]
jay-hodgson: YES!

nickgros (approved):

PR #849, opened 2024-03-22
PORTALS-3004

jay-hodgson:


jay-hodgson:
Screen Shot 2024-03-22 at 10 15 31 AM

And existing NF portal (that has background assets) renders properly:
Screen Shot 2024-03-22 at 10 14 37 AM

packages/synapse-react-client/src/components/Goals/Goals.tsx
```diff
@@ -69,24 +69,21 @@ export const Goals: React.FC<GoalsProps> = (props: GoalsProps) => {
         const assets = (queryResultBundle?.queryResult!.queryResults.rows.map(
           el => el.values[assetColumnIndex],
         ) ?? []) as string[]
-        if (assets.some(asset => asset === null)) {
```
jay-hodgson, line 72: We filter out all nulls, so this warning no longer makes sense

hallieswan (approved):
Looks great!

PR #851, opened 2024-03-22
SWC-6728 - Prompt for 2FA if necessary when changing password

nickgros:
- Remove ChangePassword component in SageAccountWeb and use synapse-react-client component
- Refactor login/2FA components to create `OneTimePasswordForm` to be reused by `ChangePassword`
- `ChangePassword` now handles 2FA case
- `ChangePassword` now uses react-query hooks
- Move SynapseToastContainer to be within theme so toasts are styled correctly
- Update logo/favicon for SageAccountWeb


packages/synapse-react-client/src/components/Authentication/LoginForm.tsx
```diff
@@ -62,29 +58,18 @@ export default function LoginForm(props: Props) {
...
-            onClick={() => onStepChange('RECOVERY_CODE')}
-          >
-            Use a backup code instead
-          </Link>
-        </>
-      )}
-      {step === 'RECOVERY_CODE' && (
-        <RecoveryCodeForm
+      {(step === 'VERIFICATION_CODE' || step === 'RECOVERY_CODE') && (
+        <OneTimePasswordForm
```
nickgros, line 62: Encapsulated both TOTPForm and RecoveryCodeForm in OneTimePasswordForm

packages/synapse-react-client/src/components/ChangePassword/ChangePassword.integration.test.tsx
```diff
@@ -0,0 +1,440 @@
...
+      newPasswordField,
+      confirmPasswordField,
+      submitButton,
+    } = setUp()
+
+    await user.type(currentPasswordField, currentPassword)
+    await user.type(newPasswordField, newPassword)
+    await user.type(confirmPasswordField, newPassword)
+
+    await userEvent.click(submitButton)
```
hallieswan, line 95: Does `userEvent` need to be used here or could `user` be used instead? If `user` works, then could update the other instances in this test suite as well

```suggestion
    await user.click(submitButton)
```

packages/synapse-react-client/src/components/ChangePassword/ChangePassword.integration.test.tsx
```diff
@@ -0,0 +1,440 @@
...
+          'org.sagebionetworks.repo.model.auth.ChangePasswordWithCurrentPassword',
+        username: mockUserProfileData.userName,
+        newPassword: newPassword,
+        currentPassword: currentPassword,
+      })
+    })
+
+    // Type the code. Once the code is entered, the form should submit automatically
+    for (let i = 0; i < otpInputs.length; i++) {
+      await userEvent.type(otpInputs[i], String(i + 1))
```
hallieswan, line 257: Is it possible to use `user` here as well?

```suggestion
      await user.type(otpInputs[i], String(i + 1))
```

packages/synapse-react-client/src/components/ChangePassword/ChangePassword.integration.test.tsx
```diff
@@ -0,0 +1,440 @@
...
+    await userEvent.click(useRecoveryCodeButton)
+
+    const recoveryCode = '1234-5678-abcd-edcb'
+    const recoveryCodeTextbox = await screen.findByPlaceholderText(
+      'Enter backup code',
+    )
+    await user.type(recoveryCodeTextbox, recoveryCode)
+    await user.click(screen.getByRole('button', { name: 'Submit' }))
+
+    // Type the code. Once the code is entered, the form should submit automatically
```
hallieswan, line 328: Looks like this comment may refer to the TOTP form, since the recovery code form requires the user to click submit. Should the comment be removed or updated?

packages/synapse-react-client/src/components/ChangePassword/ChangePassword.integration.test.tsx
```diff
@@ -0,0 +1,440 @@
...
+        username: mockUserProfileData.userName,
+        newPassword: newPassword,
+        currentPassword: currentPassword,
+      })
+    })
+
+    // Type the code. Once the code is entered, the form should submit automatically
+    for (let i = 0; i < otpInputs.length; i++) {
+      await userEvent.type(otpInputs[i], String(i + 1))
+    }
```
hallieswan, line 417: Could consider pulling this into a method that can be reused in this test and in the 'supports entering a TOTP if the server requires 2FA' test, but up to you!

hallieswan (approved):
Looks great, super clear tests! Just a couple small comments

PR #852, opened 2024-03-25
SWC-6751

nickgros:
- Fix issue where child nodes could be added to the `VirtualizedTree` component
- Attempt to simplify tree node generator function
- Update VirtualizedTree tests to be linted/type-checked

### DIFF
--- a/packages/synapse-react-client/.storybook/main.mts
+++ b/packages/synapse-react-client/.storybook/main.mts
@@ -18,7 +18,7 @@ const config: StorybookConfig = {
   framework: {
     name: '@storybook/react-vite',
     options: {
-      strictMode: false,
+      strictMode: true,
       builder: {
         viteConfigPath: './.storybook/storybook-vite.config.mts',
       },
--- a/packages/synapse-react-client/src/components/EntityFinder/tree/VirtualizedTree.test.tsx
+++ b/packages/synapse-react-client/src/components/EntityFinder/tree/VirtualizedTree.test.tsx
@@ -8,15 +8,17 @@ import { mockAllIsIntersecting } from 'react-intersection-observer/test-utils'
 import { VariableSizeNodePublicState } from 'react-vtree'
 import { NodeComponentProps } from 'react-vtree/dist/es/Tree'
 import {
+  EntityHeaderNode,
   EntityTreeNodeType,
   getNodeData,
   getTreeWalkerFunction,
   Node,
   RootNodeConfiguration,
   TreeData,
-} from '../../../../src/components/EntityFinder/tree/VirtualizedTree'
-import { createWrapper } from '../../../../src/testutils/TestingLibraryUtils'
+} from './VirtualizedTree'
+import { createWrapper } from '../../../testutils/TestingLibraryUtils'
 import { EntityType } from '@sage-bionetworks/synapse-types'
+import { WritableDeep } from 'type-fest'
 
 describe('VirtualizedTree tests', () => {
   beforeEach(() => {
@@ -77,12 +79,12 @@ describe('VirtualizedTree tests', () => {
     })
 
     it('General test for entity header', () => {
-      const node = {
+      const node: EntityHeaderNode = {
         id: 'syn123',
         type: 'org.sagebionetworks.repo.model.Folder',
         name: 'A Custom Folder Name',
-        fetchNextPage: async () => {},
-        hasNextPage: false,
+        children: [],
+        childrenNextPageToken: null,
       }
       const actual = getNodeData({
         node,
@@ -104,7 +106,7 @@ describe('VirtualizedTree tests', () => {
           id: node.id,
           node,
           getNextPageOfChildren,
-          isLeaf: false,
+          isLeaf: true,
           isOpenByDefault: true,
           nestingLevel,
           setSelectedId,
@@ -120,12 +122,12 @@ describe('VirtualizedTree tests', () => {
     it('isOpenByDefault matches the value of autoExpand', () => {
       const autoExpandTrue = jest.fn().mockReturnValue(true)
       const autoExpandFalse = jest.fn().mockReturnValue(false)
-      const node = {
+      const node: EntityHeaderNode = {
         id: 'syn123',
         name: 'A Custom Folder Name',
+        type: 'org.sagebionetworks.repo.model.Folder',
         children: [],
-        fetchNextPage: async () => {},
-        hasNextPage: false,
+        childrenNextPageToken: null,
       }
 
       let isOpenByDefault = getNodeData({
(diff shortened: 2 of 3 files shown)


packages/synapse-react-client/src/components/EntityFinder/tree/VirtualizedTree.test.tsx
```diff
@@ -77,12 +79,12 @@ describe('VirtualizedTree tests', () => {
     })
 
     it('General test for entity header', () => {
-      const node = {
+      const node: EntityHeaderNode = {
```
nickgros, line 82: Fixed types/lint warnings in this test file and moved it so it will be type checked and linted.

I did not add a new test to capture the fix; the existing code structure is just too hard to test the changed function.

packages/synapse-react-client/.storybook/main.mts
```diff
@@ -18,7 +18,7 @@ const config: StorybookConfig = {
   framework: {
     name: '@storybook/react-vite',
     options: {
-      strictMode: false,
+      strictMode: true,
```
nickgros, line 21: StrictMode seems to be working, so enable it and hopefully we can catch more bugs like this

packages/synapse-react-client/src/components/EntityFinder/tree/VirtualizedTree.tsx
```diff
@@ -100,6 +104,18 @@ function isLeafNode(node: TreeNode) {
...
+  node: TreeNode,
+): node is RootNodeConfiguration | EntityHeaderNode {
+  if (isPaginationNode(node)) {
+    return false
+  } else if (isRootNodeConfiguration(node)) {
+    return node.children.length > 0 || node.hasNextPage
+  } else {
+    return !isLeafNode(node)
+  }
+}
```
nickgros, line 117: Add utility for generator function

packages/synapse-react-client/src/components/EntityFinder/tree/VirtualizedTree.tsx
```diff
@@ -77,7 +81,7 @@ function isRootNodeConfiguration(
 
 function hasMoreChildren(node: TreeNode) {
   if (isPaginationNode(node)) {
-    return true
+    return false
```
nickgros, line 84: Our "pagination nodes" (empty/placeholders that contain an intersection observer just used to fetch more data) do not have children, so this was confusing. Today, this function is never called with a pagination node, so this shouldn't change anything.

packages/synapse-react-client/src/components/EntityFinder/tree/VirtualizedTree.tsx
```diff
@@ -402,33 +418,29 @@ export function getTreeWalkerFunction(
       // Yielding `undefined` indicates to react-vtree that we've finished iterating children from the last time we were in the loop.
       const parentMeta = yield
```
nickgros, line 420: This generator tells `react-vtree` how to render the tree. I made small changes to the logic/flow to try to make it a bit easier to follow.

packages/synapse-react-client/src/components/EntityFinder/tree/VirtualizedTree.tsx
```diff
@@ -523,27 +538,44 @@ export const VirtualizedTree = (props: VirtualizedTreeProps) => {
...
+        nextPageToken: node.childrenNextPageToken,
+        includeTypes: visibleTypes,
+      }
+      const children = await queryClient.fetchQuery({
+        queryKey: keyFactory.getEntityChildrenQueryKey(
+          entityChildrenRequest,
+          false,
+        ),
+        queryFn: () => getEntityChildren(entityChildrenRequest, accessToken),
+      })
```
nickgros, line 552: Since this runs in a callback, we can't use `useQuery`, but we can still manually use/store query data using `queryClient`.

packages/synapse-react-client/src/components/EntityFinder/tree/VirtualizedTree.tsx
```diff
@@ -523,27 +538,44 @@ export const VirtualizedTree = (props: VirtualizedTreeProps) => {
...
       // Update the node data -- add the children and store the nextPageToken
       if (node.children) {
-        node.children.push(...children.page)
+        for (const newChild of children.page) {
+          // Prevent adding duplicate entries
+          if (!node.children.find(child => child.id == newChild.id)) {
+            node.children.push(newChild)
+          }
+        }
```
nickgros, line 560: It's challenging to figure out how to prevent this from being called twice, since it's a complex mix of our code and `react-vtree`. It's comparatively much easier to just prevent appending a duplicate when it happens.

packages/synapse-react-client/src/components/EntityFinder/tree/VirtualizedTree.tsx
```diff
@@ -523,27 +538,44 @@ export const VirtualizedTree = (props: VirtualizedTreeProps) => {
...
       node.childrenNextPageToken = children.nextPageToken
 
+      // Ensure the node for which we fetched children is expanded/open
+      if (treeInstance.current) {
+        await treeInstance.current.recomputeTree({
+          [node.id]: {
+            open: true,
+          },
+        })
+      }
```
nickgros, line 573: I think there's another issue where sometimes the tree doesn't expand all the way after data is fetched. This should make that more reliable

packages/synapse-react-client/src/components/EntityFinder/tree/VirtualizedTree.tsx
hallieswan: Nice! If `useGetEntityChildren` was updated in the future, would this code need to be similarly updated? Just wondering whether that would be obvious to the dev making the change or whether it would be helpful to add a comment to remind us to update this code if `useGetEntityChildren` is updated

packages/synapse-react-client/src/components/EntityFinder/tree/VirtualizedTree.tsx
hallieswan: Would you mind adding the ticket number to this comment and/or adding some of the context you've provided here for why checking for duplicates is necessary?

hallieswan (approved):
Looks good to me! Just a few small comments

packages/synapse-react-client/src/components/EntityFinder/tree/VirtualizedTree.tsx
nickgros: This is a general issue, but I am not convinced comments would help very much.

If the `keyFactory.getEntityChildrenQueryKey` signature changes then we would catch it but that might not be the case.

One option I have thought a little bit about is that it might make sense to set up some data structure to bind all of our query keys and their corresponding query functions, and by convention only pull queries from that structure. Thought that might be awkward to use if the query functions rely on some resource (for example, an instance of a class generated by the OpenAPI spec that stores auth state and makes API calls).

hallieswan: Thanks for the context, it does seem like a tricky problem to solve! Agreed that comments probably aren't the solution

PR #855, opened 2024-03-25
GENIE: enable Send to CAVATICA

jay-hodgson:


jay-hodgson:
Confirmed integration:
Screen Shot 2024-03-25 at 1 03 45 PM

PR #868, opened 2024-04-05
Upgrade eslint plugins

nickgros:
- Major change is bumping @typescript-eslint packages from v5 to v7
  - Use @typescript-eslint/recommended-type-checked
  - Warn on @typescript-eslint/no-explicit-any (no change), @typescript-eslint/no-redundant-type-constituents (newly enabled rule)
- Update synapse-react-client to fix new lint errors


packages/synapse-react-client/src/assets/icons/Sort.tsx
```diff
@@ -21,7 +21,7 @@ const SortIcon: React.FunctionComponent<
     fill="none"
     xmlns="http://www.w3.org/2000/svg"
   >
-    {(direction as string).toUpperCase() === Direction.DESC ? (
+    {(direction as string).toUpperCase() === 'DESC' ? (
```
nickgros, line 24: New lint rule checks that we only compare enum to enum and string to string. Since `direction` is a string, we should compare it to the string `'DESC'`.

packages/synapse-react-client/src/components/AccessRequirementList/RequirementItem/RequirementItem.tsx
```diff
@@ -95,7 +95,11 @@ export default function RequirementItem(props: RequirementItemProps) {
           sx={{ gridColumnStart: 2, gridRow: '3 / span 1', mt: 2 }}
         >
           {actions.map((buttonProps, i) => (
-            <Button key={i} disabled={status === 'LOADING'} {...buttonProps} />
+            <Button
+              key={i}
+              disabled={status === RequirementItemStatus.LOADING}
```
nickgros, line 100: [@typescript-eslint/no-unsafe-enum-comparison](https://typescript-eslint.io/rules/no-unsafe-enum-comparison/) - use the enum on both sides

packages/synapse-react-client/src/components/ChallengeSubmission/ChallengeSubmissionStepper.tsx
```diff
@@ -92,7 +92,7 @@ function ChallengeSubmissionStepper({
     entityType === EntityType.DOCKER_REPO
       ? steps.SELECT_COMMIT
       : steps.SELECT_EVALUATION
-  const [step, setStep] = useState<Step>(initialStep)
+  const [step, setStep] = useState<Step<StepsEnum>>(initialStep)
```
nickgros, line 95: This was also driven by [@typescript-eslint/no-unsafe-enum-comparison](https://typescript-eslint.io/rules/no-unsafe-enum-comparison/). In unchanged code, we compare the current `step.id` to members of the `StepsEnum` that is specific to this component.

I added an optional generic argument to `Step` so in this component we can specify that the `id` must be a member of `StepsEnum`, making the comparison safer, and making eslint happy.

packages/synapse-react-client/src/components/Forum/ForumTable.tsx
```diff
@@ -65,7 +65,7 @@ export const ForumTable: React.FC<ForumTableProps> = ({
                   aria-label="Sort by Topic"
                   active={sort === DiscussionThreadOrder.THREAD_TITLE}
                   direction={
-                    sort === 'THREAD_TITLE'
+                    sort === DiscussionThreadOrder.THREAD_TITLE
```
nickgros, line 68: More [@typescript-eslint/no-unsafe-enum-comparison](https://typescript-eslint.io/rules/no-unsafe-enum-comparison/)

packages/synapse-react-client/src/components/StepperDialog/StepperDialog.tsx
```diff
@@ -5,8 +5,8 @@ import { SynapseSpinner } from '../LoadingScreen/LoadingScreen'
 import SpinnerButton from '../SpinnerButton/SpinnerButton'
 import { DialogContentProps } from '@mui/material'
 
-export type Step = {
-  id: string
+export type Step<TStepId = string> = {
+  id: TStepId
```
nickgros, line 9: Add optional generic to `Step` as described in ChallengeSubmissionStepper.tsx

packages/synapse-react-client/src/synapse-client/HttpClient.ts
```diff
@@ -32,24 +32,25 @@ const MAX_RETRY = 3
  * @throws SynapseClientError
  */
 const fetchWithExponentialTimeout = async <TResponse>(
-  url: RequestInfo,
+  requestInfo: RequestInfo,
```
nickgros, line 35: This is not always a URL, so update the parameter name to clarify

packages/synapse-react-client/src/synapse-client/HttpClient.ts
```diff
@@ -32,24 +32,25 @@ const MAX_RETRY = 3
...
 ): Promise<TResponse> => {
+  const url = typeof requestInfo === 'string' ? requestInfo : requestInfo.url
   let response
   try {
-    response = await fetch(url, options)
+    response = await fetch(requestInfo, options)
   } catch (err) {
     console.error(err)
-    throw new SynapseClientError(0, NETWORK_UNAVAILABLE_MESSAGE, url.toString())
+    throw new SynapseClientError(0, NETWORK_UNAVAILABLE_MESSAGE, url)
```
nickgros, line 45: The `url.toString()` was what triggered a new eslint error in the old code because it may not have been a string. With the changes above, `url` is now always a string.

packages/synapse-react-client/src/utils/functions/EntityTypeUtils.ts
```diff
@@ -126,41 +126,32 @@ export function entityTypeToFriendlyName(entityType: EntityType): string {
 export function convertToEntityType(
   typeString: string | ENTITY_CONCRETE_TYPE | EntityType,
 ): EntityType {
+  if (Object.values(EntityType).includes(typeString as EntityType)) {
+    return typeString as EntityType
+  }
```
nickgros, line 131: Changes in this file were also flagged by [@typescript-eslint/no-unsafe-enum-comparison](https://typescript-eslint.io/rules/no-unsafe-enum-comparison/). It didn't like that we were comparing `typeString` (a string) to `EntityType` via `switch`.

Since the goal of this function is to return an `EntityType`, change the logic to see if `typeString` is an entity type first, then do the comparison. This function is well-tested, so this should be a safe change.

hallieswan (approved):
Looks good to me!

PR #879, opened 2024-04-15
SWC-6730a: add WikiMarkdownEditor

hallieswan:
Adds an initial SRC implementation of [SWC's WikiMarkdownEditor](https://github.com/Sage-Bionetworks/SynapseWebClient/blob/develop/src/main/java/org/sagebionetworks/web/client/widget/entity/WikiMarkdownEditor.java). Will need to add ability to delete a WikiPage in the future (SWC-6774).

Note: differs from SWC implementation in that a new root wiki page is created when the user explicitly clicks the `WikiMarkdownEditorButton`, rather than automatically when the `WikiMarkdownEditor` is opened.

Root WikiPage: 
WikiMarkdownEditor_rootWikiPage

Sub WikiPage:
WikiMarkdownEditor_subWikiPage


packages/synapse-react-client/src/components/Forum/ForumThreadEditor.stories.tsx
```diff
@@ -1,15 +1,15 @@
 import { Meta, StoryObj } from '@storybook/react'
-import { ForumThreadEditor } from './ForumThreadEditor'
 import { fn } from '@storybook/test'
+import { ForumThreadEditor } from './ForumThreadEditor'
 
 const meta = {
-  title: 'Markdown/MarkdownEditor',
+  title: 'Synapse/ForumThreadEditor',
   component: ForumThreadEditor,
```
hallieswan, line 7: Create separate stories for ForumThreadEditor and MarkdownEditor to clarify each component's behavior

packages/synapse-react-client/src/components/Markdown/MarkdownEditor.tsx
```diff
@@ -156,6 +156,7 @@ export const MarkdownEditor: React.FunctionComponent<MarkdownEditorProps> = ({
       <div>
         {currentTab === 'WRITE' ? (
           <textarea
+            aria-label="markdown"
```
hallieswan, line 159: Added an aria-label so the markdown field is easier to find in tests

packages/synapse-react-client/src/components/WikiMarkdownEditor/WikiMarkdownEditor.test.tsx
```diff
@@ -0,0 +1,442 @@
...
+    expect(queryTitleField()).toBeNull()
+  })
+
+  test('displays title for non-root WikiPage', async () => {
+    setUp(defaultExistingWikiPageProps)
+    await waitForWikiLoaded()
+    expect(queryTitleField()).not.toBeNull()
+  })
+
+  test('warns user before cancelling if markdown has changed', async () => {
```
hallieswan, line 278: https://github.com/Sage-Bionetworks/synapse-web-monorepo/assets/26949006/f66f604d-03c6-42aa-94c6-f4a29e649fd5

packages/synapse-react-client/src/components/WikiMarkdownEditor/WikiMarkdownEditor.tsx
```diff
@@ -0,0 +1,256 @@
...
+  const [title, setTitle] = useState<string>('')
+  const [markdown, setMarkdown] = useState<string>('')
+  const [showConfirmCancelDialog, setShowConfirmDialog] =
+    useState<boolean>(false)
+  const [error, setError] = useState<string | null>(null)
+
+  const { data: rootWikiPageKey, error: errorLoadingRootWikiPageKey } =
+    useGetRootWikiPageKey(ownerObjectType, ownerObjectId, {
+      enabled: wikiPageId === undefined,
+    })
```
hallieswan, line 63: When a specific `wikiPageId` is not provided, get the `WikiPageKey` for the root `WikiPage`.

packages/synapse-react-client/src/components/WikiMarkdownEditor/WikiMarkdownEditor.tsx
```diff
@@ -0,0 +1,256 @@
...
+    setError(null)
+
+    // an existing root WikiPageKey was found
+    if (rootWikiPageKey) {
+      setWikiPageId(rootWikiPageKey.wikiPageId)
+    }
+
+    // root WikiPageKey was not found (so we have create permission),
+    // then create a root WikiPage for this ownerObject
+    if (rootWikiPageKey === null) {
```
hallieswan, line 85: And attempt to create a new root WikiPage, if one does not already exist

packages/synapse-react-client/src/mocks/msw/handlers/wikiHandlers.ts
```diff
@@ -1,34 +1,133 @@
...
-import { ACCESS_REQUIREMENT_WIKI_PAGE } from '../../../utils/APIConstants'
-import { FileHandleResults, WikiPage } from '@sage-bionetworks/synapse-types'
+import { WIKI_OBJECT_TYPE, WIKI_PAGE } from '../../../utils/APIConstants'
+import { MOCK_WIKI_ETAG, NEW_WIKI_PAGE_ID, mockWikiPages } from '../../mockWiki'
+import { mockWikiPageKeys } from '../../mockWikiPageKey'
+import { MOCK_USER_ID } from '../../user/mock_user_profile'
 import { SynapseApiResponse } from '../handlers'
-import { mockWikiPages } from '../../mockWiki'
 
-export const getWikiHandlers = (backendOrigin: string) => [
```
hallieswan, line 7: Updated to also handle calls for other wiki owner object types (ENTITY, EVALUATION)

packages/synapse-react-client/src/synapse-client/SynapseClient.ts
```diff
@@ -1166,13 +1168,6 @@ export const getEntityBundleV2 = <T extends EntityBundleRequest>(
   )
 }
 
-/**
- * Get a corresponding string value of ObjectType:
- **/
-function getObjectTypeToString(key: ObjectType) {
-  return ObjectType[key]
-}
-
```
hallieswan, line 1175: Moved into APIConstants

packages/synapse-react-client/src/components/WikiMarkdownEditor/WikiMarkdownEditor.tsx
hallieswan: Suspect that the SWC implementation follows this path because there are services to GET a specific WikiPage by id for [ENTITY](https://rest-docs.synapse.org/rest/GET/entity/ownerId/wiki/wikiId.html), [EVALUATION](https://rest-docs.synapse.org/rest/GET/evaluation/ownerId/wiki/wikiId.html), and [ACCESS_REQUIREMENT](https://rest-docs.synapse.org/rest/GET/access_requirement/ownerId/wiki/wikiId.html) owner object types. However, there are only services to get the root WikiPage for [ENTITY](https://rest-docs.synapse.org/rest/GET/entity/ownerId/wiki.html) and [EVALUATION](https://rest-docs.synapse.org/rest/GET/evaluation/ownerId/wiki.html), not for ACCESS_REQUIREMENT. 

So, when a specific wikiPageId is provided, we can directly GET the WikiPage. However, when a wikiPageId is not provided, then first get the root WikiPageKey (since there are services to get root WikiPageKey for [ENTITY](https://rest-docs.synapse.org/rest/GET/evaluation/ownerId/wikikey.html), [EVALUATION](https://rest-docs.synapse.org/rest/GET/entity/ownerId/wiki/wikiId.html), AND [ACCESS_REQUIREMENT](https://rest-docs.synapse.org/rest/GET/access_requirement/ownerId/wikikey.html) owner object types), then get the root WikiPage by id.

packages/synapse-react-client/src/components/WikiMarkdownEditor/WikiMarkdownEditor.tsx
```diff
@@ -0,0 +1,256 @@
...
+
+  const [wikiPageId, setWikiPageId] = useState<string | undefined>(
+    initialWikiPageId,
+  )
+  const [open, setOpen] = useState<boolean>(true)
+  const [title, setTitle] = useState<string>('')
+  const [markdown, setMarkdown] = useState<string>('')
+  const [showConfirmCancelDialog, setShowConfirmDialog] =
+    useState<boolean>(false)
+  const [error, setError] = useState<string | null>(null)
```
nickgros, line 58: Is it possible to simplify the logic in your effects by deriving the `error` in one function/hook, instead of setting it via callbacks + effects?

i.e. 
```ts

// pull the error out of the mutation hooks instead of using `onError`
const { mutate: createWikiPage, error: createError } = useCreateWikiPage({
	// ...
})

const error = useMemo(() => {
	if (errorLoadingRootWikiPageKey) {
 		return ERROR_LOADING_WIKI_FAILED // + whatever
	} else if (createError) {
		return createError.reason // or whatever
	} else if (updateError) {
		 //...
	}
	return null;
})

```

Your use memo logic might need to be slightly more complicated than this example, but it would make it easier to trace where an error comes from

packages/synapse-react-client/src/components/WikiMarkdownEditor/WikiMarkdownEditor.tsx
```diff
@@ -0,0 +1,256 @@
...
+
+  useEffect(() => {
+    setError(null)
+
+    // an existing root WikiPageKey was found
+    if (rootWikiPageKey) {
+      setWikiPageId(rootWikiPageKey.wikiPageId)
+    }
+
+    // root WikiPageKey was not found (so we have create permission),
```
nickgros, line 83: so it 403s / hook data is `undefined` if you don't have permission to create?

packages/synapse-react-client/src/components/WikiMarkdownEditor/WikiMarkdownEditor.tsx
nickgros: Why do we unconditionally create this in an effect instead of waiting for the user to try to explicitly save a wiki? Is the idea that, if this is open, you already chosen to create the wiki page?

nickgros: _If_ this effect is the right place to create the wiki (which I am not convinced), how are we assured that this hook doesn't fire twice, triggering two 'create' calls? In that case I think we should also prevent calling `createWikiPage` if the existing `createWikiPage` mutation `isPending`

packages/synapse-react-client/src/components/WikiMarkdownEditor/WikiMarkdownEditor.tsx
```diff
@@ -0,0 +1,256 @@
...
+    setError(null)
+
+    if (wikiPage) {
+      setTitle(wikiPage.title)
+      setMarkdown(wikiPage.markdown)
+    }
+    if (errorLoadingWikiPage) {
+      setError(ERROR_LOADING_WIKI_FAILED + errorLoadingWikiPage.reason)
+    }
+  }, [errorLoadingWikiPage, wikiPage])
```
nickgros, line 140: Ideally if you can derive the error instead of using a state variable as described above, this would simplify to:
```suggestion
  useEffect(() => {
    if (wikiPage) {
      setTitle(wikiPage.title)
      setMarkdown(wikiPage.markdown)
    }
  }, [wikiPage])
```

packages/synapse-react-client/src/components/WikiMarkdownEditor/WikiMarkdownEditor.test.tsx
```diff
@@ -0,0 +1,442 @@
...
+describe('WikiMarkdownEditor', () => {
+  beforeEach(() => jest.clearAllMocks())
+  beforeAll(() => server.listen())
+  afterEach(() => server.resetHandlers())
+  afterAll(() => server.close())
+
+  test('creates a root WikiPage when there was not an existing root WikiPage', async () => {
+    const { spinner } = setUp(defaultNoRootProps)
+
+    expect(spinner).toHaveLength(1)
```
nickgros, line 128: I am surprised this consistently passes, I am always worried about checking something ephemeral in UI tests. If this works, that's great.

packages/synapse-react-client/src/components/WikiMarkdownEditor/WikiMarkdownEditor.test.tsx
```diff
@@ -0,0 +1,442 @@
...
+    server.use(
+      rest.get(
+        `${getEndpoint(BackendDestinationEnum.REPO_ENDPOINT)}${WIKI_OBJECT_TYPE(
+          defaultExistingWikiPageProps.ownerObjectType,
+        )}/:ownerObjectId/wikikey`,
+        async (req, res, ctx) => {
+          return res(ctx.status(403), ctx.json(errorResponse))
+        },
+      ),
+    )
```
nickgros, line 217: I haven't been able to reproduce this behavior on a real stack instance

nickgros (commented):
Looks great for the most part, had some questions/concerns about mutations + effects and reliance on some server behavior that I couldn't reproduce

packages/synapse-react-client/src/components/WikiMarkdownEditor/WikiMarkdownEditor.tsx
hallieswan: I was following the SWC implementation, which [creates a root WikiPage](https://github.com/Sage-Bionetworks/SynapseWebClient/blob/9085d9d2f9644ba7ceb9db9624b3a48e52a269a4/src/main/java/org/sagebionetworks/web/client/widget/entity/WikiMarkdownEditor.java#L132) if one did not exist when the editor was opened. I was also surprised by that flow initially, so I'm happy to wait to create the wiki until the user explicitly tries to save the wiki instead!

Good point about triggering two 'create' calls -- if we stick with the effect, I'll add the guard against a pending `createWikiPage` call!

packages/synapse-react-client/src/components/WikiMarkdownEditor/WikiMarkdownEditor.tsx
hallieswan: I had thought this was the behavior based on [a comment](https://github.com/Sage-Bionetworks/SynapseWebClient/blob/9085d9d2f9644ba7ceb9db9624b3a48e52a269a4/src/main/java/org/sagebionetworks/web/client/widget/entity/WikiMarkdownEditor.java#L133) in the SWC implementation, but I think the response will be 404 if not found, regardless of create/edit permissions. I'll update this comment and the corresponding test!

packages/synapse-react-client/src/components/WikiMarkdownEditor/WikiMarkdownEditor.test.tsx
```diff
@@ -122,164 +91,20 @@ describe('WikiMarkdownEditor', () => {
   afterEach(() => server.resetHandlers())
   afterAll(() => server.close())
 
-  test('creates a root WikiPage when there was not an existing root WikiPage', async () => {
```
hallieswan: This behavior is now managed in `WikiMarkdownEditorButton`

packages/synapse-react-client/src/components/WikiMarkdownEditor/WikiMarkdownEditor.tsx
```diff
@@ -40,128 +31,34 @@ export const WikiMarkdownEditor: React.FunctionComponent<
...
+    open,
     ownerObjectId,
-    wikiPageId: initialWikiPageId,
+    ownerObjectType,
+    wikiPage,
     onCancel = noop,
     onSave = noop,
   } = props
 
-  const [wikiPageId, setWikiPageId] = useState<string | undefined>(
```
hallieswan: Creation of new root WikiPage was moved into `WikiPageEditorButton`

packages/synapse-react-client/src/components/WikiMarkdownEditorButton/WikiMarkdownEditorButton.test.tsx
```diff
@@ -0,0 +1,290 @@
...
+      mockEntityRootWikiPage.markdown,
+    )
+  })
+
+  test('displays error when user cannot create the root WikiPage after user clicks edit button', async () => {
+    const errorResponse: ErrorResponse = {
+      concreteType: 'org.sagebionetworks.repo.model.ErrorResponse',
+      reason: `User is not authorized to 'CREATE' a WikiPage with an ownerId: '${noRootProps.ownerObjectId}' of type: '${noRootProps.ownerObjectType}'`,
+    }
+    server.use(
```
hallieswan, line 180: Throw an error when attempting to create a new root WikiPage

packages/synapse-react-client/src/components/WikiMarkdownEditorButton/WikiMarkdownEditorButton.test.tsx
```diff
@@ -0,0 +1,290 @@
...
+    const dialog = await findEditorDialog()
+
+    const titleField = queryTitleField(dialog)
+    expect(titleField).not.toBeNull()
+    expect(titleField).toHaveValue(mockEntityWikiPage.title)
+
+    expect(getMarkdownField(dialog)).toHaveValue(mockEntityWikiPage.markdown)
+  })
+
+  test('closes editor after user clicks cancel in editor', async () => {
```
hallieswan, line 259: https://github.com/Sage-Bionetworks/synapse-web-monorepo/assets/26949006/d020d29d-1d6f-4ea9-ad4a-26032e53fded

packages/synapse-react-client/src/components/WikiMarkdownEditorButton/WikiMarkdownEditorButton.test.tsx
```diff
@@ -0,0 +1,290 @@
...
+
+    const cancelBtn = within(dialog).getByRole('button', { name: 'Cancel' })
+    await user.click(cancelBtn)
+
+    await waitFor(() => {
+      expect(dialog).not.toBeInTheDocument()
+    })
+  })
+
+  test('closes editor after user clicks save in editor', async () => {
```
hallieswan, line 275: https://github.com/Sage-Bionetworks/synapse-web-monorepo/assets/26949006/7ab59361-b78d-47f2-ae4b-9db045da2760

packages/synapse-react-client/src/components/WikiMarkdownEditorButton/WikiMarkdownEditorButton.tsx
```diff
@@ -0,0 +1,155 @@
...
+
+  const wikiPageKey = useMemo(() => {
+    const wikiPageId = initialWikiPageId || rootWikiPageKey?.wikiPageId || ''
+    const wikiPageKey: WikiPageKey = {
+      ownerObjectType,
+      ownerObjectId,
+      wikiPageId: wikiPageId,
+    }
+    return wikiPageKey
+  }, [initialWikiPageId, rootWikiPageKey, ownerObjectId, ownerObjectType])
```
hallieswan, line 70: Construct the `wikiPageKey` with either the initial wikiPageId provided by the parent or with the fetched/created `rootWikiPageKey`. If neither value exists, then use an empty string to indicate that we are not ready to GET the WikiPage.

packages/synapse-react-client/src/components/WikiMarkdownEditorButton/WikiMarkdownEditorButton.tsx
```diff
@@ -0,0 +1,155 @@
...
+  } = props
+
+  const buttonProps: Omit<ButtonProps, 'onClick' | 'disabled'> = defaults(
+    {},
+    props.buttonProps,
+    DEFAULT_BUTTON_PROPS,
+  )
+
+  const [open, setOpen] = useState<boolean>(false)
+  const [isCreatingWikiPage, setIsCreatingWikiPage] = useState<boolean>(false)
```
nickgros, line 46: For `isCreatingWikiPage` you can just use the value of `isPending` from `useCreateWikiPage`

packages/synapse-react-client/src/components/WikiMarkdownEditorButton/WikiMarkdownEditorButton.tsx
```diff
@@ -0,0 +1,155 @@
...
+        {...buttonProps}
+      />
+      {error && (
+        <Alert severity="error" sx={{ marginTop: 2 }}>
+          {error}
+        </Alert>
+      )}
+      {wikiPage && (
+        <WikiMarkdownEditor
+          key={open.toString()}
```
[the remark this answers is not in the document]
nickgros: Can you add a code comment noting this?

PR #895, opened 2024-04-18
SWC-6766 - Add ChangePasswordWithToken component, use in SageAccountWeb

nickgros:
- Add component to support changing password with a reset token
- Refactor existing ChangePassword component to share logic for 2FA via custom hook

### DIFF
--- a/apps/SageAccountWeb/src/components/ResetPassword.tsx
+++ b/apps/SageAccountWeb/src/components/ResetPassword.tsx
@@ -42,8 +42,8 @@ export const ResetPassword = (props: ResetPasswordProps) => {
         'If a matching account was found, then your password reset request has been sent. Please check your email.',
         'success',
       )
-    } catch (err: SynapseClientError) {
-      displayToast(err.reason, 'danger')
+    } catch (err: unknown) {
+      displayToast((err as SynapseClientError).reason, 'danger')
     }
   }
 
--- a/apps/SageAccountWeb/src/components/ResetPassword.tsx
+++ b/apps/SageAccountWeb/src/components/ResetPassword.tsx
@@ -1,12 +1,13 @@
-import { Box, Button, TextField, Typography } from '@mui/material'
-import { StyledFormControl } from '../components/StyledComponents'
-import React, { useEffect, useState } from 'react'
+import { Box, Button, SxProps, TextField, Typography } from '@mui/material'
+import React, { useMemo, useState } from 'react'
 import { useHistory } from 'react-router-dom'
-import { displayToast, SynapseClient } from 'synapse-react-client'
 import {
   ChangePasswordWithToken,
-  PasswordResetSignedToken,
-} from '@sage-bionetworks/synapse-types'
+  displayToast,
+  SynapseClient,
+  SynapseClientError,
+} from 'synapse-react-client'
+import { PasswordResetSignedToken } from '@sage-bionetworks/synapse-types'
 import { getSearchParam, hexDecodeAndDeserialize } from '../URLUtils'
 import { BackButton } from './BackButton'
 import { LeftRightPanel } from './LeftRightPanel'
@@ -19,20 +20,17 @@ export type ResetPasswordProps = {
 export const ResetPassword = (props: ResetPasswordProps) => {
   const history = useHistory()
   const [userName, setUserName] = useState('')
-  const [token, setToken] = useState<PasswordResetSignedToken | undefined>()
-  const [newPassword, setNewPassword] = useState<string>('')
-  const [confirmPassword, setConfirmPassword] = useState<string>('')
 
   const passwordResetTokenValue = getSearchParam('passwordResetToken')
 
-  useEffect(() => {
+  const token = useMemo(() => {
     if (passwordResetTokenValue) {
-      const hexDecodedPasswordResetToken = hexDecodeAndDeserialize(
+      return hexDecodeAndDeserialize(
         passwordResetTokenValue,
-      )
-      setToken(hexDecodedPasswordResetToken)
+      ) as PasswordResetSignedToken
     }
-  }, [])
+    return undefined
+  }, [passwordResetTokenValue])
 
   const handleResetPassword = async (
     clickEvent: React.FormEvent<HTMLElement>,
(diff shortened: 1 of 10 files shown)


apps/SageAccountWeb/src/components/ResetPassword.tsx
```diff
@@ -19,69 +20,35 @@ export type ResetPasswordProps = {
...
-      const hexDecodedPasswordResetToken = hexDecodeAndDeserialize(
+      return hexDecodeAndDeserialize(
         passwordResetTokenValue,
-      )
-      setToken(hexDecodedPasswordResetToken)
+      ) as PasswordResetSignedToken
     }
-  }, [])
+    return undefined
+  }, [passwordResetTokenValue])
```
nickgros, line 33: `token` can be derived in a `useMemo` instead of computing in a `useEffect`

apps/SageAccountWeb/src/components/ResetPassword.tsx
```diff
@@ -19,69 +20,35 @@ export type ResetPasswordProps = {
...
-        }
-        await SynapseClient.changePasswordWithToken(changeRequest)
-          .then(() => {
-            displayToast('Successfully changed', 'success')
-          })
-          .then(() => window.location.replace(props.returnToUrl))
-      }
-    } catch (err: any) {
-      displayToast(err.reason as string, 'danger')
+      displayToast(
```
nickgros, line 41: This was already using `await`, so `then/catch` is pointless. Simplify

apps/SageAccountWeb/src/components/ResetPassword.tsx
```diff
@@ -96,55 +63,12 @@ export const ResetPassword = (props: ResetPasswordProps) => {
...
-                  type="submit"
-                  fullWidth
-                  onSubmit={handleChangePasswordWithToken}
-                  sx={buttonSx}
-                  disabled={!newPassword || !confirmPassword}
-                >
-                  Change Password
-                </Button>
-              </form>
+              <ChangePasswordWithToken
```
nickgros, line 66: Use SRC component to handle changing password. This component just intercepts the token from the URLSearchParams and passes it along.

apps/SageAccountWeb/src/components/ResetPassword.tsx
```diff
@@ -154,27 +78,22 @@ export const ResetPassword = (props: ResetPasswordProps) => {
                 }}
               />
               <SourceAppLogo />
-              <StyledFormControl
```
nickgros, line 157: Remove pointless `StyledFormControl` component

apps/SageAccountWeb/src/style/_core.scss
```diff
@@ -56,10 +56,4 @@
       background-color: #f4f4f4;
     }
   }
-  .btn-container {
```
nickgros, line 59: Class is now unused

packages/synapse-react-client/src/components/ChangePassword/ChangePassword.tsx
```diff
@@ -1,52 +1,30 @@
...
-        // e.g. if the user enters the wrong 2FA code
-        setTwoFactorAuthErrorResponse(maybeTwoFactorResponse)
-      } else {
-        setTwoFactorAuthErrorResponse(undefined)
-        setOldPassword('')
-        setNewPassword('')
-        setConfirmPassword('')
-        displayToast('Password successfully changed.', 'success')
-      }
+  } = useChangePasswordFormState({
```
nickgros, line 22: Refactored to use this custom hook shared with new component in `ChangePasswordWithToken.tsx`

packages/synapse-react-client/src/components/ChangePassword/ChangePassword.tsx
```diff
@@ -55,64 +33,18 @@ export default function ChangePassword() {
...
-            }}
-            isLoading={isLoadingUserProfile || changePasswordIsPending}
-            onSubmit={handleChangePasswordWithOtp}
-          />
-          <Alert severity={'info'} sx={{ my: 2 }}>
-            Two-factor authentication is required to change your password.
-          </Alert>
-        </>
+      {promptForTwoFactorAuth ? (
+        twoFactorAuthPrompt
```
nickgros, line 47: Use the 2fa prompt returned by the custom hook, so both components have a common UI

packages/synapse-react-client/src/components/ChangePassword/ChangePassword.tsx
```diff
@@ -55,64 +33,18 @@ export default function ChangePassword() {
...
-    }
-  }
-
-  function handleChangePasswordWithOtp(
-    code: string,
-    otpType: TwoFactorAuthOtpType,
-  ) {
-    if (twoFactorAuthErrorResponse) {
-      const changeRequest: ChangePasswordWithTwoFactorAuthToken = {
+      handleChangePasswordWithCurrentPassword(
```
nickgros, line 36: Delegate creating the request to the custom hook

packages/synapse-react-client/src/components/ChangePassword/ChangePassword.tsx
```diff
@@ -161,9 +93,6 @@ export default function ChangePassword() {
               }
               variant="contained"
               type="submit"
-              onSubmit={e => {
-                handleChangePassword(e)
-              }}
```
nickgros, line 166: `onSubmit` on a button does nothing

packages/synapse-react-client/src/components/ChangePassword/ChangePasswordWithToken.tsx
```diff
@@ -0,0 +1,95 @@
...
+      onSuccess()
+    },
+  })
+
+  const handleChangePassword = (clickEvent: React.FormEvent<HTMLElement>) => {
+    clickEvent.preventDefault()
+    if (newPassword !== confirmPassword) {
+      displayToast('Passwords do not match.', 'danger')
+    } else {
+      handleChangePasswordWithResetToken(newPassword, passwordChangeToken)
```
nickgros, line 39: Delegate creating the request object to the custom hook

packages/synapse-react-client/src/components/ChangePassword/useChangePasswordFormState.tsx
```diff
@@ -0,0 +1,146 @@
...
+    }
+    changePassword(changeRequest)
+  }
+
+  function handleChangePasswordWithOtp(
+    newPassword: string,
+    code: string,
+    otpType: TwoFactorAuthOtpType,
+  ) {
+    if (twoFactorAuthErrorResponse) {
```
hallieswan, line 90: `twoFactorAuthPrompt` could be included in a component without checking whether `promptForTwoFactorAuth` was true. If `twoFactorAuthResponse` was undefined, then submitting `OneTimePasswordForm` would result in `handleChangePasswordWithOtp` being a noop. Sems like an edge case (likely related to a mistake configuring a component to use this custom hook), but I'm wondering whether it would be useful to log a message to help the developer troubleshoot that case? Up to you though!

hallieswan (approved):
Looks great to me! The custom hook is a great way to handle the shared logic and the tests are really clear

packages/synapse-react-client/src/components/ChangePassword/useChangePasswordFormState.tsx
nickgros: Updated `twoFactorAuthPrompt`  to be a function component where a fragment is rendered if `promptForTwoFactorAuth` is used. `OneTimePasswordForm` will now never be shown if `promptForTwoFactorAuth` is false

hallieswan (approved):

PR #901, opened 2024-04-23
PORTALS-2985 (donation asset) and PORTALS-3032 (ELITE portal Whats Ne…

jay-hodgson:
…w section)


packages/synapse-react-client/src/components/RssFeedCards/RssFeedCards.tsx
```diff
@@ -117,13 +117,12 @@ export class RssFeedCards extends React.Component<RssFeedCardsProps, RssState> {
...
                 // The other is to hide the large number of items in a particular feed (usually a max of 10 are returned).  See state.isShowingMoreItems
-                const isItemVisible: boolean = index < this.props.itemsToShow
+                if (index >= this.props.itemsToShow) {
+                  return
+                }
 
                 return (
-                  <div
-                    key={item.guid}
-                    className={`FeedItem ${isItemVisible ? '' : 'hidden'}`}
```
jay-hodgson, line 125: I think the '.hidden' css class has been removed, because all items with this class were still being shown in the portal.  Changed to just not include in the return (in the DOM!)

hallieswan (approved):

PR #903, opened 2024-04-23
SWC-6730: add create or update AR wizard

hallieswan:
Adds new wizard to create/update ARs. See https://github.com/Sage-Bionetworks/SynapseWebClient/pull/5368

### DIFF
--- a/packages/synapse-react-client/src/components/CreateOrUpdateAccessRequirementWizard/CreateOrUpdateAccessRequirementWizard.stories.tsx
+++ b/packages/synapse-react-client/src/components/CreateOrUpdateAccessRequirementWizard/CreateOrUpdateAccessRequirementWizard.stories.tsx
@@ -1,0 +1,87 @@
+import { Meta, StoryObj } from '@storybook/react'
+import { fn } from '@storybook/test'
+import CreateOrUpdateAccessRequirementWizard from './CreateOrUpdateAccessRequirementWizard'
+import { RestrictableObjectType } from '@sage-bionetworks/synapse-types'
+import {
+  mockACTAccessRequirement,
+  mockToUAccessRequirement,
+} from '../../mocks/mockAccessRequirements'
+
+const meta = {
+  title: 'Governance/CreateOrUpdateAccessRequirementWizard',
+  component: CreateOrUpdateAccessRequirementWizard,
+  parameters: {
+    stack: 'development',
+  },
+} satisfies Meta
+
+export default meta
+
+type Story = StoryObj<typeof meta>
+
+export const CreateTeamAR: Story = {
+  args: {
+    open: true,
+    onComplete: fn(),
+    onCancel: fn(),
+    subject: {
+      id: '3429759',
+      type: RestrictableObjectType.TEAM,
+    },
+  },
+}
+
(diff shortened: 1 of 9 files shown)


packages/synapse-react-client/src/components/CreateOrUpdateAccessRequirementWizard/CreateOrUpdateAccessRequirementWizard.stories.tsx
```diff
@@ -0,0 +1,87 @@
...
+export const EditSelfSignAR: Story = {
+  args: {
+    open: true,
+    onComplete: fn(),
+    onCancel: fn(),
+    accessRequirementId: '9602700',
+  },
+}
+
+/* For Mock Stories, updates to Wiki Instructions will not be visible until MarkdownSynapse is updated to use react-query. */
```
hallieswan, line 64: When the user updates the wiki content via `useUpdateWikiPage`, react query will update the cache for that wiki page. If a component gets a wiki page via `useGetWikiPage`, then the updated wiki page content will be returned, even when using mocks. However, MarkdownSynapse calls `SynapseClient.getEntityWiki` rather than using the react-query cache. So in mock stories, the MSW handler will intercept the GET request and return the original wiki page, rather than the updated page, and the wiki instructions will appear not to update after editing. 

Since stories against the dev stack will show the expected behavior, I held off on attempting to update our MSW handlers to accommodate this case.

packages/synapse-react-client/src/components/CreateOrUpdateAccessRequirementWizard/CreateOrUpdateAccessRequirementWizard.tsx
```diff
@@ -0,0 +1,159 @@
...
+    }
+  }
+
+  return (
+    <ConfirmationDialog
+      open={open}
+      onCancel={onCancel}
+      onConfirm={onClickPrimary}
+      confirmButtonProps={{
+        children: isLastStep(step) ? 'Save' : 'Save & Continue',
```
hallieswan, line 149: Changed primary button copy from "Next" to "Save & Continue" to reflect that each step saves the AR

packages/synapse-react-client/src/components/CreateOrUpdateAccessRequirementWizard/CreateOrUpdateAccessRequirementWizard.stories.tsx
jay-hodgson: Should MarkdownSynapse be changed to use useGetWikiPage?  Is the problem that MarkdownSynapse a an old class-based component?

hallieswan: Yes and yes! Since MarkdownSynapse is a class-based component, I was concerned that the refactor to use react-query could be complicated. Should I open a new ticket to capture the refactor work?  We could also tackle the refactor as part of https://sagebionetworks.jira.com/browse/SWC-6791

jay-hodgson (approved):

PR #905, opened 2024-04-24
SWC-6804

jay-hodgson:


packages/synapse-react-client/src/components/ProvenanceGraph/EntityNodeLabel.tsx
```diff
@@ -22,11 +22,11 @@ export const EntityNodeLabel = (data: EntityHeader) => {
         <>
           <div>
             <Tooltip title={data.name} placement="top" enterNextDelay={300}>
-              <a
+              <Link
```
jay-hodgson, line 25: Unrelated, I think a lot of material components should be used in here.

hallieswan (approved):
Looks good to me!

PR #907, opened 2024-04-25
PORTALS-3053: add custom click handler to UpsetPlot

jay-hodgson:


apps/portals/src/configurations/elportal/synapseConfigs/handleUpsetPlotClick.tsx
```diff
@@ -0,0 +1,21 @@
...
+          'org.sagebionetworks.repo.model.table.FacetColumnValuesRequest',
+        columnName: 'Assays',
+        facetValues: [name],
+      },
+    ],
+    // additionalFilters: [filter], //technically,
+  }
+  window.location.assign(
+    `/Explore/Data%20by%20Files?QueryWrapper0=${JSON.stringify(query)}`,
+  )
```
hallieswan, line 20: Could consider using SRC's [generateEncodedPathAndQueryForSelectedFacetURL](https://github.com/Sage-Bionetworks/synapse-web-monorepo/blob/main/packages/synapse-react-client/src/components/QueryWrapper/generateEncodedPathAndQueryForSelectedFacetURL.ts), if you'd like! Something like -- 

```suggestion
  const url = generateEncodedPathAndQueryForSelectedFacetURL(
    '/Explore/Data%20by%20Files',
    cohortBuilderFilesSql,
    [{facet: 'Assays', facetValue: name}]
  window.location.assign(url)
```

hallieswan (approved):
Looks good to me!

PR #909, opened 2024-04-26
Update resources.ts

cconrad8:
want to always use latest version of datasets

@anngvu @jaybee84 @allaway for visibility


jay-hodgson:
Doesn't this still require manually updating the versions in the Dataset Collection (updating each Dataset to the latest release value)?

PR #910, opened 2024-04-26
SWC-6810

jay-hodgson:


jay-hodgson:
No change to ARs with Team subject association:
Screen Shot 2024-04-26 at 2 28 27 PM

But for entities, you can now say that the AR should be based on annotations (which hides the entity selector):
Screen Shot 2024-04-26 at 2 28 38 PM
dbda-eccd-4807-a1b9-034d40ad2760">

packages/synapse-react-client/src/components/CreateOrUpdateAccessRequirementWizard/CreateOrUpdateAccessRequirementWizard.tsx
```diff
@@ -96,7 +96,7 @@ export const CreateOrUpdateAccessRequirementWizard: React.FunctionComponent<
             accessRequirementId={accessRequirementId!}
             onSaveComplete={saveSuccessful => {
               if (saveSuccessful) {
-                onComplete()
+                onComplete(accessRequirementId)
```
jay-hodgson, line 99: Pass back the AR ID that we are working with.  On the SWC side, this will redirect to a page that is focused on the single AR (not subject based), which will work for this new type of AR configuration.

packages/synapse-react-client/src/components/SetAccessRequirementCommonFields/SetAccessRequirementCommonFields.test.tsx
```diff
@@ -128,12 +139,68 @@ describe('SetAccessRequirementCommonFields', () => {
...
+      expect(onSave).toHaveBeenCalledTimes(1)
+      expect(onSave).toHaveBeenLastCalledWith(
+        MOCK_NEWLY_CREATED_AR_ID.toString(),
+        MANAGED_ACT_ACCESS_REQUIREMENT_CONCRETE_TYPE_VALUE,
+      )
+      expect(onError).not.toHaveBeenCalled()
+    })
+  })
+
+  test('creates a new managed entity AR with subjects defined by annotation', async () => {
```
jay-hodgson, line 170: Try flipping on the subjectsDefinedByAnnotationsCheckbox to verify that the field is set to true, and subjectIds array is now empty (they are mutually exclusive)

packages/synapse-react-client/src/components/SetAccessRequirementCommonFields/SetAccessRequirementCommonFields.tsx
```diff
@@ -199,26 +208,32 @@ export const SetAccessRequirementCommonFields = React.forwardRef(
...
-              onError()
-              return
+                onError()
+                return
+              }
             }
-
+            const newAccessType = subjectsDefinedByAnnotations
+              ? ACCESS_TYPE.DOWNLOAD
+              : getAccessType(subjects[0])
```
hallieswan, line 229: This is a change I should've made originally, but since we've already stored the AR's `RestrictableObjectType` in `subjectsType` , could `getAccessType` be refactored to return the `ACCESS_TYPE` given the `RestrictableObjectType` rather than given an `RestrictableObjectDescriptor`? Then I think this could be simplified to -- 

```suggestion
            const newAccessType = getAccessType(subjectsType)
```

packages/synapse-react-client/src/components/SetAccessRequirementCommonFields/SetAccessRequirementCommonFields.test.tsx
hallieswan: What do you think about adding a test for editing an AR with `subjectsDefinedByAnnotations` to make sure that the content loads and `subjectsDefinedByAnnotationsCheckbox` is checked?

packages/synapse-react-client/src/components/SetAccessRequirementCommonFields/SetAccessRequirementCommonFields.tsx
jay-hodgson: Done!

packages/synapse-react-client/src/components/SetAccessRequirementCommonFields/SetAccessRequirementCommonFields.tsx
```diff
@@ -199,26 +211,30 @@ export const SetAccessRequirementCommonFields = React.forwardRef(
...
             const hasSubjectsError = hasPendingSubjects || subjects.length === 0
-
-            if (isStillLoading || hasSubjectsError) {
-              if (hasSubjectsError && !isStillLoading) {
-                if (hasPendingSubjects) {
-                  setSubjectsError(UNSAVED_SUBJECTS_ERROR_MESSAGE(subjectsType))
-                } else if (subjects.length === 0) {
-                  setSubjectsError(EMPTY_SUBJECT_LIST_ERROR_MESSAGE)
+            if (!subjectsDefinedByAnnotations) {
+              if (isStillLoading || hasSubjectsError) {
```
hallieswan, line 215: When the parent calls `save()`, if this component `isStillLoading`, then this component should call `onError` to inform the parent that the AR wasn't saved, so I think we may need to remove the outer check for `!subjectsDefinedByAnnotations`.

What do you think about changing `hasSubjectsError` to include a check for `subjectsDefinedByAnnotations`, so that it's clearer that there can only be a subjects error when the AR's subjects aren't defined by annotations? Something like -- 

```ts
const hasSubjectsError = !subjectsDefinedByAnnotations && (hasPendingSubjects || subjects.length === 0)
if (isStillLoading || hasSubjectsError) {
  ... 
  onError()
  return
}
```

packages/synapse-react-client/src/components/SetAccessRequirementCommonFields/SetAccessRequirementCommonFields.test.tsx
jay-hodgson: Good idea.  Added this test

packages/synapse-react-client/src/components/SetAccessRequirementCommonFields/SetAccessRequirementCommonFields.tsx
jay-hodgson: Oops, made this change with the intent of revisiting later.  Made your suggested change to the logic

hallieswan (approved):
Looks good to me!

PR #916, opened 2024-05-01
CPO-105

jay-hodgson:
May also help IT-3607, if it turns out to be a category instead of a tag


packages/synapse-react-client/src/components/RssFeedCards/RssFeedCards.tsx
```diff
@@ -15,9 +14,14 @@ type RssState = {
...
 }
 
+export type RssFilter = {
+  type?: 'tag' | 'category'
+  value: string
+}
+
 export type RssFeedCardsProps = {
   url: string
-  lockedColumn?: LockedColumn // optional tag to filter by, typically set by using this component on a DetailsPage
```
jay-hodgson, line 20: Unclear why it was using the LockedColumn type, since it didn't really make sense here.  The new type has the option to set the filter type to either 'tag' (which is default) or 'category'.  This determines the WordPress feed URL to use

hallieswan (approved):
Looks good to me!

PR #917, opened 2024-05-01
PORTALS-3057

jay-hodgson:


apps/portals/src/types/portal-config.ts
```diff
@@ -254,6 +254,8 @@ type Metadata = {
   // applied to outer most container of the object
   containerClassName?: string
   outsideContainerClassName?: string
+  // if set, will show a help icon next to the title with this text
+  helpText?: string
```
hallieswan, line 258: Could consider changing this prop name to `helpMarkdown`, since `HelpPopover` will display markdown as well as text, but up to you!

hallieswan (approved):
Looks good to me!

PR #919, opened 2024-05-02
SWC-6737 - Update SSO flow to support passing `twoFaResetCode` in state

nickgros:
This change will make it possible to use a `twoFaResetCode` in conjunction with login through an external identity provider. Implementation of the 2FA reset feature will be added in a later PR.

- Add types for 2FA reset
- Update OAuth2 SSO flow so that we can pass the twoFaResetCode to the SSO provider in the `state` field


packages/synapse-react-client/src/components/Authentication/AuthenticationMethodSelection.tsx
```diff
@@ -15,6 +15,7 @@ type AuthenticationMethodSelectionProps = {
   /* Invoked before redirecting to Google. Useful in portals where we may want to store the current URL to redirect back here. */
   onBeginOAuthSignIn?: () => void
   onSelectUsernameAndPassword: () => void
+  state?: OAuth2State
```
nickgros, line 18: This component generates the OAuth2 URL (by making a call to the backend). Pass the OAuth2 state object into the component so that it can be included in the backend.

packages/synapse-react-client/src/synapse-client/SynapseClient.ts
```diff
@@ -634,12 +640,16 @@ export function loginWith2fa(
...
   endpoint = BackendDestinationEnum.REPO_ENDPOINT,
 ) => {
   return doPost<{ authorizationUrl: string }>(
     '/auth/v1/oauth2/authurl',
-    { provider, redirectUrl, state },
+    {
+      provider,
+      redirectUrl,
+      state: state ? encodeURIComponent(JSON.stringify(state)) : undefined,
+    },
```
nickgros, line 652: State is now an object that, if defined, is now stringified and URL-encoded

packages/synapse-react-client/src/utils/hooks/useDetectSSOCode.ts
```diff
@@ -91,10 +101,22 @@ export default function useDetectSSOCode(
...
+                // The user logged in with OAuth while attempting to disable 2FA using an emailed signed token
+                state &&
+                state.twoFaResetToken &&
+                onTwoFactorAuthResetTokenPresent
+              ) {
+                // Let the app handle redirecting to the 2FA reset page
+                onTwoFactorAuthResetTokenPresent(
+                  response,
+                  state.twoFaResetToken,
+                )
```
nickgros, line 118: This callback will be implemented in a later PR

packages/synapse-react-client/src/utils/types/OAuth2State.ts
```diff
@@ -0,0 +1,10 @@
+/**
+ * Type that represents the data that is encoded and sent to/returned from OAuth2 providers
+ */
+export type OAuth2State = {
+  /* If provided, the user is registering for a new account with this username */
+  registrationUsername?: string
+  /* If present, the user has signed in to attempt resetting 2FA on their account. This is the encoded signed token that
+   is required to disable 2FA, in conjunction with a twoFaToken returned by the authentication request. */
+  twoFaResetToken?: string
+}
```
nickgros, line 10: We could later append other fields, such as the URL to redirect to after completing SSO (we currently capture this in localStorage)

apps/synapse-oauth-signin/src/test/App.test.tsx
```diff
@@ -34,11 +38,12 @@ const mockConsentToOAuth2Request = vi.mocked(
 const { ACCESS_TOKEN_COOKIE_KEY, POST_SSO_REDIRECT_URL_LOCALSTORAGE_KEY } =
   SynapseConstants
 function createParams(prompt?: string) {
+  const state: OAuth2State = { registrationUsername: '1ga1bi0qm' }
   const params = new URLSearchParams()
   params.set('response_type', 'code')
   params.set('client_id', '1234')
   params.set('scope', 'openid')
-  params.set('state', '1ga1bi0qm')
```
nickgros, line 41: I think this was originally meaningless (not a registration username), but the test broke when parsing since it's no longer an object

packages/synapse-react-client/src/utils/hooks/useLogin.ts
```diff
@@ -83,6 +83,8 @@ export default function useLogin(
         setStep('VERIFICATION_CODE')
       }
     }
+    // We do NOT want to rerun this effect on step change. It should only run once, when we get the error.
+    // eslint-disable-next-line react-hooks/exhaustive-deps
   }, [twoFaErrorResponse])
```
nickgros, line 88: Add comment to remove lint warning

jay-hodgson (approved):

PR #921, opened 2024-05-02
PORTALS-3061

jay-hodgson:


apps/portals/src/configurations/cancercomplexity/routesConfig.ts
```diff
@@ -55,6 +58,7 @@ const routes: GenericRoute[] = [
         centerTitle: true,
         props: {
           onPointClick,
+          onIndividualThemeBarPlotPointClick,
```
jay-hodgson, line 61: Add a new custom click handler for the individual theme bar charts

packages/synapse-react-client/src/components/Plot/ThemesPlot.tsx
```diff
@@ -338,9 +342,18 @@ export function ThemesPlot({
                           optionsConfig={optionsConfig}
                           plotData={sideBarPlotData}
                           isTop={false}
-                          xMax={xMaxForSideBarPlot}
+                          // PORTALS-3061: No longer use a global x max for the side bar (where the bar size of 1 grant is consistent),
+                          //    Instead, each bar chart is independent (the size of 1 grant differs across bar charts).
+                          xMax={totalsByDotPlotY[i].count}
```
jay-hodgson, line 347: It sounds like once this portal has a lot of data, they may want to go back to the original implementation.  Left breadcrumbs on how to revert this change (by sending you to the related issue, and finding this PR).

nickgros (approved):

PR #922, opened 2024-05-03
PORTALS-3071

jay-hodgson:


apps/portals/src/configurations/genie/synapseConfigs/genieBPCData.ts
```diff
@@ -13,11 +13,11 @@ const genieBPCData: SynapseConfig = {
     isRowSelectionVisible: true,
     rowSelectionPrimaryKey: ['id'],
     visibleColumnCount: 6,
+    fileIdColumnName: 'id',
```
jay-hodgson, line 16: Consolidate entity ID column name and version column name.

packages/synapse-react-client/src/components/QueryWrapper/QueryWrapper.tsx
```diff
@@ -52,6 +52,11 @@ export type QueryWrapperProps = React.PropsWithChildren<{
   isRowSelectionUIFloating?: boolean
   isInfinite?: boolean
   combineRangeFacetConfig?: CombineRangeFacetConfig
+  /** If provided, will use the value in this column instead of the rowID for the access column, download column, etc */
```
jay-hodgson, line 55: Move these to the QueryWrapper, and set atoms instead of passing plumbing them through the hierarchy.

jay-hodgson:
Verified functionality on NF files (Materialized View), ELITE file data (Virtual Table), and ARK (File View, via AddToDownloadV2 SynapseTable renderer which pulls the file ID column from the atom).  All work well when pointing to the Synapse staging backend (498), but not when pointing to production today (497).

packages/synapse-react-client/src/components/download_list/TableQueryDownloadConfirmation.test.tsx
```diff
@@ -44,14 +44,24 @@ const mockUseGetQueryResultBundle = jest.mocked(
   useGetQueryResultBundleWithAsyncStatus,
 )
 const mockUseAddQueryToDownloadList = jest.mocked(useAddQueryToDownloadList)
-
+const ID_COLUMN_ID = 11112
```
jay-hodgson, line 47: Set up a couple of ColumnModels with the response data - one column for 'id' and one for 'currentVersion'.  Then verify that when fileIdColumnName and fileVersionColumnName are set, the selectFileColumn and selectFileVersionColumn are set in the AddToDownloadListRequest

packages/synapse-react-client/src/components/download_list/TableQueryDownloadConfirmation.test.tsx
```diff
@@ -265,9 +287,8 @@ describe('TableQueryDownloadConfirmation', () => {
       false,
     )
   })
-
-  it('setting the row primary key indicates that the row version should not be used', async () => {
-    await setUp('fileSynapseID')
```
nickgros, line 270: Test for this behavior is lost, but seems ok since it was added for PORTALS-2954, which was a workaround now resolved by using the feature under test

nickgros (approved):
LGTM. Do you think `entityIdColumnName` would be a more appropriate name than `fileIdColumnName` (and so on) since this is used for more than just download (e.g. access)? I could go either way since this is really primarily used for files.

PR #927, opened 2024-05-07
IT-3611

jay-hodgson:
Added a backdrop while the session is being detected (or not found).  In this video, I have an artificial sleep for 2s.  Namely this code in ApplicationSessionManager.refreshSession():
`await new Promise(r => setTimeout(r, 2000))`

https://github.com/Sage-Bionetworks/synapse-web-monorepo/assets/1864447/c3177a43-c05d-4862-9402-1234b2370543

### DIFF
--- a/apps/synapse-oauth-signin/src/OAuth2Form.tsx
+++ b/apps/synapse-oauth-signin/src/OAuth2Form.tsx
@@ -1,4 +1,11 @@
-import { Button, Link, Paper, Typography } from '@mui/material'
+import {
+  Backdrop,
+  Button,
+  CircularProgress,
+  Link,
+  Paper,
+  Typography,
+} from '@mui/material'
 import {
   AccessCodeResponse,
   FileHandleAssociateType,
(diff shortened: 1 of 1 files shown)


jay-hodgson:
Bruce found (in [IT-3611](https://sagebionetworks.jira.com/browse/IT-3611)) that he could cause the error by clicking on the Google Login button quickly (before the session was detected).  I'd like to stop this from happening in the future, so I'm proposing a backdrop that is shown before the session has been initialized.

apps/synapse-oauth-signin/src/OAuth2Form.tsx
```diff
@@ -319,6 +330,9 @@ export function OAuth2Form() {
 
   return (
     <StyledOuterContainer>
+      <Backdrop open={!hasInitializedSession} sx={{ zIndex: 5 }}>
```
jay-hodgson, line 333: zIndex required because the MUI buttons were above the backdrop (everything else was below)

nickgros (approved):

PR #928, opened 2024-05-07
PORTALS-3073 - do not parse state during OAuth sign in

nickgros:
Should fix PORTALS-3073 where we should return an arbitrary `state` searchParam to an OAuth2 client that is using Synapse as an OIDC identity provider

image

### DIFF
--- a/packages/synapse-react-client/src/utils/hooks/useDetectSSOCode.test.ts
+++ b/packages/synapse-react-client/src/utils/hooks/useDetectSSOCode.test.ts
@@ -8,7 +8,10 @@ import {
 } from '@sage-bionetworks/synapse-types'
 import { BackendDestinationEnum } from '../functions'
 import { SynapseClientError } from '../SynapseClientError'
-import { OAUTH2_PROVIDERS } from '../SynapseConstants'
+import {
+  LOGIN_METHOD_OAUTH2_GOOGLE,
+  OAUTH2_PROVIDERS,
+} from '../SynapseConstants'
 import { MOCK_CONTEXT_VALUE } from '../../mocks/MockSynapseContext'
 import { createWrapper } from '../../testutils/TestingLibraryUtils'
 import { SynapseContextType } from '../context'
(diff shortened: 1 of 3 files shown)


packages/synapse-react-client/src/utils/hooks/useDetectSSOCode.test.ts
```diff
@@ -384,4 +387,47 @@ describe('useDetectSSOCode tests', () => {
       consoleSpy.mockReset()
     },
   )
+
+  test('Does not attempt to use the state param if we are acting as identity provider', async () => {
```
nickgros, line 391: This is kind of a weird test, but it checks that we maintain the existing behavior.

In this scenario, we are both acting as an IdP and simultaneously using an external IdP to sign into a Synapse account. In this case we do not do anything with the `state` parameter. 

The 'proper' behavior would probably be that we capture and serialize client's `state` data in our own `state` object that we send to the external IdP (along with any other data we might want to put in `state`), then deserialize the data and send it back to the client, but that's a more complex change than necessary to fix the regression.

I captured that we should test this more thoroughly in https://sagebionetworks.jira.com/browse/PORTALS-2999?focusedCommentId=207029

packages/synapse-react-client/src/utils/hooks/useDetectSSOCode.ts
```diff
@@ -59,11 +59,33 @@ export default function useDetectSSOCode(
...
+        console.error(
+          'Error parsing state param:\n',
+          e,
+          '\nEncoded value:\n',
+          encodedState,
+        )
+      }
+    }
+    return null
+  }, [isHandlingSynapseOAuthSignIn, searchParams])
```
nickgros, line 88: WDYT of this logic? Try to guess if we're in the middle of an OAuth sign-in attempt where we are the IdP, and if so, do not parse the state.

jay-hodgson: I wonder if we can separate out the concerns for this app.  I think this is fine for now, but what if in the future we delegated establishing a Synapse session to OneSage?  The redirectURL that we would send OneSage would be the current URL (including the state that the client sent to the Synapse IdP).  Would that simplify this solution (if we took out the login component)?  How bad do you think the UX would be with this additional redirect (if a session is not discovered)?  I like the idea that this sign-in app would not need to worry about other login complexities (like the 2FA reset workflow!).  WDYT?

jay-hodgson (approved):

packages/synapse-react-client/src/utils/hooks/useDetectSSOCode.ts
nickgros: I agree, I think separating concerns per-app would be ideal. It's a lot of redirects, but I think it's pretty standard for systems that federate identity + consent. We could consider combining both the OAuth signin + OneSage login+account management into an individual SPA to save a few full page loads, but then we'd have to put OneSage on the signin.synapse.org domain.

PR #939, opened 2024-05-14
Prototype for ROR using MUI/Synapse components

nickgros:
https://github.com/Sage-Bionetworks/synapse-web-monorepo/assets/17580037/cc70bb8e-b282-422d-9815-34f86971776a


nickgros:
integrated in #1760

PR #925, opened 2024-05-06
PORTALS-2521 - Single workspace project per portal

nickgros:
This PR splits each portal into its own workspace project. Instead of storing each portal in a configuration folder (`apps/portals/src/config/<portal>`), we now have a project for the portals framework (`apps/synapse-portals-framework`) and individual projects for each portals that depend on the framework (`apps/portals/<portal>`).

Making this change results in the following benefits
* Building, linting, and testing portals is now easier and can be done with nx, e.g. `nx run adknowledgeportal:build`. These jobs will utilize the nx cache like other projects in the monorepo.
* Any portal can be built without moving configuration files. Multiple portals can be built in a single command without reconfiguring any project.
* Building a portal no longer depends on symlinks. Symlinks have caused problems for us as we've migrated between bundlers (webpack 4 -> webpack 5 -> vite) as each bundler has handled them differently, with varying levels of cross-platform support.
* Shared portals code no longer requires preparing a configuration to run typecheck or test jobs.

Other changes:

- Created Vite apps for each portal. The bulk of the code for each portal is the configuration, which were mostly copied without changes from the original configuration folders.
- Moved shared `portals` code to a new project, `synapse-portals-framework`. Each portal passes its configuration to `synapse-portals-framework` to render the content of the portal.
  - Main change in `synapse-portals-framework` is that the `Portal` component accepts portal configuration data as props instead of importing the data directly from the config files. Each portal will create its own config object and pass it to `Portal` to render a portal.
- Fix `generate-sitemap` to work with configs with `.tsx` extension

PR #940, opened 2024-05-14
PORTALS-2521 - Run prettier on portals project

nickgros:
Some unformatted files complicated the diff for #925. Let's reformat those first to try to simplify the diff.


jay-hodgson (approved):

jay-hodgson:
Merging so I can make changes after these updates

PR #941, opened 2024-05-15
SWC-6839

jay-hodgson:


packages/synapse-react-client/src/utils/hooks/useLogin.test.tsx
```diff
@@ -148,7 +148,7 @@ describe('useLogin tests', () => {
   test.each([
     ['Regular path', '/'],
     // The hashbang URL is used by SWC but isn't normally compatible with URL.searchParams
```
nickgros, line 150: Comment should be changed/removed

packages/synapse-react-client/src/components/dataaccess/ReviewerDashboard.tsx
```diff
@@ -40,15 +40,15 @@ function LinkTab(props: {
 }
 
 type ReviewerDashboardProps = {
-  /** Used to determine the base path for the component. Default is #!DataAccessManagement:default */
+  /** Used to determine the base path for the component. Default is #ataAccessManagement:default */
```
nickgros, line 43:
```suggestion
  /** Used to determine the base path for the component. Default is DataAccessManagement:default */
```

hallieswan (approved):

PR #947, opened 2024-05-17
PORTALS-3081, PORTALS-3082

jay-hodgson:


apps/SageAccountWeb/src/AppInitializer.tsx
```diff
@@ -36,34 +34,13 @@ function AppInitializer(props: { children?: React.ReactNode }) {
     }
   }, [])
 
-  useEffect(() => {
```
jay-hodgson, line 39: Remove query parameter redirectURL, and use appURL in SourceApp configuration found in the source table. https://www.synapse.org/#!Synapse:syn45291362/tables/

apps/SageAccountWeb/src/components/RegisterAccount1.tsx
```diff
@@ -33,7 +33,7 @@ import {
   StyledInnerContainer,
   StyledOuterContainer,
 } from './StyledComponents'
-import { DEFAULT_SOURCE_APP_ID, useSourceApp } from './useSourceApp'
+import { SYNAPSE_SOURCE_APP_ID, useSourceApp } from './useSourceApp'
```
jay-hodgson, line 36: Deleted DEFAULT_SOURCE_APP_ID, and now using SYNAPSE_SOURCE_APP_ID everywhere for clarity (they were both constants set to 'synapse.org')

hallieswan (approved):

hallieswan (approved):

PR #949, opened 2024-05-20
PORTALS-3086

jay-hodgson:


packages/synapse-react-client/src/utils/hooks/useDetectSSOCode.ts
```diff
@@ -87,105 +89,108 @@ export default function useDetectSSOCode(
     return null
   }, [isHandlingSynapseOAuthSignIn, searchParams])
 
-  const { accessToken } = useSynapseContext()
   const [isLoading, setIsLoading] = useState(!!(code && provider))
 
   useEffect(() => {
-    if (code && provider) {
-      const redirectUrl = `${redirectURL}?provider=${provider}`
+    if (hasInitializedSession) {
```
jay-hodgson, line 95: Wrapped existing code with `if (hasInitializedSession)`

packages/synapse-react-client/src/utils/hooks/useDetectSSOCode.ts
```diff
@@ -87,105 +89,108 @@ export default function useDetectSSOCode(
...
-                onTwoFactorAuthResetTokenPresent
-              ) {
-                // Let the app handle redirecting to the 2FA reset page
-                onTwoFactorAuthResetTokenPresent(
-                  response,
-                  state.twoFaResetToken,
-                )
-              }
+        //If user is already logged in, and the provider is ORCID, then try to bind this OAuth provider to the account.
+        if (OAUTH2_PROVIDERS.ORCID == provider && token !== undefined) {
```
jay-hodgson, line 100: The assumption here is that the access token has been initialized by this time (when this is mounted), which may not be true!  So I updated this hook to only execute after the session has been initialized so we can determine what action to take if the provider is ORCID.  If an access token is present, then try to bind to the account.  If not present, then try to log in (utilizing the previously bound alias).

packages/synapse-react-client/src/utils/hooks/useDetectSSOCode.ts
```diff
@@ -87,105 +89,108 @@ export default function useDetectSSOCode(
...
 
   useEffect(() => {
-    if (code && provider) {
-      const redirectUrl = `${redirectURL}?provider=${provider}`
+    if (hasInitializedSession) {
+      if (code && provider) {
+        const redirectUrl = `${redirectURL}?provider=${provider}`
 
-      //If user is already logged in, and the provider is ORCID, then try to bind this OAuth provider to the account.
-      if (OAUTH2_PROVIDERS.ORCID == provider && accessToken !== undefined) {
```
jay-hodgson, line 98: And look at prop.token instead of SynapseContext.accessToken

packages/synapse-react-client/src/utils/hooks/useDetectSSOCode.test.ts
```diff
@@ -49,23 +47,11 @@ const mockBindOAuthProviderToAccount = jest.spyOn(
...
-const AUTHENTICATED_CONTEXT: SynapseContextType = {
-  ...MOCK_CONTEXT_VALUE,
-}
-
-const UNAUTHENTICATED_CONTEXT: SynapseContextType = {
-  ...MOCK_CONTEXT_VALUE,
-  accessToken: undefined,
-}
-
 describe('useDetectSSOCode tests', () => {
```
jay-hodgson, line 50: No longer relies on the ApplicationContext for the access token - it is given the token directly by the ApplicationSessionManager.

packages/synapse-react-client/src/utils/hooks/useDetectSSOCode.ts
nickgros: Is `token` passed as an arg ever different from `accessToken` retrieved through context? In `ApplicationSessionManager`, these are derived from the exact same variable

packages/synapse-react-client/src/utils/hooks/useDetectSSOCode.ts
```diff
@@ -33,6 +32,8 @@ export type UseDetectSSOCodeOptions = {
     resp: TwoFactorAuthErrorResponse,
     encodedTwoFaResetToken: string,
   ) => void
+  hasInitializedSession?: boolean
```
nickgros, line 35: Since this hook won't do anything unless this is true, and the default is false, consider one of the following:
* Making `hasInitializedSession` required
* Making `hasInitializedSession` default true
* Keeping the default `false`, but flipping the variable to represent the inverse

packages/synapse-react-client/src/utils/hooks/useDetectSSOCode.ts
```diff
@@ -33,6 +32,8 @@ export type UseDetectSSOCodeOptions = {
...
+  token?: string
```
nickgros, line 36: I am not convinced we need this over context, but if we do, add a comment to clarify that it's not an OAuth provider-related token
```suggestion
  /* Synapse access token, if the user is currently logged in */
  token?: string
```

PR #953, opened 2024-05-23
PORTALS-3088

jay-hodgson:


jay-hodgson:
Screen Shot 2024-05-23 at 10 12 57 AM

Screen Shot 2024-05-23 at 11 10 37 AM

Screen Shot 2024-05-23 at 11 12 43 AM

Screen Shot 2024-05-23 at 11 14 45 AM

Screen Shot 2024-05-23 at 11 16 16 AM

Screen Shot 2024-05-23 at 11 18 06 AM

apps/SageAccountWeb/src/components/AccountSettings.tsx
```diff
@@ -572,8 +585,30 @@ export const AccountSettings = () => {
                 </div>
                 <div className="credential-partition">
                   <h4>Profile Validation</h4>
-                  <CompletionStatus isComplete={verified} />
-                  <p>
+                  {(verificationState === undefined ||
+                    verificationState.state ===
+                      VerificationStateEnum.APPROVED) && (
+                    <CompletionStatus isComplete={verified} />
+                  )}
```
nickgros, line 592: Is this appropriate to show if `verificationState === undefined`? I would expect a skeleton (or nothing)

nickgros (approved):

apps/SageAccountWeb/src/components/AccountSettings.tsx
jay-hodgson: This is the case before a submission has been created, so verified is false (and this UI would show Incomplete in red).

nickgros: I completely missed the `isComplete` prop (which didn't even change). Totally makes sense!

PR #954, opened 2024-05-24
SWC-6844 - Features that rely on CHANGE_PASSWORD_2FA_CHECK_BYPASS feature flag should point to OneSage 

nickgros:
See SWC-6844

### DIFF
--- a/apps/SageAccountWeb/src/components/AccountSettings.tsx
+++ b/apps/SageAccountWeb/src/components/AccountSettings.tsx
@@ -177,6 +177,8 @@ export const AccountSettings = () => {
   useEffect(() => {
     if (hash === '#trust') {
       handleScroll(trustCredentialRef)
+    } else if (hash === '#ChangePassword') {
+      handleScroll(changePasswordRef)
     }
   }, [hash])
 
--- a/packages/synapse-react-client/src/components/Authentication/UsernamePasswordForm.tsx
+++ b/packages/synapse-react-client/src/components/Authentication/UsernamePasswordForm.tsx
@@ -1,6 +1,5 @@
 import { Button, Link } from '@mui/material'
 import React, { useState } from 'react'
-import { BackendDestinationEnum, getEndpoint } from '../../utils/functions'
 import TextField from '../TextField/TextField'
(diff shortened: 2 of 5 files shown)


apps/SageAccountWeb/src/components/AccountSettings.tsx
```diff
@@ -167,6 +167,8 @@ export const AccountSettings = () => {
   useEffect(() => {
     if (hash === '#trust') {
       handleScroll(trustCredentialRef)
+    } else if (hash === '#ChangePassword') {
+      handleScroll(changePasswordRef)
```
nickgros, line 171: Scroll to the change password section if the hash is provided in the link

packages/synapse-react-client/src/components/Authentication/UsernamePasswordForm.tsx
```diff
@@ -13,9 +12,7 @@ type UsernamePasswordFormProps = {
 
 export default function UsernamePasswordForm(props: UsernamePasswordFormProps) {
   const {
-    resetPasswordUrl = `${getEndpoint(
-      BackendDestinationEnum.PORTAL_ENDPOINT,
-    )}PasswordReset:0`,
+    resetPasswordUrl = `https://accounts.sagebionetworks.synapse.org/resetPassword`,
```
nickgros, line 15: We will have to think about how we want to handle these links for each stack:
1. Should these links ensure you maintain in contact with the same backend environment?
2. Should links on e.g. staging.synapse.org send you to the staging accounts site, or always prod?

packages/synapse-react-client/src/components/ChangePassword/useChangePasswordFormState.tsx
```diff
@@ -44,8 +46,9 @@ export default function useChangePasswordFormState(
         // e.g. if the user enters the wrong 2FA code
         setTwoFactorAuthErrorResponse(maybeTwoFactorResponse)
       } else {
+        setSuccessfullyChangedPassword(true)
```
nickgros, line 49: We have to maintain this in a separate state variable, because `isSuccess` returned by the mutation will be true when the server issues a 2FA challenge

PR #955, opened 2024-05-24
PORTALS-3089 - Show error if param(s) are missing

nickgros:
See PORTALS-3089

image

### DIFF
--- a/apps/synapse-oauth-signin/src/OAuth2Form.tsx
+++ b/apps/synapse-oauth-signin/src/OAuth2Form.tsx
@@ -120,29 +120,51 @@ export function OAuth2Form() {
     )
   }
 
+  /**
+   * Parse the URL search params to construct an OIDC Authorization Request
+   *
+   * @returns OIDCAuthorizationRequest if the request is valid
+   * @returns undefined if the request is invalid. `onError` will be called with an error message.
+   */
   const oidcAuthorizationRequestFromSearchParams:
     | OIDCAuthorizationRequest
     | undefined = useMemo(() => {
-    const clientId = queryParams.get('client_id')
-    const scope = queryParams.get('scope')
-    const responseType = queryParams.get('response_type')
-    const redirectUri = queryParams.get('redirect_uri')
+    const missingParams: string[] = []
 
-    if (
-      clientId == null ||
-      scope == null ||
-      responseType == null ||
-      redirectUri == null
-    ) {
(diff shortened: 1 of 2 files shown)


apps/synapse-oauth-signin/src/OAuth2Form.tsx
```diff
@@ -120,37 +120,59 @@ export function OAuth2Form() {
...
-      // We don't have the params to construct the request
+    if (missingParams.length > 0) {
+      // We don't have the params required to construct the request
+      onError(
+        new Error(
+          `Invalid request. Missing required parameter(s): ${missingParams.join(
+            ', ',
+          )}`,
+        ),
+      )
```
nickgros, line 159: Previously, the error was handled server-side. The regression occurred because I added client-side validation to prevent sending an invalid request to the server, but didn't show an error if the validation failed. Now, an error is shown if any required params are missing.

apps/synapse-oauth-signin/src/test/App.test.tsx
```diff
@@ -32,24 +32,45 @@ const overrideWaitForOptions: waitForOptions = {
...
+  const params = new URLSearchParams()
+  Object.entries(paramsObject).forEach(([key, value]) => {
+    if (value == null) {
+      params.delete(key)
+    } else {
+      params.set(key, value)
+    }
+  })
   return params
 }
```
nickgros, line 68: Update test helpers to allow overriding required params like `response_type`

apps/synapse-oauth-signin/src/test/App.test.tsx
```diff
@@ -429,4 +450,22 @@ describe('App integration tests', () => {
...
+
+      await waitFor(() => {
+        const alert = screen.getByRole('alert')
+        within(alert).findByText(
+          `Invalid request. Missing required parameter(s): ${missingParam}`,
+        )
+      })
+      expect(window.location.replace).not.toHaveBeenCalled()
+    },
+  )
```
nickgros, line 470: Verify the error is shown

apps/synapse-oauth-signin/src/test/App.test.tsx
```diff
@@ -429,4 +450,22 @@ describe('App integration tests', () => {
...
```
nickgros, line 470: Verify the alert is shown

jay-hodgson (approved):

PR #957, opened 2024-05-24
PORTALS-2692: add title, and hide (but still keep in DOM) if no conte…

jay-hodgson:
…nt after markdown processing


jay-hodgson:
Integrated into the ADKP study detail page.  There's no content for the current wiki, so the citations section does not appear:
Screen Shot 2024-05-24 at 1 00 42 PM

PR #959, opened 2024-05-24
PORTALS-3091: refactor to share file configuration

jay-hodgson:


jay-hodgson:
NF portal config change only, will be verified by validator of PORTALS-3091

PR #960, opened 2024-05-28
PORTALS-3087: retrieve description field feature flag in src

jinjunoh:


packages/synapse-react-client/src/synapse-queries/featureflags/useGetFeatureFlags.ts
```diff
@@ -0,0 +1,16 @@
...
+  options?: Partial<UseQueryOptions<FeatureFlags, SynapseClientError>>,
+) {
+  const { keyFactory } = useSynapseContext()
+  return useQuery<FeatureFlags, SynapseClientError>({
+    staleTime: Infinity,
+    ...options,
+    queryKey: keyFactory.getFeatureFlagQueryKey(),
+    queryFn: () => SynapseClient.getFeatureFlags(),
+  })
+}
```
nickgros, line 16: Could do something similar to SWC where you make a custom UseQuery hook with the signature `useGetFeatureFlag(String FeatureFlagEnum)`. I think it would be pretty simple, something like this:

```suggestion
}

export function useGetFeatureFlag(
  featureFlag: FeatureFlagEnum
  options?: Partial<UseQueryOptions<boolean, SynapseClientError>>,
) {
  return useGetFeatureFlags({
    select: featureFlags => featureFlags[featureFlag]
  })
}
```

packages/synapse-react-client/src/components/entity/page/title_bar/TitleBarProperties.integration.test.tsx
```diff
@@ -71,6 +71,25 @@ function useDoiAssociationOverride(doiAssociation: DoiAssociation | null) {
...
+          ctx.status(200),
+          ctx.json({
+            'mock feature flag': true,
+          }),
+          ctx.set('Content-Type', 'application/json'), // Ensure header names are valid
+        )
+      },
+    ),
+  )
+}
```
nickgros, line 91: Can you put a reusable version of this in `mocks/msw/handlers/featureFlagHandlers.ts`, then use that in this file?

packages/synapse-react-client/src/components/CreateTableViewWizard/CreateTableViewWizard.integration.test.tsx
```diff
@@ -70,6 +70,25 @@ async function getLatestCreatedColumnModelIdsFromSpy(
   ).list.map(cm => cm.id)
 }
 
+function useGetFeatureFlagsOverride() {
```
nickgros, line 73: Can you put a reusable version of this in `mocks/msw/handlers/featureFlagHandlers.ts`, then use that in this file?

packages/synapse-react-client/src/components/CreateTableViewWizard/TableNameForm.tsx
```diff
@@ -16,7 +18,7 @@ type TableNameFormProps = {
 export default function TableNameForm(props: TableNameFormProps) {
   const { name, setName, description, setDescription } = props
   const { isInExperimentalMode } = useSynapseContext()
-
+  const { data: featureFlags } = useGetFeatureFlags()
```
nickgros, line 21: Could use the suggested `useGetFeatureFlag(FeatureFlagEnum.DESCRIPTION_FIELD)` (see comment on `useGetFeatureFlags.ts`)

packages/synapse-react-client/src/components/CreateTableViewWizard/TableNameForm.tsx
```diff
@@ -28,7 +30,8 @@ export default function TableNameForm(props: TableNameFormProps) {
         }}
         fullWidth
       />
-      {isInExperimentalMode && (
+      {(isInExperimentalMode ||
```
nickgros, line 33: With the suggested solution, we can also get rid of the `isInExperimentalMode` check, since that hook should do it for us

packages/synapse-react-client/src/synapse-queries/featureflags/useGetFeatureFlags.ts
```diff
@@ -0,0 +1,20 @@
+import { useQuery, UseQueryOptions } from '@tanstack/react-query'
+import SynapseClient from '../../synapse-client'
+import { SynapseClientError, useSynapseContext } from '../../utils'
+import { FeatureFlags, FeatureFlagEnum } from '@sage-bionetworks/synapse-types'
+
+export function useGetFeatureFlags(
```
nickgros, line 6: Rename if this hook only provides one flag value
```suggestion
export function useGetFeatureFlag(
```

nickgros (approved):

PR #962, opened 2024-05-28
PORTALS-3090: OneSage Synapse Resources page update

jay-hodgson:


apps/SageAccountWeb/src/components/AccountSettings.tsx
```diff
@@ -236,7 +236,9 @@ export const AccountSettings = () => {
                 </Typography>
                 <Typography variant={'body1'}>
                   This information is reused across all{' '}
-                  <RouterLink to="/sageresources">Sage products.</RouterLink>
+                  <RouterLink to="/sageresources">
+                    Synapse resources.
+                  </RouterLink>
```
nickgros, line 241: "Synapse resources" to refer to these apps feels awkward. I would expect "Synapse resources" to mean "content in Synapse"

nickgros (approved):

apps/SageAccountWeb/src/components/AccountSettings.tsx
jay-hodgson: Yes, we refer to them as products and resources (see Account Created Page above).  I'd be happy to change to products, but I just want them to be consistent.

PR #966, opened 2024-05-30
SWC-6845: migrate all experimental mode features to use feature flags

jinjunoh:


packages/synapse-types/src/FeatureFlags.ts
```diff
@@ -2,5 +2,8 @@ export type FeatureFlags = Record<FeatureFlagEnum, boolean>
 
 export enum FeatureFlagEnum {
   DESCRIPTION_FIELD = 'Description field',
-  PROVENANCE_V2_VISUALIZATION = 'Provenance v2 visualization',
+  CHANGE_PASSWORD_2FA_CHECK = 'Change password 2FA check',
+  VIRTUALTABLE_SUPPORT = 'VirtualTable support',
+  JSONSCHEMA_VALIDATION_STATUS = 'JSON Schema validation status',
+  JSONSCHEMA_ANNOTATION_VALIDATION_STATUS = 'JSON Schema-aware annotation editor',
```
nickgros, line 8: you can consolidate these into one. we won't release one without the other

nickgros, line 8: As a convention can we have the actual values always be SCREAMING_SNAKE_CASE ? That would reduce likelihood of error transcribing between AppConfig and source code
```suggestion
  CHANGE_PASSWORD_2FA_CHECK = 'CHANGE_PASSWORD_2FA_CHECK',
  VIRTUALTABLE_SUPPORT = 'VIRTUALTABLE_SUPPORT',
  JSONSCHEMA_VALIDATION_STATUS = 'JSONSCHEMA_VALIDATION_STATUS',
  JSONSCHEMA_ANNOTATION_VALIDATION_STATUS = 'JSONSCHEMA_ANNOTATION_VALIDATION_STATUS',
```

nickgros (approved):

PR #975, opened 2024-06-04
SWC-6861: Convert internal fragment-based Synapse GWT place links to …

jay-hodgson:
…path-based relative link

### DIFF
--- a/packages/markdown-it-synapse/src/index.ts
+++ b/packages/markdown-it-synapse/src/index.ts
@@ -337,6 +337,11 @@ function init_markdown_it(
         tokens[idx].attrs[hrefIndex][1],
         '#!',
       )
+      if (isInternalSynapseLink) {
+        tokens[idx].attrs[hrefIndex][1] = `/${tokens[idx].attrs[
+          hrefIndex
+        ][1].substring(2)}`
+      }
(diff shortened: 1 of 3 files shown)


packages/markdown-it-synapse/src/index.ts
```diff
@@ -337,6 +337,11 @@ function init_markdown_it(
         tokens[idx].attrs[hrefIndex][1],
         '#!',
       )
+      if (isInternalSynapseLink) {
+        tokens[idx].attrs[hrefIndex][1] = `/${tokens[idx].attrs[
+          hrefIndex
+        ][1].substring(2)}`
+      }
```
nickgros, line 344: is it a non-goal to make these links work in non-SWC apps at this time?

jay-hodgson: It's a fair point.  
I don't want to make too many assumptions about the environment that the wiki is rendered in.  
So if a portal owner added a relative link that points to a portal specific route (in a wiki), for example, it should work in the environment that they are targeting (on the portal), and be broken on www.synapse.org.  If they want it to work in multiple environments, then I think a full URL is required.
In this specific case, where we find a legacy '#!' fragment relative link, I think we can assume this was intended for our legacy GWT routing.

PR #977, opened 2024-06-05
PORTALS-3101

jay-hodgson:


apps/SageAccountWeb/src/components/ConfigureEmail.tsx
```diff
@@ -69,6 +69,7 @@ export const ConfigureEmail = (props: ConfigureEmailProps) => {
           accessToken,
         ).then(() => {
           displayToast('Email has been successfully added', 'success')
+          refetchCurrentProfile()
```
jay-hodgson, line 72: PORTALS-3101: Refetch after new email has been successfully added to the account

apps/SageAccountWeb/src/components/ConfigureEmail.tsx
```diff
@@ -99,6 +100,7 @@ export const ConfigureEmail = (props: ConfigureEmailProps) => {
     try {
       await SynapseClient.deleteEmail(accessToken, email)
       refetchCurrentProfile()
+      displayToast('Email has been successfully removed', 'success')
```
jay-hodgson, line 103: Noticed that there was no notification when an email was removed

packages/synapse-react-client/src/synapse-client/SynapseClient.ts
```diff
@@ -4256,7 +4256,7 @@ export const addEmailAddressStep2 = (
  */
 export const deleteEmail = (accessToken: string | undefined, email: string) => {
   return doDelete(
-    `/repo/v1/email?email=${email}`,
+    `/repo/v1/email?email=${encodeURIComponent(email)}`,
```
jay-hodgson, line 4259: And an email with a '+' was not encoded correctly in this call

apps/SageAccountWeb/src/components/ConfigureEmail.tsx
hallieswan: Could consider adding a notification to `changePrimaryEmail` when the primary email was successfully changed, but up to you!

hallieswan (approved):
Looks good to me!

apps/SageAccountWeb/src/components/ConfigureEmail.tsx
jay-hodgson: Good idea

hallieswan (approved):

PR #996, opened 2024-06-10
PORTALS-3139: fix URL construction bug found in One Sage prod

jay-hodgson:


jay-hodgson:
Merging to fix blocker bug on prod (to allow for some testing)

PR #997, opened 2024-06-10
PORTALS-3131: clean up certification quiz (removes bootstrap)

jay-hodgson:
https://sagebionetworks.jira.com/browse/PORTALS-3131

### DIFF
--- a/apps/SageAccountWeb/src/components/CertificationQuiz.tsx
+++ b/apps/SageAccountWeb/src/components/CertificationQuiz.tsx
@@ -2,29 +2,31 @@ import React from 'react'
 import { Link as RouterLink } from 'react-router-dom'
 import { MuiContainer, CertificationQuiz as Quiz } from 'synapse-react-client'
 import ReturnArrow from '../assets/ReturnArrow.svg?url'
-import { Link } from '@mui/material'
+import { Link, Paper } from '@mui/material'
 
 export const CertificationQuiz = () => {
   return (
     <div className="ProfileValidation blue-background">
       <MuiContainer style={{ padding: '40px' }}>
-        <Link
-          color="primary"
-          component={RouterLink}
-          to="/authenticated/myaccount"
-          sx={{
-            position: 'absolute',
-            top: '25px',
-            left: '30px',
-            paddingTop: '15px',
-            paddingBottom: '15px',
-            textAlign: 'center',
-          }}
-        >
-          <img className="arrow-icon" src={ReturnArrow} alt="return arrow" />
-          Return to Account Settings
-        </Link>
-        <Quiz />
+        <Paper sx={{ padding: '35px' }}>
+          <Link
+            color="primary"
+            component={RouterLink}
+            to="/authenticated/myaccount"
+            sx={{
+              position: 'absolute',
+              top: '25px',
+              left: '30px',
+              paddingTop: '15px',
+              paddingBottom: '15px',
+              textAlign: 'center',
+            }}
+          >
+            <img className="arrow-icon" src={ReturnArrow} alt="return arrow" />
+            Return to Account Settings
+          </Link>
+          <Quiz />
+        </Paper>
       </MuiContainer>
     </div>
   )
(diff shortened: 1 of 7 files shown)


jay-hodgson:
Screen Shot 2024-06-10 at 4 17 23 PM

Screen Shot 2024-06-10 at 4 18 07 PM

apps/SageAccountWeb/src/components/CertificationQuiz.tsx
```diff
@@ -1,19 +1,31 @@
...
 import ReturnArrow from '../assets/ReturnArrow.svg?url'
+import { Link } from '@mui/material'
 
 export const CertificationQuiz = () => {
   return (
-    <div className="ProfileValidation bootstrap-4-backport blue-background">
-      <Container style={{ padding: '40px' }}>
-        <Link className="return-link" to="/authenticated/myaccount">
+    <div className="ProfileValidation blue-background">
+      <MuiContainer style={{ padding: '40px' }}>
```
jay-hodgson, line 10: Replace bootstrap Container with our own drop-in replacement that uses MUI

jay-hodgson: Removed bootstrap-4-backport!

apps/SageAccountWeb/src/components/CertificationQuiz.tsx
```diff
@@ -1,19 +1,31 @@
...
+        <Link
```
jay-hodgson, line 11: Fix link style by using a MUI Link (still using the router Link under the hood)

packages/synapse-react-client/src/components/CertificationQuiz/CertificationQuiz.tsx
```diff
@@ -191,16 +192,32 @@ const CertificationQuiz: React.FunctionComponent = () => {
...
-              Help
+              <Box sx={{ display: 'flex', alignItems: 'center' }}>
+                <HelpOutlineTwoTone
+                  className="HelpButton"
+                  style={{ marginRight: '4px', color: 'white' }}
+                />
+                <>Help</>
+              </Box>
             </Button>
-            <div dangerouslySetInnerHTML={{ __html: quiz.header }}></div>
```
jay-hodgson, line 203: We need to style this section, but I'm sure this is controversial.  
We could modify the quiz header to not contain the quiz name or html, and just contain a plain text version of the description.  I suggest we implement the next time the quiz description needs to change.  By that time, the UI will be pointing at this snapshot in history, and we can modify the quiz value without impacting users.  Then we can switch this code to pull in the quiz.header again

packages/synapse-react-client/src/components/MuiContainer.tsx
```diff
@@ -0,0 +1,31 @@
+import { StyledComponent } from '@emotion/styled'
+import { Paper } from '@mui/material'
+import { PaperProps, styled } from '@mui/material'
+
+/**
+ * Drop in MUI substitute class for Bootstrap Container component
+ */
+export const MuiContainer: StyledComponent<PaperProps> = styled(Paper, {
```
jay-hodgson, line 8: New StyledComponent (based on Paper) that sets the width dynamically based on the current theme breakpoint

nickgros: May want to base this on `Box` instead of `Paper`, and just wrap in `Paper` (with `px: 15px`) when appropriate (with `px: 15px`)

nickgros: You get slightly different behavior if you use [MUI's built-in](https://mui.com/material-ui/api/container/) `Container` with the `fixed` prop, but would that be 'good enough' as a drop-in replacement?

packages/synapse-react-client/src/components/CertificationQuiz/CertificationQuiz.tsx
nickgros: As long as we remember to make that change, sounds good to me!

nickgros (approved):
This looks so much better! Had a few comments about the custom styled component, but nothing blocking approval. There also appears to be a test failure

PR #1000, opened 2024-06-11
SWC-6871 - fix reset2FA link to point to OneSage

nickgros:


packages/synapse-react-client/src/components/Authentication/LoginForm.tsx
```diff
@@ -36,6 +35,9 @@ type Props = {
 
 export default function LoginForm(props: Props) {
   const defaultRegistrationUrl = useOneSageURL('/register1')
+  const defaultTwoFactorAuthResetUrl = useOneSageURL(
+    '/reset2FA?twoFAResetToken=',
+  )
```
nickgros, line 40: Sort of causes a regression where staging.accounts.synapse.org will now create an email that points to accounts.synapse.org, but I think that's ok. We can fix that in SWC-6870

jay-hodgson (approved):

PR #1003, opened 2024-06-11
PORTALS-3140

jay-hodgson:
Add cbioportal linkify rule, and additional column to NF dataset card


hallieswan (approved):

jay-hodgson:
Also double-checked that each value in the String_list will be processed.
Screen Shot 2024-06-11 at 4 01 08 PM

PR #1004, opened 2024-06-12
SWC-6872 - Fix case where 2FA could not be reset during OAuth

nickgros:
- Fix case where 2FA could not be reset during OAuth
- Release synapse-react-client v3.2.22

### DIFF
--- a/packages/synapse-react-client/src/utils/hooks/useLogin.test.tsx
+++ b/packages/synapse-react-client/src/utils/hooks/useLogin.test.tsx
@@ -46,6 +46,22 @@ function renderUseLogin(
   })
 }
 
+async function verify2FAResetStarted(
+  result: ReturnType<typeof renderUseLogin>['result'],
+  resetEndpoint: string,
+) {
+  await waitFor(() => {
+    expect(mockReset2FA).toHaveBeenCalledWith({
+      userId: twoFactorAuthErrorResponse.userId,
+      twoFaToken: twoFactorAuthErrorResponse.twoFaToken,
+      twoFaResetEndpoint: resetEndpoint,
+    })
+    expect(mockSetAccessTokenCookie).not.toHaveBeenCalled()
+    expect(onSuccessfulLoginFn).not.toHaveBeenCalled()
+    expect(result.current.step).toBe('DISABLE_2FA_PROMPT')
+  })
+}
+
 describe('useLogin tests', () => {
   beforeEach(() => {
     jest.clearAllMocks()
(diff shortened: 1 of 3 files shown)


packages/synapse-react-client/src/utils/hooks/useLogin.test.tsx
```diff
@@ -403,21 +419,92 @@ describe('useLogin tests', () => {
...
     })
 
+    // Change the step so the UI shows the correct prompt
+    act(() => {
+      result.current.onStepChange('DISABLE_2FA_PROMPT')
+    })
+    await waitFor(() => {
+      expect(result.current.step).toBe('DISABLE_2FA_PROMPT')
+    })
+
```
nickgros, line 429: Update the test to simulate what the components are doing. This alone doesn't catch the bug, because the screen flashes back after the `expect` completes.

packages/synapse-react-client/src/utils/hooks/useLogin.test.tsx
```diff
@@ -46,6 +46,22 @@ function renderUseLogin(
...
+) {
+  await waitFor(() => {
+    expect(mockReset2FA).toHaveBeenCalledWith({
+      userId: twoFactorAuthErrorResponse.userId,
+      twoFaToken: twoFactorAuthErrorResponse.twoFaToken,
+      twoFaResetEndpoint: resetEndpoint,
+    })
+    expect(mockSetAccessTokenCookie).not.toHaveBeenCalled()
+    expect(onSuccessfulLoginFn).not.toHaveBeenCalled()
+    expect(result.current.step).toBe('DISABLE_2FA_PROMPT')
```
nickgros, line 61: This set of assertions was extracted. I added an assertion `expect(result.current.step).toBe('DISABLE_2FA_PROMPT')` which fails when it encounters this bug.

packages/synapse-react-client/src/utils/hooks/useLogin.test.tsx
```diff
@@ -403,21 +419,92 @@ describe('useLogin tests', () => {
...
+
+    // Trigger sending the email
     act(() => {
       result.current.beginTwoFactorAuthReset(resetEndpoint)
     })
 
+    await verify2FAResetStarted(result, resetEndpoint)
+  })
+
+  it('Allows the user to begin the 2fa reset process when a 2FA token is passed via props', async () => {
```
nickgros, line 438: Add additional tests to handle OAuth cases

jay-hodgson (approved):

nickgros:
Tests are failing even after rolling back changes. Merging and will open a new ticket to investigate test failures

PR #1008, opened 2024-06-13
SWC-6872 - Fix issue with URL generation including malformed searchParams

nickgros:


packages/synapse-react-client/src/components/Authentication/LoginForm.tsx
```diff
@@ -33,11 +33,22 @@ type Props = {
...
+    // Already has params, so use '&'
+    return `${url.toString()}&${queryParam}=`
+  }
+}
+
 export default function LoginForm(props: Props) {
   const defaultRegistrationUrl = useOneSageURL('/register1')
-  const defaultTwoFactorAuthResetUrl = useOneSageURL(
-    '/reset2FA?twoFAResetToken=',
-  )
```
nickgros, line 40: Previously this appended the appId, which would end up causing the twoFAResetToken to be in the wrong place after the server appended the token.

jay-hodgson (approved):

PR #1018, opened 2024-06-17
PORTALS-3155

jay-hodgson:


apps/SageAccountWeb/src/components/SourceAppImage.tsx
```diff
@@ -12,25 +15,24 @@ const SourceAppImage: React.FC<SourceAppImageProps> = (
...
-      })
-    }
-  }, [fileHandleId])
-  return svg ? (
+  const fha: FileHandleAssociation = {
+    associateObjectId: 'syn45291362',
+    associateObjectType: FileHandleAssociateType.TableEntity,
+    fileHandleId: fileHandleId ?? '',
+  }
+  const { data: svg } = SynapseQueries.useGetPresignedUrlContentFromFHA(
```
jay-hodgson, line 23: Use react query

packages/synapse-react-client/src/synapse-queries/KeyFactory.ts
```diff
@@ -518,11 +531,19 @@ export class KeyFactory {
...
-      forceAnonymous,
-    )
+    if (forceAnonymous) {
+      return this.getKeyAnonymous(
+        'presignedUrlContentFromFHA',
+        fileHandleAssociation,
+        forceAnonymous,
+      )
+    } else {
+      return this.getKey(
```
jay-hodgson, line 541: getKey always includes the accessKey (if set)

packages/synapse-react-client/src/synapse-queries/file/useFiles.ts
```diff
@@ -13,6 +13,50 @@ import {
...
+  useEffect(() => {
+    if (error) {
+      console.error(
+        `Failed to fetch file object. See network log for details: FileHandleAssociation=${JSON.stringify(
+          fileHandleAssociation,
+        )}`,
+      )
+    }
+  }, [error, fileHandleAssociation])
+  return useCreateUrlForData(blob)
```
nickgros, line 58: I like that you used `useCreateUrlForData` to handle the entire resource-from-presigned URL lifecycle here. I suggest also including the object returned by `useQuery` so that loading/error states can be tracked and handled by the component calling this hook.

```suggestion
  return { dataUrl: useCreateUrlForData(blob), queryResult }
```

nickgros (approved):
LGTM, had one suggested change. The tangential improvements to query key logic / tests are greatly appreciated

nickgros (approved):

PR #1025, opened 2024-06-24
PORTALS-3168 - Upgrade TypeScript to v5.5

nickgros:
- Upgrade TypeScript to v5.5.2
 - Update tsconfig.base.json to use new `configDir` feature
 - Fix new lint error in SummaryTable.tsx

### DIFF
--- a/apps/portals/package.json
+++ b/apps/portals/package.json
@@ -12,6 +12,7 @@
     "@sage-bionetworks/react-base-table": "^1.13.4",
     "@sage-bionetworks/synapse-types": "workspace:*",
     "@tanstack/react-query": "5.22.2",
+    "@upsetjs/react": "^1.11.0",
     "clone-deep": "^4.0.1",
     "dayjs": "^1.11.10",
     "katex": "^0.16.10",
@@ -27,8 +28,7 @@
     "react-spinners": "^0.13.8",
     "react-switch": "^7.0.0",
     "sass": "^1.71.1",
-    "synapse-react-client": "workspace:*",
-    "@upsetjs/react": "^1.11.0"
+    "synapse-react-client": "workspace:*"
   },
   "scripts": {
     "clean": "rimraf coverage",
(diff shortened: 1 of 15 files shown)


apps/portals/package.json
```diff
@@ -12,6 +12,7 @@
     "@sage-bionetworks/react-base-table": "^1.13.4",
     "@sage-bionetworks/synapse-types": "workspace:*",
     "@tanstack/react-query": "5.22.2",
+    "@upsetjs/react": "^1.11.0",
```
nickgros, line 15: relocated (automatically by pnpm)

packages/synapse-react-client/src/components/ChangePassword/ChangePassword.integration.test.tsx
```diff
@@ -520,6 +520,9 @@ describe('ChangePassword tests', () => {
 
     await user.click(submitButton)
 
+    // Wait for the UI to update before we change the server behavior
+    await getTOTPInputs()
+
```
nickgros, line 525: Fixes flaky test. This test makes two calls to the same service. With this change we now wait for the UI to update after making the first call (that has a 2FA challenge) before preparing the mock server with the second mock response. Before, I think sometimes the server behavior would update before the first API call was even made.

packages/synapse-react-client/src/components/ChangePassword/ChangePasswordWithToken.tsx
```diff
@@ -4,7 +4,7 @@ import { PasswordResetSignedToken } from '@sage-bionetworks/synapse-types'
 import { displayToast } from '../ToastMessage'
 import useChangePasswordFormState from './useChangePasswordFormState'
 
-export type ChangePasswordWithTokenProps = {
```
nickgros, line 7: For whatever reason this export is triggering warnings in our tests. We're not using it anywhere, so for now just remove the export

jay-hodgson (approved):

PR #1031, opened 2024-06-26
SWC-6826: must set to sub domain to share cookie

jay-hodgson:
Try to set the cookie domain

### DIFF
--- a/apps/SageAccountWeb/src/components/AccountSettings.tsx
+++ b/apps/SageAccountWeb/src/components/AccountSettings.tsx
@@ -105,7 +105,7 @@ export const AccountSettings = () => {
     cookies.set(SynapseConstants.DATETIME_UTC_COOKIE_KEY, isUTCTime, {
       path: '/',
       expires: nextYear,
-      domain: hostname.endsWith('.synapse.org') ? '.synapse.org' : undefined,
+      domain: hostname.endsWith('.synapse.org') ? 'synapse.org' : undefined,
     })
   }, [isUTCTime])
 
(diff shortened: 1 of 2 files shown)


apps/SageAccountWeb/src/components/AccountSettings.tsx
```diff
@@ -101,9 +101,11 @@ export const AccountSettings = () => {
     const current = new Date()
     const nextYear = new Date()
     nextYear.setFullYear(current.getFullYear() + 1)
+    const hostname = window.location.hostname.toLowerCase()
     cookies.set(SynapseConstants.DATETIME_UTC_COOKIE_KEY, isUTCTime, {
       path: '/',
       expires: nextYear,
+      domain: hostname.endsWith('.synapse.org') ? '.synapse.org' : undefined,
```
nickgros, line 108: [MDN](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Set-Cookie#domaindomain-value)
> Contrary to earlier specifications, leading dots in domain names (.example.com) are ignored.
>
> Multiple host/domain values are not allowed, but if a domain is specified, then subdomains are always included.

```suggestion
      domain: hostname.endsWith('.synapse.org') ? 'synapse.org' : undefined,
```

packages/synapse-react-client/src/utils/hooks/useCookiePreferences.ts
```diff
@@ -58,9 +58,11 @@ export const useCookiePreferences = (): [
       const current = new Date()
       const nextYear = new Date()
       nextYear.setFullYear(current.getFullYear() + 1)
+      const hostname = window.location.hostname.toLowerCase()
       cookies.set(COOKIES_AGREEMENT_COOKIE_KEY, prefs, {
         path: '/',
         expires: nextYear,
+        domain: hostname.endsWith('.synapse.org') ? '.synapse.org' : undefined,
```
nickgros, line 65:
```suggestion
        domain: hostname.endsWith('.synapse.org') ? 'synapse.org' : undefined,
```

nickgros (approved):

PR #1036, opened 2024-06-27
PORTALS-3176

jay-hodgson:


packages/synapse-react-client/src/components/DatasetJsonLdScript.tsx
```diff
@@ -0,0 +1,84 @@
...
+    version = versionFromProps
+  } else if (searchParams) {
+    entityId = searchParams['id']
+    if (searchParams['version']) {
+      version = parseInt(searchParams['version'])
+    }
+  }
+  const { data: entityBundle } = useGetEntityBundle(entityId!, version)
+  const { data: wikiPage } = useWikiPage(entityId!)
+  const html = md.renderInline(wikiPage?.markdown ?? '')
```
nickgros, line 54: I could see us re-using and gradually improving a utility to get plain text from Synapse wiki markdown, so you may consider extracting some of this logic into its own file

packages/synapse-react-client/src/synapse-queries/wiki/useWikiPage.ts
```diff
@@ -0,0 +1,20 @@
+import { ObjectType, WikiPage } from '@sage-bionetworks/synapse-types'
+import { useQuery, UseQueryOptions } from '@tanstack/react-query'
+import { SynapseClientError } from '../../utils/SynapseClientError'
+import { useSynapseContext } from '../../utils'
+import SynapseClient from '../../synapse-client'
+
+export function useWikiPage(
```
nickgros, line 7: Is this different from useWiki.ts `useGetWikiPage`?

packages/synapse-react-client/src/utils/hooks/useJsonLdScriptElement.ts
```diff
@@ -0,0 +1,21 @@
+import { useEffect } from 'react'
+import { Thing, WithContext } from 'schema-dts'
```
nickgros, line 2: These JSON-LD types seem useful to have

nickgros (approved):

packages/synapse-react-client/src/synapse-queries/wiki/useWikiPage.ts
jay-hodgson: Nice catch.  I pulled in changes from an old PR, perhaps useGetWikiPage was unavailable at that time.  Replaced with useWiki.useGetWikiPage()!

PR #1044, opened 2024-07-03
PORTALS-2519 - disable running msw tests in serial

nickgros:
Test passed 4 times in CI consecutively, job time decreased from ~16 min to ~11 min.


jay-hodgson:
Nice!

jay-hodgson (approved):

PR #1047, opened 2024-07-03
SWC-6721

nickgros:
Fix content overflow in EntityFinderModal

### DIFF
--- a/packages/synapse-react-client/src/components/EntityFinder/EntityFinderModal.tsx
+++ b/packages/synapse-react-client/src/components/EntityFinder/EntityFinderModal.tsx
@@ -56,8 +56,16 @@ export const EntityFinderModal = (props: EntityFinderModalProps) => {
           }
         }}
         content={
-          <Box width={props.configuration.treeOnly ? 'min-content' : undefined}>
-            <Typography variant="body1" mb="10px">
+          <Box>
+            <Typography
+              variant="body1"
+              mb="10px"
+              sx={{
+                // Do not allow the help text to expand the dialog width
+                minWidth: '100%',
+                width: 0,
+              }}
+            >
               {props.promptCopy}
             </Typography>
             <EntityFinder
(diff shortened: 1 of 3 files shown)


packages/synapse-react-client/src/style/components/entity_finder/_entity-finder.scss
```diff
@@ -57,15 +56,9 @@ $-splitter-width: 12px;
     height: $-finder-height + ($-border-size * 2);
   }
 
-  &__MainPanel.MainPanelSearch {
-    width: 85vw;
```
nickgros, line 61: These `vw` units caused overflow when 85% of the viewport width was greater than the maximum MUI dialog content width

packages/synapse-react-client/src/components/EntityFinder/EntityFinderModal.tsx
```diff
@@ -56,8 +56,16 @@ export const EntityFinderModal = (props: EntityFinderModalProps) => {
...
+          <Box>
+            <Typography
+              variant="body1"
+              mb="10px"
+              sx={{
+                // Do not allow the help text to expand the dialog width
+                minWidth: '100%',
+                width: 0,
+              }}
+            >
```
nickgros, line 68: The `width: min-content` style was designed to prevent this scenario, where the paragraph text would expand the dialog width even when it should be narrow:

image

The styles above directly applied to the paragraph solve this in a more predictable way that doesn't affect other EntityFinder elements

packages/synapse-react-client/src/style/components/entity_finder/_details-view.scss
```diff
@@ -27,6 +26,7 @@ $row-hovered-background-color: unset;
   height: $-finder-height;
   width: 100%;
   background-color: SRC.$background-color-gray-light;
+  overflow-y: hidden;
```
nickgros, line 29: Apply `overflow-y: hidden` in all cases. This causes the overflow to be applied to the table instead of the container wrapping the table, so now the container that holds the search field does not stretch to match the width of the outer container.

jay-hodgson (approved):

PR #1050, opened 2024-07-03
Bump ADTR fileview version for portal

ryaxley:
Increment fileview version


nickgros (approved):

nickgros:
Going to try re-running the e2e tests. At first glance, it seems like it's an issue with data loading rather than an issue with the view.

PR #1058, opened 2024-07-10
SWC-6933

nickgros:
- Make ID utilities case-insensitive
- Add error boundary to AR subjects editor

### DIFF
--- a/packages/synapse-react-client/src/components/SetAccessRequirementCommonFields/SetAccessRequirementCommonFields.tsx
+++ b/packages/synapse-react-client/src/components/SetAccessRequirementCommonFields/SetAccessRequirementCommonFields.tsx
@@ -24,11 +24,12 @@ import {
   useGetAccessRequirements,
   useUpdateAccessRequirement,
 } from '../../synapse-queries'
-import { SynapseClientError } from '../../utils/SynapseClientError'
+import { SynapseClientError } from '../../utils'
 import EntitySubjectsSelector from '../EntitySubjectsSelector'
 import TeamSubjectsSelector from '../TeamSubjectsSelector'
 import { RadioGroup } from '../widgets/RadioGroup'
 import { Checkbox } from '../widgets/Checkbox'
+import { SynapseErrorBoundary } from '../error'
 
 export const EMPTY_SUBJECT_LIST_ERROR_MESSAGE =
   'Please select at least one resource for this Access Requirement to be associated with.'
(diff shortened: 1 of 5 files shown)


packages/synapse-react-client/src/components/SetAccessRequirementCommonFields/SetAccessRequirementCommonFields.tsx
```diff
@@ -311,7 +312,7 @@ export const SetAccessRequirementCommonFields = React.forwardRef(
             />
           </>
         )}
-        {selectorContent}
+        <SynapseErrorBoundary>{selectorContent}</SynapseErrorBoundary>
```
nickgros, line 315: Wrap in error boundary to limit the impact of such an error in the future

packages/synapse-react-client/src/utils/functions/EntityTypeUtils.ts
```diff
@@ -305,8 +305,8 @@ export function getVersionDisplay(entity: Entity): string {
  * @param entityId
  */
 export function normalizeSynPrefix(entityId: string) {
-  if (entityId.startsWith('syn')) {
-    return entityId
+  if (entityId.toLowerCase().startsWith('syn')) {
+    return entityId.toLowerCase()
```
nickgros, line 309: Fix the issue -- before `Syn123` would be transformed into `synSyn123`

packages/synapse-react-client/src/utils/functions/RegularExpressions.ts
```diff
@@ -17,7 +18,7 @@ export const DOI_REGEX = /^10.\d{4,9}\/[-._;()/:a-z0-9]+$/i
  * > result[2]
  * '9'
  */
-export const SYNAPSE_ENTITY_ID_REGEX = /^(syn\d+)(?:\.(\d+))?$/
+export const SYNAPSE_ENTITY_ID_REGEX = /^(syn\d+)(?:\.(\d+))?$/i
```
nickgros, line 21: make this regex case insensitive, hoping that this will preemptively solve other issues that may be lurking

packages/synapse-react-client/src/utils/functions/RegularExpressions.test.ts
```diff
@@ -0,0 +1,14 @@
+import { parseSynId } from './RegularExpressions'
+
+describe('RegularExpressions', () => {
+  describe('parseSynId', () => {
+    test.each([
+      ['syn123', { targetId: 'syn123', targetVersionNumber: undefined }],
+      ['syn123.4', { targetId: 'syn123', targetVersionNumber: 4 }],
+      ['Syn123.4', { targetId: 'syn123', targetVersionNumber: 4 }],
+      ['123.4', null],
```
nickgros, line 9: For `parseSynId`, I am explicitly not handling the case with no syn prefix because that's the current behavior.

jay-hodgson (approved):

PR #1062, opened 2024-07-10
SWC-6927 - Do not allow adding Dataset Collection query to the download cart

nickgros:


packages/synapse-react-client/src/components/QueryContext/QueryContext.tsx
```diff
@@ -9,16 +9,6 @@ import {
...
 
-/*
-  For details page: to lock a column (e.g. study, grant) so that the facet values and active filters
-  will not appear on the details page. The facet name is given by the URL search param.
-  The type is defined here so that other child components in SRC won't generate type errors.
- */
-export type LockedColumn = {
-  columnName?: string
-  value?: string
-}
```
nickgros, line 20: In this PR I updated queryUtils.ts but I noticed that `queryUtils` imported this `LockedColumn` type. To prevent circular dependencies, `utils/` should never import from `components/`. So I moved `LockedColumn` to `utils/types`

packages/synapse-react-client/src/utils/functions/queryUtils.ts
```diff
@@ -201,7 +206,8 @@ export function canTableQueryBeAddedToDownloadList<T extends Table = Table>(
 ) {
   return Boolean(
     entity &&
-      ((isEntityView(entity) && isFileView(entity)) || isDataset(entity)),
+      ((isEntityView(entity) && isFileView(entity)) || isDataset(entity)) &&
+      !isDatasetCollection(entity),
```
nickgros, line 210: The actual change is just this

packages/synapse-react-client/src/utils/functions/queryUtils.ts
```diff
@@ -196,12 +202,29 @@ export function hasResettableFilters(
   return hasFacetFilters || hasAdditionalFilters
 }
 
+/**
+ * Returns true iff a table query can be added to the download list.
+ * @param entity
+ * @param entityColumnId
+ */
 export function canTableQueryBeAddedToDownloadList<T extends Table = Table>(
```
nickgros, line 210: Refactor to put all logic to determine if we can do this in this utility

packages/synapse-react-client/src/utils/functions/queryUtils.ts
```diff
@@ -196,12 +202,29 @@ export function hasResettableFilters(
...
   entity?: T,
+  entityColumnId?: string,
 ) {
+  const viewCannotIncludeFiles =
+    // EntityViews without the file bit mask cannot contain files
+    (entity && isEntityView(entity) && !hasFilesInView(entity)) ||
+    // DatasetCollections cannot contain files
+    (entity && isDatasetCollection(entity))
```
nickgros, line 218: Dataset collections should never allow you to add a query to the DL list

packages/synapse-react-client/test/containers/queryUtils.test.ts
```diff
@@ -309,6 +352,29 @@ describe('facet support', () => {
         true,
       )
     })
+    test('dataset collection cannot be added to download list', () => {
```
jay-hodgson, line 355: May want to add a comment that points to SWC-6927

jay-hodgson (approved):

packages/synapse-react-client/test/containers/queryUtils.test.ts
nickgros:
```suggestion
    test('dataset collection cannot be added to download list (SWC-6927)', () => {
```

PR #1073, opened 2024-07-15
SWC-6800 - Sort resource access in ACL editor 

nickgros:
- Update `useUpdateAcl` to sort access requirements by permission/username. Data is not re-sorted as ACL is edited.
- Update `AclEditorProps` to allow each ACL editor to pick their own default ACCESS_TYPE(s) to assign to newly added entries on an ACL


packages/synapse-react-client/src/components/AccessRequirementAclEditor/AccessRequirementAclEditor.tsx
```diff
@@ -166,7 +166,9 @@ export const AccessRequirementAclEditor = React.forwardRef(
           isLoading={isLoadingOriginalAcl}
           isInEditMode={true}
           emptyText={EMPTY_RESOURCE_ACCESS_LIST_TEXT}
-          addResourceAccessItem={addResourceAccessItem}
+          onAddPrincipalToAcl={id =>
+            addResourceAccessItem(id, [ACCESS_TYPE.REVIEW_SUBMISSIONS])
+          }
```
nickgros, line 171: change this callback so each implementation of the AclEditor can supply its own default access type (e.g. for ARs, the default resource access set is `['REVIEW_SUBMISSIONS']`, for non-public usergroups on entities it will be `['READ', 'DOWNLOAD']`)

packages/synapse-react-client/src/components/AclEditor/useSortResourceAccessList.ts
```diff
@@ -0,0 +1,68 @@
...
+    data: userGroupHeadersOnResourceAccessList,
+    isLoading,
+    error,
+  } = useGetUserGroupHeaders(principalIdsOnResourceAccessList.map(String))
+
+  useEffect(() => {
+    if (error) {
+      console.error('Error fetching user group headers: ', error)
+    }
+  }, [error])
```
nickgros, line 31: This error is not a showstopper, so just log

packages/synapse-react-client/src/components/AclEditor/useSortResourceAccessList.ts
```diff
@@ -0,0 +1,68 @@
...
+      obj => Boolean(obj.userGroupHeader),
+    )
+
+    if (!allEntriesHaveUserGroupHeader) {
+      console.warn(
+        'Some ACL entries do not have a corresponding UserGroupHeader. The ResourceAccess list will not be sorted. Missing entries: ',
+        joinedWithUserGroupHeaders.filter(obj => !obj.userGroupHeader),
+      )
+      return resourceAccessList
+    }
```
nickgros, line 57: I have a feeling this could happen if the ACL is very large (not sure if we have limits today)

packages/synapse-react-client/src/components/AclEditor/ResourceAccessAndUserGroupHeader.test.ts
```diff
@@ -0,0 +1,146 @@
...
+      }
+      const b = {
+        resourceAccess: {
+          principalId: 2,
+          accessType: [ACCESS_TYPE.CHANGE_PERMISSIONS],
+        },
+        userGroupHeader: userGroupHeaderB,
+      }
+
+      expect(compareResourceAccessAndUserGroupHeader(a, b)).toBe(-1)
```
jay-hodgson, line 67: Just for readability, it might be helpful to define an enum.  Something like:
```
enum CompareResult {
  A_FIRST = -1,
  B_FIRST = 1
}
```

packages/synapse-react-client/src/components/AclEditor/useSortResourceAccessList.ts
jay-hodgson: We tried to sort but failed - they will still have the resource access list.  Might be difficult to manage without the user group headers

packages/synapse-react-client/src/components/AclEditor/useUpdateAcl.ts
```diff
@@ -12,15 +13,33 @@ type UseUpdateAclOptions = {
...
+
+  // While the form has not been edited, sort the resourceAccessList
+  const { sortedResourceAccessList, isLoading: isLoadingSortedList } =
+    useSortResourceAccessList(resourceAccessList)
+  useEffect(() => {
+    if (!isDirty && !isLoadingSortedList && !hasSorted) {
+      setResourceAccessList(sortedResourceAccessList)
+      // `useSortResourceAccessList` will return a new array every time `resourceAccessList` changes, so we need to
+      // track if we've already sorted to prevent an infinite loop
+      setHasSorted(true)
```
jay-hodgson, line 30: Thank you for stopping the cycle

jay-hodgson (approved):

PR #1086, opened 2024-07-20
SWC-6983, and point to production by default

jay-hodgson:


packages/synapse-react-client/src/components/SynapseHomepageV2/SynapseHomepageNavBar.tsx
```diff
@@ -67,7 +67,7 @@ export const SynapseHomepageNavBar: React.FunctionComponent<
         alignItems: 'center',
         columnGap: '5px',
         p: {
-          xs: '10px 0px',
+          xs: '10px 5px',
```
jay-hodgson, line 70: Removed the padding that storybook adds when wrapping the component, and realized the synapse logo was bumping against the left side of the viewport.  Really only noticeable on mobile

packages/synapse-react-client/src/components/SynapseHomepageV2/SynapseHomepageV2.stories.ts
```diff
@@ -35,7 +35,7 @@ export const DemoVersion2: Story = {
     },
   },
   parameters: {
-    stack: 'mock',
+    stack: 'production',
```
jay-hodgson, line 38: This should fix the many bugs that were created because users were looking at cached results from the mock service

PR #1087, opened 2024-07-21
SWC-6954

jay-hodgson:
Move the SynapseInAction images into the SynapseInActionItem, and change the layout (to a flex grid).  
And instead of fading in when the element is in view, set the desktop image opacity based on the element position in the viewport!


nickgros (approved):
seems pretty straightforward!

PR #1091, opened 2024-07-22
SWC-6985 - Fix bug where Safari users + 2FA cannot log in / reset 2FA

nickgros:
Fix case where trailing hash was transformed and then added to the twoFaToken

### DIFF
--- a/packages/synapse-react-client/src/utils/hooks/useLogin.ts
+++ b/packages/synapse-react-client/src/utils/hooks/useLogin.ts
@@ -101,7 +101,6 @@ export default function useLogin(opts: UseLoginOptions): UseLoginReturn {
     if (searchParams) {
       const userId = searchParams.get('userId')
       const twoFaToken = searchParams.get('twoFaToken')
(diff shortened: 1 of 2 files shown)


packages/synapse-react-client/src/utils/hooks/useLogin.ts
```diff
@@ -95,13 +95,13 @@ export default function useLogin(opts: UseLoginOptions): UseLoginReturn {
...
   useEffect(() => {
-    // Replace the hash with a slash to handle SWC, which uses a hash in the URL. Otherwise, the searchParams aren't found
-    const fullUrl: URL = new URL(window.location.href.replaceAll('#', '/'))
+    const fullUrl: URL = new URL(window.location.href)
 
     const { searchParams } = fullUrl
     if (searchParams) {
       const userId = searchParams.get('userId')
       const twoFaToken = searchParams.get('twoFaToken')
+      console.log('twoFaToken search param:', twoFaToken)
```
nickgros, line 104:
```suggestion
```

PR #1102, opened 2024-07-24
SWC-6800 - EntityAclEditor

nickgros:
Still needs light refactoring + SWC integration

### DIFF
--- a/packages/synapse-react-client/src/components/AclEditor/AclEditor.test.tsx
+++ b/packages/synapse-react-client/src/components/AclEditor/AclEditor.test.tsx
@@ -1,4 +1,4 @@
-import { render, screen, waitFor } from '@testing-library/react'
+import { render, screen, waitFor, within } from '@testing-library/react'
 import userEvent from '@testing-library/user-event'
 import React from 'react'
 import { server } from '../../mocks/msw/server'
@@ -20,9 +20,16 @@ import { PermissionLevel } from '../../utils/PermissionLevelToAccessType'
 import {
   addUserToAcl,
   confirmItem,
+  queryForAddUserCombobox,
   removeItem,
   updatePermissionLevel,
 } from './AclEditorTestUtils'
+import {
+  ANONYMOUS_PRINCIPAL_ID,
+  AUTHENTICATED_PRINCIPAL_ID,
+  PUBLIC_PRINCIPAL_ID,
+  PUBLIC_PRINCIPAL_IDS,
+} from '../../utils/SynapseConstants'
 
 const DEFAULT_RESOURCE_ACCESS: ResourceAccess[] = [
   {
@@ -59,10 +66,12 @@ const defaultProps: AclEditorProps = {
   availablePermissionLevels: DEFAULT_AVAILABLE_PERMISSION_LEVELS,
   isLoading: false,
   emptyText: DEFAULT_EMPTY_TEXT,
-  isInEditMode: true,
+  canEdit: true,
   onAddPrincipalToAcl: mockAddResourceAccessItem,
   updateResourceAccessItem: mockUpdateResourceAccessItem,
   removeResourceAccessItem: mockRemoveResourceAccessItem,
+  showNotifyCheckbox: false,
+  showAddRemovePublicButton: false,
 }
 
 function renderComponent(props: AclEditorProps) {
@@ -81,10 +90,16 @@ async function setUp(
   const user = userEvent.setup()
   const { component } = renderComponent(props)
 
+  const numberOfPublicPrincipalsInResourceAccess =
+    props.resourceAccessList.filter(ra =>
+      PUBLIC_PRINCIPAL_IDS.includes(ra.principalId),
+    ).length
+
   // wait for UserOrTeamBadge(s) to finish loading
   await waitFor(() => {
     expect(screen.queryAllByRole('link')).toHaveLength(
-      expectedResourceAccessItems,
+      // public principals do not get a link, so subtract that amount from the expected count
+      expectedResourceAccessItems - numberOfPublicPrincipalsInResourceAccess,
     )
   })
 
@@ -158,8 +173,8 @@ describe('AclEditor', () => {
     expect(mockRemoveResourceAccessItem).toHaveBeenCalledWith(MOCK_TEAM_ID)
   })
 
-  it('does not show edit controls when isInEditMode is false', async () => {
-    const { itemRows } = await setUp({ isInEditMode: false })
+  it('does not show edit controls when canEdit is false', async () => {
+    const { itemRows } = await setUp({ canEdit: false })
 
     // Data should still be present
     confirmItem(itemRows[0], mockTeamData.name, 'Can Review')
@@ -173,5 +188,184 @@ describe('AclEditor', () => {
     // No menus or buttons should be shown
     expect(screen.queryByRole('combobox')).not.toBeInTheDocument()
     expect(screen.queryByRole('button')).not.toBeInTheDocument()
+  })
+
+  it('allows customizing which entries are editable by passing a function to canEdit', async () => {
+    function allowEditingOnlyTeam1(resourceAccess: ResourceAccess) {
+      return resourceAccess.principalId === MOCK_TEAM_ID
+    }
+
+    const { itemRows } = await setUp({
+      canEdit: allowEditingOnlyTeam1,
+      canRemoveEntry: true,
+    })
+
+    // Data should still be present
+    confirmItem(itemRows[0], mockTeamData.name, 'Can Review')
+    confirmItem(itemRows[1], mockTeamData2.name, 'Exempt Eligible')
+    confirmItem(
+      itemRows[2],
+      `@${mockUserData1.userProfile?.userName}`,
+      'Can Review & Exempt Eligible',
+    )
+
+    // team 1 is editable and deletable
+    within(itemRows[0]).getByRole('combobox')
+    within(itemRows[0]).getByRole('button')
+    // team2 and user1 are deletable but not editable
+    expect(within(itemRows[1]).queryByRole('combobox')).not.toBeInTheDocument()
+    within(itemRows[1]).getByRole('button')
+    expect(within(itemRows[2]).queryByRole('combobox')).not.toBeInTheDocument()
+    within(itemRows[2]).getByRole('button')
+
+    // Users can still be added
+    expect(queryForAddUserCombobox()).toBeInTheDocument()
+  })
+
+  it('respects the value of canRemoveEntry (boolean)', async () => {
+    const { itemRows } = await setUp({
+      canEdit: true,
+      canRemoveEntry: false,
+    })
+
+    // Data should still be present
+    confirmItem(itemRows[0], mockTeamData.name, 'Can Review')
+    confirmItem(itemRows[1], mockTeamData2.name, 'Exempt Eligible')
+    confirmItem(
+      itemRows[2],
+      `@${mockUserData1.userProfile?.userName}`,
+      'Can Review & Exempt Eligible',
+    )
+
+    // all rows are editable but not deletable
+    within(itemRows[0]).getByRole('combobox')
+    expect(within(itemRows[0]).queryByRole('button')).not.toBeInTheDocument()
+    within(itemRows[1]).getByRole('combobox')
+    expect(within(itemRows[1]).queryByRole('button')).not.toBeInTheDocument()
+    within(itemRows[2]).getByRole('combobox')
+    expect(within(itemRows[2]).queryByRole('button')).not.toBeInTheDocument()
+
+    // Users can still be added
+    expect(queryForAddUserCombobox()).toBeInTheDocument()
+  })
+
+  it('respects the value of canRemoveEntry (function)', async () => {
+    function allowRemovingOnlyTeam1(resourceAccess: ResourceAccess) {
+      return resourceAccess.principalId === MOCK_TEAM_ID
+    }
+
+    const { itemRows } = await setUp({
+      canEdit: true,
+      canRemoveEntry: allowRemovingOnlyTeam1,
+    })
+
+    // Data should still be present
+    confirmItem(itemRows[0], mockTeamData.name, 'Can Review')
+    confirmItem(itemRows[1], mockTeamData2.name, 'Exempt Eligible')
+    confirmItem(
+      itemRows[2],
+      `@${mockUserData1.userProfile?.userName}`,
+      'Can Review & Exempt Eligible',
+    )
+
+    // team 1 is editable and deletable
+    within(itemRows[0]).getByRole('combobox')
+    within(itemRows[0]).getByRole('button')
+    // team2 and user1 are editable but not deletable
+    within(itemRows[1]).getByRole('combobox')
+    expect(within(itemRows[1]).queryByRole('button')).not.toBeInTheDocument()
+    within(itemRows[2]).getByRole('combobox')
+    expect(within(itemRows[2]).queryByRole('button')).not.toBeInTheDocument()
+
+    // Users can still be added
+    expect(queryForAddUserCombobox()).toBeInTheDocument()
+  })
+
+  it('allows overriding a displayed permission value', async () => {
+    const overrideText = 'Can take over the world'
+    function displayedPermissionLevelOverride(resourceAccess: ResourceAccess) {
+      if (resourceAccess.principalId === MOCK_TEAM_ID) {
+        return overrideText
+      }
+      return undefined
+    }
+    const { itemRows } = await setUp({
+      canEdit: false,
+      displayedPermissionLevelOverride,
+    })
+
+    // Team 1 gets the override
+    confirmItem(itemRows[0], mockTeamData.name, overrideText)
+
+    // Remaining teams are unchanged
+    confirmItem(itemRows[1], mockTeamData2.name, 'Exempt Eligible')
+    confirmItem(
+      itemRows[2],
+      `@${mockUserData1.userProfile?.userName}`,
+      'Can Review & Exempt Eligible',
+    )
+  })
+
+  it('shows a checkbox to notify users', async () => {
+    const onCheckboxChange = jest.fn()
+    const { user } = await setUp({
+      showNotifyCheckbox: true,
+      notifyCheckboxValue: true,
+      onNotifyCheckboxChange: onCheckboxChange,
+    })
+
+    const checkbox = screen.getByLabelText('Notify people via email')
+    expect(checkbox).toHaveAttribute('value', 'true')
+
+    await user.click(checkbox)
+
+    expect(onCheckboxChange).toHaveBeenLastCalledWith(false)
+  })
+
+  it('shows a button to add public', async () => {
+    const { user } = await setUp({
+      showAddRemovePublicButton: true,
+    })
+
+    const makePublicButton = screen.getByRole('button', { name: 'Make Public' })
+    await user.click(makePublicButton)
+
+    expect(mockAddResourceAccessItem).toHaveBeenCalledWith(
+      String(PUBLIC_PRINCIPAL_ID),
+    )
+    expect(mockAddResourceAccessItem).toHaveBeenCalledWith(
+      String(AUTHENTICATED_PRINCIPAL_ID),
+    )
+  })
+
+  it('shows a button to remove public if any public principals are in the ACL', async () => {
+    const { user } = await setUp({
+      resourceAccessList: [
+        {
+          principalId: PUBLIC_PRINCIPAL_ID,
+          accessType: [ACCESS_TYPE.REVIEW_SUBMISSIONS],
+        },
+        {
+          principalId: AUTHENTICATED_PRINCIPAL_ID,
+          accessType: [ACCESS_TYPE.REVIEW_SUBMISSIONS],
+        },
+      ],
+      showAddRemovePublicButton: true,
+    })
+
+    const removePublicAccessButton = screen.getByRole('button', {
+      name: 'Remove Public Access',
+    })
+    await user.click(removePublicAccessButton)
+
+    expect(mockRemoveResourceAccessItem).toHaveBeenCalledWith(
+      PUBLIC_PRINCIPAL_ID,
+    )
+    expect(mockRemoveResourceAccessItem).toHaveBeenCalledWith(
+      AUTHENTICATED_PRINCIPAL_ID,
+    )
+    expect(mockRemoveResourceAccessItem).toHaveBeenCalledWith(
+      ANONYMOUS_PRINCIPAL_ID,
+    )
   })
 })
--- a/packages/synapse-react-client/src/components/AclEditor/AclEditorSkeleton.tsx
+++ b/packages/synapse-react-client/src/components/AclEditor/AclEditorSkeleton.tsx
@@ -1,0 +1,17 @@
+import React from 'react'
+import { Skeleton, Stack } from '@mui/material'
+import { SkeletonParagraph, SkeletonTable } from '../Skeleton'
+
+export function AclEditorSkeleton() {
+  return (
+    <Stack gap={2} role={'progressbar'}>
+      <Skeleton height={50} width={200} />
+      <SkeletonParagraph numRows={2} />
+
+      <SkeletonTable numCols={2} numRows={4} rowHeight={'30px'} />
+
+      <Skeleton height={50} width={100} />
+      <SkeletonParagraph numRows={2} />
+    </Stack>
+  )
+}
--- a/packages/synapse-react-client/src/components/AclEditor/AclEditorTestUtils.ts
+++ b/packages/synapse-react-client/src/components/AclEditor/AclEditorTestUtils.ts
@@ -2,12 +2,65 @@ import userEvent from '@testing-library/user-event'
 import { screen, waitFor, within } from '@testing-library/react'
 import { REMOVE_BUTTON_LABEL } from './ResourceAccessItem'
 
+/**
+ * Find a row in the ACL editor that contains the specified principal name.
+ * @param rows
+ * @param principalName
+ */
+export function queryForRowWithPrincipalName(
+  rows: HTMLElement[],
+  principalName: string,
+) {
+  return rows.find(row => {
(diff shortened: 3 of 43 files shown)


packages/synapse-react-client/src/components/AclEditor/AclEditor.test.tsx
```diff
@@ -174,4 +189,183 @@ describe('AclEditor', () => {
     expect(screen.queryByRole('combobox')).not.toBeInTheDocument()
     expect(screen.queryByRole('button')).not.toBeInTheDocument()
   })
+
+  it('allows customizing which entries are editable by passing a function to canEdit', async () => {
```
nickgros, line 193: In the Entity ACL editor, we must prevent users from editing the permission level of the public group ('Can view' only, unless it supports anonymous download access, in which case it is 'Can download' only).

Verify that passing a function to `canEdit` allows the parent component to apply such a restriction.

packages/synapse-react-client/src/components/AclEditor/AclEditor.test.tsx
```diff
@@ -174,4 +189,183 @@ describe('AclEditor', () => {
...
+    within(itemRows[1]).getByRole('combobox')
+    expect(within(itemRows[1]).queryByRole('button')).not.toBeInTheDocument()
+    within(itemRows[2]).getByRole('combobox')
+    expect(within(itemRows[2]).queryByRole('button')).not.toBeInTheDocument()
+
+    // Users can still be added
+    expect(queryForAddUserCombobox()).toBeInTheDocument()
+  })
+
+  it('allows overriding a displayed permission value', async () => {
```
nickgros, line 284: E.g. the parent component can change the text to 'Can download' in the case where the public group has READ ('Can view') and the entity allows anonymous download.

packages/synapse-react-client/src/components/AclEditor/AclEditorSkeleton.tsx
```diff
@@ -0,0 +1,17 @@
+import React from 'react'
+import { Skeleton, Stack } from '@mui/material'
+import { SkeletonParagraph, SkeletonTable } from '../Skeleton'
+
+export function AclEditorSkeleton() {
```
nickgros, line 5: Create a more accurate skeleton

packages/synapse-react-client/src/components/AclEditor/AclEditorTestUtils.ts
```diff
@@ -2,12 +2,65 @@ import userEvent from '@testing-library/user-event'
...
+ */
 export function confirmItem(
   row: HTMLElement,
   principalName: string,
   accessTypeLabel: string,
 ) {
-  expect(within(row).getByRole('link')).toHaveTextContent(principalName)
+  expect(within(row).queryByRole('link') ?? row).toHaveTextContent(
+    principalName,
+  )
```
nickgros, line 63: Public/authenticated groups don't have links, so we have to adjust this selector

packages/synapse-react-client/src/components/AclEditor/AclEditorTestUtils.ts
```diff
@@ -67,3 +142,33 @@ export async function addUserToAcl(
...
+
+  let rows: HTMLElement[] = []
+  let publicRow: HTMLElement | undefined = undefined
+  let authenticatedUsersRow: HTMLElement | undefined = undefined
+  await waitFor(() => {
+    rows = screen.getAllByRole('row')
+    publicRow = queryForRowWithPrincipalName(rows, 'Anyone on the web')
+    authenticatedUsersRow = queryForRowWithPrincipalName(
+      rows,
+      'All registered Synapse users',
```
nickgros, line 164: TODO: use constants

packages/synapse-react-client/src/components/AclEditor/useSortResourceAccessList.ts
```diff
@@ -22,7 +22,9 @@ export default function useSortResourceAccessList(
     data: userGroupHeadersOnResourceAccessList,
     isLoading,
     error,
-  } = useGetUserGroupHeaders(principalIdsOnResourceAccessList.map(String))
+  } = useGetUserGroupHeaders(principalIdsOnResourceAccessList.map(String), {
+    enabled: principalIdsOnResourceAccessList.length > 0,
```
nickgros, line 26: prevent attempting to fetch usergroupheaders in the case where the parent component passes an empty resource access list

packages/synapse-react-client/src/components/AclEditor/useSortResourceAccessList.ts
```diff
@@ -32,7 +34,7 @@ export default function useSortResourceAccessList(
 
   const sortedResourceAccessList = useMemo(() => {
     if (!userGroupHeadersOnResourceAccessList) {
-      return resourceAccessList
+      return null
```
nickgros, line 37: Return null so it's clear to the caller if the sort was actually successful.

packages/synapse-react-client/src/components/AclEditor/useUpdateAcl.test.ts
```diff
@@ -165,5 +165,20 @@ describe('useUpdateAcl', () => {
         { principalId: MOCK_USER_ID_2, accessType: [ACCESS_TYPE.READ] },
       ]),
     )
+
+    // Verify that if we reset the dirty state, the list is re-sorted
+    act(() => {
+      result.current.resetDirtyState()
```
nickgros, line 171: We may want to reset the state and re-sort in various circumstances:
* After a save + close 
* After the user adds/removes a local ACL (the editor should reset and we should sort the inherited ACL)

packages/synapse-react-client/src/components/AclEditor/useUpdateAcl.ts
```diff
@@ -35,57 +40,64 @@ export default function useUpdateAcl(options: UseUpdateAclOptions) {
...
-        const updatedResourceAccessList = [
-          ...resourceAccessList,
-          newResourceAccess,
-        ]
-        setResourceAccessList(updatedResourceAccessList)
+  const addResourceAccessItem = useCallback(
+    (newReviewerId: string | null, accessTypes: ACCESS_TYPE[]) => {
+      setIsDirty(true)
+      if (newReviewerId) {
+        setResourceAccessList(resourceAccessList => {
```
nickgros, line 47: use setState callback for all of these so consecutive calls are properly batched

packages/synapse-react-client/src/components/AclEditor/useUpdateAcl.ts
```diff
@@ -35,57 +40,64 @@ export default function useUpdateAcl(options: UseUpdateAclOptions) {
...
+        raListItem => raListItem.principalId !== principalId,
+      ),
     )
-    setResourceAccessList(updatedResourceAccessList)
-  }
+  }, [])
+
+  const resetDirtyState = useCallback(() => {
+    setIsDirty(false)
+  }, [])
```
nickgros, line 93: TODO: Add a return type and comment so it's clear that `resetDirtyState` basically exists to re-sort.

packages/synapse-react-client/src/components/EntityAclEditor/EntityAclEditor.stories.tsx
```diff
@@ -0,0 +1,65 @@
+import { Meta, StoryObj } from '@storybook/react'
+import mockProject from '../../mocks/entity/mockProject'
+import mockFileEntity from '../../mocks/entity/mockFileEntity'
+import EntityAclEditorModal, {
+  EntityAclEditorModalProps,
+} from './EntityAclEditorModal'
+import { mockFileEntityCurrentUserCannotEdit } from '../../mocks/entity/mockFileEntityACLVariants'
+
+const meta: Meta<EntityAclEditorModalProps> = {
```
nickgros, line 9: TODO: Replace with only mocks

packages/synapse-react-client/src/components/EntityAclEditor/EntityAclEditor.test.tsx
```diff
@@ -0,0 +1,612 @@
...
+  await waitFor(() => {
+    itemRows = screen.queryAllByRole('row')
+    expect(itemRows).toHaveLength(expectedResourceAccessItems)
+  })
+
+  return { ref, component, itemRows, user }
+}
+
+async function checkNotifyUsers(user: ReturnType<(typeof userEvent)['setup']>) {
+  const notifyCheckbox = screen.getByLabelText('Notify people via email')
```
nickgros, line 78: constant

packages/synapse-react-client/src/components/EntityAclEditor/EntityAclEditor.test.tsx
```diff
@@ -0,0 +1,612 @@
...
+                'READ',
+                'DOWNLOAD',
+                'UPDATE',
+                'DELETE',
+              ]),
+            },
+          ],
+        },
+        MOCK_ACCESS_TOKEN,
+      )
```
nickgros, line 150: try to simplify assertion (maybe with helper)

packages/synapse-react-client/src/components/EntityAclEditor/EntityAclEditor.test.tsx
```diff
@@ -0,0 +1,612 @@
...
+    await screen.findByText('This is anonymous access data.')
+    await screen.findByText(
+      'Anyone can download it, even if they aren’t logged in to Synapse.',
+    )
+
+    // Verify that the public and authenticated are displayed as 'Can download'
+    await confirmItemViaQuery(itemRows, 'Anyone on the web', 'Can download')
+    await confirmItemViaQuery(
+      itemRows,
+      'All registered Synapse users',
```
nickgros, line 424: constant

packages/synapse-react-client/src/components/EntityAclEditor/EntityAclEditor.tsx
```diff
@@ -0,0 +1,352 @@
...
+  )}Synapse:${entityId}`
+}
+
+const EntityAclEditor = React.forwardRef(function EntityAclEditor(
+  props: EntityAclEditorProps,
+  ref: React.ForwardedRef<EntityAclEditorHandle>,
+) {
+  const { entityId, onCanSaveChange, onUpdateSuccess } = props
+
+  const { data: ownProfile } = useSuspenseGetCurrentUserProfile()
```
nickgros, line 74: I decided to use `Suspense` and `useSuspenseQuery` here. The wrapped `Suspense` component (which is an ACL editor skeleton component) will be shown until these successfully resolve, so we don't need to handle loading states for these queries--IMO removing that loading state logic is worth the trade off in a component like this that is already complex.

If any of the queries fail, the error will be propagated to the nearest parent error boundary.

packages/synapse-react-client/src/components/EntityAclEditor/EntityAclEditor.tsx
```diff
@@ -0,0 +1,352 @@
...
+    entityId,
+    undefined,
+    ALL_ENTITY_BUNDLE_FIELDS,
+    { staleTime: Infinity },
+  )
+  const isProject = EntityType.PROJECT == entityBundle.entityType
+
+  const parentId = entityBundle.entity.parentId
+  const { data: parentAcl } = useGetEntityBenefactorACL(parentId!, {
+    enabled: parentId != null && !isProject,
```
nickgros, line 85: consider making suspense because I think this will work even for projects -- add a comment that it does work for projects.

packages/synapse-react-client/src/components/EntityAclEditor/EntityAclEditor.tsx
```diff
@@ -0,0 +1,352 @@
...
+    : false
+
+  function canDeleteResourceAccess(resourceAccess: ResourceAccess): boolean {
+    // Users cannot delete their own permissions
+    const isSelf = ownProfile.ownerId === String(resourceAccess.principalId)
+    if (isSelf) {
+      return false
+    }
+    return canEdit
+  }
```
nickgros, line 174: pull logic out of component & unit test

packages/synapse-react-client/src/components/EntityAclEditor/EntityAclEditor.tsx
```diff
@@ -0,0 +1,352 @@
...
+  function getDisplayedPermissionLevelOverride(
+    resourceAccess: ResourceAccess,
+  ): string | undefined {
+    if (resourceAccess.principalId === PUBLIC_PRINCIPAL_ID) {
+      return isOpenData
+        ? permissionLevelToLabel['CAN_DOWNLOAD']
+        : permissionLevelToLabel['CAN_VIEW']
+    }
+    return undefined
+  }
```
nickgros, line 185: pull logic out of component & unit test

packages/synapse-react-client/src/components/EntityAclEditor/OpenData.tsx
```diff
@@ -0,0 +1,75 @@
...
+})(({ theme }) => ({
+  background:
+    theme.palette.mode === 'light'
+      ? theme.palette.grey[100]
+      : tinycolor(theme.palette.background.paper).desaturate(1).toString(),
+  padding: `${theme.spacing(2.5)} ${theme.spacing(4)}`,
+  border:
+    theme.palette.mode === 'light'
+      ? `1px solid ${theme.palette.grey[300]}`
+      : 'none',
```
nickgros, line 18: remove light/dark mode for now, move to another changeset

packages/synapse-react-client/src/components/EntityAclEditor/useNotifyNewACLUsers.test.ts
```diff
@@ -100,6 +104,7 @@ describe('useNotifyNewACLUsers', () => {
         principalId: mockUserData2.id,
         accessType: getAccessTypeFromPermissionLevel('CAN_EDIT_DELETE'),
       },
+      // removed user 3
```
nickgros, line 107: Modify this test to also verify that users _removed_ from an ACL are also not sent an email notification

packages/synapse-react-client/src/components/TeamBadge.tsx
```diff
@@ -29,11 +30,11 @@ export default function TeamBadge(props: TeamBadgeProps) {
     disableHref = true
   }
 
-  const Tag = disableHref ? 'span' : 'a'
+  const Tag = disableHref ? 'span' : Link
 
   return (
-    <span>
-      <IconSvg icon={icon} />
+    <Box component={'span'} display={'inline-flex'} alignItems={'center'}>
```
nickgros, line 36: modify style to improve icon alignment with text

packages/synapse-react-client/src/mocks/entity/mockFileEntity.ts
```diff
@@ -146,34 +144,7 @@ const mockFileEntityBundle: EntityBundle = {
...
-        principalId: 273948,
-        accessType: [ACCESS_TYPE.READ],
-      },
-      {
-        principalId: 273949,
-        accessType: [ACCESS_TYPE.READ],
-      },
-    ],
-  },
+  benefactorAcl: mockProjectEntityData.bundle.accessControlList!,
```
nickgros, line 147: update this mock file entity so it directly uses the parent mock project's ACL

packages/synapse-react-client/src/mocks/entity/mockFileEntity.ts
```diff
@@ -236,17 +203,6 @@ const mockFileEntityHeader: EntityHeader = {
   isLatestVersion: true,
 }
 
-const generatedFileEntityData: MockEntityData<FileEntity>[] = times(50).map(
```
nickgros, line 239: moved generating entities to a different file to break a dependency cycle

packages/synapse-react-client/src/mocks/entity/mockProject.ts
```diff
@@ -84,7 +114,7 @@ const mockProjectEntityBundle: EntityBundle = {
     canPublicRead: true,
     canModerate: true,
     canMove: true,
-    isEntityOpenData: false,
+    isEntityOpenData: true,
```
nickgros, line 117: Change back to false

packages/synapse-react-client/src/mocks/msw/handlers/entityHandlers.ts
```diff
@@ -324,4 +324,85 @@ export const getEntityHandlers = (backendOrigin: string) => [
     }
     return res(ctx.status(200), ctx.json(response))
   }),
+
+  rest.post(
```
nickgros, line 328: Handlers for Entity ACL CRUD

packages/synapse-react-client/src/mocks/msw/handlers/fileHandlers.ts
```diff
@@ -29,5 +32,38 @@ export function getFileHandlers(backendOrigin: string) {
 
       return res(ctx.status(201), ctx.json(response))
     }),
+
+    rest.post(
```
nickgros, line 36: Handlers for file upload (needed because the message service uploads a file)

packages/synapse-react-client/src/synapse-queries/entity/useEntity.ts
```diff
@@ -414,6 +461,70 @@ export function useUpdateEntityACL(
...
+  }
+}
+
+/**
+ * Retrieve the ACL of an entity. This call will succeed even for entities where the caller
+ * does not have READ permission.
+ * @param entityId
+ * @param options
+ */
+export function useGetEntityBenefactorACL(
```
nickgros, line 511: Create a dedicated hook for fetching the `benefactorACL` via the entity bundle. This is valuable because this call will always succeed regardless of your permissions on an entity. 

If you were to request any additional bundle fields on an entity where you do not have READ access, the call will fail.

This is useful for the case where you can CHANGE_PERMISSIONS on an entity but you cannot READ the parent container. You can still remove local sharing settings on the file and see the parent benefactor ACL using this call.

packages/synapse-react-client/src/synapse-queries/entity/useEntityBundle.ts
```diff
@@ -7,17 +11,15 @@ import {
   EntityBundleRequest,
 } from '@sage-bionetworks/synapse-types'
 
-export function useGetEntityBundle<
+export function useGetEntityBundleQueryOptions<
```
nickgros, line 14: Extract logic to return query options, which can be passed into either `useQuery` or `useSuspenseQuery`

packages/synapse-react-client/src/synapse-queries/entity/useEntityBundle.ts
```diff
@@ -31,6 +33,47 @@ export function useGetEntityBundle<
         version,
         accessToken,
       ),
+  }
+}
+
+export function useGetEntityBundle<
```
nickgros, line 39: Existing method, same functionality, just uses refactored hook

packages/synapse-react-client/src/synapse-queries/entity/useEntityBundle.ts
```diff
@@ -31,6 +33,47 @@ export function useGetEntityBundle<
...
+    version,
+    bundleRequest,
+  )
+  return useQuery<EntityBundle<T>, SynapseClientError, TSelect>({
+    ...options,
+    ...queryOptions,
+  })
+}
+
+export function useSuspenseGetEntityBundle<
```
nickgros, line 61: Get an entity bundle but using [useSuspenseQuery](https://tanstack.com/query/latest/docs/framework/react/guides/suspense#suspense)

packages/synapse-react-client/src/components/EntityAclEditor/OpenData.tsx
nickgros: need to add button contrast text from theme changes

packages/synapse-react-client/src/components/EntityAclEditor/EntityAclEditor.tsx
```diff
@@ -0,0 +1,352 @@
...
+    isPendingDeleteAcl ||
+    isPendingSendMessageToNewACLUsers
+
+  const hasAclChanged = useMemo(() => {
+    return (
+      originalIsInherited != updatedIsInherited ||
+      !isEqualTreatArraysAsSets(
+        originalResourceAccess,
+        updatedResourceAccessList,
+      )
```
nickgros, line 229: It would be nice if we had a(n OpenAPI generated?) serialized class/type that used more 'correct' data types, like `Set`s. But if we're just working with JSON data, we need to use something like this

nickgros: I investigated this and the generator we would likely use (`typescript-fetch` from `openapi-generator`) does support this in the latest version, but the Synapse OpenAPI spec does not currently indicate which fields are sets. Here's an issue I created: https://sagebionetworks.jira.com/browse/PLFM-8516

packages/synapse-react-client/src/components/EntityAclEditor/EntityAclEditor.tsx
```diff
@@ -0,0 +1,352 @@
...
+    isPendingDeleteAcl ||
+    isPendingSendMessageToNewACLUsers
+
+  const hasAclChanged = useMemo(() => {
+    return (
+      originalIsInherited != updatedIsInherited ||
+      !resourceAccessListIsEqual(
+        originalResourceAccess,
+        updatedResourceAccessList,
+      )
```
nickgros, line 229: It would be nice if we had a(n OpenAPI generated?) serialized class/type that used more 'correct' data types, like `Set`s, which wouldn't require creating a utility just to check equality.

I investigated this and the generator we would likely use (`typescript-fetch` from `openapi-generator`) does support this in the latest version, but the Synapse OpenAPI spec does not currently indicate which fields are sets. Here's an issue I created: https://sagebionetworks.jira.com/browse/PLFM-8516

packages/synapse-react-client/src/synapse-queries/entity/useEntity.ts
```diff
@@ -414,6 +462,70 @@ export function useUpdateEntityACL(
...
+  }
+}
+
+/**
+ * Retrieve the ACL of an entity. This call will succeed even for entities where the caller
+ * does not have READ permission.
+ * @param entityId
+ * @param options
+ */
+export function useSuspenseGetEntityBenefactorACL(
```
nickgros, line 512: Created a dedicated hook for fetching the `benefactorACL` via the entity bundle. This is valuable because this call will always succeed regardless of your permissions on an entity. 

If you were to request any additional bundle fields on an entity where you do not have READ access, the call will fail.

This is useful for the case where you can CHANGE_PERMISSIONS on an entity but you cannot READ the parent container. You can still remove local sharing settings on the file and see the parent benefactor ACL using this call.

packages/synapse-react-client/src/components/AclEditor/useUpdateAcl.ts
```diff
@@ -23,7 +52,12 @@ export default function useUpdateAcl(options: UseUpdateAclOptions) {
   const { sortedResourceAccessList, isLoading: isLoadingSortedList } =
     useSortResourceAccessList(resourceAccessList)
   useEffect(() => {
-    if (!isDirty && !isLoadingSortedList && !hasSorted) {
+    if (
+      !isDirty &&
+      !isLoadingSortedList &&
+      !hasSorted &&
+      sortedResourceAccessList != null
```
jay-hodgson, line 59: And there's the check looking for your possibly null return value

packages/synapse-react-client/src/components/DialogBase.tsx
```diff
@@ -53,7 +53,9 @@ export function DialogBaseTitle(props: DialogBaseTitleProps) {
     <DialogTitle>
       <Stack direction="row" alignItems={'center'} gap={'5px'}>
         {title}
-        {titleHelpPopoverProps && <HelpPopover {...titleHelpPopoverProps} />}
+        <Box component={'span'} fontSize={'14px'}>
+          {titleHelpPopoverProps && <HelpPopover {...titleHelpPopoverProps} />}
+        </Box>
```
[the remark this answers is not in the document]
jay-hodgson: Eek, this might be an issue in other corners of Synapse.org this week...

packages/synapse-react-client/src/synapse-queries/entity/useEntity.ts
jay-hodgson: Interesting!

packages/synapse-react-client/src/synapse-queries/entity/useEntityBundle.ts
jay-hodgson: This is a clean pattern, do you recommend we start using this across the library?

jay-hodgson (approved):

packages/synapse-react-client/src/synapse-queries/entity/useEntityBundle.ts
nickgros: I think in many cases, it's fantastic, and 100% worth adopting!

We cannot use it in cases where the query is dependent (i.e. the `enabled` option may be `false`), and it can cause pretty bad request waterfalls if we have many nested components and poorly-targeted `Suspense` wrappers (since nested components do not mount until the parents' suspense queries resolve) . In the latter cases, I think the react-query docs recommend pre-fetching data into the cache as a way to avoid long delays.

packages/synapse-react-client/src/components/EntityAclEditor/CreateOrDeleteLocalSharingSettingsButton.tsx
```diff
@@ -0,0 +1,55 @@
...
+        >
+          {DELETE_LOCAL_SHARING_SETTINGS}
+        </Button>
+      </div>
+    )
+  }
+  return (
+    <div>
+      <Typography variant={'body1'} mt={2} mb={1}>
+        By default the sharing settings are inherited from the parent folder or
```
jay-hodgson, line 39: Could you update this text, @nickgros  per https://sagebionetworks.jira.com/browse/SWC-6925 ?

> Sharing settings are initially inherited from the parent folder or project by default. To customize settings for a specific file, folder, or table, you must create and adjust local sharing settings.

PR #1103, opened 2024-07-25
clean up Synapse Homepage V1

jay-hodgson:


nickgros (approved):
Can you also remove the dependency `@brainhubeu/react-carousel`?

nickgros (approved):

PR #1109, opened 2024-07-30
SWC-6945: remove markdown-it-emphasis-alt

jay-hodgson:


packages/synapse-react-client/package.json
```diff
@@ -68,7 +68,6 @@
     "markdown-it": "^14.1.0",
     "markdown-it-br": "^1.0.0",
     "markdown-it-center-text": "^1.0.4",
-    "markdown-it-emphasis-alt": "^1.0.8",
```
nickgros, line 71: You can also remove the type definition from markdown-it-plugins.d.ts

nickgros (approved):

nickgros (approved):

PR #1113, opened 2024-07-31
revert changes made for PORTALS-3194 

jay-hodgson:


jay-hodgson:
Unclear if we want to keep this facet value sorting behavior, so staging this PR so that it's ready if needed.

PR #1116, opened 2024-08-01
SWC-7011

jay-hodgson:


packages/synapse-react-client/src/utils/hooks/useScrollFadeTransition.ts
```diff
@@ -6,7 +6,8 @@ const calculateOpacity = (rect: DOMRect): number => {
   const middleRangeStart = viewportHeight * 0.4
   const middleRangeEnd = viewportHeight * 0.6
 
-  if (elementCenterY >= middleRangeStart && elementCenterY <= middleRangeEnd) {
+  if (elementCenterY <= middleRangeEnd) {
+    // && elementCenterY >= middleRangeStart) {
```
nickgros, line 10: was this commented out on purpose?

nickgros (approved):

packages/synapse-react-client/src/utils/hooks/useScrollFadeTransition.ts
jay-hodgson: Yes.  Adam would like to slightly change the fade in behavior (keep it solid/opaque as you scroll down the page past the element)

nickgros (approved):

PR #1117, opened 2024-08-01
PORTALS-2970 - allow resetting debounce timer without recreating query

nickgros:
Tracking component rerenders demo:

Before:

https://github.com/user-attachments/assets/e48d0458-bb1f-44e7-ad91-c3a90a68fe73

After:

https://github.com/user-attachments/assets/7a0699b0-644d-47ff-8b13-84100723167d

### DIFF
--- a/packages/synapse-react-client/src/components/widgets/query-filter/EnumFacetFilter/EnumFacetFilter.tsx
+++ b/packages/synapse-react-client/src/components/widgets/query-filter/EnumFacetFilter/EnumFacetFilter.tsx
@@ -44,6 +44,7 @@ export function EnumFacetFilter(props: EnumFacetFilterProps) {
     removeSelectedFacet,
     removeValueFromSelectedFacet,
     executeQueryRequest,
+    resetDebounceTimer,
   } = useQueryContext()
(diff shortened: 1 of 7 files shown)


packages/synapse-react-client/src/components/widgets/query-filter/EnumFacetFilter/EnumFacetFilter.tsx
```diff
@@ -162,12 +163,7 @@ export function EnumFacetFilter(props: EnumFacetFilterProps) {
       hideCollapsible={hideCollapsible}
       onHoverOverValue={() => {
         // SWC-6698: delay the query execution (via the debounce) when an item is hovered over
-        executeQueryRequest(
-          request => {
-            return cloneDeep(request)
-          },
-          { debounce: true },
-        )
```
nickgros, line 170: Calling `executeQueryRequest` would recreate the query object and trigger unnecessary rerenders every time an option was hovered.

Replace this call a function that does the bare minimum to avoid unnecessarily rerendering most of the QueryWrapper tree.

PR #1120, opened 2024-08-02
SWC-7014

jay-hodgson:


packages/synapse-react-client/src/components/SynapseHomepageV2/SynapseInActionItem.tsx
```diff
@@ -150,13 +153,20 @@ export const SynapseInActionItem: React.FunctionComponent<
...
           alignSelf: 'end',
-          opacity: opacity,
         }}
       >
-        <ImageFromSynapseTable
-          tableId={tableId}
-          fileHandleId={imageFileHandleId}
-        />
+        <Fade in={inView} timeout={transitionTimeoutMs}>
+          <div>
```
jay-hodgson, line 159: Div wrapper to avoid open MUI issue https://github.com/mui/material-ui/issues/27154

packages/synapse-react-client/src/utils/hooks/useFadeTransition.ts
```diff
@@ -1,28 +0,0 @@
-import { useState, useEffect } from 'react'
```
jay-hodgson, line 1: useFadeTransition() was to support a very specific case, where a Fade was being reused (when the image file handle changed, which we tracked in "state").  This is no longer the design, and I was confused by the purpose of this hook (so I can't imagine anyone else will use it!).

nickgros (approved):

PR #1121, opened 2024-08-02
SWC-7016: add wrapper container to set max width of 1500px

jay-hodgson:


jay-hodgson:
Terrible diffs for this PR, my apologies.  I added a Box wrapper around 2 sections.  I moved most of the sx from the original Box wrapper to the parent (for padding, background color, etc), and introduced new sx to the child Box that would limit the width and center (maxWidth: 1500px and margin: auto)

nickgros (approved):

PR #1127, opened 2024-08-13
Fix issue where handlers had a dependency cycle that would cause Storybook to not work

nickgros:
server.ts -> handlers.ts -> featureFlagHandlers.ts -> server.ts

### DIFF
--- a/packages/synapse-react-client/src/mocks/msw/handlers/featureFlagHandlers.ts
+++ b/packages/synapse-react-client/src/mocks/msw/handlers/featureFlagHandlers.ts
@@ -1,5 +1,4 @@
 import { rest } from 'msw'
-import { server } from '../server'
 import { FeatureFlagEnum, FeatureFlags } from '@sage-bionetworks/synapse-types'
 import { BackendDestinationEnum, getEndpoint } from '../../../utils/functions'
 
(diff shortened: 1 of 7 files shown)


packages/synapse-react-client/src/mocks/msw/handlers/featureFlagHandlers.ts
```diff
@@ -1,5 +1,4 @@
 import { rest } from 'msw'
-import { server } from '../server'
```
nickgros, line 2: `server.ts` imports `handlers.ts` which imports this file, so we cannot import from `server.ts` here.

packages/synapse-react-client/src/mocks/msw/handlers/featureFlagHandlers.ts
```diff
@@ -30,9 +29,3 @@ export function getFeatureFlagsOverride(
     )
   })
 }
-
-export function useGetFeatureFlagsOverride(
-  options?: FeatureFlagHandlerOptions,
-) {
-  server.use(getFeatureFlagsOverride(options))
-}
```
nickgros, line 38: Remove this, callers can use `server` themselves.

jay-hodgson (approved):

PR #1128, opened 2024-08-13
SWC-7026: optional callback on toggling experimental mode

jay-hodgson:


jay-hodgson:
Love to avoid a full page reload, but is there another way that would work in both a react app as well as our GWT-based app?

PR #1133, opened 2024-08-15
SWC-7010

nickgros:
Fixes SWC-7010


packages/synapse-react-client/src/components/dataaccess/SubmissionPage.integration.test.tsx
```diff
@@ -290,8 +292,7 @@ describe('Submission Page tests', () => {
     ).not.toBeInTheDocument()
   })
 
-  // Flaky in TravisCI
```
nickgros, line 293: I ran this entire suite 50 times locally and it never failed, so I think this is no longer flaky

jay-hodgson (approved):

PR #1137, opened 2024-08-15
and update the header

jay-hodgson:


jay-hodgson:
(based on content in the updated figma design file)

PR #1142, opened 2024-08-20
PORTALS-3052

jay-hodgson:


packages/synapse-react-client/src/components/FullTextSearch.tsx
```diff
@@ -36,23 +42,50 @@ export const FullTextSearch: React.FunctionComponent<FullTextSearchProps> = ({
...
-        }
-        // PORTALS-2093: does this additional filter already exist?
-        const found = additionalFilters.find(
-          filter =>
-            filter.concreteType == textMatchesQueryFilter.concreteType &&
-            filter.searchExpression == textMatchesQueryFilter.searchExpression,
-        )
-        if (found) {
-          return request
+        const idColumnModel = getColumnModel('id')
```
nickgros, line 45: Maybe use `getFileColumnModelId` (in SynapseTableUtils), unless that doesn't work for some known cases.

packages/synapse-react-client/src/components/FullTextSearch.tsx
```diff
@@ -36,23 +42,50 @@ export const FullTextSearch: React.FunctionComponent<FullTextSearchProps> = ({
...
+        const isSynapseID = searchText.match(SYNAPSE_ENTITY_ID_REGEX)
+        if (
+          isSynapseID &&
+          idColumnModel &&
+          idColumnModel.columnType == ColumnTypeEnum.ENTITYID
```
nickgros, line 50: Can remove this last check if you use `getFileColumnModelId`

packages/synapse-react-client/src/components/FullTextSearch.tsx
```diff
@@ -36,23 +42,50 @@ export const FullTextSearch: React.FunctionComponent<FullTextSearchProps> = ({
...
+        ) {
+          const singleValueQueryFilter: ColumnSingleValueQueryFilter = {
+            concreteType:
+              'org.sagebionetworks.repo.model.table.ColumnSingleValueQueryFilter',
+            columnName: 'id',
+            operator: ColumnSingleValueFilterOperator.IN,
+            values: [searchText],
+          }
+          const matchingFilter = additionalFilters.find(
```
nickgros, line 59: Nice touch to replace a filter with a different term on the same column, just want to a comment to make it clear what's happening
```suggestion
          // Replace the active filter on the column, if one exists
          const matchingFilter = additionalFilters.find(
```

nickgros (approved):

packages/synapse-react-client/src/components/FullTextSearch.tsx
jay-hodgson: Done

packages/synapse-react-client/src/components/FullTextSearch.tsx
jay-hodgson: Done

PR #1145, opened 2024-08-20
PORTALS-3204

jay-hodgson:


jay-hodgson:
image

packages/synapse-react-client/src/components/SynapseChat/SynapseChat.stories.ts
```diff
@@ -0,0 +1,34 @@
...
+export default meta
+type Story = StoryObj<typeof meta>
+
+export const ChatWithSynapse: Story = {
+  args: { initialMessage: 'hello' },
+  parameters: {
+    stack: 'mock',
+    design: {
+      type: 'figma',
+      url: 'https://www.figma.com/',
```
nickgros, line 24: :eyes:

packages/synapse-react-client/src/components/SynapseChat/SynapseChat.tsx
```diff
@@ -0,0 +1,131 @@
...
+  InputAdornment,
+  IconButton,
+} from '@mui/material'
+import { useTheme } from '@mui/material'
+import { ColorPartial } from '@mui/material/styles/createPalette'
+import { OutlinedInput } from '@mui/material'
+import { ArrowUpward } from '@mui/icons-material'
+import SynapseChatInteraction from './SynapseChatInteraction'
+import { SkeletonParagraph } from '../Skeleton'
+import { useCreateAgentSession } from 'src/synapse-queries/chat/useChat'
```
nickgros, line 17: surprised this non-relative import works

packages/synapse-react-client/src/components/SynapseChat/SynapseChat.tsx
```diff
@@ -0,0 +1,131 @@
...
+
+  useEffect(() => {
+    // on mount, create a new agent session!
+    if (createAgentSession) {
+      createAgentSession({
+        agentAccessLevel: AgentAccessLevel.PUBLICLY_ACCESSIBLE,
+        agentId: agentId,
+      })
+    }
+  }, [createAgentSession])
```
nickgros, line 48:
```suggestion
  useEffect(() => {
    // on mount, create a new agent session!
    if (createAgentSession && initialMessage && !agentSessionId) {
      createAgentSession({
        agentAccessLevel: AgentAccessLevel.PUBLICLY_ACCESSIBLE,
        agentId: agentId,
      })
    }
  }, [createAgentSession, initialMessage, agentSessionId])
```

Just in case something changes later on and we render this before the user has typed a message, let's check that a message is present and that we haven't already started a session.

IMO this would be more useful when accessible from any page, not just a single chat page, so while that won't be how it works in this initial version, I feel like there's a decent chance we'll eventually end up there

packages/synapse-react-client/src/components/SynapseChat/SynapseChat.tsx
```diff
@@ -0,0 +1,131 @@
...
+  initialMessage: string
+  agentId?: string // if provided, use this agent
+}
+
+export const SynapseChat: React.FunctionComponent<SynapseChatProps> = ({
+  initialMessage,
+  agentId,
+}) => {
+  // Create a new Agent Session on mount, add ability to initialize this component with a starting message from the user.
+  const [agentSessionId, setAgentSessionId] = useState('')
```
nickgros, line 30: You could just use `data` returned by useMutation

packages/synapse-react-client/src/components/SynapseChat/SynapseChat.tsx
```diff
@@ -0,0 +1,131 @@
...
+  const { mutate: createAgentSession } = useCreateAgentSession({
+    onSuccess: newAgentSession => {
+      setAgentSessionId(newAgentSession.sessionId)
+    },
+  })
+  const theme = useTheme()
+  const [messages, setMessages] = useState([initialMessage])
```
nickgros, line 37: I am very surprised that the set of messages is not stored and provided by the server

packages/synapse-react-client/src/components/SynapseChat/SynapseChat.tsx
```diff
@@ -0,0 +1,131 @@
...
+    if (currentMessage.trim()) {
+      setMessages([...messages, currentMessage])
+      setCurrentMessage('')
+    }
+  }
+
+  const handleKeyDown: KeyboardEventHandler<
+    HTMLInputElement | HTMLTextAreaElement
+  > = event => {
+    if (event.key === 'Enter') {
```
nickgros, line 60: does this allow Shift+Enter to start a new line?

packages/synapse-react-client/src/components/SynapseChat/SynapseChat.tsx
```diff
@@ -0,0 +1,131 @@
...
+                  <ArrowUpward />
+                </IconButton>
+              </InputAdornment>
+            }
+            placeholder={'Message SynapseChat'}
+            sx={{ borderRadius: 96.6 }}
+            onChange={e => setCurrentMessage(e.target.value)}
+            onKeyDown={handleKeyDown}
+          />
+        </FormControl>
```
nickgros, line 125: `TextField` can do all of this, I think this should be pretty close to what you have:

```suggestion
        <TextField
          fullWidth
          value={currentMessage}
          onChange={e => setCurrentMessage(e.target.value)}
          onKeyDown={handleKeyDown}
          placeholder={'Message SynapseChat'}
          sx={{ m: 1 }}
          InputProps={{
            sx: { borderRadius: 96.6 },
            endAdornment: (
              <IconButton
                disabled={!isMessage}
                onClick={handleSendMessage}
                sx={{
                  ml: '7px',
                  mr: '13px',
                  color: sendMessageButtonColor,
                  borderStyle: 'solid',
                  borderWidth: isMessage ? '2px' : '1px',
                  borderColor: isMessage ? sendMessageButtonColor : 'gray',
                }}
              >
                <ArrowUpward />
              </IconButton>
            ),
          }}
        />
```

packages/synapse-react-client/src/components/SynapseChat/SynapseChatInteraction.tsx
```diff
@@ -0,0 +1,57 @@
...
+      sessionId: sessionId,
+    }
+  }, [userMessage])
+  const { data: chatResponse, isLoading } =
+    useGetAgentChatWithAsyncStatus(agentChatRequest)
+
+  const chatResponseText = chatResponse?.responseBody?.responseText
+  return (
+    <>
+      <ListItem sx={{ justifyContent: 'flex-end' }}>
```
nickgros, line 29: Rendering a list aligned with flex is clever, I like it

packages/synapse-react-client/src/components/SynapseChat/SynapseChatInteraction.tsx
```diff
@@ -0,0 +1,57 @@
...
+export const SynapseChatInteraction: React.FunctionComponent<
+  SynapseChatInteractionProps
+> = ({ sessionId, userMessage }) => {
+  const theme = useTheme()
+  const agentChatRequest: AgentChatRequest = useMemo(() => {
+    return {
+      chatText: userMessage,
+      sessionId: sessionId,
+    }
+  }, [userMessage])
```
nickgros, line 22:
```suggestion
  }, [userMessage, sessionId])
```

packages/synapse-react-client/src/synapse-queries/chat/useChat.ts
```diff
@@ -0,0 +1,82 @@
...
+    ...options,
+    queryKey: keyFactory.getAgentChatWithAsyncStatusQueryKey(request),
+
+    queryFn: () =>
+      SynapseClient.getAgentChatAsyncJobResults(
+        request,
+        accessToken,
+        setCurrentAsyncStatus,
+      ),
+  })
```
nickgros, line 81: I suspect this service is not idempotent, and will query the LLM each time it's called, so this should probably be a `useMutation` (unless I'm wrong!)

packages/synapse-react-client/src/components/SynapseHomepageV2/SynapseHomepageChatSearch.tsx
```diff
@@ -0,0 +1,96 @@
...
+                onChange={v => {
+                  setMode(v.target.value as SearchMode)
+                  //restore search or chat value
+                }}
+              >
+                <MenuItem value={SearchMode.SEARCH}>Search</MenuItem>
+                <MenuItem value={SearchMode.CHAT}>Chat</MenuItem>
+              </Select>
+            </InputAdornment>
+          }
```
nickgros, line 79: feature flag check to hide chat feature?

packages/synapse-react-client/src/components/SynapseChat/SynapseChatInteraction.tsx
```diff
@@ -0,0 +1,57 @@
...
+  const { data: chatResponse, isLoading } =
+    useGetAgentChatWithAsyncStatus(agentChatRequest)
```
nickgros, line 24: mentioned elsewhere, this should probably be a mutation (you might need to move this to the outer component)

packages/synapse-react-client/src/components/SynapseChat/SynapseChat.tsx
jay-hodgson: me too!  I'll fix it (I'm sure it will break at some point)

packages/synapse-react-client/src/components/SynapseChat/SynapseChat.tsx
jay-hodgson: Like the agentSessionId check, let's not create a new agent session if we already have one!  Not sure about the initialMessage in this logic.  Whether we have an initial message or not, I think we should establish the agent session ID.

packages/synapse-react-client/src/components/SynapseChat/SynapseChat.tsx
jay-hodgson: Good point.  Added this comment to the LLM doc (asking John):
> Nick brought up a good point after looking at my client prototype implementation.  Will there be a service call to ask for all existing request chatText and response responseText values in the history of this chat?  Right now I’m storing these on the client as the conversation happens, but if the client is eventually supposed to restore a previous session, I think this history is needed.

packages/synapse-react-client/src/components/SynapseChat/SynapseChat.tsx
jay-hodgson: Good call

jay-hodgson: Storybook pops out to an iframe link, but I think this will work in the apps.

packages/synapse-react-client/src/components/SynapseChat/SynapseChat.tsx
jay-hodgson: That works for me!

packages/synapse-react-client/src/components/SynapseChat/SynapseChat.tsx
```diff
@@ -1,41 +1,60 @@
...
+    // when we have both a message and response, add a new interaction and reset
+    if (currentMessage && currentResponse) {
+      setInteractions([
+        ...interactions,
+        { userMessage: currentMessage, chatResponseText: currentResponse },
+      ])
+      setCurrentResponse('')
+      setCurrentMessage('')
+    }
+  }, [currentMessage, currentResponse])
```
nickgros, line 57: Why not just do this in the `onSuccess` callback passed to `useGetAgentChatWithAsyncStatus`? And maybe just disable the TextField while the mutation is pending?

packages/synapse-react-client/src/synapse-queries/chat/useChat.ts
```diff
@@ -51,32 +49,33 @@ export function useCreateAgentSession(
...
-  TData = AsynchronousJobStatus<AgentChatRequest, AgentChatResponse>,
->(
-  request: AgentChatRequest,
-  options?: Partial<
-    UseQueryOptions<
-      AsynchronousJobStatus<AgentChatRequest, AgentChatResponse>,
-      SynapseClientError,
-      TData
-    >
+export function useGetAgentChatWithAsyncStatus(
```
nickgros, line 52: Super nitpicky, consider renaming to `useSendChatMessageToAgent` to make the purpose of this mutation more obvious.

packages/synapse-react-client/src/components/SynapseChat/SynapseChat.tsx
```diff
@@ -1,70 +1,122 @@
...
-  const { mutate: createAgentSession } = useCreateAgentSession({
-    onSuccess: newAgentSession => {
-      setAgentSessionId(newAgentSession.sessionId)
+  const { data: agentSession, mutate: createAgentSession } =
+    useCreateAgentSession()
+  const theme = useTheme()
+  // When both the current message and current response are available, add a new ChatInteraction to the array
+  const [interactions, setInteractions] = useState<ChatInteraction[]>([])
+  const [currentInteraction, setCurrentInteraction] =
+    useState<ChatInteraction>()
```
nickgros, line 41: I wonder if it might be cleaner to set it up like:

1. Display the historical set of chat interactions & responses
2. When the user submits an interaction, store it in this state variable (rename to `pendingInteraction`) and send the message to the agent.
3. While we have a `pendingInteraction`, show the interaction + a loader for the response
4. Once the response comes back, clear out the `pendingInteraction` state and append the interaction + response to the cached history object (as a sort of optimistic update)

A state variable for `pendingInteraction` might not even be necessary if that state is also tracked and provided by the `useSendChatMessageToAgent` return object.

Thoughts?

packages/synapse-react-client/src/components/SynapseChat/SynapseChat.tsx
```diff
@@ -1,70 +1,122 @@
...
+      setInitialMessageProcessed(true)
+    }
+  }, [agentSession, initialMessage, initialMessageProcessed])
+
+  useEffect(() => {
+    // on mount, read a page worth of previous messages
+    if (agentSession && getSessionHistory) {
+      getSessionHistory(agentSession.sessionId)
+    }
+  }, [agentSession, getSessionHistory])
```
nickgros, line 96: would expect this to use `useQuery`

If you decide to try the optimistic update approach I suggested earlier, you may need to give it infinite `staleTime`

packages/synapse-react-client/src/components/SynapseChat/SynapseChat.tsx
```diff
@@ -73,56 +125,108 @@ export const SynapseChat: React.FunctionComponent<SynapseChatProps> = ({
...
+              ),
+            }}
           />
-        </FormControl>
+          <Typography
+            variant="smallText1"
+            sx={{ pt: '8px', textAlign: 'center' }}
+          >
+            {chatbotName} can make mistakes.
+          </Typography>
```
nickgros, line 228: I wish I could add this disclaimer to everything I do :wink:

packages/synapse-react-client/src/components/SynapseChat/SynapseChatInteraction.tsx
```diff
@@ -1,45 +1,74 @@
...
+              mt: '10px',
+              height: '31px',
+              borderRadius: '50%',
+              borderStyle: 'solid',
+              borderWidth: '1px',
+              borderColor: (theme.palette.grey as ColorPartial)[300],
+            }}
+          >
+            <SmartToyTwoTone
+              sx={{ color: (theme.palette.secondary as ColorPartial)[500] }}
```
nickgros, line 69: Since these are in `sx` you can just use strings
```suggestion
              borderColor: 'grey.300',
            }}
          >
            <SmartToyTwoTone
              sx={{ color: 'secondary.main' }}
```

packages/synapse-react-client/src/synapse-queries/chat/useChat.ts
```diff
@@ -37,46 +30,60 @@ export function useCreateAgentSession(
...
+    mutationFn: (request: AgentChatRequest) => {
+      return SynapseClient.getAgentChatAsyncJobResults(
         request,
         accessToken,
         setCurrentAsyncStatus,
-      ),
+      )
+    },
+    onSuccess: (data, variables, ctx) => {
+      if (options?.onSuccess && data.responseBody) {
```
nickgros, line 64: May optimistically update the history cache (once it uses useQuery). It might get weird with pagination, but hopefully that isn't too complicated to solve.

packages/synapse-react-client/src/components/SynapseChat/SynapseChat.tsx
jay-hodgson: Definitely renaming currentInteraction to pendingInteraction.  The setup that you describe is what is implemented for steps 1-3, but it does not append the interaction to the cached history object. 
I think I want to keep the server history paginated response separate from the client-side tracked interactions (new interaction data) for now.  This keeps things organized in my head.  In the unsupported case where a session was being interacted with on two (or more) browsers, each browser chat will be isolated and consistent (instead of attempting to keep all in sync with the actual chat history).  But this use case is a terrible reason to keep the existing approach.

nickgros (approved):

packages/synapse-react-client/src/components/SynapseHomepageV2/SynapseHomepageChatSearch.tsx
jay-hodgson: https://github.com/Sage-Bionetworks/synapse-web-monorepo/pull/1190

PR #1132, opened 2024-08-15
PORTALS-2970 - Refactor + changes to QueryWrapper to improve perceived loading experience

nickgros:
- Remove Query-related data from Jotai store and use react-query. This data changed infrequently, and rerendering most of the component tree is appropriate when it does change, so it did not benefit from being in Jotai. Furthermore, the data already existed in the react-query QueryClient cache.
  - Utilize react-query features at the component/hook level such as suspending while the query data is fetching, and using loading states, placeholder data, etc. to finely control loading states for each component.
  - UseQueryOptions for rows + metadata are now stored in QueryContext for convenience.
- Decouple card/table rendering logic from pagination logic.
   - CardContainer and SynapseTable components are now more simple and should be easier to mock/stub for UI testing and stories.
  - Renamed FilterAndView to RowSetView. RowSetView now fetches a RowSet to pass to SynapseTable/CardContainer and handles pagination separately.
  - Removed PaginatedQueryContext and InfiniteQueryContext.
 - In paginated mode, prefetch a page of data when the user hovers over the button to navigate to that page
 - Minor style tweaks to improve UX when resizing columns or clicking facet chips

PR #1148, opened 2024-08-22
PORTALS-2970 - Fix issue that broke result editing in SWC, and fixed tests to catch regression

nickgros:
Follow up to / depends on #1132 

The row editor functionality in SWC depends on QueryWrapperPlotNav passing the entire QueryResultBundle in the `onQueryResultChanged` callback. Fix this regression.


packages/synapse-react-client/src/components/QueryWrapper/QueryWrapper.test.tsx
```diff
@@ -66,22 +69,19 @@ const initialQueryRequest: QueryBundleRequest = {
 }
 
 describe('QueryWrapper', () => {
+  beforeAll(() => server.listen())
```
nickgros, line 72: Update tests to use MSW. Since our MSW handler utilizes the partMask, these changes catch the regression.

jay-hodgson (approved):

PR #1150, opened 2024-08-22
PORTALS-2970 - changes from review

nickgros:
- Update UserCardList to prevent merging the bundle
- Add comments/conditions to clarify strange-looking loader JSX in RowSetView
- Update FilterAndView classname to be RowSetView
- Add "just-in-case" suspense to FullContextProvider

### DIFF
--- a/packages/synapse-react-client/src/components/CardContainer/CardContainer.tsx
+++ b/packages/synapse-react-client/src/components/CardContainer/CardContainer.tsx
@@ -114,18 +114,7 @@ function _CardContainer(props: CardContainerProps) {
     }
     const listIds = dataRows.map(el => el.values[userIdColumnIndex])
     cards = (
-      <UserCardList
-        data={{
-          // TODO: UserCardList should be refactored to not require passing the full QueryResultBundle. Until then, we must reconstruct it here.
-          ...queryMetadata,
-          queryResult: {
-            queryResults: rowSet!,
-            concreteType: 'org.sagebionetworks.repo.model.table.QueryResult',
-          },
-        }}
-        list={listIds}
-        size={MEDIUM_USER_CARD}
-      />
+      <UserCardList rowSet={rowSet} list={listIds} size={MEDIUM_USER_CARD} />
     )
   } else {
     // render the cards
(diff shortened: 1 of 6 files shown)


packages/synapse-react-client/src/components/CardContainer/CardContainer.tsx
```diff
@@ -114,18 +114,7 @@ function _CardContainer(props: CardContainerProps) {
     }
     const listIds = dataRows.map(el => el.values[userIdColumnIndex])
     cards = (
-      <UserCardList
-        data={{
-          // TODO: UserCardList should be refactored to not require passing the full QueryResultBundle. Until then, we must reconstruct it here.
```
nickgros, line 119: It was actually easy enough to refactor, so I went ahead and did it

packages/synapse-react-client/src/components/QueryWrapperPlotNav/RowSetView.tsx
```diff
@@ -53,27 +53,29 @@ export function RowSetView(props: RowSetViewProps) {
...
-              isLoadingNewPage={isLoadingNewPage}
-            />
-          )}
-          {cardConfiguration && rowSet && (
-            <CardContainer {...cardConfiguration} rowSet={rowSet} />
-          )}
-        </Suspense>
-
+        {!isLoading && rowSet && (
+          // even after the rowSet is loaded, a child component may suspend
```
nickgros, line 61: Added conditions/comments to clarify why we declared two of the same loader.

packages/synapse-react-client/src/utils/context/FullContextProvider.tsx
```diff
@@ -41,7 +41,7 @@ export function FullContextProvider(props: FullContextProviderProps) {
     <QueryClientProvider client={queryClient ?? defaultQueryClient}>
       <ThemeProvider theme={theme}>
         <SynapseContextProvider synapseContext={synapseContext}>
-          {children}
+          <Suspense fallback={null}>{children}</Suspense>
```
nickgros, line 44: Added a Suspense to the wrapper object used by SWC, just in case an unwrapped component ends up being used in SWC.

nickgros: Random thought: it would be so nice if we had static checking for Suspense. I want TypeScript to force me to wrap my component in Suspense or tell the parent component to deal with it the same way Java makes me handle checked exceptions or force the caller to deal with it!

jay-hodgson (approved):

PR #1168, opened 2024-08-30
SWC-7051

jay-hodgson:


jay-hodgson:
Style changes for an unreleased component (home page v2)

PR #1169, opened 2024-09-03
PORTALS-3215

jay-hodgson:


jay-hodgson:
May want to wrap in a dialog, or just have on a new (initially unlinked) page in the NF portal.  Still some things to work out, but this PR provides a very simple wrapper around the rjsf library.
image

packages/synapse-react-client/src/components/DynamicForm/DynamicForm.tsx
```diff
@@ -0,0 +1,124 @@
...
+        const uiSchemaData = await uiSchemaResponse.json()
+        setUiSchema(uiSchemaData)
+      } catch (error) {
+        displayToast(`Error fetching UI Schema: ${error}`, 'danger')
+      }
+    }
+
+    fetchSchema()
+    fetchUiSchema()
+  }, [schemaUrl, uiSchemaUrl])
```
nickgros, line 54: Can you use `useQuery` instead of `useEffect` + `useState`? I'm ok with inlining it rather than having custom hook(s) since these won't be reused

packages/synapse-react-client/src/components/DynamicForm/DynamicForm.tsx
```diff
@@ -0,0 +1,124 @@
...
+      } else {
+        const responseText = await response.text()
+        displayToast(`Form submission failed: ${responseText}`, 'danger')
+      }
+    } catch (error) {
+      displayToast(`Form submission failed: ${JSON.stringify(error)}`, 'danger')
+    } finally {
+      setIsSubmitting(false)
+    }
+  }
```
nickgros, line 82: Ditto but `useMutation` for this one instead of `useQuery`

packages/synapse-react-client/src/components/DynamicForm/DynamicForm.tsx
```diff
@@ -0,0 +1,124 @@
...
+
+  if (!schema || !uiSchema) {
```
nickgros, line 84: Use `isLoading` returned by `useQuery`.

Is it true that there will always be a `uiSchema`? It's optional in RJSF

nickgros (changes requested):

packages/synapse-react-client/src/components/DynamicForm/DynamicForm.tsx
jay-hodgson: Will do!

packages/synapse-react-client/src/components/DynamicForm/DynamicForm.tsx
jay-hodgson: For NF cases so far, UI schema is absolutely necessary.  If we encounter a case where it is not needed we can update

packages/synapse-react-client/src/components/DynamicForm/DynamicForm.tsx
```diff
@@ -0,0 +1,148 @@
...
+    onError: error => {
+      displayToast(`Error submitting form: ${error.message}`, 'danger')
+    },
+  })
+
+  const handleSubmit = (formData: any) => {
+    submitFormData(formData)
+  }
+
+  if (!schema || !uiSchema) {
```
nickgros, line 96: Need to explicitly check loading state. Since you're just checking the data, this will evaluate to true even if one or both queries are in an error state.
```suggestion
  if (_isSchemaLoading || _isUiSchemaLoading) {
```

packages/synapse-react-client/src/components/DynamicForm/DynamicForm.tsx
```diff
@@ -0,0 +1,148 @@
...
+    )
+  }
+
+  if (isSchemaError) {
+    displayToast(`Error fetching Schema: ${schemaError.message}`, 'danger')
+  }
+
+  if (isUiSchemaError) {
+    displayToast(`Error fetching UI Schema: ${uiSchemaError.message}`, 'danger')
+  }
```
nickgros, line 111: This will push 1-2 toasts every time the component re-renders and the query is in error. I would recommend returning an `Alert` component with the error message. If you really want to push a toast (and still try to render the form), either push the toast 1) in the query function or 2) in a `useEffect` that monitors the error

packages/synapse-react-client/src/components/DynamicForm/DynamicForm.tsx
```diff
@@ -0,0 +1,148 @@
...
+    error: schemaError,
+  } = useQuery({
+    queryKey: ['DynamicForm', 'schema', schemaUrl],
+    queryFn: async () => {
+      const response = await fetch(schemaUrl)
+      if (!response.ok) {
+        throw new Error('Error fetching schema')
+      }
+      return response.json()
+    },
```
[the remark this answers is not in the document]
jay-hodgson: Good call, will do

packages/synapse-react-client/src/components/DynamicForm/DynamicForm.tsx
jay-hodgson: These are the loading flags that did not seem to be set properly.  Schema variables were undefined when these loading flags are false

packages/synapse-react-client/src/components/DynamicForm/DynamicForm.tsx
jay-hodgson: Good call, will do

apps/portals/nf/src/config/routesConfig.ts
```diff
@@ -186,6 +186,24 @@ const routes: GenericRoute[] = [
...
+            'https://raw.githubusercontent.com/nf-osi/nf-research-tools-schema/cconrad8-patch-1/NF-Tools-Schemas/SubmitObservationSchema.json',
+          uiSchemaUrl:
+            'https://raw.githubusercontent.com/nf-osi/nf-research-tools-schema/cconrad8-patch-1/NF-Tools-Schemas/SubmitObservationUiSchema.json',
+          postUrl: 'https://submit-form.com/KwZ46H4T',
+        },
+        isOutsideContainer: false,
+      },
+    ],
+  },
   {
```
jay-hodgson, line 207: Integrate unlinked page into the NF portal:
image

packages/synapse-react-client/src/components/DynamicForm/DynamicForm.tsx
```diff
@@ -119,37 +119,39 @@ function DynamicForm(props: DynamicFormProps) {
...
-      uiSchema={uiSchema}
-      formData={formData}
-      validator={validator}
-      onChange={({ formData }) => setFormData(formData)}
-      onSubmit={({ formData }) => handleSubmit(formData)}
-      disabled={isPending}
-    >
-      <Box sx={{ display: 'flex', gap: '10px', justifyContent: 'flex-end' }}>
-        {onCancel && (
+    <Box sx={{ p: '30px 10px' }}>
```
jay-hodgson, line 122: Wrapped the form in a Box for some additional padding

nickgros (approved):

PR #1180, opened 2024-09-10
PORTALS-3217: update based on current table schema and confluence doc…

jay-hodgson:
… for ELITE Portal Study Schema


jay-hodgson:
ELITE portal e2e tests passed, merging

PR #1195, opened 2024-09-13
PORTALS-3243, PORTALS-3244

jay-hodgson:
Change the NFSurveyToast into a more generalized component, and add one for ADKP and GENIE


jay-hodgson:
Portal component refactor and config updates

PR #1196, opened 2024-09-13
PORTALS-3214 - Fix bug where chat history was reversed in place on every render

nickgros:


packages/synapse-react-client/src/components/SynapseChat/SynapseChat.tsx
```diff
@@ -157,21 +171,14 @@ export const SynapseChat: React.FunctionComponent<SynapseChatProps> = ({
           }}
         >
           {sessionHistory &&
-            sessionHistory.pages.map(sessionHistoryResponse => {
-              {
-                /* Note: session history is ordered from most recent to least recent in each page, so reverse the order when restoring the chat interface */
-              }
-              return sessionHistoryResponse.page
-                .reverse()
```
nickgros, line 165: Reverse is in-place, so the history order was reversed on each render

packages/synapse-react-client/src/components/SynapseChat/SynapseChat.tsx
```diff
@@ -63,6 +68,15 @@ export const SynapseChat: React.FunctionComponent<SynapseChatProps> = ({
...
     )
 
+  const sessionHistory = useMemo(
+    () =>
+      sessionHistoryInfiniteData?.pages
+        .flatMap(page => page.page)
+        /* Note: session history is ordered from most recent to least recent in each page, so reverse the order when restoring the chat interface */
+        .reverse() ?? [],
+    [sessionHistoryInfiniteData],
+  )
```
nickgros, line 78: Could have just replaced the `reverse` with the new [Array.prototype.toReversed](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/toReversed) function, but to support multiple pages you would have needed to add this useMemo, so might as well just call `reverse` here.

packages/synapse-react-client/src/components/SynapseChat/SynapseChat.tsx
jay-hodgson: Oops!

jay-hodgson (approved):

PR #1205, opened 2024-09-16
SWC-6923 - Handle suspension case

nickgros:


packages/synapse-react-client/src/components/dataaccess/RejectProfileValidationRequestModal.tsx
```diff
@@ -39,18 +40,35 @@ export function RejectProfileValidationRequestModal(
...
+      case VerificationStateEnum.APPROVED:
+        newState = VerificationStateEnum.SUSPENDED
+        break
+      case VerificationStateEnum.REJECTED:
+      case VerificationStateEnum.SUSPENDED:
+        console.error(
+          'Cannot reject a request where the current state is already rejected or suspended',
+        )
+        onClose()
+        return
```
nickgros, line 65: This should never happen

jay-hodgson (approved):

PR #1207, opened 2024-09-16
SWC-7062

jay-hodgson:
Refactor, and add new UI that uses the same Source Configs hook


jay-hodgson:
In the new Synapse Homepage:
chrome_Qu9Sih8jCM

Behind a feature flag in all portals!
chrome_9639RGgd84

packages/synapse-react-client/src/components/error/ErrorPage.tsx
```diff
@@ -148,7 +148,7 @@ const ErrorPage: React.FunctionComponent<ErrorPageProps> = props => {
     () => setSendMessageToAdminDialogOpen(false),
     [],
   )
-
+  console.log('Entity ID', entityId)
```
nickgros, line 151: remove

nickgros, line 151:
```suggestion
```

packages/synapse-types/src/FeatureFlags.ts
```diff
@@ -15,4 +15,7 @@ export enum FeatureFlagEnum {
 
   // If enabled, show the chatbot entrypoint from the new Synapse Homepage
   HOMEPAGE_CHATBOT = 'HOMEPAGE_CHATBOT',
+
+  // If enabled, show the Portals dropdown on the portal site (to discover other portals)
+  PORTALS_DROPDOWN = 'PORTALS_DROPDOWN',
```
nickgros, line 20: Please also make sure this is added to the stack builder in https://github.com/Sage-Bionetworks/Synapse-Stack-Builder/blob/develop/src/main/resources/templates/repo/appconfig-config.json

nickgros (approved):

packages/synapse-types/src/FeatureFlags.ts
jay-hodgson: https://github.com/Sage-Bionetworks/Synapse-Stack-Builder/pull/662

PR #1209, opened 2024-09-17
PORTALS-3249

jay-hodgson:


packages/synapse-react-client/src/utils/AppUtils/session/ApplicationSessionManager.tsx
```diff
@@ -148,6 +148,11 @@ export function ApplicationSessionManager(
     } else {
       refreshSession()
     }
+    // PORTALS-3249: Set up an interval to call refreshSession every 60 seconds (60000 milliseconds)
+    const intervalId = setInterval(refreshSession, 60000)
```
nickgros, line 152: Lint error:
```
    152:36  error    Promise returned in function argument where a void return was expected                                        @typescript-eslint/no-misused-promises
```

This might fix it?
```suggestion
    const intervalId = setInterval(() => {
      refreshSession()
    }, 60000)
```

nickgros (approved):

PR #1210, opened 2024-09-18
Upgrade Nx, set up Nx cache sharing for CI job steps

nickgros:
Save the Nx build cache into a GitHub actions cache to reduce the total number of builds. 

The motivation is that when building the new `openapi-generator`-based Synapse client, I am frequently seeing 403s when the openapi-generator NPM package attempts to pull the openapi-generator JAR from Maven central. An issue to track this is here: https://github.com/OpenAPITools/openapi-generator-cli/issues/802 , but we can reduce the number of times this can fail by storing the Nx cache in a GitHub actions cache.

Note that Nx warns that this may leave us susceptible to cache poisoning: https://nx.dev/troubleshooting/unknown-local-cache , but I think if we only use this cache for checks, and not for deployment, this is low risk.

This should also reduce the total number of minutes of GitHub actions we use, since runners aren't recomputing build tasks. The total CI wall clock time will not be faster, though, since the operations waiting on `build` steps will still have to wait for the build step to finish.

### DIFF
--- a/.github/actions/pnpm-setup-action/action.yml
+++ b/.github/actions/pnpm-setup-action/action.yml
@@ -1,5 +1,10 @@
 name: 'Set up pnpm and configure environment'
 description: 'Installs Node, pnpm, fetches cache (if exists), and runs pnpm install'
+inputs:
+  STORE_NX_CACHE:
+    description: 'Whether to store/load the Nx cache in the action cache'
+    required: false
+    default: 'false'
 runs:
   using: 'composite'
   steps:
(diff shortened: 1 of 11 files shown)


.github/actions/pnpm-setup-action/action.yml
```diff
@@ -27,14 +32,19 @@ runs:
...
+      # See: https://nx.dev/recipes/troubleshooting/unknown-local-cache#you-share-cache-with-another-machine-using-a-network-drive
+      if: ${{ inputs.STORE_NX_CACHE == 'true' }}
+      uses: actions/cache@v3
+      with:
+        path: .nx/cache
+        key: ${{ runner.os }}-nx-${{ github.ref }}-${{ env.HEAD_COMMIT }}
+        restore-keys: |
+          ${{ runner.os }}-nx-${{ github.ref }}-${{ env.HEAD_COMMIT }}
+          ${{ runner.os }}-nx-${{ github.ref }}
+          ${{ runner.os }}-nx
```
nickgros, line 47:
```suggestion
        restore-keys: |
          ${{ runner.os }}-nx-${{ github.ref }}-${{ env.HEAD_COMMIT }}
          ${{ runner.os }}-nx-${{ github.ref }}-
          ${{ runner.os }}-nx-
```

.github/actions/pnpm-setup-action/action.yml
```diff
@@ -27,14 +32,19 @@ runs:
         restore-keys: |
           ${{ runner.os }}-pnpm-store-
     #    TODO: Storing the Nx cache is failing builds because the cache is not meant to be shared between machines.
```
nickgros, line 34:
```suggestion
```

.gitignore
```diff
@@ -48,6 +48,7 @@ build-storybook.log
 
 # Nx cache
 .nx/cache
+.nx/workspace-data
```
nickgros, line 51: Upgraded Nx now stores some data in this folder, should be gitignored

apps/SageAccountWeb/package.json
```diff
@@ -73,7 +73,6 @@
     "start": "vite",
     "build": "vite build",
     "test": "vitest",
-    "test:ci": "vitest run --coverage",
```
nickgros, line 76: Remove test:ci jobs, and just pass `--coverage` to each script

packages/vite-config/src/vitest-config.ts
```diff
@@ -7,6 +7,7 @@ export default mergeConfig(config, {
     include: ['@vitest/utils', 'vitest/browser'],
   },
   test: {
+    watch: false,
```
nickgros, line 10: Default behavior for Vitest is to watch, which would makes it appear like it hangs forever if you just run `pnpm test` in the root directory. For all of our Vitest projects, make the default behavior not to watch. You can still run `vitest watch` or `vitest dev` if you want to use it locally.

jay-hodgson (approved):

PR #1213, opened 2024-09-18
SWC-6662 - Add OpenAPI-generated TypeScript client for Synapse

nickgros:
- Add synapse-client package, which is created using the Synapse OpenAPI specification and the openapi-generator tool
- Refactor to move `fetchWithExponentialTimeout` and `SynapseClientError` to the new package


apps/SageAccountWeb/src/components/TwoFactorAuth/ResetTwoFactorAuth.tsx
```diff
@@ -9,10 +9,8 @@ import {
   useApplicationSessionContext,
   useQuerySearchParam,
 } from 'synapse-react-client'
-import {
-  TwoFactorAuthErrorResponse,
-  TwoFactorAuthResetToken,
-} from '@sage-bionetworks/synapse-types'
+import { TwoFactorAuthResetToken } from '@sage-bionetworks/synapse-types'
+import { TwoFactorAuthErrorResponse } from 'synapse-client'
```
nickgros, line 13: All error types have been updated to use the generated package

packages/synapse-client/.gitignore
```diff
@@ -0,0 +1 @@
+src/generated/
```
nickgros, line 1: Don't commit the generated TypeScript files to git

packages/synapse-client/out/modelGeneric.mustache
```diff
@@ -0,0 +1,168 @@
+import { mapValues } from '../runtime{{importFileExtension}}';
```
nickgros, line 1: This template file defines the generated code for a model defined in the OpenAPI spec. Most of this code is copied/generated from the original template source.

The lone change is denoted by `START CUSTOM TEMPLATING FOR synapse-client`

packages/synapse-client/out/modelGeneric.mustache
```diff
@@ -0,0 +1,168 @@
...
+    {{#required}}
+    if (!('{{name}}' in value) || value['{{name}}'] === undefined) return false;
+    {{! START CUSTOM TEMPLATING FOR synapse-client }}
+    {{! if the property is a required enum, check that the value is a valid enum value }}
+    {{#isEnum}}
+      if (!Object.values({{classname}}{{enumName}}).includes(value["{{name}}"] as {{classname}}{{enumName}})) {
+        return false
+      }
+    {{/isEnum}}
+    {{! END CUSTOM TEMPLATING FOR synapse-client }}
```
nickgros, line 34: In the `instanceOf` method, if a model's property is a required enum, but the property's value does not match a known enum value, return `false`. This is especially useful because it will use our `concreteType` property, which is now always defined as an enum following PLFM-8257

Here is an example of generated code that includes this logic. See that the generated code checks that the `concreteType` matches one of the allowed enumeration values, which in this case is only `'org.sagebionetworks.repo.model.FileEntity'`.

```typescript
/**
 * Check if a given object implements the FileEntity interface.
 */
export function instanceOfFileEntity(value: object): value is FileEntity {
  if (!('concreteType' in value) || value['concreteType'] === undefined)
    return false
  if (
    !Object.values(FileEntityConcreteTypeEnum).includes(
      value['concreteType'] as FileEntityConcreteTypeEnum,
    )
  ) {
    return false
  }
  return true
}
```

packages/synapse-client/src/SynapseClient.ts
```diff
@@ -0,0 +1,182 @@
...
+import { fetchWithExponentialTimeout } from './util/fetchWithExponentialTimeout'
+
+const DEFAULT_CONFIG = new Configuration({
+  fetchApi: fetchWithExponentialTimeout,
+})
+
+/**
+ * Creates one class that encapsulates all sets of Synapse API services.
+ */
+export class SynapseClient extends BaseAPI {
```
nickgros, line 57: The generator creates one API client class per Java controller. I created this convenience wrapper over every service.

packages/synapse-client/src/generatedSpecTest.test.ts
```diff
@@ -0,0 +1,70 @@
+import { describe, expect, expectTypeOf, test } from 'vitest'
+import {
+  EntityFromJSONTyped,
+  FileEntity,
+  instanceOfFileEntity,
+} from './generated'
+
+describe('Tests for the generated specification', () => {
```
nickgros, line 8: Tests to verify that our template changes were successful

packages/synapse-client/src/codegen/ShortenCanonicalName.ts
```diff
@@ -0,0 +1,123 @@
...
+ * Output:
+ * ```
+ *  {
+ *    'org.sagebionetworks.foo.Bar': 'FooBar',
+ *    'org.sagebionetworks.foo.Baz': 'Baz',
+ *    'org.sagebionetworks.bar.Bar': 'BarBar',
+ *  }
+ * ```
+ */
+export function shortenCanonicalNames(
```
nickgros, line 47: This function is used to generate the names of the models so we don't have to use model names like `OrgSagebionetworksRepoModelFileEntity`

packages/synapse-client/src/util/fetchWithExponentialTimeout.ts
```diff
@@ -0,0 +1,95 @@
+import { BaseError } from '../generated/models/BaseError'
```
nickgros, line 1: Moved from synapse-react-client

packages/synapse-client/package.json
```diff
@@ -0,0 +1,53 @@
+{
+  "name": "synapse-client",
```
nickgros, line 2: `synapse-client` or `@sage-bionetworks/synapse-client`?

packages/synapse-client/src/SynapseClient.ts
```diff
@@ -0,0 +1,182 @@
...
+  public tableServicesClient: TableServicesApi
+  public teamServicesClient: TeamServicesApi
+  public userGroupServicesClient: UserGroupServicesApi
+  public userProfileServicesClient: UserProfileServicesApi
+  public verificationServicesClient: VerificationServicesApi
+  public webhookServicesClient: WebhookServicesApi
+  public wikiPageServices2Client: WikiPageServices2Api
+  public wikiPageServicesClient: WikiPageServicesApi
+
+  constructor(protected configuration = DEFAULT_CONFIG) {
```
nickgros, line 102: It was a little weird to me to have the `protected` keyword in the constructor args, but this is how the parent class does it

packages/synapse-client/src/spec/openapispecification.json
nickgros, line 1: This is just the latest Synapse OpenAPI spec. As explained in the README, we commit it for stability because it's a live document.

packages/synapse-client/src/util/SynapseClientError.ts
```diff
@@ -1,28 +1,31 @@
...
+  | BaseError
+  | DrsErrorResponse
+  | OAuthErrorResponse
+  | TwoFactorAuthErrorResponse
 
+/**
+ * Error message returned by the Synapse backend joined with the
+ * HTTP status code.
+ */
 export class SynapseClientError extends Error {
```
nickgros, line 18: Moved this class into `synapse-client` because our `fetch` implementation (`fetchWithExponentialTimeout`) may create an instance of this class.

packages/synapse-client/package.json
```diff
@@ -0,0 +1,74 @@
...
+    "vite-config": "workspace:*",
+    "vitest": "^1.6.0"
+  },
+  "scripts": {
+    "format-spec": "prettier --write src/spec/openapispecification.json",
+    "get-spec:staging": "npx tsx src/codegen/downloadSpec.ts --stack staging && pnpm format-spec",
+    "get-spec:production": "npx tsx src/codegen/downloadSpec.ts --stack production && pnpm format-spec",
+    "get-spec": "pnpm get-spec:production",
+    "configure": "openapi-generator-cli version-manager set 7.8.0",
+    "generate": "MODEL_NAME_MAPPINGS=$(npx tsx src/codegen/generateModelNameMappings.ts); pnpm configure && rimraf src/generated && openapi-generator-cli generate -i src/spec/openapispecification.json -c config.json -o src/generated -t out/ --model-name-mappings $MODEL_NAME_MAPPINGS",
```
nickgros, line 39: Script to generate the TS code. First run the `generateModelNameMappings` script, which will scan the OpenAPI doc to modify the model names (e.g. `org.sagebionetworks.repo.model.FileEntity` -> `FileEntity`). Then run the openapi-generator, passing the generated mappings as an argument

eslint.config.mjs
```diff
@@ -113,4 +113,7 @@ export default [
       '@typescript-eslint/no-floating-promises': 'error',
     },
   },
+  {
+    ignores: ['synapse-client/src/generated/'],
+  },
```
nickgros, line 118: Don't lint the generated files

packages/synapse-client/src/codegen/downloadSpec.ts
```diff
@@ -0,0 +1,35 @@
...
+switch (stack) {
+  case 'staging':
+    restDocsUrl = STAGING_SPEC_URL
+    break
+  case 'production':
+  default:
+    restDocsUrl = PROD_SPEC_URL
+}
+
+request(restDocsUrl).then(({ body }) => {
```
nickgros, line 31: `request` is first-party fetch for Nodejs

packages/synapse-client/tsconfig.build.json
```diff
@@ -0,0 +1,9 @@
+{
+  "extends": "./tsconfig.json",
+  "compilerOptions": {
+    "noEmit": false,
+    "types": ["vitest/globals"]
+  },
+  "include": ["src/*.ts", "src/generated/**/*.ts", "src/util/**/*.ts"],
```
nickgros, line 7: This will be built with ts, but don't emit the `src/codegen` folder, which are internal utilities that other packages should not use

apps/SageAccountWeb/src/components/TwoFactorAuth/ResetTwoFactorAuth.tsx
```diff
@@ -9,10 +9,8 @@ import {
   useApplicationSessionContext,
   useQuerySearchParam,
 } from 'synapse-react-client'
-import {
-  TwoFactorAuthErrorResponse,
-  TwoFactorAuthResetToken,
-} from '@sage-bionetworks/synapse-types'
+import { TwoFactorAuthResetToken } from '@sage-bionetworks/synapse-types'
+import { TwoFactorAuthErrorResponse } from 'synapse-client/generated/models/TwoFactorAuthErrorResponse'
```
nickgros, line 13: All of the error types now come from the new `synapse-client` package, so imports had to be updated everywhere. It's 99% an interface change, so it should be low risk

packages/synapse-client/out/modelGeneric.mustache
jay-hodgson: Oh I see!  Thanks for the example output, very clever

packages/synapse-client/package.json
jay-hodgson: Seems like we have adopted @sage-bionetworks/name based on packages like @sage-bionetworks/synapse-types.   What do you think?

packages/synapse-client/src/SynapseClient.ts
jay-hodgson: Is this something that we will need to extend as additional controllers are added?

packages/synapse-client/src/codegen/ShortenCanonicalName.ts
```diff
@@ -0,0 +1,123 @@
...
+  /* Find any duplicate names and iterate until they are removed
+  e.g.
+  {
+    'org.sagebionetworks.repo.model.one.Foo': 'foo',
+    'org.sagebionetworks.repo.model.one.Bar': 'one.bar',
+    'org.sagebionetworks.repo.model.two.Bar': 'two.bar',
+    'org.sagebionetworks.repo.model.two.Baz': 'baz',
+  }
+  */
+  let duplicateNames = getDuplicateNames(Object.values(caseInsensitiveMapping))
```
jay-hodgson, line 76: I was just going to ask about this case :)

jay-hodgson (approved):

packages/synapse-client/src/SynapseClient.ts
nickgros: Yes.

packages/synapse-client/package.json
nickgros: Works for me! I'll make the change in a follow-up PR, since this isn't going to be published yet.

PR #1214, opened 2024-09-19
Fix duplicate plugins in vite config, all projects should use ConfigBuilder

nickgros:


packages/vite-config/src/baseConfig.ts
```diff
@@ -1,34 +1,23 @@
...
-  plugins: [
-    react(),
-    svgr({
-      svgrOptions: {
-        plugins: ['@svgr/plugin-jsx'],
-        ref: true,
-        exportType: 'named',
-      },
-      // Explicitly exclude SVG imports that end in a query (such as ?url) - Vite can already handle these
-      include: /^.*\.svg$/,
```
nickgros, line 28: React plugins were erroneously included in the base config, so they were duplicated in the final app configs, causing errors in development mode

packages/vite-config/src/index.ts
```diff
@@ -1,15 +1,4 @@
...
-export {
-  viteConfig,
-  vitestConfig,
-  portalsViteConfig,
-  viteLibraryConfig,
-  ConfigBuilder,
-}
-
-export default viteConfig
+export { portalsViteConfig, ConfigBuilder }
```
nickgros, line 4: Only export ConfigBuilder and the shared portals config

jay-hodgson (approved):

PR #1215, opened 2024-09-19
PORTALS-3241

jay-hodgson:


packages/synapse-react-client/src/components/SharePageLinkButton/SharePageLinkButton.tsx
```diff
@@ -0,0 +1,70 @@
...
+
+  return (
+    <Button
+      variant="contained"
+      onClick={() => {
+        createShortUrl()
+      }}
+      {...buttonProps}
+    >
+      <IconSvg icon="contentCopy" wrap={false} sx={{ mr: '5px' }} />
```
nickgros, line 64: move this into the `startIcon` ButtonProp (and remove the margin)

nickgros (approved):
Super clean, just called out one small change

packages/synapse-react-client/src/components/SharePageLinkButton/SharePageLinkButton.tsx
jay-hodgson: Right, will do!

PR #1219, opened 2024-09-20
SWC-6662

nickgros:
- Add generated SynapseClient to context
- Use SynapseClient for OAuth client operations as proof of concept
- Rename package to `@sage-bionetworks/synapse-client` and publish to NPM


apps/SageAccountWeb/package.json
```diff
@@ -23,7 +23,7 @@
     "react-easy-crop": "^5.0.7",
     "react-router-dom": "^5.3.4",
     "sass": "^1.71.1",
-    "synapse-client": "workspace:*",
+    "synapse-client": "workspace:@sage-bionetworks/synapse-client@*",
```
nickgros, line 26: With this alias we can import using the shorter `'synapse-client'` while still publishing under the @sage-bionetworks namespace

packages/synapse-react-client/src/components/OAuthClientManagement/CreateOAuthClient.tsx
nickgros, line 1: As a proof of concept, use the new client in this component

packages/synapse-react-client/src/components/OAuthClientManagement/CreateOAuthClient.tsx
```diff
@@ -105,31 +102,37 @@ export const CreateOAuthModal: React.FunctionComponent<
...
+        | 'JSON'
+        | OIDCSigningAlgorithm,
     )
     setTosUri(client?.tos_uri ?? '')
   }, [isShowingModal, client])
 
   useDebouncedEffect(
     () => {
-      if (accessToken && oAuthClient.client_id) {
+      if (oAuthClient.client_id) {
```
nickgros, line 123: You should never be able to see this without an access token, so I'm not worried about missing the check

packages/synapse-react-client/src/components/OAuthClientManagement/CreateOAuthClient.integration.test.tsx
```diff
@@ -44,10 +43,14 @@ const defaultProps: CreateOAuthModalProps = {
...
-const createOAuthClientSpy = jest.mocked(SynapseClient.createOAuthClient)
-const updateOAuthClientSpy = jest.mocked(SynapseClient.updateOAuthClient)
+const createOAuthClientSpy = jest.spyOn(
+  MOCK_CONTEXT_VALUE.synapseClient.openIDConnectServicesClient,
+  'postAuthV1Oauth2Client',
+)
+const updateOAuthClientSpy = jest.spyOn(
+  MOCK_CONTEXT_VALUE.synapseClient.openIDConnectServicesClient,
+  'putAuthV1Oauth2ClientId',
+)
```
nickgros, line 53: In test, I want to verify that we sent the new data to the server, without testing the implementation of the mock service. We can still `spyOn` the method, but we need to get the specific instance of `synapseClient` used to make the call.

packages/synapse-react-client/test/components/download_list/FolderDownloadConfirmation.test.tsx
```diff
@@ -11,7 +11,7 @@ import {
 import * as DownloadConfirmationUIModule from '../../../src/components/download_list/DownloadConfirmationUI'
 import { DownloadConfirmationUIProps } from '../../../src/components/download_list/DownloadConfirmationUI'
 import { FolderDownloadConfirmation } from '../../../src/components/download_list'
-import { SynapseClientError } from 'synapse-client/uti'
```
nickgros, line 14: This is just used as a type in a non-type-checked test file, which explains why this passed in CI

.github/workflows/npmpublish.yml
```diff
@@ -16,6 +16,7 @@ jobs:
             markdown-it-synapse,
             markdown-it-synapse-table,
             '@sage-bionetworks/markdown-it-container',
+            '@sage-bionetworks/synapse-client',
```
nickgros, line 19: publish the generated client code to npm

apps/SageAccountWeb/package.json
```diff
@@ -23,12 +25,10 @@
     "react-easy-crop": "^5.0.7",
     "react-router-dom": "^5.3.4",
     "sass": "^1.71.1",
-    "synapse-client": "workspace:*",
     "synapse-react-client": "workspace:*",
     "universal-cookie": "^4.0.4"
   },
   "devDependencies": {
-    "@sage-bionetworks/synapse-types": "workspace:*",
```
nickgros, line 31: synapse-types has some runtime code, so it may not truly be a 'devDependency'. I moved it to `dependencies`, but this is just for clarity and won't affect anything

packages/synapse-react-client/src/components/OAuthClientManagement/CreateOAuthClient.tsx
jay-hodgson: Yay!

jay-hodgson (approved):

PR #1221, opened 2024-09-20
Trigger release of synapse-client v0.0.1

nickgros:


packages/synapse-client/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@sage-bionetworks/synapse-client",
-  "version": "0.0.1",
+  "version": "0.0.0",
```
nickgros, line 3:
```suggestion
  "version": "0.0.1",
```

PR #1233, opened 2024-09-24
PORTALS-2923 - A few fixes for the OpenAPI client

nickgros:
- Use ConfigurationParameters instead of Configuration to simplify merging defaults
- Update fetchApi implementation to not consume the response
- Add middleware to rethrow SynapseClientErrors, which is how the handwritten SynapseClient behaves.
- In synapse-react-client, set the BaseURL based on the result of `getEndpoint`


packages/synapse-client/src/SynapseClient.test.ts
```diff
@@ -0,0 +1,35 @@
+import { SynapseClient } from './SynapseClient'
+import { fetchResponseWithExponentialTimeout } from './util/fetchWithExponentialTimeout'
+import { SynapseClientError } from './util/SynapseClientError'
+
+vi.mock('./util/fetchWithExponentialTimeout', () => {
+  return { fetchResponseWithExponentialTimeout: vi.fn() }
+})
+const mockFetchResponseWithExponentialTimeout = vi.mocked(
+  fetchResponseWithExponentialTimeout,
+)
```
nickgros, line 10: Mock the `fetchApi` implementation passed to the client

packages/synapse-client/src/SynapseClient.ts
```diff
@@ -41,17 +41,56 @@ import { VerificationServicesApi } from './generated/apis/VerificationServicesAp
...
+} from './generated/runtime'
+import { fetchResponseWithExponentialTimeout } from './util/fetchWithExponentialTimeout'
+import { SynapseClientError } from './util/SynapseClientError'
 
-const DEFAULT_CONFIG = new Configuration({
-  fetchApi: fetchWithExponentialTimeout,
-})
+const DEFAULT_CONFIG_PARAMETERS: ConfigurationParameters = {
+  fetchApi: fetchResponseWithExponentialTimeout,
+  middleware: [
```
nickgros, line 55: The generated client throws its own error class, which breaks all of our existing error handling code. This middleware will inspect the error and throw (or rethrow) a SynapseClientError when appropriate.

packages/synapse-client/src/SynapseClient.ts
```diff
@@ -41,17 +41,56 @@ import { VerificationServicesApi } from './generated/apis/VerificationServicesAp
...
+      },
+    },
+  ],
+}
 
 /**
  * Creates one class that encapsulates all sets of Synapse API services.
  */
-export class SynapseClient extends BaseAPI {
+export class SynapseClient {
```
nickgros, line 93: No need to extend BaseAPI

packages/synapse-client/src/SynapseClient.ts
```diff
@@ -96,15 +135,13 @@ export class SynapseClient extends BaseAPI {
   public wikiPageServices2Client: WikiPageServices2Api
   public wikiPageServicesClient: WikiPageServicesApi
 
-  constructor(protected configuration = DEFAULT_CONFIG) {
+  constructor(configurationParameters?: ConfigurationParameters) {
```
nickgros, line 138: Update the constructor to take `ConfigurationParameters`, which is easy to merge with defaults, unlike a `Configuration` class instance

packages/synapse-client/src/util/fetchWithExponentialTimeout.ts
```diff
@@ -58,6 +61,26 @@ export const fetchWithExponentialTimeout = async <TResponse>(
...
 
+  return response
+}
+
+/**
+ * Fetches data, retrying if the HTTP status code indicates that it could be retried. Contains custom logic for
+ * handling errors returned by the Synapse backend.
+ * @throws SynapseClientError
+ */
+export const fetchWithExponentialTimeout = async <TResponse>(
```
nickgros, line 72: Refactor to use new method, behavior is the same to prevent breaking changes for other places where this function is used.

packages/synapse-client/src/util/fetchWithExponentialTimeout.ts
```diff
@@ -25,16 +25,19 @@ export function delay(t: number) {
...
- * Fetches data, retrying if the HTTP status code indicates that it could be retried. Contains custom logic for
- * handling errors returned by the Synapse backend.
+ * Fetches data, retrying if the HTTP status code indicates that it could be retried.
+ * To use it in our generated client, this function must be SIDE-EFFECT FREE, so it does not
+ * consume the response body.
+ *
  * @throws SynapseClientError
  */
-export const fetchWithExponentialTimeout = async <TResponse>(
+export const fetchResponseWithExponentialTimeout = async (
```
nickgros, line 36: Refactor to create a new method that does not consume the response stream, which caused problems when the generated code also tried to consume it.

packages/synapse-react-client/src/mocks/MockSynapseContext.tsx
```diff
@@ -14,11 +14,9 @@ export const MOCK_CONTEXT_VALUE: SynapseContextType = {
   downloadCartPageUrl: '/DownloadCart',
   withErrorBoundary: false,
   keyFactory: new KeyFactory(MOCK_ACCESS_TOKEN),
-  synapseClient: new SynapseClient(
-    new Configuration({
-      apiKey: `Bearer ${MOCK_ACCESS_TOKEN}`,
-    }),
-  ),
+  synapseClient: new SynapseClient({
+    accessToken: MOCK_ACCESS_TOKEN,
```
nickgros, line 18: The generated code will automatically add `'Bearer '` if we use `accessToken` instead of `apiKey`

packages/synapse-react-client/src/components/SharePageLinkButton/SharePageLinkButton.test.tsx
```diff
@@ -34,8 +34,10 @@ describe('SharePageLinkButton', () => {
     await userEvent.click(
       screen.getByRole('button', { name: 'Share Page Link' }),
     )
-    expect(navigator.clipboard.writeText).toHaveBeenCalledWith(
-      MOCK_SHORT_IO_URL,
-    )
+    await waitFor(() => {
```
nickgros, line 37: test is flaky

packages/synapse-client/src/SynapseClient.test.ts
```diff
@@ -0,0 +1,102 @@
...
+      .fn()
+      .mockImplementation(original.fetchResponseWithExponentialTimeout),
+  }
+})
+
+const fetchResponseWithExponentialTimeoutSpy = vi.mocked(
+  fetchResponseWithExponentialTimeout,
+)
+
+describe('SynapseClient', () => {
```
nickgros, line 25: add a few tests with mock service worker

packages/synapse-client/src/util/fetchWithExponentialTimeout.ts
```diff
@@ -25,24 +25,20 @@ export function delay(t: number) {
 const RETRY_STATUS_CODES = [0, 429, 502, 503, 504]
 const MAX_RETRY_STATUS_CODES = [502, 503]
 const MAX_RETRY = 3
+
 /**
- * Fetches data, retrying if the HTTP status code indicates that it could be retried. Contains custom logic for
- * handling errors returned by the Synapse backend.
+ * Fetches data, retrying if the HTTP status code indicates that it could be retried.
+ * To use it in our generated client, this function must be SIDE-EFFECT FREE, so it does not
+ * consume the response body.
```
nickgros, line 32:
```suggestion
 * To use it in our generated client, this function must NOT consume the response body.
```

packages/synapse-client/src/SynapseClient.ts
```diff
@@ -41,17 +41,63 @@ import { VerificationServicesApi } from './generated/apis/VerificationServicesAp
...
+            )
+          }
+        }
+      },
+      // Convert error objects to our SynapseClientError
+      onError(context: ErrorContext): Promise<Response | void> {
+        const { response, error } = context
+        console.error(error)
+        throw new SynapseClientError(
+          0,
```
jay-hodgson, line 88: This looks familiar (like the network failure handling from the old SynapseClient), but I just want to confirm that we are not losing response.status information here

jay-hodgson (approved):

packages/synapse-client/src/SynapseClient.ts
nickgros: I believe this callback only runs when `fetch` throws an error, in which case there is no response object. This should give us identical behavior to the previous implementation.

PR #1234, opened 2024-09-24
PORTALS-3252

jay-hodgson:


packages/synapse-react-client/src/components/QueryWrapperPlotNav/QueryWrapperPlotNav.tsx
```diff
@@ -69,6 +69,7 @@ type QueryWrapperPlotNavOwnProps = {
   >
   facetsToPlot?: string[]
   availableFacets?: FacetFilterControlsProps['availableFacets']
+  facetValueSortConfigs?: FacetFilterControlsProps['facetValueSortConfigs']
```
nickgros, line 72: Could this name imply that the facet data is sorted in other places? The only other place I can think of is the plots (i.e. pie plot slices are ordered by sort rather than size). Consider a rename or comment to clarify that this is not the case, but maybe not a big deal.

packages/synapse-react-client/src/components/widgets/query-filter/EnumFacetFilter/EnumFacetFilter.tsx
```diff
@@ -137,8 +145,13 @@ function _EnumFacetFilter(props: EnumFacetFilterProps) {
...
     const valueNotSetFacetArray = partitions[0]
     const restOfFacetValuesArray = partitions[1]
+    const sortedValues = sortBy(restOfFacetValuesArray, fv =>
+      fv.displayText.toLowerCase(),
+    )
+    //PORTALS-3252: provide way to sort in descending order on the client-side
+    const sortDescending = sortConfig && sortConfig.direction == Direction.DESC
     return [
-      ...sortBy(restOfFacetValuesArray, fv => fv.displayText.toLowerCase()),
+      ...(sortDescending ? sortedValues.reverse() : sortedValues),
```
nickgros, line 154: Use of `reversed` is probably ok since `sortedValues` is new each render

packages/synapse-react-client/src/components/widgets/query-filter/EnumFacetFilter/EnumFacetFilter.tsx
```diff
@@ -137,8 +145,13 @@ function _EnumFacetFilter(props: EnumFacetFilterProps) {
     )
     const valueNotSetFacetArray = partitions[0]
     const restOfFacetValuesArray = partitions[1]
+    const sortedValues = sortBy(restOfFacetValuesArray, fv =>
+      fv.displayText.toLowerCase(),
+    )
```
nickgros, line 150: May want to check if the facet column is a number/integer, and if so, parse before sorting (don't forget a special case for null). Otherwise we'll end up with issues like these (though unlikely to happen for most real date data) 

```javascript
_.sortBy(['2001', '2010', '2003', '20030', '2011', '20110'], x => x.toLowerCase())
//   -> [ '2001', '2003', '20030', '2010', '2011', '20110' ]
```

packages/synapse-react-client/src/components/QueryWrapperPlotNav/QueryWrapperPlotNav.tsx
jay-hodgson: Added a lengthy comment for now

packages/synapse-react-client/src/components/widgets/query-filter/EnumFacetFilter/EnumFacetFilter.tsx
jay-hodgson: parsing number for simple (not list) number type columns

packages/synapse-react-client/src/components/widgets/query-filter/EnumFacetFilter/EnumFacetFilter.tsx
```diff
@@ -137,8 +156,19 @@ function _EnumFacetFilter(props: EnumFacetFilterProps) {
     )
     const valueNotSetFacetArray = partitions[0]
     const restOfFacetValuesArray = partitions[1]
+    let sortedValues: RenderedFacetValue[]
+    if (isNumberColumnType) {
+      sortedValues = sortBy(restOfFacetValuesArray, fv => new Number(fv.value))
```
nickgros, line 161: Any reason to not use `Number(...)` which returns a number, instead of `new Number(...)`, which returns an object?
```suggestion
      sortedValues = sortBy(restOfFacetValuesArray, fv => Number(fv.value))
```

nickgros (approved):

PR #1235, opened 2024-09-24
SWC-7091

jay-hodgson:


packages/synapse-react-client/src/components/SynapseChat/SynapseChat.tsx
```diff
@@ -85,7 +87,19 @@ export const SynapseChat: React.FunctionComponent<SynapseChatProps> = ({
       },
     },
     (status: AsynchronousJobStatus<AgentChatRequest, AgentChatResponse>) => {
-      setCurrentProgressMessage(status?.progressMessage)
+      setCurrentlyProcessingJobId(status.jobId)
```
jay-hodgson, line 90: when an agent chat message job is processing, enable the polling query for the trace events

packages/synapse-react-client/src/components/SynapseChat/SynapseChat.tsx
```diff
@@ -122,8 +136,8 @@ export const SynapseChat: React.FunctionComponent<SynapseChatProps> = ({
       setInteractions([...interactions, pendingInteraction])
       setCurrentResponse('')
       setCurrentResponseError('')
-      setCurrentProgressMessage('')
```
jay-hodgson, line 125: Move progress UI into SynapseChat (from SynapseChatInteraction), and look at the trace events instead of the async job status

packages/synapse-react-client/src/components/SynapseChat/SynapseChat.tsx
```diff
@@ -144,6 +158,7 @@ export const SynapseChat: React.FunctionComponent<SynapseChatProps> = ({
       sendChatMessageToAgent({
         chatText: initialMessage,
         sessionId: agentSession!.sessionId,
+        enableTrace: true,
```
jay-hodgson, line 161: Must tell the backend to provide the trace info

packages/synapse-react-client/src/components/SynapseChat/SynapseChat.tsx
```diff
@@ -245,13 +261,48 @@ export const SynapseChat: React.FunctionComponent<SynapseChatProps> = ({
...
+                <Box
+                  sx={{
+                    display: 'flex',
+                    flexDirection: 'column',
+                    justifyContent: 'center',
+                  }}
+                >
+                  <SynapseSpinner size={40} />
+                  {/* Show the current message, as well as the full trace log in a tooltip */}
+                  {traceEvents && traceEvents.page && (
```
jay-hodgson, line 282: Currently we ask for the full page of trace events (every second), but we could technically set the timestamp of the message we saw last and store them in another state variable.  Current implementation surfaces the latest trace event message, and surfaces the full (most recent) page of trace events in the tooltip

packages/synapse-types/src/Chat.ts
```diff
@@ -120,3 +122,40 @@ export type SessionHistoryResponse = {
...
+export type TraceEventsRequest = {
+  /**
+   * The job ID issued when the agent chat request job was started.
+   */
+  jobId: string
+
+  /**
+   * When a time stamp value is provided, only trace events that occurred after the provided time stamp will be included in the results.
+   */
+  olderThanTimestamp?: number // Optional since the JSON definition allows for it to be omitted
```
nickgros, line 160: I commented on John's doc but also calling out here I think this should be `newerThanTimestamp` or something like that

packages/synapse-react-client/src/mocks/msw/handlers/chatHandlers.ts
```diff
@@ -47,6 +52,20 @@ export const getChatbotHandlers = (
...
+    async (_req, res, ctx) => {
+      //mock showing progress (increasing number of items)
+      traceCallCount++
+      if (traceCallCount == 1) {
+        return res(ctx.status(201), ctx.json(mockTraceEventsResponse1))
+      } else if (traceCallCount == 2) {
+        return res(ctx.status(201), ctx.json(mockTraceEventsResponse2))
+      } else return res(ctx.status(201), ctx.json(mockTraceEventsResponse3))
+    },
+  ),
```
nickgros, line 68: With react-query (and probably also strictmode), we can't guarantee only one request is made before we should expect to see more data. However, I'm not sure if there's an easy way to change this, so I think this is OK for tests

packages/synapse-react-client/src/components/SynapseChat/SynapseChat.tsx
```diff
@@ -66,26 +69,39 @@ export const SynapseChat: React.FunctionComponent<SynapseChatProps> = ({
...
+    useState<string>()
   const [currentResponse, setCurrentResponse] = useState('')
-  const [currentProgressMessage, setCurrentProgressMessage] = useState<
-    string | undefined
-  >()
   const [currentResponseError, setCurrentResponseError] = useState('')
   // Keep track of the text that the user is currently typing into the textfield
   const [userChatTextfieldValue, setUserChatTextfieldValue] = useState('')
   const [initialMessageProcessed, setInitialMessageProcessed] = useState(false)
+  // const delay = (ms: number) => new Promise(resolve => setTimeout(resolve, ms))
```
nickgros, line 79: Remove, unless you still need these to test something

packages/synapse-react-client/src/components/SynapseChat/SynapseChat.tsx
nickgros: Since you aren't checking if the jobState is PROCESSING, I think you could remove this state variable and just use `isPending` returned by useSendChatMessageToAgent

packages/synapse-react-client/src/components/SynapseChat/SynapseChat.tsx
```diff
@@ -66,26 +69,39 @@ export const SynapseChat: React.FunctionComponent<SynapseChatProps> = ({
...
   const { mutate: sendChatMessageToAgent } = useSendChatMessageToAgent(
     {
       onSuccess: data => {
         // whenever the response is returned, set the last interaction response text
+        // await delay(6000)
```
nickgros, line 84:
```suggestion
```

packages/synapse-react-client/src/components/SynapseChat/SynapseChat.tsx
nickgros:
```suggestion
```

nickgros (approved):
Had a few suggestions

packages/synapse-react-client/src/components/SynapseChat/SynapseChat.tsx
jay-hodgson: The trace requires the job ID.  Is there another (preferable) way to access the job ID?
Unclear if I need to wait for the job to be PROCESSING.  I think with the current code, the call to ask for the trace events would fail, but I don't think the failure will have an impact on the UI (am I wrong?).

packages/synapse-types/src/Chat.ts
jay-hodgson: Nice catch.  I would expect it to be named newerThanTimestamp (if my understanding about the purpose of this field is correct)

packages/synapse-react-client/src/components/SynapseChat/SynapseChat.tsx
```diff
@@ -245,13 +263,51 @@ export const SynapseChat: React.FunctionComponent<SynapseChatProps> = ({
...
+                            })}
+                        </div>
+                      }
+                    >
+                      <Typography
+                        sx={{ textAlign: 'center' }}
+                        variant="body1Italic"
+                      >
+                        {traceEvents.page[0].message}
+                      </Typography>
```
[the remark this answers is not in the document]
jay-hodgson: Funny that you mentioned that, because I was going down that path already!  It introduced some additional complexity to the PR, so I wanted to implement in a follow-up.

packages/synapse-react-client/src/components/SynapseChat/SynapseChat.tsx
jay-hodgson: Is there an easier way to introduce some delay in this response during testing?  I need this delay to test the progress trace output.

packages/synapse-react-client/src/components/SynapseChat/SynapseChat.tsx
nickgros: Totally missed that you're pulling the jobId out of the status, so that makes sense.

I think it should handle the failure case just fine, but we'll see

packages/synapse-react-client/src/components/SynapseChat/SynapseChat.tsx
nickgros: You could add the delay in mock service worker: https://v1.mswjs.io/docs/api/context/delay

packages/synapse-react-client/src/components/SynapseChat/SynapseChat.tsx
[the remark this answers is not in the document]
jay-hodgson: The use of a TransitionGroup should really simplify (if it's based on mount/unmount).  This means that I will need to mount/unmount based on the current progress text.  Then it's a matter of positioning (since I don't want the old message to be fading out and taking up vertical space while the new message is fading in)

packages/synapse-react-client/src/components/SynapseChat/SynapseChat.tsx
jay-hodgson: Tried adding this option to our asyncJobHandlers.generateAsyncJobHandlers() helper function, but with the delay it never returns.  Oh well, I'll just add it when I need to.

jay-hodgson: Introducing in our asyncJobHandlers.generateAsyncJobHandlers() utility function caused the result to never return.  Not important, so I'm just going to clean up the "delay" code for now

nickgros (approved):

nickgros (approved):

PR #1237, opened 2024-09-25
SWC-7091: use newerThanTimestamp in TraceAgentsRequest, do not revers…

jay-hodgson:
…e the trace events (they are added in ascending order), and rework mock responses to align with expectations (return a new event on some calls, but an empty array for most of the status checks)

### DIFF
--- a/packages/synapse-react-client/src/components/SynapseChat/SynapseChat.tsx
+++ b/packages/synapse-react-client/src/components/SynapseChat/SynapseChat.tsx
@@ -17,6 +17,7 @@ import {
   AgentChatResponse,
   AgentSession,
   AsynchronousJobStatus,
+  TraceEvent,
 } from '@sage-bionetworks/synapse-types'
 import { TextField } from '@mui/material'
 import { useSynapseContext } from '../../utils'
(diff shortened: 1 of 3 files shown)


jay-hodgson:
With simulated delay in processing:
chrome_NzxyNzcNus

packages/synapse-react-client/src/components/SynapseChat/SynapseChat.tsx
```diff
@@ -279,46 +302,40 @@ export const SynapseChat: React.FunctionComponent<SynapseChatProps> = ({
...
-                            }
+                  <Box sx={{ position: 'relative', pt: '35px' }}>
+                    <TransitionGroup>
+                      {/* The key is based on the current text so when the text changes, the Fade component will remount */}
+                      <Fade key={latestTraceEventMessage} timeout={500}>
+                        <Tooltip
+                          placement="bottom"
+                          title={
+                            <div style={{ textAlign: 'center' }}>
+                              <Typography>Initializing...</Typography>
```
nickgros, line 313: Probably only show this node if `allTraceEvents` is empty

nickgros (approved):

packages/synapse-react-client/src/components/SynapseChat/SynapseChat.tsx
jay-hodgson: nice catch

jay-hodgson:
Ignoring SageAccountWeb test failure in ResetTwoFactorAuth.test.tsx (ReferenceError: window is not defined)

PR #1239, opened 2024-09-26
SWC-7061

nickgros:
- Fix issue where the portal endpoint was missing a trailing slash
- Fix issue where SCSS-styled links were not clearly links. This should not affect most of SageAccountWeb, which is styled with MUI, but it does affect the EntityFinder, which was recently surfaced via the webhooks feature.

### DIFF
--- a/apps/SageAccountWeb/src/style/_variables.scss
+++ b/apps/SageAccountWeb/src/style/_variables.scss
@@ -1,4 +1,4 @@
 // Variables used in Portals SCSS should be defined in this file.
 
-$primary-action-color: #2a2a2a !default;
+$primary-action-color: #395979 !default;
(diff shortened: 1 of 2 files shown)


apps/SageAccountWeb/src/style/_variables.scss
```diff
@@ -1,4 +1,4 @@
 // Variables used in Portals SCSS should be defined in this file.
 
-$primary-action-color: #2a2a2a !default;
+$primary-action-color: #395979 !default;
```
nickgros, line 3: This styles the links in the EntityFinder. Use the Synapse primary color instead of a dark gray color.

jay-hodgson (approved):

PR #1240, opened 2024-09-26
SWC-7091: fix trace endpoint, increase tooltip width, and attempt to …

jay-hodgson:
…extract 'thinking' text for a more friendly progress message


jay-hodgson:
chrome_tivq8ydEHK

nickgros (approved):
What a bizarre format to return a message. I would recommend removing the tooltip. More technical users working on their own agent could follow the network log (or fetch the trace themselves), and I don't think it's very useful for non-technical users.

nickgros (approved):

PR #1247, opened 2024-10-01
SWC-7100

jay-hodgson:


packages/synapse-react-client/src/synapse-queries/github/useGithub.ts
```diff
@@ -0,0 +1,32 @@
+import { useQuery, UseQueryOptions } from '@tanstack/react-query'
+import { SynapseClientError, useSynapseContext } from '../../utils'
+
+const GITHUB_API_URL = 'https://api.github.com/repos'
+
+export function useLatestTag(
```
jay-hodgson, line 6: This react-query wrapped fetch to get the latest tag will be reused in the new ToS workflow (where we will need to provide this to the backend).  We will also reuse the MarkdownGithub component (to get and display the new ToS content).

packages/synapse-react-client/src/components/Markdown/MarkdownGithub.tsx
```diff
@@ -0,0 +1,70 @@
+import React, { useEffect, useState } from 'react'
+import { SkeletonParagraph } from '../Skeleton'
+import MarkdownSynapse from './MarkdownSynapse'
+import { Box } from '@mui/material'
+import { useLatestTag } from '../../synapse-queries/github/useGithub'
+
+export type MarkdownGithubProps = {
```
jay-hodgson, line 7: SWC to use this component to show all documents from the Governance TrustCenter repo.

packages/synapse-react-client/src/components/Markdown/MarkdownGithub.tsx
```diff
@@ -0,0 +1,70 @@
...
+          const content = await fileResponse.text()
+          setFileContent(content)
+        } catch (error) {
+          console.error(error)
+        }
+      }
+    }
+
+    fetchFileContent()
+  }, [repoOwner, repoName, latestTag])
```
nickgros, line 45: Should also use react-query for this request

packages/synapse-react-client/src/components/Markdown/MarkdownGithub.tsx
```diff
@@ -0,0 +1,70 @@
...
+
+  const loadingUI = (
+    <Box sx={{ width: '100%' }}>
+      <SkeletonParagraph rowHeight="30px" numRows={70} />
+    </Box>
+  )
+  return (
+    <Box
```
nickgros, line 53: If you're just trying to center and constrain the width, you could consider using [Container](https://v5.mui.com/material-ui/react-container/) instead of adding all of these flex styles.

packages/synapse-react-client/src/components/Markdown/MarkdownGithub.tsx
```diff
@@ -0,0 +1,70 @@
...
+      sx={{
+        display: 'flex',
+        maxWidth: '1100px',
+        alignItems: 'center',
+        justifyContent: 'center',
+        p: { xs: '5px', sm: '30px' },
+        ml: { xs: '5px', md: 'auto' },
+        mr: { xs: '5px', md: 'auto' },
```
nickgros, line 61: Shorthand `mx` applies both left and right margins
```suggestion
        mx: { xs: '5px', md: 'auto' },
```

packages/synapse-react-client/src/synapse-queries/github/useGithub.ts
```diff
@@ -0,0 +1,32 @@
...
+    if (!response.ok) {
+      throw new Error(`Error fetching tags: ${response.statusText}`)
+    }
+    const tags = await response.json()
+    if (tags.length === 0) {
+      throw new Error('No tags found.')
+    }
+
+    const latestTag = tags[0].name
+    return latestTag
```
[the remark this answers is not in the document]
jay-hodgson: We should keep the latest tag logic though, for use in other contexts

nickgros (approved):

PR #1252, opened 2024-10-03
SWC-6533

jay-hodgson:


apps/synapse-oauth-signin/src/OAuth2Form.tsx
```diff
@@ -84,9 +79,6 @@ export function OAuth2Form() {
   )
 
   const clientId = useMemo(() => {
-    const code = queryParams.get('code')
-    if (code) return // we're in the middle of a SSO, do not attempt to get OAuthClient info yet
```
nickgros, line 88: Removal of all the registration/login logic, got it

apps/synapse-portal-framework/src/README.md
```diff
@@ -72,3 +72,11 @@ The route resolver handles taking a URL and matching it the config. Read more in
...
 The detail pages ([e.g.](https://staging.adknowledgeportal.synapse.org/Explore/Studies/DetailsPage?Study=syn5550404])) give
 a deeper dive into a particular portal section. Read more in the [component](portal-components/DetailsPage/DetailsPage.tsx).
+
+# How to Log In to a Portal Running on Localhost
+
+To log in to the portal running on localhost, follow these steps:
+
+1. Start the SageAccountWeb application locally by navigating to the app directory and running pnpm start (located in apps/SageAccountWeb).
+2. Once the application is running, log in to the local Accounts site.
+3. After successfully logging in, refresh the portal page to complete the process.
```
nickgros, line 82: We can probably add some orchestration with Nx so if you run `nx run myportal:start` we can automatically run `nx run SageAccountWeb:start`. We might want to give SageAccountWeb a unique port so we can reliably redirect to it.

packages/synapse-react-client/src/utils/AppUtils/AppUtils.ts
```diff
@@ -1,11 +1,39 @@
...
+  // save current URL in a cookie that One Sage will use to send you back to the correct page
+  const domainValue = window.location.hostname
+    .toLowerCase()
+    .endsWith('.synapse.org')
+    ? '.synapse.org'
+    : undefined
+  cookies.set(ONE_SAGE_REDIRECT_COOKIE_KEY, window.location.href, {
+    path: '/',
+    domain: domainValue,
+  })
```
nickgros, line 24: I was thinking we should expire the cookie, but the only case I think that would matter is this sequence:

1. I am redirected to log in from App A, which sets the cookie
2. I do not complete authentication for whatever reason
3. After a while, I use App B, which does NOT set the cookie and it redirects me to log in
4. I log in to Synapse and it sends me to App A

Kind of an edge case, WDYT? IMO it doesn't hurt to have it expire after a few hours.

packages/synapse-react-client/src/utils/AppUtils/AppUtils.ts
```diff
@@ -1,11 +1,39 @@
...
+}
+
+export function processRedirectURLInOneSage() {
+  if (cookies.get(ONE_SAGE_REDIRECT_COOKIE_KEY)) {
+    const href = cookies.get(ONE_SAGE_REDIRECT_COOKIE_KEY)
+    cookies.remove(ONE_SAGE_REDIRECT_COOKIE_KEY)
+    window.location.assign(href)
+    return true
```
nickgros, line 32: This will never return

nickgros (approved):

packages/synapse-react-client/src/utils/AppUtils/AppUtils.ts
jay-hodgson: No harm in setting the expiration.  The "session" cookie could be deleted by that time, and App B (all of our apps) _should_ set the cookie, but it's a good idea to guard against this case.  Thanks!

packages/synapse-react-client/src/utils/AppUtils/AppUtils.ts
jay-hodgson: That's what I thought, but I wasn't sure if this was an async process or not.  Back in the day, I saw a lot of instances of this line (which I think did have the possibility of executing the following lines):
window.location.href = "http://site.com"

jay-hodgson:
Let's start moving everyone to One Sage for login...

PR #1277, opened 2024-10-14
SWC-7114: make the charts real

jay-hodgson:
image


packages/synapse-react-client/src/components/SynapseHomepageV2/SynapseHomepageV2.tsx
```diff
@@ -27,7 +27,7 @@ import { SynapseHomepageChatSearch } from './SynapseHomepageChatSearch'
 
 export const synapseInActionTable = 'syn61670075'
 export const past30DaysDownloadMetricsTable = 'syn61597084'
-export const generalStatsMetricsTable = 'syn61588163'
+export const generalStatsMetricsTable = 'syn61588123'
```
jay-hodgson, line 30: This other table has historical data, not just data for the most recent month

packages/synapse-react-client/src/components/SynapseHomepageV2/SynapseByTheNumbersItem.tsx
```diff
@@ -1,15 +1,86 @@
...
+      y: yArrayNumbers,
+      hoverinfo: 'skip', //hide hover over tool tips
+    },
+  ]
+
+  // calculate the y axis range to show off the trend
+  const maxY = Math.max(...yArrayNumbers)
+  const minY = Math.min(...yArrayNumbers)
+  // we're showing a trend, so set the y-axis range minimum to 5% below the minimum point
+  const rangeMinYValue = minY - minY * 0.05
```
jay-hodgson, line 56: Setting the range minimum value is key in showing the trend rather than the raw y values

packages/synapse-react-client/src/components/SynapseHomepageV2/SynapseByTheNumbersItem.tsx
```diff
@@ -1,15 +1,86 @@
...
+
+  const plotData: Partial<Plotly.PlotData>[] = [
+    {
+      type: 'scatter',
+      mode: 'lines',
+      fill: 'tozeroy',
+      fillcolor: '#336663',
+      line: {
+        color: '#6ba89d',
+        shape: 'spline', //smooth line, will need to do a more complex cubic spline interpolation (introducing new values between the data points) if we want to really smooth this out
```
jay-hodgson, line 43: it still doesn't look smooth.  My hope is that with more data, this will look better over time.  If we need to do something more complex (to generate points between the actual data), then I have some notes here.

nickgros (approved):
looks great and is super straightforward!

PR #1279, opened 2024-10-15
Portals-3265: Search bar on tools database homepage is undersized and difficult to see

kianamcc:
Jira Ticket: https://sagebionetworks.jira.com/browse/PORTALS-3265?atlOrigin=eyJpIjoiZTU4MWFhMDY1OWE3NDA3YjliNmQwMWY0ZTQ0Njg1MmEiLCJwIjoiaiJ9

Some styling options from Adam include:
- white background, light gray outline to distinguish it from the background
- white background with a little drop of NF blue color in it

I went with the white background and light gray outline.

Screenshot 2024-10-14 at 4 36 52 PM
Screenshot 2024-10-14 at 4 36 46 PM


apps/synapse-portal-framework/src/components/nf/NFBrowseToolsPage.tsx
```diff
@@ -162,7 +168,14 @@ const NFBrowseToolsPage = (props: NFBrowseToolsPageProps) => {
           <div className="searchToolsRow">
             <div className="searchInputWithIcon">
               <TextField
```
jay-hodgson, line 170: I just realized that there's already a copy/paste of this search component in the ELBrowseToolsPage.  Could you refactor and reuse?  I know you asked about this, and at the time I didn't think it was needed, but you were right.

jay-hodgson:
This looks so much better, thank you!

apps/synapse-portal-framework/src/components/Search.tsx
```diff
@@ -0,0 +1,59 @@
+import React from 'react'
+import { TextField, Button, InputAdornment } from '@mui/material'
+
+type SearchProps = {
+  searchText: string
+  setSearchText: (value: string) => void
+  onSearch: (searchText: string) => void
+}
+
+const Search = ({ searchText, setSearchText, onSearch }: SearchProps) => {
```
jay-hodgson, line 10: If you moved the searchText state variable to this new Search component, would you only need to pass in onSearch()?

kianamcc: Whoops, yeah, defining that state in every component that uses the Search component would be a pain :sweat_smile:

kianamcc:
I tested a few words for the Search component on both nf and elportal!

jay-hodgson (approved):

PR #1280, opened 2024-10-15
PORTALS-3195 - SRC Utilities, finished SageAccountWeb and synapse-oauth-signin

nickgros:


apps/SageAccountWeb/src/style/_Navbar.scss
```diff
@@ -12,7 +12,7 @@
     text-decoration: none;
   }
 
-  .isSelected {
+  .active {
```
nickgros, line 15: Use the default NavLink `.active` class for styling

apps/synapse-oauth-signin/src/App.tsx
```diff
@@ -1,81 +1,43 @@
...
   return (
     <div className="App">
-      <BrowserRouter>
-        <StyledEngineProvider injectFirst>
-          <QueryClientProvider client={queryClient}>
-            <ThemeProvider theme={theme}>
-              <AppInitializer>
-                <SynapseErrorBoundary>
-                  <Switch>
-                    <Route exact path="/" render={() => <OAuth2Form />} />
```
nickgros, line 40: Routes extracted to new file

apps/synapse-oauth-signin/src/AppInitializer.tsx
```diff
@@ -6,18 +7,18 @@ import { handleErrorRedirect } from './URLUtils'
 function AppInitializer(
   props: React.PropsWithChildren<Record<string, unknown>>,
 ) {
-  const urlSearchParams = new URLSearchParams(window.location.search)
-  const prompt = urlSearchParams.get('prompt')
+  const [searchParams] = useSearchParams()
```
nickgros, line 10: Use react-router's `useSearchParams` hook. This is required since we're using `MemoryRouter` in tests--the `window.location` object provided by JSDOM is no longer modified.

apps/synapse-oauth-signin/src/LoginComponent.tsx
```diff
@@ -0,0 +1,15 @@
+import React, { useEffect } from 'react'
+import { useSearchParams } from 'react-router-dom'
+import { SynapseClient, useSynapseContext } from 'synapse-react-client'
+
+export function LoginComponent() {
```
nickgros, line 5: Existing component, just pulled out into its own file. I don't think it's used, but I'm trying not to make many functional changes in this PR.

apps/synapse-oauth-signin/src/LogoutComponent.tsx
```diff
@@ -0,0 +1,10 @@
+import React, { useEffect } from 'react'
+import { SynapseClient, useSynapseContext } from 'synapse-react-client'
+
+export function LogoutComponent() {
```
nickgros, line 4: Existing component, just pulled out into its own file. I don't think it's used, but I'm trying not to make many functional changes in this PR.

apps/synapse-oauth-signin/src/test/App.test.tsx
```diff
@@ -346,7 +356,7 @@ describe('App integration tests', () => {
     expect(window.location.replace).not.toHaveBeenCalled()
   })
 
-  test('Supports OAuth2 login with 2FA', async () => {
+  test.skip('Supports OAuth2 login with 2FA', async () => {
```
nickgros, line 359: Couldn't get this to work, but it will be skipped in #1252, so I decided to not spend much time on it.

packages/synapse-react-client/src/components/GenericCard/GenericCard.tsx
```diff
@@ -163,15 +163,34 @@ export const getValueOrMultiValue = ({
...
+      return findIndexResult
+    }
+  }
+  if (columnModels) {
+    const findIndexResult = columnModels.findIndex(el => el.name === columnName)
+    if (findIndexResult != -1) {
+      return findIndexResult
+    }
+  }
+  return undefined
```
nickgros, line 193: Fix edge cases found while testing portals

packages/synapse-react-client/src/components/Markdown/MarkdownSynapse.tsx
```diff
@@ -156,7 +156,9 @@ const MarkdownSynapse: MarkdownSynapseComponent = class MarkdownSynapse extends
         event.preventDefault()
         // handle table of contents widget
         const idOfContent = anchor.getAttribute('data-anchor')
-        const goTo = this.markupRef.current!.querySelector(`#${idOfContent}`)
+        const goTo = this.markupRef.current!.querySelector(
+          `[id="${idOfContent}"]`,
+        )
```
nickgros, line 161: In portals, there was a case where the `id` could contain a space--in those cases, `querySelector('#some id')` won't work because it ends up querying for a descendent of `#some` with tag `id`. This new query fixes that edge case.

The component that had this bug is changed in a PR coming later, but I decided to search for other usages and found this case, so I'm not sure if this code ever led to any bugs in the wild.

packages/synapse-react-client/src/components/QueryWrapperPlotNav/QueryWrapperPlotNav.tsx
```diff
@@ -107,9 +107,7 @@ type QueryWrapperPlotNavOwnProps = {
   Pick<QueryContextType, 'combineRangeFacetConfig'>
 
 export type SearchParams = {
-  searchParams?: {
-    facetValue: string
-  }
+  searchParams?: Record<string, string>
```
nickgros, line 110: This type was wrong.

jay-hodgson:
Refactor looks great.  As discussed, we agreed to hold off on merging this change until after the critical Synapse ToS changes

jay-hodgson (approved):

PR #1296, opened 2024-10-16
PORTALS-3290

jay-hodgson:


packages/synapse-react-client/src/components/ExperimentalMode/ExperimentalMode.tsx
```diff
@@ -35,16 +35,15 @@ const ExperimentalMode: React.FunctionComponent<ExperimentalModeProps> = ({
...
-    cookies.set(EXPERIMENTAL_MODE_COOKIE, { path: '/' })
+    cookies.set(EXPERIMENTAL_MODE_COOKIE, 'true', { path: '/' })
     setIsExperimentalModeOn(true)
     if (onExperimentalModeToggle) {
       onExperimentalModeToggle(true)
     }
   }
 
   const deleteExperimentalModeCookie = () => {
-    // cookie.remove requires to re-calculate date obj each time, this is more straightforward
```
jay-hodgson, line 46: Maybe this comment was true at some point, but it certainly isn't now.  Just use cookies.remove() (I suspect that this will fix the bug)

packages/synapse-react-client/src/components/ExperimentalMode/ExperimentalMode.tsx
```diff
@@ -35,16 +35,15 @@ const ExperimentalMode: React.FunctionComponent<ExperimentalModeProps> = ({
   }, [])
 
   const createExperimentalModeCookie = () => {
-    cookies.set(EXPERIMENTAL_MODE_COOKIE, { path: '/' })
+    cookies.set(EXPERIMENTAL_MODE_COOKIE, 'true', { path: '/' })
```
jay-hodgson, line 38: The second parameter is the cookie value, the third parameter contain the options

jay-hodgson:
I built the SRC and tested in SageAccountWeb (One Sage) to verify that the Experimental Mode slider still functioned properly (locally)

sagely1 (approved):
LGTM!

PR #1297, opened 2024-10-17
should be converted to hooks, but for now fix for development environ…

jay-hodgson:
…ment


jay-hodgson:
This logic should be refactored into a hook that utilizes react-query

kianamcc:
This does the trick!

kianamcc (approved):

PR #1301, opened 2024-10-18
PORTALS-3190: from testing, clear the session if we encounter an invalid_token message

jay-hodgson:


jay-hodgson:
Test failure is due to the ELITE portal Computational Tools being empty (for some reason).  I informed Milan of this situation

xschildw (approved):

packages/synapse-react-client/src/components/QueryWrapperErrorBoundary.tsx
```diff
@@ -22,8 +26,8 @@ export function QueryWrapperErrorBoundary({
     [getCurrentQueryRequest],
   )
   const { accessToken } = useSynapseContext()
+  const { clearSession } = useApplicationSessionContext()
```
nickgros, line 29: This is going to break in SWC where there is no wrapping `ApplicationSessionContext`

jay-hodgson: Let's see if the polling fixes this issue (as you hypothesized here: https://sagebionetworks.jira.com/browse/PORTALS-3190?focusedCommentId=223776)

PR #1302, opened 2024-10-21
SWC-7123: When on localhost, links to OneSage should also be on localhost

nickgros:
https://sagebionetworks.jira.com/browse/SWC-7123

### DIFF
--- a/apps/SageAccountWeb/package.json
+++ b/apps/SageAccountWeb/package.json
@@ -72,6 +72,7 @@
   "scripts": {
     "clean": "rimraf build coverage",
     "start": "vite",
+    "preview": "vite preview",
     "build": "vite build",
     "test": "vitest",
     "type-check": "tsc --noEmit"
--- a/packages/synapse-react-client/src/components/SynapseHomepageV2/SynapseHomepageNavBar.tsx
+++ b/packages/synapse-react-client/src/components/SynapseHomepageV2/SynapseHomepageNavBar.tsx
@@ -87,7 +87,12 @@ export const SynapseHomepageNavBar: React.FunctionComponent<
   const handleClosePortalResources = () => {
     setPortalResourcesAnchorEl(null)
   }
-  const oneSageURL = useOneSageURL()
+
+  const onLoginClicked = () => {
+    // Navigate to the LoginPlace, which will set a cookie that will be used to redirect back from OneSage
+    gotoPlace(LOGIN_LINK)
+  }
+
   return (
     <Box
       sx={{
@@ -140,7 +145,7 @@ export const SynapseHomepageNavBar: React.FunctionComponent<
               variant="outlined"
               color="secondary"
               sx={navButtonSx}
-              href={oneSageURL.toString()}
+              onClick={onLoginClicked}
             >
               Login
             </Button>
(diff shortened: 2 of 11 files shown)


packages/synapse-react-client/src/utils/hooks/useOneSageURL.ts
```diff
@@ -20,10 +20,16 @@ export function useOneSageURL(
...
-        : ONE_SAGE_STAGING_URL
+    let targetURL = ONE_SAGE_PRODUCTION_URL
+    if (window.location.hostname == 'staging.synapse.org') {
+      targetURL = ONE_SAGE_STAGING_URL
+    } else if (
+      window.location.hostname == 'localhost' ||
+      window.location.hostname == '127.0.0.1'
+    ) {
+      targetURL = `http://${window.location.hostname}:3000/`
+    }
```
nickgros, line 31: e.g. the login button on SWC uses this hook to determine the login URL. It should point to localhost in those cases

apps/SageAccountWeb/package.json
```diff
@@ -72,6 +72,7 @@
   "scripts": {
     "clean": "rimraf build coverage",
     "start": "vite",
+    "preview": "vite preview",
```
nickgros, line 75: Shortcut to locally serve the built output (so I can just run `nx run SageAccountWeb:preview --port 3000` before running SWC e2e tests locally)

jay-hodgson (approved):

packages/synapse-react-client/src/components/SynapseHomepageV2/SynapseHomepageNavBar.tsx
```diff
@@ -87,7 +87,12 @@ export const SynapseHomepageNavBar: React.FunctionComponent<
   const handleClosePortalResources = () => {
     setPortalResourcesAnchorEl(null)
   }
-  const oneSageURL = useOneSageURL()
+
+  const onLoginClicked = () => {
+    // Navigate to the LoginPlace, which will set a cookie that will be used to redirect back from OneSage
+    gotoPlace(LOGIN_LINK)
+  }
```
nickgros, line 94: Setting this cookie is crucial to allow login in the dev environment, since the [syn45291362 table](https://www.synapse.org/Synapse:syn45291362/tables/) doesn't exist when using another backend stack.

packages/synapse-react-client/src/utils/hooks/useOneSageURL.ts
```diff
@@ -1,11 +1,70 @@
...
+    currentHostname === 'synapse.org' ||
+    currentHostname.endsWith('.synapse.org')
+  ) {
+    if (currentBackendEndpoint === SYNAPSE_BACKEND_PRODUCTION_URL) {
+      return ONE_SAGE_PRODUCTION_URL
+    }
+    if (currentBackendEndpoint === SYNAPSE_BACKEND_STAGING_URL) {
+      return ONE_SAGE_STAGING_URL
+    }
+  }
```
nickgros, line 27: Before, we were checking each SWC domain. This didn't work for non-SWC sites. For example, if OneSage _staging_ discovers at login time that your password is weak, it redirects you to OneSage _production_.

This should prevent stack hopping and work more reliably across all sites, including OneSage & portals.

jay-hodgson (approved):

PR #1304, opened 2024-10-21
SWC-7063

jay-hodgson:


jay-hodgson:
chrome_GCBQ11e1Qd

packages/synapse-react-client/src/components/ProjectStorage/ProjectStorage.stories.ts
```diff
@@ -0,0 +1,77 @@
...
+    isAuthenticated: true,
+  },
+} satisfies Meta
+export default meta
+type Story = StoryObj<typeof meta>
+
+export const ProjectDataUnderLimit: Story = {
+  args: {
+    projectId: UNDER_LIMIT_PROJECT_ID,
+    sx: { backgroundColor: '#375574' },
```
jay-hodgson, line 33: Note, I'm setting the background color in the story (not in the component)

packages/synapse-react-client/src/components/ProjectStorage/ProjectDataAvailability.tsx
```diff
@@ -0,0 +1,101 @@
...
+  const synapseStorageUsage =
+    projectDataUsageArray?.length == 1 ? projectDataUsageArray[0] : undefined
+  if (!synapseStorageUsage) {
+    return <></>
+  }
+  const { sumFileBytes = 0, maxAllowedFileBytes = 1 } = synapseStorageUsage
+  const usageBarFilledPx = Math.min(
+    Math.round((sumFileBytes / maxAllowedFileBytes) * usageBarWidth),
+    usageBarWidth,
+  )
```
nickgros, line 37: What happens if `maxAllowedFileBytes` is 0?

packages/synapse-react-client/src/mocks/projectStorage/mockProjectStorageLimits.ts
```diff
@@ -0,0 +1,41 @@
...
+      storageLocationId: `${SYNAPSE_STORAGE_LOCATION_ID}`,
+      sumFileBytes: 500000000,
+      maxAllowedFileBytes: 1073741824,
+      isOverLimit: false, // Under the limit
+    },
+    {
+      storageLocationId: 'location-2',
+      sumFileBytes: 1000,
+      maxAllowedFileBytes: 1073741824,
+      isOverLimit: true,
```
nickgros, line 38:
```suggestion
      isOverLimit: false,
```

packages/synapse-react-client/src/components/ProjectStorage/ProjectStorage.stories.ts
nickgros: You could instead wrap the story in a `decorator` that applies the background (e.g. EnumFacetFilter.stories.tsx)

Then remove the sx prop if this is the only place you plan to use it.

jay-hodgson: I initially implemented a decorator, but thought it would be more flexible if this component supported SxProps!  Happy to change to a decorator if you think it's more appropriate

packages/synapse-react-client/src/components/ProjectStorage/ProjectDataAvailability.tsx
jay-hodgson: Bad things with the current code, but what should be shown in this case?  Showing a bar with a max of 0 seems confusing, so maybe we should just show nothing.

nickgros (approved):

PR #1307, opened 2024-10-22
PORTALS-3269: staging only: "submit an observation" at bottom of tool detail pages needs changing 

kianamcc:
Jira Ticket: https://sagebionetworks.jira.com/browse/PORTALS-3269?atlOrigin=eyJpIjoiY2QzZDFjZmZjYTExNDU0NjliNDAxMzJiODQ4N2JkNTIiLCJwIjoiaiJ9

Made a new wiki for "Submit Observation" as this change is for staging only at the moment: https://www.synapse.org/Synapse:syn26338068/wiki/629918

Before:
Screenshot 2024-10-22 at 10 51 00 AM

After:
Screenshot 2024-10-22 at 10 50 53 AM


jay-hodgson (approved):
Approved but should not yet be merged (unless we want to revert before the production release)

kianamcc:
> Approved but should not yet be merged (unless we want to revert before the production release)

Will until after the production release!

PR #1309, opened 2024-10-22
PORTALS-3293: Fix access token issue, as well as discrepancy in API d…

jay-hodgson:
…esign doc (field renamed during backend implementation).

During this set of testing, I was able to log in with an up-to-date ToU agreement.
Then (as an admin), I set the ToU requirement (PUT /termsOfUse2/requirements) on staging to a few days in the future.
{
     "requirementDate": "2024-10-25T22:03:55.000Z",
     "minimumTermsOfServiceVersion": "1.0.1"
}
On OneSage login I was then sent to the ToU page, where I could skip.  So I did!  I verified that I was not prompted to agree to the new terms in this browser session.  I then verified that in a new browser session, I was again prompted to agree to the updated terms.  I agreed to the terms, and verified the ToU status response was updated appropriately.
This seems to work!

### DIFF
--- a/packages/synapse-react-client/src/synapse-queries/termsOfService/useTermsOfService.ts
+++ b/packages/synapse-react-client/src/synapse-queries/termsOfService/useTermsOfService.ts
@@ -29,9 +29,10 @@ export function useTermsOfServiceInfo(
 }
 
 export function useTermsOfServiceStatus(
+  accessToken?: string, //usually we can fetch the access token from the context, but this hook is used by ApplicationSessionManager which sets the access token in the context (ApplicationSessionContextProvider)!
   options?: Partial<UseQueryOptions<TermsOfServiceStatus, SynapseClientError>>,
 ) {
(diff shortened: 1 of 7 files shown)


packages/synapse-react-client/src/synapse-queries/termsOfService/useTermsOfService.ts
```diff
@@ -29,9 +29,10 @@ export function useTermsOfServiceInfo(
 }
 
 export function useTermsOfServiceStatus(
+  accessToken?: string, //usually we can fetch the access token from the context, but this hook is used by ApplicationSessionManager which sets the access token in the context (ApplicationSessionContextProvider)!
```
jay-hodgson, line 32: note the note

nickgros: noted :)

PR #1310, opened 2024-10-23
When the ToU are not accepted, redirect to ToU page, not app.

nickgros:


apps/SageAccountWeb/src/App.tsx
```diff
@@ -1,37 +1,77 @@
...
+
+function LoggedInRedirector() {
+  const { accessToken } = useSynapseContext()
+  const appContext = useAppContext()
+  const { data: tosStatus } = SynapseQueries.useTermsOfServiceStatus(
+    accessToken,
+    {
+      enabled: !!accessToken,
+    },
+  )
```
nickgros, line 47: Add a check on the ToS. If you MUST agree to the ToS, send you to the ToS page instead of the app you tried to log into

apps/SageAccountWeb/src/App.tsx
```diff
@@ -1,37 +1,77 @@
...
+    // User is logged in and visiting the root, redirect (unless in the SSO Flow)
+    if (!isInSSOFlow && accessToken) {
+      if (
+        tosStatus &&
+        tosStatus.userCurrentTermsOfServiceState ===
+          TermsOfServiceState.MUST_AGREE_NOW
+      ) {
+        history.push('/authenticated/signUpdatedTermsOfUse')
+      } else if (tosStatus) {
+        // take user back to page they came from in the source app, if stored in a cookie
```
jay-hodgson, line 64: Right, we don't want to redirect before tosStatus is retrieved

apps/SageAccountWeb/src/App.tsx
jay-hodgson: Not sure why the AppInitializer is not properly handling, there must be a race condition when accessing the root.  Thanks for adding this!

jay-hodgson (approved):

apps/SageAccountWeb/src/components/SignUpdatedTermsOfUsePage.tsx
```diff
@@ -50,7 +53,11 @@ export const SignUpdatedTermsOfUsePage = (
           },
           {
             onSuccess: () => {
-              restoreLastPlace()
+              // take user back to page they came from in the source app, if stored in a cookie
+              const isProcessed = processRedirectURLInOneSage()
+              if (!isProcessed) {
+                restoreLastPlace(history)
+              }
```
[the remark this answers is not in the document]
jay-hodgson: Do we set the cookie when we send the user to their account settings page in SWC (for example)?  If so, they will be sent back to SWC after signing instead of to the account settings (which is not ideal)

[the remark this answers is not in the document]
nickgros: Perhaps not. The redirect issue may only affect localhost, which is annoying but maybe not that bad. I will remove the cookie check for now.

apps/SageAccountWeb/src/App.tsx
nickgros: Slipped my mind that you added this logic to the AppInitializer. I'm going to remove my redirect and let it handle yours, but keep the logic to skip the race condition.

apps/SageAccountWeb/src/App.tsx
```diff
@@ -32,44 +30,34 @@ import TwoFactorAuthBackupCodesPage from './components/TwoFactorAuth/TwoFactorAu
...
-    },
-  )
 
   const isCodeSearchParam = getSearchParam('code') !== undefined
   const isProviderSearchParam = getSearchParam('provider') !== undefined
   const isInSSOFlow = isCodeSearchParam && isProviderSearchParam
-  const history = useHistory()
+
+  const { mayRedirect: mayRedirectToSignToS } =
+    useMaybeRedirectToSignTermsOfService()
```
nickgros, line 47: Extracted the ToS redirect into a shared hook, which returns a `mayRedirect` flag that we can use to decide if we should run the redirect at the root route

jay-hodgson (approved):

PR #1313, opened 2024-10-23
SWC-7127, SWC-7128

jay-hodgson:


apps/SageAccountWeb/src/components/SignUpdatedTermsOfUsePage.tsx
```diff
@@ -108,6 +110,8 @@ export const SignUpdatedTermsOfUsePage = (
                 onClick={() => {
                   sessionStorage.setItem('skippedSigningToS', 'true')
                   restoreLastPlace(history)
+                  //if we did not redirect to a page in this app, then look for the redirect cookie
+                  processRedirectURLInOneSage()
```
jay-hodgson, line 114: SWC-7127: Fix the redirect issue after skipping or agreeing to the new ToS

packages/synapse-react-client/src/utils/hooks/useOneSageURL.ts
```diff
@@ -14,14 +14,19 @@ export function getOneSageBaseUrl(
   currentBackendEndpoint: string,
 ) {
   // If we're on synapse.org, then go to the accounts site that matches the current backend stack.
+  //backend endpoint may contain a trailing slash, so let's remove it if present:
```
jay-hodgson, line 17: SWC-7128: Fix the mapping issue

packages/synapse-react-client/src/utils/hooks/useOneSageURL.ts
```diff
@@ -31,11 +36,11 @@ export function getOneSageBaseUrl(
   }
 
   if (currentHostname === 'localhost' || currentHostname === '127.0.0.1') {
-    return `http://${currentHostname}:3000/`
+    return `http://${currentHostname}:3000`
```
jay-hodgson, line 39: Just to make consistent with the other URLs that are returned (ONE_SAGE_STAGING_URL and ONE_SAGE_PRODUCTION_URL found in SynapseConstants, which do not have a trailing forward slash)

jay-hodgson:
Merging now due to the tight deadline, but would like this to be reviewed

apps/SageAccountWeb/src/components/SignUpdatedTermsOfUsePage.tsx
nickgros: One minor change suggested in #1314

PR #1314, opened 2024-10-24
SWC-7127 - edge case

nickgros:
redirect within OneSage may have happened without triggering a page reload

### DIFF
--- a/apps/SageAccountWeb/src/components/SignUpdatedTermsOfUsePage.tsx
+++ b/apps/SageAccountWeb/src/components/SignUpdatedTermsOfUsePage.tsx
@@ -19,11 +19,7 @@ import {
 } from 'synapse-react-client'
 import { StyledOuterContainer } from './StyledComponents'
 
-export type SignUpdatedTermsOfUsePageProps = {}
-
-export const SignUpdatedTermsOfUsePage = (
-  props: SignUpdatedTermsOfUsePageProps,
-) => {
+export function SignUpdatedTermsOfUsePage() {
   const [isLoading, setIsLoading] = useState(false)
   const [isCheckboxSelected, setIsCheckboxSelected] = useState(false)
   const { accessToken } = SynapseContextUtils.useSynapseContext()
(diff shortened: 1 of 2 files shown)


apps/SageAccountWeb/src/components/SignUpdatedTermsOfUsePage.tsx
```diff
@@ -37,6 +33,14 @@ export const SignUpdatedTermsOfUsePage = (
     },
   )
 
+  const redirectAfterSignOrSkip = () => {
+    const didRedirect = restoreLastPlace(history)
+    if (!didRedirect) {
+      // if we did not redirect to a page in this app, then look for the redirect cookie
+      processRedirectURLInOneSage()
+    }
```
nickgros, line 41: minor change: if we redirected you within OneSage (e.g. to your account settings page), react-router would handle it, so `processRedirectURLInOneSage` would run unexpectedly.

Updated `restoreLastPlace` to return a boolean if a redirect happened, and use it to decide whether to redirect using the cookie.

packages/synapse-react-client/src/utils/AppUtils/AppUtils.ts
```diff
@@ -40,15 +40,26 @@ export function processRedirectURLInOneSage() {
...
+): boolean {
   // go back to original route after successful SSO login
   const originalUrl = localStorage.getItem(LAST_PLACE_LOCALSTORAGE_KEY)
   localStorage.removeItem(LAST_PLACE_LOCALSTORAGE_KEY)
   const redirectUrl = originalUrl ?? fallbackRedirectUrl
-  if (redirectUrl) {
+  if (
+    redirectUrl &&
+    window.location.href != redirectUrl &&
+    window.location.href.substring(window.location.origin.length) != redirectUrl
```
nickgros, line 61: Do not redirect if the redirect URL matches the current URL

jay-hodgson (approved):

PR #1315, opened 2024-10-24
fix cases where you may not be redirected on the root route

nickgros:
useMaybeRedirectToSignTermsOfService should return a `useState` variable to trigger a re-render, rather than the current value of a ref.

This fixes SWC e2e tests. 10/10 passed in chromium & ff, 9/10 passed in webkit with one timeout

### DIFF
--- a/apps/SageAccountWeb/src/hooks/useMaybeRedirectToSignTermsOfService.ts
+++ b/apps/SageAccountWeb/src/hooks/useMaybeRedirectToSignTermsOfService.ts
@@ -1,5 +1,5 @@
 import { TermsOfServiceState } from '@sage-bionetworks/synapse-types'
-import { useEffect, useRef } from 'react'
+import { useEffect, useState } from 'react'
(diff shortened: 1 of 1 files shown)


apps/SageAccountWeb/src/hooks/useMaybeRedirectToSignTermsOfService.ts
```diff
@@ -23,7 +23,7 @@ export default function useMaybeRedirectToSignTermsOfService(): UseMaybeRedirect
   const location = useLocation()
 
   // true until we confirm we will not redirect
-  const mayRedirect = useRef(true)
+  const [mayRedirect, setMayRedirect] = useState(true)
```
nickgros, line 26: by using `useRef`, we avoided triggering re-renders when `mayRedirect` was updated. We DO want the re-render, so the redirect behavior in the LoginRedirector component is retriggered when this finally resolves to `false`.

PR #1322, opened 2024-10-28
PORTALS-3272: Add multi-select dropdown menu to observation pages to filter

kianamcc:
Jira Ticket: https://sagebionetworks.jira.com/browse/PORTALS-3272?atlOrigin=eyJpIjoiMjFjMGU0MDJiZWZmNDU2NmI1NmE3YTBlYzRkOGVhY2UiLCJwIjoiaiJ9


packages/synapse-react-client/src/components/CardContainer/CardContainer.tsx
```diff
@@ -90,72 +96,168 @@ function _CardContainer(props: CardContainerProps) {
     ids,
     type: 'ENTITY_HEADER',
   })
+
+  const observationColIndex = rowSet?.headers?.findIndex(
```
jay-hodgson, line 100: I don't think this logic belongs here.  Am I correct in assuming that in this case it will render many ObservationCard components?  Could this logic be moved to the ObservationCard (if it should be filtered out, then may it returns an empty react fragment?)?

packages/synapse-react-client/src/components/TimelinePlot/TimelinePhase.tsx
```diff
@@ -215,11 +219,54 @@ const TimelinePhase = ({
...
+        })
+      },
+    })),
+  ]
+
+  const filteredRows = rowData.filter(row =>
+    selectedObservationTypes.every(type =>
+      JSON.parse(row.values[schema.observationType] ?? '[]').includes(type),
+    ),
+  )
```
nickgros, line 268: Instead of filtering client-side, we should make this new UI control apply a facet filter and pass it to Synapse in the QueryBundleRequest.

To actually get the valid observationType, the [Observations MaterializedView](https://www.synapse.org/Synapse:syn51735464/tables/) should be updated to make `observationType` a faceted column. Once that's done, Synapse will give us the set of observationType values in the QueryResultBundle.facets object.

packages/synapse-react-client/src/components/CardContainer/CardContainer.tsx
nickgros: At the very least, this logic should be less specific to ObservationCards, and should utilize the Synapse tables faceted column features to get filter values and apply the filter to the rows server-side.

nickgros: I think a good example of a similar component is the `QuerySortSelector`

apps/portals/nf/src/config/synapseConfigs/tools.ts
```diff
@@ -254,23 +254,13 @@ export const toolDetailsPageConfig: DetailsPageProps = {
...
-        //   name: 'CardContainerLogic',
-        //   props: {
-        //     sql: `${observationsSql} WHERE observationTime IS NOT NULL ORDER BY observationTime DESC`,
-        //     type: SynapseConstants.OBSERVATION_CARD,
-        //     limit: 3,
-        //   },
-        //   title: 'Natural History Observations',
-        //   tableSqlKeys: ['resourceId'],
-        //   columnName: 'resourceId',
-        // },
```
nickgros, line 267: Commented out on purpose?

packages/synapse-react-client/src/components/CardContainer/CardContainer.tsx
```diff
@@ -90,65 +90,71 @@ function _CardContainer(props: CardContainerProps) {
...
+    selectColumns: rowSet.headers,
+    columnModels: queryMetadata.columnModels,
+    tableEntityConcreteType: tableEntityConcreteType[0]?.type,
+    tableId: rowSet.tableId,
+    queryContext,
+    queryVisualizationContext,
+    ...rest,
+  })
+
+  const renderCards = (rows: any) =>
```
nickgros, line 122: Is the only change here to extract this `renderCards` method and change the `if`/`else` to `switch`? Is this necessary?

packages/synapse-react-client/src/components/CardContainerLogic/CardContainerLogic.tsx
```diff
@@ -219,6 +221,9 @@ export function CardContainerLogic(props: CardContainerLogicProps) {
       >
         <QueryWrapperErrorBoundary>
           {sortConfig && <QuerySortSelector sortConfig={sortConfig} />}
+          {filterColumnName && (
+            <ColumnFilter filterColumnName={filterColumnName} />
+          )}
```
nickgros, line 226: Nice! Super simple addition here.

packages/synapse-react-client/src/components/ColumnFilter.tsx
```diff
@@ -0,0 +1,129 @@
...
+  ...rest
+}: ControlProps<any, boolean, GroupBase<any>>) => {
+  return (
+    <components.Control {...rest} className="form-control">
+      {children}
+    </components.Control>
+  )
+}
+
+const ColumnFilter: React.FC<FilterProps> = props => {
```
nickgros, line 32: Prefer using `function` and removing `React.FC` (which I think is deprecated)
```suggestion
function ColumnFilter (props: FilterProps) {
```

packages/synapse-react-client/src/components/ColumnFilter.tsx
```diff
@@ -0,0 +1,129 @@
...
+export type FilterProps = {
+  filterColumnName?: string
+}
+
+export const Control = ({
+  children,
+  ...rest
+}: ControlProps<any, boolean, GroupBase<any>>) => {
+  return (
+    <components.Control {...rest} className="form-control">
```
nickgros, line 26: Could you convert this to use MUI's `Autocomplete`  instead of using Bootstrap 4 + react-select? In general we are trying to migrate away from Bootstrap, and while we still use react-select, prefer to use MUI when possible.

packages/synapse-react-client/src/components/ColumnFilter.tsx
```diff
@@ -0,0 +1,129 @@
...
+
+      if (newFilter) {
+        updatedFilters.push(newFilter)
+      }
+
+      request.query.additionalFilters = updatedFilters
+
+      setSelectedTypes(selectedValues)
+      return request
+    })
```
nickgros, line 97: I think you can just do something like
```
removeSelectedFacet(columnName)
values.forEach(value => addValueToSelectedFacet(columnName, value))
```

Where `removeSelectedFacet` and `addValueToSelectedFacet` can be retrieved from the queryContext object.

packages/synapse-react-client/src/components/ColumnFilter.tsx
```diff
@@ -0,0 +1,129 @@
...
+
+const ColumnFilter: React.FC<FilterProps> = props => {
+  const queryContext = useQueryContext()
+  const {
+    executeQueryRequest,
+    queryMetadataQueryOptions,
+    getCurrentQueryRequest,
+  } = queryContext
+  const { data: queryMetadata } = useQuery(queryMetadataQueryOptions)
+  const [selectedTypes, setSelectedTypes] = useState<string[]>([])
```
nickgros, line 40: Instead of storing the selection in state, we can make the source of truth the current query and eliminate this state variable.

You can do something like
```
const selectedFacet = getCorrespondingSelectedFacet({columnName}, currentQuery.query.selectedFacets)
const selectedValues = (selectedFacet as FacetColumnValuesRequest).facetValues
```

`getCorrespondingSelectedFacet` is a utility function you can import.

packages/synapse-react-client/src/components/ColumnFilter.tsx
nickgros: Could you add some tests for this component? The QuerySortSelector has a corresponding QuerySortSelector.test.tsx file that you can use as a base. Unfortunately, that one is older, so it's not type-checked and co-located with the QuerySortSelector component file.

For the `ColumnFilter.test.tsx`, could you co-locate the file with your new component file, and ensure it passes typechecking?

packages/synapse-react-client/src/components/TimelinePlot/TimelinePhase.tsx
```diff
@@ -235,6 +234,31 @@ const TimelinePhase = ({
...
+    sql = `SELECT * FROM syn51735464
+    WHERE observationText IN (${observationTexts.join(', ')})`
+  } else {
+    sql = `SELECT * FROM syn51735464
+    WHERE observationId IN (${observationIds.join(', ')})${
+      observationTexts.length > 0
+        ? ` OR observationText IN (${observationTexts.join(', ')})`
+        : ''
+    }`
+  }
```
nickgros, line 260: Why do you need this SQL? Sorry, I don't totally understand the feature

nickgros (changes requested):
New file is very close to what I was expecting and is so much less code than the original implementation! I had some questions and pointed out opportunities to clean up

packages/synapse-react-client/src/components/TimelinePlot/TimelinePhase.tsx
nickgros: We can use the `ROW_ID` meta-column, so here we can collect the row IDs
```suggestion
  const selectedRowIds = selectedRows.map(row => row.rowId!)
```

packages/synapse-react-client/src/components/TimelinePlot/TimelinePhase.tsx
```diff
@@ -260,23 +284,19 @@ const TimelinePhase = ({
...
+        onClose={() => setClickEvent(undefined)}
+        open={selectedRows && selectedRows.length > 0}
+      >
+        <DialogContent>
+          <CardContainerLogic
+            filterColumnName="observationType"
+            sql={sql}
+            type={OBSERVATION_CARD}
+            initialLimit={3}
+          />
```
nickgros, line 297: We can add additional filters using the `searchParams` prop (which needs a better name, but that would be a massive change).

We add the `sqlOperator` prop to tell our code to add the searchParams as a filter with the `IN` clause.

And we can also add the `lockedColumn` to hide any UI that would indicate the ROW_ID filter (it may not make a difference now, but we can add it to be safe)

```suggestion
          <CardContainerLogic
            filterColumnName="observationType"
            sql={observationsSql}
            searchParams={{["ROW_ID"]: selectedRowIds}}
            sqlOperator="IN"
            lockedColumn={{ columnName: "ROW_ID" }}
            type={OBSERVATION_CARD}
            initialLimit={3}
          />
```

packages/synapse-react-client/src/components/ColumnFilter/ColumnFilter.test.tsx
```diff
@@ -0,0 +1,125 @@
...
+}))
+jest.mock('@tanstack/react-query', () => ({
+  useQuery: jest.fn(),
+  QueryClient: jest.fn().mockImplementation(() => ({
+    queryCache: { clear: jest.fn() },
+    setQueryData: jest.fn(),
+    getQueryData: jest.fn(),
+  })),
+}))
+
```
nickgros, line 20: I like how you approached making this a unit test. That said, we've tended to avoid mocking the entire react-query module because it becomes difficult to create mocks in components that contain multiple instances of `useQuery`. If we were to extend this component with another query, we'd have to totally change this mock approach, building dynamic logic into `useQuery` mock.

Instead, I think you could continue to extend the QueryContext mock by making the queryMetadataQueryOptions something like this:

```ts
{
  queryKey: ['queryMetadataQueryOptions']
  queryFn: () => data
}
```

It looks like we've actually done this in another test, FullTextSearch.test.ts

nickgros: We don't typically do it, but I like how you mocked `useQueryContext` instead of using the real one, so no need to change that.

packages/synapse-react-client/src/components/ColumnFilter/ColumnFilter.test.tsx
```diff
@@ -0,0 +1,125 @@
...
+const data = syn20337467Json as QueryResultBundle
+
+describe('ColumnFilter tests', () => {
+  const executeQueryRequest = jest.fn()
+  const addValueToSelectedFacet = jest.fn()
+  const removeSelectedFacet = jest.fn()
+
+  beforeEach(() => {
+    ;(useQueryContext as jest.Mock).mockReturnValue({
```
nickgros, line 29: You can get type-safe mocks with the following pattern:

const mockUseQueryContext = jest.mocked(useQueryContext)

```suggestion


describe('ColumnFilter tests', () => {
  const executeQueryRequest = jest.fn()
  const addValueToSelectedFacet = jest.fn()
  const removeSelectedFacet = jest.fn()

  beforeEach(() => {
   mockUseQueryContext.mockReturnValue({
```

Please remove all of these unnecessary casts where you could use `jest.mocked`

packages/synapse-react-client/src/components/ColumnFilter/ColumnFilter.test.tsx
```diff
@@ -0,0 +1,125 @@
...
+    fireEvent.click(openIcon)
+  }
+
+  const ColumnFilterComponent = (
+    <ColumnFilter
+      filterColumnName="program"
+      removeSelectedFacet={removeSelectedFacet}
+      addValueToSelectedFacet={addValueToSelectedFacet}
+    />
+  )
```
nickgros, line 55: You should recreate the React element in each test. You can share props between tests if you want to keep the verbosity down, e.g.


```ts
const props = (
    filterColumnName: "program",
    removeSelectedFacet: removeSelectedFacet,
    addValueToSelectedFacet: addValueToSelectedFacet
}

```
```ts
render(<ColumnFilter {...props} />)
```

packages/synapse-react-client/src/components/ColumnFilter/ColumnFilter.test.tsx
nickgros: When you use the real react-query, you'll need to wrap the component under test in a QueryContext. We created a context wrapper for all tests, so when you call `render`,  you can just do

```tsx
render(<ColumnFilter {...props} />, { wrapper: createWrapper() })
```

`createWrapper` will come from TestingLibraryUtils.tsx

packages/synapse-react-client/src/components/ColumnFilter/ColumnFilter.tsx
```diff
@@ -0,0 +1,81 @@
...
+    getCurrentQueryRequest,
+    addValueToSelectedFacet,
+    removeSelectedFacet,
+  } = queryContext
+  const { data: queryMetadata } = useQuery(queryMetadataQueryOptions)
+  const { filterColumnName } = props
+
+  // const addValueToSelectedFacet = queryContext.addValueToSelectedFacet
+  // const addValueToSelectedFacet = queryContext.removeSelectedFacet
+
```
nickgros, line 34:
```suggestion

```

packages/synapse-react-client/src/components/ColumnFilter/ColumnFilter.tsx
```diff
@@ -0,0 +1,81 @@
...
+import { getCorrespondingSelectedFacet } from '../../utils/functions/queryUtils'
+
+export type FilterProps = {
+  filterColumnName?: string
+  addValueToSelectedFacet?: (
+    facet: { columnName: string },
+    value: string,
+  ) => void
+  removeSelectedFacet?: (facet: { columnName: string }) => void
+}
```
nickgros, line 18: `filterColumnName` will always be defined.

Also, remove the other unused props for the functions that are coming from context

```suggestion
export type FilterProps = {
  filterColumnName: string
}
```

packages/synapse-react-client/src/components/ColumnFilter/ColumnFilter.tsx
```diff
@@ -0,0 +1,81 @@
...
+
+  const activeFacet = (queryMetadata?.facets ?? []).find(
+    facet => facet.columnName === `${filterColumnName}`,
+  ) as FacetColumnResultValues | undefined
+
+  // Map the facet values to format for MUI Autocomplete options
+  const filterOptions = activeFacet?.facetValues.map(({ value }) => value) ?? []
+
+  const selectedFacetFromQuery = getCorrespondingSelectedFacet(
+    { columnName: filterColumnName || '' },
```
nickgros, line 45: `filterColumnName` is always defined (after you update the prop type)
```suggestion
    { columnName: filterColumnName },
```

packages/synapse-react-client/src/components/ColumnFilter/ColumnFilter.tsx
```diff
@@ -0,0 +1,81 @@
...
+    currentQuery.query.selectedFacets,
+  ) as FacetColumnValuesRequest | undefined
+
+  const selectedValues = selectedFacetFromQuery?.facetValues || []
+
+  const onChange = (
+    event: React.SyntheticEvent,
+    values: string[], // New value for the selected options
+  ) => {
+    const facetIdentifier = { columnName: filterColumnName || '' }
```
nickgros, line 55: `filterColumnName` is always defined (after you update the prop type)
```suggestion
    const facetIdentifier = { columnName: filterColumnName }
```

packages/synapse-react-client/src/components/ColumnFilter/ColumnFilter.tsx
```diff
@@ -0,0 +1,81 @@
...
+
+    removeSelectedFacet(facetIdentifier)
+
+    values.forEach(value => addValueToSelectedFacet(facetIdentifier, value))
+
+    executeQueryRequest(request => request)
+  }
+
+  return (
+    <Box data-testid="column-filter">
```
nickgros, line 65: If you aren't using MUI-specific styling logic, `div` is fine. The testid is unnecessary after you make suggested text changes

```suggestion
    <div>
```

packages/synapse-react-client/src/components/ColumnFilter/ColumnFilter.tsx
```diff
@@ -0,0 +1,81 @@
...
+          multiple
+          options={filterOptions}
+          value={selectedValues}
+          onChange={onChange}
+          renderInput={params => (
+            <TextField {...params} placeholder="Select Filter(s)" />
+          )}
+        />
+      )}
+    </Box>
```
nickgros, line 77:
```suggestion
    </div>
```

packages/synapse-react-client/src/components/TimelinePlot/TimelinePhase.tsx
```diff
@@ -201,11 +203,11 @@ const TimelinePhase = ({
   schema,
   widthPx,
 }: TimelinePhaseProps) => {
+  const observationsSql = 'SELECT * FROM syn51735464'
```
nickgros, line 206: This should not be hard-coded, it should be a prop that ultimately comes from `observationsSql` in the NF portal resources.ts

packages/synapse-react-client/src/components/ColumnFilter/ColumnFilter.tsx
```diff
@@ -0,0 +1,73 @@
...
+  FacetColumnResultValues,
+  FacetColumnValuesRequest,
+} from '@sage-bionetworks/synapse-types'
+import { useQuery } from '@tanstack/react-query'
+import { Autocomplete, TextField } from '@mui/material'
+import { getCorrespondingSelectedFacet } from '../../utils/functions/queryUtils'
+
+export type FilterProps = {
+  filterColumnName: string
+}
```
nickgros, line 13: This slipped my mind before, but since this is a facet column, can we update this:

```suggestion
export type FilterProps = {
  topLevelEnumeratedFacetToFilter: UniqueFacetIdentifier
}
```

Not 100% sure about `topLevelEnumeratedFacetToFilter` but I want to make it clear that this must be an enumerated facet, and providing the value adds the 'top-level filter' to the UI

packages/synapse-react-client/src/components/ColumnFilter/ColumnFilter.tsx
```diff
@@ -0,0 +1,73 @@
...
+    removeSelectedFacet,
+  } = queryContext
+  const { data: queryMetadata } = useQuery(queryMetadataQueryOptions)
+  const { filterColumnName } = props
+
+  const currentQuery = getCurrentQueryRequest()
+
+  const activeFacet = (queryMetadata?.facets ?? []).find(
+    facet => facet.columnName === `${filterColumnName}`,
+  ) as FacetColumnResultValues | undefined
```
nickgros, line 31:
```suggestion
  const activeFacet = (queryMetadata?.facets ?? []).find(
    facet => facetObjectMatchesDefinition(facet, topLevelEnumeratedFacetToFilter)
  ) as FacetColumnResultValues | undefined
```

packages/synapse-react-client/src/components/ColumnFilter/ColumnFilter.tsx
```diff
@@ -0,0 +1,79 @@
...
+  ) => {
+    const facetIdentifier = {
+      columnName: topLevelEnumeratedFacetToFilter.columnName,
+    }
+
+    removeSelectedFacet(facetIdentifier)
+
+    values.forEach(value => addValueToSelectedFacet(facetIdentifier, value))
+
+    executeQueryRequest(request => request)
```
nickgros, line 59: 1. Since we changed the prop to be a UniqueFacetIdentifier, you can just pass it in here
2. You shouldn't need to call `executeQueryRequest` after calling `removeSelectedFacet`/`addValueToSelectedFacet`

```suggestion
    removeSelectedFacet(topLevelEnumeratedFacetToFilter)

    values.forEach(value => addValueToSelectedFacet(topLevelEnumeratedFacetToFilter, value))
```

packages/synapse-react-client/src/components/ColumnFilter/ColumnFilter.tsx
```diff
@@ -0,0 +1,79 @@
...
+
+  const activeFacet = (queryMetadata?.facets ?? []).find(facet =>
+    facetObjectMatchesDefinition(facet, topLevelEnumeratedFacetToFilter),
+  ) as FacetColumnResultValues | undefined
+
+  // Map the facet values to format for MUI Autocomplete options
+  const filterOptions = activeFacet?.facetValues.map(({ value }) => value) ?? []
+
+  const selectedFacetFromQuery = getCorrespondingSelectedFacet(
+    { columnName: topLevelEnumeratedFacetToFilter.columnName },
```
nickgros, line 41: You can just pass this since it is a facet identifier
```suggestion
    topLevelEnumeratedFacetToFilter,
```

nickgros (changes requested):

nickgros (approved):
fantastic work!

PR #1327, opened 2024-10-29
PORTALS-3270: Tool detail page spacing between lines is too large

kianamcc:
Jira Ticket: https://sagebionetworks.jira.com/browse/PORTALS-3270?atlOrigin=eyJpIjoiODI4OTQwMzhjYmNhNDZkNGE5NzQzODg1NmE2NDA0ZGYiLCJwIjoiaiJ9


packages/synapse-react-client/src/components/SubsectionRowRenderer/SubsectionRowRenderer.tsx
```diff
@@ -121,15 +121,15 @@ const SubsectionRowRenderer: React.FunctionComponent<
...
+              key={`header-${colIndex}`}
               className="SubsectionRowRenderer__item"
               role="table"
             >
               <Typography
                 variant={
                   columnNameIsSectionTitle ? 'sectionTitle' : 'subsectionHeader'
                 }
-                role="heading"
+                role="subHeading"
```
jay-hodgson, line 132: Is "subHeading" a valid ARIA role?  Heading is... https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/heading_role
For the spacing change, would it be appropriate to define sx properties in this Typography component instead of relying on the css?

kianamcc: @jay-hodgson subHeading is indeed not a valid ARIA role. I just reverted it back to "heading". If I use sx I have to use !important to override the global styles but if I use inline css, I can override it!

jay-hodgson (approved):

PR #1329, opened 2024-10-29
PORTALS-3299: only respect maxAge if we did not just log in

jay-hodgson:


nickgros (approved):

apps/synapse-oauth-signin/src/AppInitializer.tsx
```diff
@@ -12,18 +12,17 @@ const cookies = new UniversalCookies()
...
-  const prompt =
-    accountSitePrompted ? 'none' : urlSearchParams.get('prompt')
+  const prompt = accountSitePrompted ? 'none' : urlSearchParams.get('prompt')
 
   let maxAge = undefined
   // check max age when re-establishing the session, not to auto-consent.
   const maxAgeURLParam = urlSearchParams.get('max_age')
   // SWC-5597: if max_age is defined, then return if the user last authenticated more than max_age seconds ago
-  if (maxAgeURLParam && parseInt(maxAgeURLParam)) {
+  if (!accountSitePrompted && maxAgeURLParam && parseInt(maxAgeURLParam)) {
```
jay-hodgson, line 25: This is the only real change (the other changes are formatting, which are auto-formatted during commit so I'm not sure why they differ)

PR #1331, opened 2024-10-30
PORTALS-3299

jay-hodgson:


apps/synapse-oauth-signin/src/OAuth2Form.tsx
```diff
@@ -197,6 +199,9 @@ export function OAuth2Form() {
         // done!  redirect with access code.
         setShowPendingRedirectUI(true)
         const redirectUri = queryParams.get('redirect_uri')!
+        cookies.remove(
+          SynapseConstants.ACCOUNT_SITE_PROMPTED_FOR_LOGIN_COOKIE_KEY,
```
jay-hodgson, line 203: Cleanup.  Just before redirecting, also remove the "account site prompted" cookie state

packages/synapse-react-client/src/utils/AppUtils/AppUtils.ts
```diff
@@ -14,20 +14,20 @@ const cookies = new UniversalCookies()
 export const ONE_SAGE_REDIRECT_COOKIE_KEY =
   'org.sagebionetworks.cookies.redirect-after-login'
 
-export function storeRedirectURLForOneSageLoginAndGotoURL(href: string) {
-  // save current URL in a cookie that One Sage will use to send you back to the correct page
-  const domainValue = window.location.hostname
-    .toLowerCase()
-    .endsWith('.synapse.org')
+const getCookieDomain = () => {
```
jay-hodgson, line 17: pulled out getCookieDomain logic for reuse

packages/synapse-react-client/src/utils/AppUtils/AppUtils.ts
```diff
@@ -39,17 +39,24 @@ export function processRedirectURLInOneSage() {
...
-    domain: hostname.endsWith('.synapse.org') ? 'synapse.org' : undefined,
+    domain: getCookieDomain(),
   })
 
   if (cookies.get(ONE_SAGE_REDIRECT_COOKIE_KEY)) {
     const href = cookies.get(ONE_SAGE_REDIRECT_COOKIE_KEY)
-    cookies.remove(ONE_SAGE_REDIRECT_COOKIE_KEY)
-    window.location.assign(href)
+    // instead of removing, set the expiration to 10 seconds to avoid race condition with SageAccountWeb LoggedInRedirector
+    const tenSecondsFromNow = new Date()
```
jay-hodgson, line 51: This was a tricky bug, and it may only be happening in react strict mode (double render), but the LoggedInRedirector was calling the window.location.replace() (with the Source Config app URL) in parallel to this method calling window.location.assign() (with the cookie value that contains the One Sage redirect URL).

packages/synapse-react-client/src/utils/AppUtils/session/ApplicationSessionManager.tsx
```diff
@@ -91,7 +91,7 @@ export function ApplicationSessionManager(
       initAnonymousUserState()
       return
     }
-    if (maxAge) {
+    if (maxAge !== undefined && !!token) {
```
jay-hodgson, line 94: To be safe, let's only use the maxAge parameter to force logout iff maxAge is defined and we are logged in

packages/synapse-react-client/src/utils/AppUtils/session/ApplicationSessionManager.tsx
```diff
@@ -101,8 +101,8 @@ export function ApplicationSessionManager(
       )
       const now = dayjs.utc()
       if (now.diff(lastAuthenticatedOn, 'seconds') > maxAge) {
-        // Don't set the token so the user must re-authenticate to use this app
-        setHasInitializedSession(true)
+        // Invalidate the token (if present) so the user must re-authenticate to use this app
+        initAnonymousUserState()
```
jay-hodgson, line 105: explicitly signout the user (invalidate the token), set the token to undefined, and set hasInitializedSession to true (all done in initAnonymousUserState())

packages/synapse-react-client/src/utils/AppUtils/AppUtils.ts
jay-hodgson: Before this change, the workflow would usually (in my environment) send me to Synapse.org (the source app URL) rather than the URL stored in the cookie.  
The request to get the correct ONE_SAGE_REDIRECT_COOKIE_KEY value was in the network activity, but reported (in Chrome) as "cancelled".

apps/synapse-portal-framework/src/components/RedirectWithQuery.tsx
```diff
@@ -9,7 +9,6 @@ import { Redirect, RedirectProps, useLocation } from 'react-router-dom'
 export default function RedirectWithQuery(props: RedirectProps) {
   const { search } = useLocation()
   const hash = window.location.hash
-
```
jay-hodgson, line 12: I had a lot of debugger statements ;)

packages/synapse-react-client/src/utils/AppUtils/session/ApplicationSession.test.tsx
```diff
@@ -371,8 +371,8 @@ describe('ApplicationSessionManager tests', () => {
     await waitFor(() => {
       expect(mockGetAccessToken).toHaveBeenCalled()
       expect(mockAuthenticatedOn).toHaveBeenCalled()
+      expect(signOutSpy).toHaveBeenCalled()
```
jay-hodgson, line 374: maxAge violated, we should sign out!

packages/synapse-react-client/src/utils/AppUtils/AppUtils.ts
nickgros: Should we have a unique appId for prod OAuth signin?

nickgros: What do you think of having a dedicated route on OneSage (`/oauth` ?) that we redirect to from the OAuth sign in app, that doesn't use the LoginRedirector?

jay-hodgson: It's a good idea.  Would we also need to make updates to handle SSO (Google and ORCiD) login (update the supported redirect endpoints in those apps)?

jay-hodgson: Wait, the LoggedInRedirector is the one responsible for processing the redirect cookie (or falling back to redirecting based on the Source App redirect URL).  I suppose we could write a new LoggedInRedirector that never attempts to use the appContext.redirectURL.  Is this what you were envisioning?

nickgros: A different component similar to LoggedInRedirector was what I had in mind. I think you might be right about the SSO issues--but maybe that's not a big deal.

Another idea is to add appIds for the production oauth signin app (+ one for each environment) and treat it like any other app--would that work with the OAuth search parameters?

jay-hodgson: Search parameters are currently not stored/restored on app redirect

nickgros (approved):

PR #1335, opened 2024-10-30
PORTALS-3299: prompt=login should not call clearSession, clearSession…

jay-hodgson:
… will also reload the page to get rid of any sensitive data


packages/synapse-react-client/src/utils/AppUtils/session/ApplicationSessionManager.tsx
```diff
@@ -142,11 +139,7 @@ export function ApplicationSessionManager(
 
   /** Call refreshSession on mount */
   useEffect(() => {
-    if (forceRelogin) {
-      clearSession()
```
jay-hodgson, line 146: We just never reached this code because we didn't wait for the app to be initialized before redirecting to One Sage.  Now we are hitting this code in the OAuth sign in app, which causes the page to reload (infinite).  Idea of this PR is to utilize maxAge in order to force re-login.

nickgros (commented):
Nice catch!

PR #1353, opened 2024-11-04
SWC-7064 - Utilities for pre-upload

nickgros:


packages/synapse-react-client/src/utils/hooks/useUploadFileEntity/getFileEntityIdWithSameName.ts
```diff
@@ -0,0 +1,54 @@
...
+/**
+ * Gets the ID of the file entity with the given name whose parent has the given ID.
+ *
+ * @param name The name of the entity to find.
+ * @param parentId The ID of the parent that the found entity must have.
+ * @return The ID of the file entity with the given name and parent ID, or null if the entity does not exist
+ * @throws Error If an entity with given name and parent ID was found, but that entity
+ *         was not a File Entity.
+ */
+export async function getFileEntityIdWithSameName(
```
nickgros, line 17: In the uploader, this function will tell us if we should ask the user if a file should be uploaded as a new version of an existing entity.

packages/synapse-react-client/src/utils/hooks/useUploadFileEntity/useCreateFolderPath.ts
```diff
@@ -0,0 +1,68 @@
...
+import {
+  convertToEntityType,
+  entityTypeToFriendlyName,
+} from '../../functions/EntityTypeUtils'
+
+/**
+ * Given the ID of a container entity and a path string array, create the folder structure as necessary and return the ID of the final folder in the path.
+ * @returns The ID of the final folder in the path.
+ */
+export function useCreateFolderPath() {
```
nickgros, line 14: When uploading an entire folder, we may need to create many nested folders. This hook handles creating those folders.

jay-hodgson (approved):

PR #1354, opened 2024-11-05
SWC-7064 - Hook to prepare uploading a list of files

nickgros:


packages/synapse-react-client/src/testutils/ReactQueryMockUtils.ts
```diff
@@ -119,5 +117,5 @@ export function getUseMutationMock<TData, TError, TVariables>(
     failureReason: null,
     isPending: false,
     submittedAt: 0,
-  }
+  } satisfies UseMutationResult<TData, TError, TVariables>
```
nickgros, line 120: Using `satisfies` here gives us the type-checking we want while preserving the actual type.

The utility of this is that callers will now see that `mutate`/`mutateAsync`/`reset` properties are Jest mock functions, and can be interacted with as such without a type cast

packages/synapse-react-client/src/utils/hooks/useUploadFileEntity/usePrepareFileEntityUpload.ts
```diff
@@ -0,0 +1,113 @@
...
+ * 1. Create the necessary Folder entities in which the files to be uploaded (e.g. if the user uploaded a folder)
+ * 2. Check if any of the files to be uploaded already exist in the target parent folder
+ *
+ * The mutation will return two lists of files and their destination parentIds:
+ *    - newFileEntities: New files that do not have corresponding file entities in the target parent folder
+ *    - updatedFileEntities: Files that have corresponding file entities in the target parent folder. The user
+ *        should be prompted to accept or reject the creation of a new version of the file.
+ *
+ * In the future, this sequence could be amended to check if the storage location has enough space to accommodate all the
+ * new files, and return an error if not.
```
nickgros, line 32: Holding off on making this check for now while I try to achieve feature parity. I think it makes sense to put this check here as the first step.

packages/synapse-react-client/src/utils/hooks/useUploadFileEntity/usePrepareFileEntityUpload.ts
```diff
@@ -0,0 +1,113 @@
...
+
+      const newFileEntities = filesPreparedForUpload.filter(
+        f => f.existingEntityId == null,
+      )
+
+      const updatedFileEntities = filesPreparedForUpload.filter(
+        f => f.existingEntityId != null,
+      )
+
+      return { newFileEntities, updatedFileEntities }
```
nickgros, line 110: This wasn't captured in any design, but I'm slightly changing how we do this in SWC.

In SWC, each file is handled serially, so the process is something like (pseudocode)

```
for file in files:
  createDirectories(file)
  const fileEntityExists = checkIfFileEntityExists(file)
  if fileEntityExists:
    promptUserToUpdateFile()
  upload(file)
```

Where now it will be something like
```
const userPrompts = []
for file in files:
  createDirectories(file)
  const fileEntityExists = checkIfFileEntityExists(file)
  if fileEntityExists:
    userPrompts.push(file)

promptUserToUpdateFiles(userPrompts)
upload(files)
```

When we prompt the user "Do you want to create a new version of file xyz?" we can give the options "Yes", "Yes to all", "No", "Cancel uploads", so their upload won't be interrupted unless there's an error.

packages/synapse-react-client/src/utils/hooks/useUploadFileEntity/usePrepareFileEntityUpload.ts
```diff
@@ -0,0 +1,113 @@
...
+ *    - newFileEntities: New files that do not have corresponding file entities in the target parent folder
+ *    - updatedFileEntities: Files that have corresponding file entities in the target parent folder. The user
+ *        should be prompted to accept or reject the creation of a new version of the file.
+ *
+ * In the future, this sequence could be amended to check if the storage location has enough space to accommodate all the
+ * new files, and return an error if not.
+ *
+ * @param options
+ */
+export function usePrepareFileEntityUpload(
```
nickgros, line 36: This wasn't captured in any design, but I'm slightly changing how we do this in SWC.

In SWC, each file is handled serially, so the process is something like (pseudocode)

```
for file in files:
  createDirectories(file)
  const fileEntityExists = checkIfFileEntityExists(file)
  if fileEntityExists:
    promptUserToUpdateFile()
  upload(file)
```

Where now it will be something like
```
const userPrompts = []
for file in files:
  createDirectories(file)
  const fileEntityExists = checkIfFileEntityExists(file)
  if fileEntityExists:
    userPrompts.push(file)

promptUserToUpdateFiles(userPrompts)
upload(files)
```

When we prompt the user "Do you want to create a new version of file xyz?" we can give the options "Yes", "Yes to all", "No", "Cancel uploads", so their upload won't be interrupted unless there's an error, and no uploads start until they have gone through each of these prompts.

WDYT about this? I think in the happy-path scenario this is much more user-friendly. A major downside to this approach is that a bunch of directories might be created unnecessarily

packages/synapse-react-client/src/utils/hooks/useUploadFileEntity/usePrepareFileEntityUpload.test.ts
```diff
@@ -0,0 +1,171 @@
...
+    const useCreatePathsAndGetParentIdMockResult = getUseMutationMock(null)
+    useCreatePathsAndGetParentIdMockResult.mutateAsync.mockImplementation(
+      ({ file }) => ({ file, parentId }),
+    )
+    mockUseCreatePathsAndGetParentId.mockReturnValue(
+      useCreatePathsAndGetParentIdMockResult,
+    )
+
+    mockGetFileEntityIdWithSameName.mockImplementation(fileName => {
+      if (fileName === 'newFile.txt') {
```
jay-hodgson, line 45: Prefer logic to be reversed - look for "existingFile.txt" and return the updatedEntityId, but by default it's a new file

packages/synapse-react-client/src/utils/hooks/useUploadFileEntity/usePrepareFileEntityUpload.ts
jay-hodgson: There have been many requests to move away from serially processing, so that users do not have to babysit the upload (with this new approach, the user has the option to say "Yes to all" and walk away!).  I think the side effect creating the full directory tree up front is acceptable (given the user has not spent time uploading actual file content, which is the bulk of the operation, and they can always move the entire tree into the trash can if they don't want it).

jay-hodgson (approved):

PR #1355, opened 2024-11-05
SWC-7064 - Hook to manage state of multiple file uploads

nickgros:


packages/synapse-react-client/src/utils/hooks/useUploadFileEntity/useTrackFileUploads.ts
```diff
@@ -0,0 +1,182 @@
...
+      entry.abortController.abort()
+    }
+    setStatus(file, 'CANCELED_BY_USER')
+  }
+
+  function pauseUpload(file: File) {
+    const entry = trackedUploadProgress.get(file)
+    if (entry != null) {
+      entry.abortController.abort()
+    }
```
nickgros, line 130: Because multipart uploads can be resumed, we can just cancel the upload with an AbortController (which will passed to the upload method by a hook that uses this one). 

The only difference between an upload that is PAUSED and one that is CANCELED_BY_USER is the UI.

packages/synapse-react-client/src/utils/hooks/useUploadFileEntity/useTrackFileUploads.ts
```diff
@@ -0,0 +1,182 @@
...
+    setStatus(file, 'PAUSED')
+  }
+
+  function removeUpload(file: File) {
```
jay-hodgson, line 134: Could use a comment about behavior (no-op if file status isn't cancelled or failed)

jay-hodgson (approved):

packages/synapse-react-client/src/utils/hooks/useUploadFileEntity/useTrackFileUploads.ts
nickgros, line 134:
```suggestion
  /**
   * Removes the file from the list if the upload state is CANCELED_BY_USER or FAILED. Otherwise, a noop.
   */
  function removeUpload(file: File) {
```

PR #1362, opened 2024-11-08
SWC-7064 - Hook to upload one or more files to Synapse / S3

nickgros:
- Add `useUploadFileEntities`, which uploads and tracks the entire process of uploading a file to a Synapse-managed bucket or directly uploading to an S3 bucket over HTTP
- Add `useSynapseMultipartUpload`, a simple `useMutation` wrapper over our existing upload logic
- Update `SynapseClient` upload logic to accept an AbortController in place of a callback to cancel the upload. Callback argument preserved to avoid a breaking change.
- Update `SynapseClient.calculateMd5` to be memoized based on the Blob. Also simplified internal code without logical changes for modern browsers.
- Add `useConfirmItems`, a utility hook that tracks a set of items in state that the user must individually or collectively confirm.
- Updated React, ReactDOM types and @testing-library packages to fix lint warnings where `act` had type `any`
- Added `p-limit` package to enable limiting concurrent promises (such as the number of concurrent file uploads)


packages/synapse-react-client/src/synapse-client/SynapseClient.ts
```diff
@@ -2074,6 +2075,8 @@ export const uploadFile = (
   contentType: string = file.type,
   progressCallback?: (progress: ProgressCallback) => void,
   getIsCancelled?: () => boolean,
+  onMd5Computed?: () => void,
```
nickgros, line 2078: Add `onMD5Computed` so we can tell the user when a file is finished 'processing' and starts being uploaded

packages/synapse-react-client/src/synapse-client/SynapseClient.ts
```diff
@@ -2074,6 +2075,8 @@ export const uploadFile = (
...
+  abortController?: AbortController,
```
nickgros, line 2079: We have the `getIsCancelled` callback, but it's kind of an awkward pattern compared to just passing in an AbortController. I added an ability to pass one but kept `getIsCancelled` to avoid a breaking change.

packages/synapse-react-client/src/synapse-client/SynapseClient.ts
```diff
@@ -2099,54 +2105,57 @@ export const uploadFile = (
...
       reject(fileReader.error)
     }
 
     const loadNext = () => {
       const start = currentChunk * chunkSize,
-        end = start + chunkSize >= file.size ? file.size : start + chunkSize
+        end = start + chunkSize >= blob.size ? blob.size : start + chunkSize
 
-      fileReader.readAsArrayBuffer(blobSlice.call(file, start, end))
+      fileReader.readAsArrayBuffer(blob.slice(start, end))
```
nickgros, line 2152: I think this code sample was originally taken from https://github.com/satazor/js-spark-md5?tab=readme-ov-file#hash-a-file-incrementally

In that example, `blobSlice.call` is used because `blobSlice` was assigned to work with different vendors' prefixed functions (`slice`/`mozSlice`/`webkitSlice`). We don't have that logic here, so since we're always using `Blob.prototype.slice`, we can just call `slice` on the instance.

packages/synapse-react-client/src/synapse-client/SynapseClient.ts
```diff
@@ -2099,54 +2105,57 @@ export const uploadFile = (
...
     }
     loadNext()
   })
-}
+})
 
-const processFilePart = (
+const processFilePart = async (
```
nickgros, line 2158: Rewritten to use async/await instead of Promise.then. No logical changes other than passing `getIsCancelled` (missing) and `AbortController` (new) through to the recursive call on retry.

packages/synapse-react-client/src/synapse-client/SynapseClient.ts
```diff
@@ -2172,73 +2182,73 @@ const processFilePart = (
...
-              multipartUploadStatus,
-              clientSidePartsState,
-              accessToken,
-              fileName,
-              file,
-              request,
-              fileUploadResolve,
-              fileUploadReject,
-              updateProgress,
-            )
```
nickgros, line 2235: note we were missing
```ts
      getIsCancelled,
      abortController,
```

packages/synapse-react-client/jest.config.cjs
```diff
@@ -4,6 +4,8 @@ const esModules = [
   'lodash-es',
   'nanoid',
   'mui-one-time-password-input',
+  'p-limit',
+  'yocto-queue',
```
nickgros, line 8: p-limit is a new dependency. yocto-queue is an esm-only dependency pulled in by yocto-queue

packages/synapse-react-client/src/utils/hooks/useConfirmItems.ts
nickgros, line 1: Utility hook to track a list of items that the user must confirm. In this case, it will be used to track which file uploads the user must confirm to be new versions of existing entities.

packages/synapse-react-client/src/utils/hooks/useUploadFileEntity/useUploadFileEntities.ts
```diff
@@ -0,0 +1,423 @@
...
+  isPrecheckingUpload: boolean
+  activePrompts: Prompt[]
+  initiateUpload: (files: File[]) => void
+  activeUploadCount: number
+  uploadProgress: FileUploadProgress[]
+}
+
+// Limit the number of concurrent uploads to avoid overwhelming the browser
+// Note that within one upload operation, multiple parts could be uploaded in parallel
+const limitConcurrentUploads = pLimit(8)
```
nickgros, line 78: For large set of files, this should limit the number of concurrently computed MD5s (this slowed down my browser with hundreds of files), and prioritize completing file uploads before starting new ones, which IMO is a better user experience. At the same time, I'm trying to avoid adding an artificial throttle.

This is totally unsophisticated, and we may need to tweak this and/or add other limits later.

packages/synapse-react-client/src/utils/hooks/useUploadFileEntity/useUploadFileEntities.ts
```diff
@@ -0,0 +1,423 @@
...
+    isUploadComplete,
+    isUploading,
+  ])
+
+  const { mutateAsync: createEntityWithNewFile } = useCreateEntity()
+  const { mutateAsync: updateEntityWithNewFile } = useUpdateEntity()
+  const { mutateAsync: uploadFile } = useSynapseMultipartUpload()
+  const { mutateAsync: uploadFileDirectlyToS3 } = useDirectUploadToS3()
+
+  const uploadPreparedFile = useCallback(
```
nickgros, line 157: General flow for how the functions in this hook are used is 

1. `initiateUpload(files)` - called by the caller of the hook (e.g. the new upload component)
2. `prepareUpload(files)` - sets up the folder paths, checks for existing entities with the same name
3. `postPrepareUpload(files)` - may prompt the user to confirm uploading new versions of files
4. `startUpload(files)` - starts tracking file uploads, for each file, calls `uploadPreparedFile`
5. `uploadPreparedFile(file)` - uploads the file and creates/updates the entity

packages/synapse-react-client/src/utils/hooks/useUploadFileEntity/useUploadFileEntities.ts
```diff
@@ -0,0 +1,423 @@
...
+    isLoadingUploadDestination,
+    isUploadComplete,
+    isUploading,
+  ])
+
+  const { mutateAsync: createEntityWithNewFile } = useCreateEntity()
+  const { mutateAsync: updateEntityWithNewFile } = useUpdateEntity()
+  const { mutateAsync: uploadFile } = useSynapseMultipartUpload()
+  const { mutateAsync: uploadFileDirectlyToS3 } = useDirectUploadToS3()
+
```
nickgros, line 156: General flow for how the functions in this hook are used is 

1. `initiateUpload(files)` - called by the caller of the hook (e.g. the new upload component)
2. `prepareUpload(files)` - sets up the folder paths, checks for existing entities with the same name
3. `postPrepareUpload(files)` - may prompt the user to confirm uploading new versions of files
4. `startUpload(files)` - starts tracking file uploads, for each file, calls `uploadPreparedFile`
5. `uploadPreparedFile(file)` - uploads the file and creates/updates the entity

packages/synapse-react-client/src/utils/hooks/useConfirmItems.test.ts
```diff
@@ -0,0 +1,78 @@
+import { renderHook, act } from '@testing-library/react'
+import useConfirmItems from './useConfirmItems'
+
+describe('useConfirmItems', () => {
```
nickgros, line 4: fyi, the tests in this suite were ~90% written by GitHub Copilot, with minor fixes

packages/synapse-react-client/src/utils/hooks/useUploadFileEntity/useUploadFileEntities.ts
```diff
@@ -0,0 +1,423 @@
...
+        },
+        pause: () => {
+          pauseUpload(file)
+        },
+        resume: () => {
+          prepareUpload(
+            { files: [file], parentId: progress.parentId },
+            {
+              onSuccess: result => {
+                postPrepareUpload(result, true)
```
nickgros, line 385:
```suggestion
                // the file was already checked before it was paused, so we can skipChecks
                postPrepareUpload(result, true)
```

packages/synapse-react-client/package.json
```diff
@@ -214,6 +214,7 @@
     "memfs": "^3.5.3",
     "msw": "^1.3.2",
     "msw-storybook-addon": "^1.10.0",
+    "p-limit": "^6.1.0",
```
jay-hodgson, line 217: Good idea!  This popular MIT licensed package seems like a great pick

packages/synapse-react-client/src/synapse-client/SynapseClient.ts
```diff
@@ -2277,14 +2289,15 @@ const uploadFilePart = async (
...
 ) => {
   const controller = new AbortController()
-  const signal = controller.signal
+  const signal = abortController?.signal || controller.signal
 
   const checkIsCancelled = () => {
     if (getIsCancelled) {
       const isCancelled = getIsCancelled()
-      if (isCancelled) {
+      if (isCancelled && !controller.signal.aborted) {
```
jay-hodgson, line 2300: should this be `!signal.aborted` rather than `!controller.signal.aborted`?

packages/synapse-react-client/src/synapse-client/SynapseClient.ts
```diff
@@ -2277,14 +2289,15 @@ const uploadFilePart = async (
...
         controller.abort()
```
jay-hodgson, line 2301: Similarly, should this use abortController || controller ?

packages/synapse-react-client/src/utils/hooks/useConfirmItems.test.ts
jay-hodgson: I asked ChatGPT review, and it said it looks ok ;)

packages/synapse-react-client/src/synapse-client/SynapseClient.ts
nickgros: Yes, nice catch. Thanks!

packages/synapse-react-client/src/utils/hooks/useUploadFileEntity/useUploadFileEntities.ts
jay-hodgson: This overview comment could be really useful in the code (comment for useUploadFileEntities).  So cool to see it all come together like this.

packages/synapse-react-client/src/utils/hooks/useUploadFileEntity/useUploadFileEntities.ts
```diff
@@ -0,0 +1,434 @@
...
+   * Upload the list of prepared files to Synapse.
+   */
+  const startUpload = useCallback(
+    (...preparedFiles: FilePreparedForUpload[]) => {
+      clearPendingFiles()
+
+      const newTrackedUploadProgress = trackNewFiles(...preparedFiles)
+
+      // TODO: We already got the storage location from the root, but what if one of the sub-folders already existed
+      //   with a different storage location? should we use that storage location?
```
jay-hodgson, line 271: Good question.  I think it's understandable if the web client does not have this sophisticated logic at this time (to recognize new folder storage locations during the upload of subfolder files)

jay-hodgson (approved):

PR #1363, opened 2024-11-09
PORTALS-3271: Observation cards design tweaks

kianamcc:
Jira Ticket: https://sagebionetworks.jira.com/browse/PORTALS-3271?atlOrigin=eyJpIjoiZTViZDBmYzY5MGI1NDdhMTlhYjEwNjMyNGVlNzNiMjUiLCJwIjoiaiJ9


packages/synapse-react-client/src/components/CardContainer/CardContainer.tsx
```diff
@@ -39,6 +39,17 @@ const releaseCardMediumListSx = {
   gap: '10px',
   mb: '30px',
 }
+const multiCardListSx = {
+  display: 'grid',
+  gridTemplateColumns: 'repeat(auto-fit, minmax(300px, 1fr))',
```
jay-hodgson, line 44: This is an excellent responsive grid definition (IMO)

packages/synapse-react-client/src/components/CardContainer/CardContainer.tsx
```diff
@@ -39,6 +39,17 @@ const releaseCardMediumListSx = {
...
+  gridColumnGap: '20px',
+  marginBottom: '20px',
+  marginTop: '20px',
```
jay-hodgson, line 47: Note - In the codebase you may see the shorthand `mb` used for `marginBottom`, `mt` for `marginTop`, and `m` for `margin`

packages/synapse-react-client/src/components/CardContainer/CardContainer.tsx
```diff
@@ -158,7 +172,13 @@ function _CardContainer(props: CardContainerProps) {
     <>
       <Box
         role="list"
-        sx={isReleaseCardMediumList ? releaseCardMediumListSx : defaultListSx}
+        sx={
+          isReleaseCardMediumList
+            ? releaseCardMediumListSx
+            : multiCardList && isMultipleCards
+            ? multiCardListSx
+            : defaultListSx
```
jay-hodgson, line 180: Instead of multiple ternary operators, could you write a series of if/else statements above to figure out this value?
I realize this is more verbose, but I think it improves readability and can be easily extended.
Something like
```
let cardListSx = defaultListSx
if (isReleaseCardMediumList) {
  cardListSx = releaseCardMediumListSx
} else if (multiCardList && isMultipleCards) {
  cardListSx = multiCardListSx
}
```

jay-hodgson: @nickgros / @sagely1 , what are your thoughts/opinions?

kianamcc: I agree! Using if/else statements would make it a lot easier to extend.

jay-hodgson (approved):

PR #1365, opened 2024-11-11
fix cookie on dev domain

nickgros:
When the accounts site will be created, the cookie will be set on a domain like `accounts.dev.sagebase.org`. Manually set the domain to `.dev.sagebase.org` so it will work on `portal-dev.dev.sagebase.org`.

### DIFF
--- a/packages/synapse-react-client/src/utils/AppUtils/AppUtils.ts
+++ b/packages/synapse-react-client/src/utils/AppUtils/AppUtils.ts
@@ -15,9 +15,13 @@ export const ONE_SAGE_REDIRECT_COOKIE_KEY =
   'org.sagebionetworks.cookies.redirect-after-login'
 
 export const getCookieDomain = () => {
-  return window.location.hostname.toLowerCase().endsWith('.synapse.org')
-    ? '.synapse.org'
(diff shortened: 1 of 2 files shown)


packages/synapse-react-client/src/utils/AppUtils/AppUtils.ts
```diff
@@ -15,9 +15,13 @@ export const ONE_SAGE_REDIRECT_COOKIE_KEY =
   'org.sagebionetworks.cookies.redirect-after-login'
 
 export const getCookieDomain = () => {
-  return window.location.hostname.toLowerCase().endsWith('.synapse.org')
-    ? '.synapse.org'
-    : undefined
+  if (window.location.hostname.toLowerCase().endsWith('.synapse.org')) {
+    return '.synapse.org'
+  }
+  if (window.location.hostname.toLowerCase().endsWith('dev.sagebase.org')) {
```
nickgros, line 21:
```suggestion
  if (window.location.hostname.toLowerCase().endsWith('.dev.sagebase.org')) {
```

jay-hodgson (approved):
Nice catch

PR #1371, opened 2024-11-12
SWC-7064 - Handle case where a FileEntity is directly updated

nickgros:
- Handle case where a FileEntity is directly updated (e.g. from the entity page) and the user should not be prompted.
- Minor refactoring

### DIFF
--- a/packages/synapse-react-client/src/utils/hooks/useUploadFileEntity/useCreateFolderPath.test.ts
+++ b/packages/synapse-react-client/src/utils/hooks/useUploadFileEntity/useCreateFolderPath.test.ts
@@ -38,7 +38,7 @@ describe('useCreateFolderPath', () => {
     const { result: hook } = renderHook()
 
     const result = await hook.current.mutateAsync({
-      parentId: 'syn123',
+      rootContainerId: 'syn123',
       path: ['folder'],
     })
 
@@ -62,7 +62,7 @@ describe('useCreateFolderPath', () => {
     const { result: hook } = renderHook()
 
     const result = await hook.current.mutateAsync({
-      parentId: 'syn123',
+      rootContainerId: 'syn123',
       path: ['folder'],
     })
 
@@ -85,7 +85,7 @@ describe('useCreateFolderPath', () => {
     const { result: hook } = renderHook()
 
     const result = await hook.current.mutateAsync({
-      parentId: 'syn123',
+      rootContainerId: 'syn123',
       path: [],
     })
 
@@ -115,7 +115,7 @@ describe('useCreateFolderPath', () => {
     const { result: hook } = renderHook()
 
     const result = await hook.current.mutateAsync({
-      parentId: 'syn123',
+      rootContainerId: 'syn123',
       path: ['parentFolder', 'childFolder'],
     })
 
@@ -142,7 +142,7 @@ describe('useCreateFolderPath', () => {
 
     await expect(
       hook.current.mutateAsync({
-        parentId: 'syn123',
+        rootContainerId: 'syn123',
         path: ['some_name'],
       }),
     ).rejects.toThrow(
@@ -173,7 +173,7 @@ describe('useCreateFolderPath', () => {
     const { result: hook } = renderHook()
     await expect(
       hook.current.mutateAsync({
-        parentId: 'syn123',
+        rootContainerId: 'syn123',
         path: ['folder'],
       }),
     ).rejects.toThrow(`Forbidden`)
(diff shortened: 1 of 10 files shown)


packages/synapse-react-client/src/utils/hooks/useUploadFileEntity/useCreateFolderPath.test.ts
```diff
@@ -38,7 +38,7 @@ describe('useCreateFolderPath', () => {
     const { result: hook } = renderHook()
 
     const result = await hook.current.mutateAsync({
-      parentId: 'syn123',
+      rootContainerId: 'syn123',
```
nickgros, line 41: renamed parameter for clarity (explained later)

packages/synapse-react-client/src/utils/hooks/useUploadFileEntity/useCreatePathsAndGetParentId.ts
```diff
@@ -9,24 +9,24 @@ export function useCreatePathsAndGetParentId() {
   const { mutateAsync: createFolderPath } = useCreateFolderPath()
 
   return useMutation({
-    mutationFn: (args: { file: File; parentId: string }) => {
-      const { file, parentId } = args
+    mutationFn: (args: { file: File; rootContainerId: string }) => {
+      const { file, rootContainerId } = args
```
nickgros, line 13: This function has a return type `{ file: File; parentId: string }`, so I changed the argument type to `{ file: File; rootContainerId: string }` to make it more clear that the `parentId` may be different from the `rootContainerId`

packages/synapse-react-client/src/utils/hooks/useUploadFileEntity/usePrepareFileEntityUpload.ts
```diff
@@ -4,17 +4,24 @@ import { useSynapseContext } from '../../context/index'
 import { getFileEntityIdWithSameName } from './getFileEntityIdWithSameName'
 import { useCreatePathsAndGetParentId } from './useCreatePathsAndGetParentId'
 
-export type FilePreparedForUpload = {
-  file: File
-  parentId: string
-  existingEntityId: string | null
-}
```
nickgros, line 11: split this type up so `parentId` and `existingEntityId` are mutually exclusive.

packages/synapse-react-client/src/utils/hooks/useUploadFileEntity/usePrepareFileEntityUpload.ts
```diff
@@ -4,17 +4,24 @@ import { useSynapseContext } from '../../context/index'
...
+
+export type FilePreparedForUpload =
+  | NewEntityFileUploadReturn
+  | UpdateEntityFileUpload
 
 export type PrepareDirsForUploadReturn = {
-  newFileEntities: FilePreparedForUpload[]
-  updatedFileEntities: FilePreparedForUpload[]
+  filesReadyForUpload: FilePreparedForUpload[]
+  filesToPromptForNewVersion: FilePreparedForUpload[]
```
nickgros, line 17: Renamed these properties to more accurately capture the purpose of this hook. With the changes in this PR, the old names are now inaccurate. If a file is explicitly uploaded as a new version of an entity, then there's no need to show a prompt.

packages/synapse-react-client/src/utils/hooks/useUploadFileEntity/useUploadFileEntities.test.ts
```diff
@@ -344,9 +340,105 @@ describe('useUploadFileEntities', () => {
     expect(hook.current.state).toBe('COMPLETE')
   })
 
+  test('upload a new version of a specified FileEntity', async () => {
```
nickgros, line 343: new test to capture the explicit update case

packages/synapse-react-client/src/utils/hooks/useUploadFileEntity/useUploadFileEntities.ts
```diff
@@ -284,27 +313,20 @@ export function useUploadFileEntities(
    * decisions to continue the upload.
    */
   const postPrepareUpload = useCallback(
-    (
-      preparedFiles: PrepareDirsForUploadReturn,
-      /** If true, skip checks such as prompting the user before creating a new version.
-       * This is useful when resuming an upload that has already been pre-checked. */
-      skipChecks = false,
```
nickgros, line 291: Removed `skipChecks` parameter; refactoring has made it unnecessary.

packages/synapse-react-client/src/utils/hooks/useUploadFileEntity/useUploadFileEntities.ts
```diff
@@ -375,28 +394,20 @@ export function useUploadFileEntities(
...
-          prepareUpload(
-            { files: [file], parentId: progress.parentId },
-            {
-              onSuccess: result => {
-                // the file was already checked before it was paused, so we can skipChecks
-                postPrepareUpload(result, true)
-              },
-            },
-          )
+          startUpload(trackedProgress.filePreparedForUpload)
```
nickgros, line 410: since we now store the `filePreparedForUpload` in the TrackedUploadProgress object, we can just call `startUpload` on resume, and remove the `skipChecks` parameter from `postPrepareUpload`

PR #1372, opened 2024-11-12
SWC-7143

jay-hodgson:


packages/synapse-react-client/src/components/EntityBadgeIcons/EntityBadgeIcons.tsx
```diff
@@ -104,6 +105,11 @@ export const EntityBadgeIcons = (props: EntityBadgeIconsProps) => {
 
   const { ref, inView } = useInView()
 
+  const { data: entityData } = useGetJson(entityId, versionNumber, true, {
```
jay-hodgson, line 108: Get the derived annotations!

packages/synapse-react-client/src/components/EntityBadgeIcons/EntityBadgeIcons.tsx
```diff
@@ -104,6 +105,11 @@ export const EntityBadgeIcons = (props: EntityBadgeIconsProps) => {
...
+    enabled: inView,
+    staleTime: 60 * 1000, // 60 seconds
```
nickgros, line 110: Consider removing the `staleTime` override here and in the call to `useGetEntityBundle`. I don't think these are serving any purpose
```suggestion
```

packages/synapse-react-client/src/components/EntityBadgeIcons/EntityBadgeIcons.tsx
```diff
@@ -118,7 +124,25 @@ export const EntityBadgeIcons = (props: EntityBadgeIconsProps) => {
...
+  ) => {
+    if (typeof value === 'string') {
+      return value
+    } else if (Array.isArray(value)) {
+      return value.join(', ')
+    } else {
+      // Handle number and boolean cases if needed
+      return value?.toString()
+    }
+  }
```
nickgros, line 145: you may be able to reuse / modify `getDisplayedAnnotation` in AnnotationsTable.tsx

nickgros (approved):
nice! a few suggestions

packages/synapse-react-client/src/components/EntityBadgeIcons/EntityBadgeIcons.tsx
jay-hodgson: Reusing, which improves the rendering of datetimes.  Great idea

nickgros (approved):

PR #1377, opened 2024-11-14
SWC-7064 - Entity Upload: add Link to URL tab

nickgros:
image

### DIFF
--- a/packages/synapse-react-client/src/components/EntityUpload/LinkToURL.tsx
+++ b/packages/synapse-react-client/src/components/EntityUpload/LinkToURL.tsx
@@ -1,0 +1,93 @@
+import { Alert, Box, Stack, TextField, Typography } from '@mui/material'
+import { FileEntity } from '@sage-bionetworks/synapse-types'
+import { noop } from 'lodash-es'
+import React, { useEffect, useImperativeHandle, useState } from 'react'
+import { useGetEntity } from '../../synapse-queries/index'
+import { useLinkFileEntityToURL } from '../../utils/hooks/useUploadFileEntity/useLinkFileEntityToURL'
+import { SynapseSpinner } from '../LoadingScreen/LoadingScreen'
+
+type LinkToURLProps = {
+  /** The ID of the FileEntity that should be updated, or container entity in which a new FileEntity should be created. */
+  entityId: string
+  /** Invoked when the validity of the form changes (e.g. to disable/activate a submit button) */
+  onIsValidChanged?: (isValid: boolean) => void
+  /** Invoked when the call to create or update the entity succeeds. */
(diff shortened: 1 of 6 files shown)


packages/synapse-react-client/src/synapse-queries/entity/useEntity.ts
```diff
@@ -46,21 +47,29 @@ import { getNextPageParamForPaginatedResults } from '../InfiniteQueryUtils'
...
+    queryOptions<T, SynapseClientError>({
+      queryKey: keyFactory.getEntityVersionQueryKey(entityId, versionNumber),
+      queryFn: () =>
+        SynapseClient.getEntity<T>(
+          accessToken,
+          entityId,
+          versionNumber?.toString(),
+        ),
+    })
+}
```
nickgros, line 62: Extracted to allow inline fetch with `queryClient.fetchQuery(options)`

packages/synapse-react-client/src/utils/hooks/useUploadFileEntity/useLinkFileEntityToURL.ts
```diff
@@ -0,0 +1,106 @@
...
+      const firstQuestionMark = path.indexOf('?', lastSlash)
+      if (firstQuestionMark > -1) {
+        fileName = path.substring(lastSlash + 1, firstQuestionMark)
+      } else {
+        fileName = path.substring(lastSlash + 1)
+      }
+    }
+  }
+  return fileName
+}
```
nickgros, line 29: copied from SWC

packages/synapse-react-client/src/components/EntityUpload/LinkToURL.tsx
```diff
@@ -0,0 +1,93 @@
...
+  const { data: entity } = useGetEntity(entityId)
+  const isFileEntity =
+    entity?.concreteType === 'org.sagebionetworks.repo.model.FileEntity'
+
+  const { mutate, error } = useLinkFileEntityToURL({
+    onSuccess,
+  })
+
+  useEffect(() => {
+    const isValid = url.trim().length > 0
```
jay-hodgson, line 40: Any additional validation needed here?  I'm not sure how much sanitation we do on the rendering side for this URL value throughout the web apps.

jay-hodgson: We can't assume this is set via the web interface, so I think this is fine.

nickgros: This is the same behavior as SWC -- only check that the URL is a nonempty string and let the backend validate syntax

jay-hodgson (approved):

PR #1383, opened 2024-11-15
SWC-7064 - Project storage limit alert

nickgros:
- UI + logic for storage limit alerts
- Update types for ProjectStorage objects
- Use constant for Sage Offerings help docs links

### DIFF
--- a/packages/synapse-react-client/src/components/EntityUpload/ProjectStorageLimitAlert.tsx
+++ b/packages/synapse-react-client/src/components/EntityUpload/ProjectStorageLimitAlert.tsx
@@ -1,0 +1,83 @@
+import { ProjectStorageLocationUsage } from '@sage-bionetworks/synapse-types'
+import React from 'react'
+import { calculateFriendlyFileSize } from '../../utils/functions/calculateFriendlyFileSize'
+import { SAGE_OFFERINGS_HELP_URL } from '../../utils/SynapseConstants'
+import FullWidthAlert from '../FullWidthAlert'
+
+export type ProjectStorageLimitAlertProps = {
+  didUploadsExceedLimit: boolean
+  usage: ProjectStorageLocationUsage
+}
+
+// When the remaining storage is less than this percentage, show a warning
+const PERCENT_REMAINING_STORAGE_TO_SHOW_WARNING = 0.25
+
+export function ProjectStorageLimitAlert(props: ProjectStorageLimitAlertProps) {
+  const { usage, didUploadsExceedLimit } = props
+
+  if (usage.maxAllowedFileBytes == null) {
+    // No limit!
+    return null
+  }
+
+  const warningThresholdBytes =
+    usage.maxAllowedFileBytes * (1 - PERCENT_REMAINING_STORAGE_TO_SHOW_WARNING)
+
+  const remainingStorageBytes = usage.maxAllowedFileBytes - usage.sumFileBytes
+
+  if (!didUploadsExceedLimit && usage.sumFileBytes < warningThresholdBytes) {
+    // No need to show a warning
+    return null
+  }
+
+  if (usage.isOverLimit) {
+    return (
+      <FullWidthAlert
+        variant="danger"
+        isGlobal={false}
+        title={'You have no data availability'}
+        description={
+          'You do not have any remaining of data within this project. In order to upload more files, you must edit your file list or upgrade your plan.'
+        }
+        primaryButtonConfig={{
+          text: 'Learn more',
+          href: SAGE_OFFERINGS_HELP_URL,
+        }}
+      ></FullWidthAlert>
+    )
+  }
+
+  if (didUploadsExceedLimit) {
+    return (
+      <FullWidthAlert
+        variant="danger"
+        isGlobal={false}
+        title={'Selected files exceed available storage'}
+        description={`This project has only ${calculateFriendlyFileSize(
+          remainingStorageBytes,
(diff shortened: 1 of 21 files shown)


packages/synapse-react-client/src/components/EntityUpload/ProjectStorageLimitAlert.tsx
```diff
@@ -0,0 +1,83 @@
...
+        }
+        primaryButtonConfig={{
+          text: 'Learn more',
+          href: SAGE_OFFERINGS_HELP_URL,
+        }}
+      ></FullWidthAlert>
+    )
+  }
+
+  if (didUploadsExceedLimit) {
```
nickgros, line 50: image

packages/synapse-react-client/src/components/EntityUpload/ProjectStorageLimitAlert.tsx
```diff
@@ -0,0 +1,83 @@
...
+          remainingStorageBytes,
+          1,
+        )} of storage remaining. To continue, you must upgrade your storage, or add custom storage.`}
+        primaryButtonConfig={{
+          text: 'Learn more',
+          href: SAGE_OFFERINGS_HELP_URL,
+        }}
+      ></FullWidthAlert>
+    )
+  }
```
nickgros, line 66: image

packages/synapse-react-client/src/components/EntityUpload/ProjectStorageLimitAlert.tsx
```diff
@@ -0,0 +1,83 @@
...
+      description={`Your project has less than ${
+        PERCENT_REMAINING_STORAGE_TO_SHOW_WARNING * 100
+      }% remaining of available storage left. Upgrade your storage, or bring your own custom storage.`}
+      primaryButtonConfig={{
+        text: 'Learn more',
+        href: SAGE_OFFERINGS_HELP_URL,
+      }}
+    ></FullWidthAlert>
+  )
+}
```
nickgros, line 83: image

packages/synapse-react-client/src/components/EntityUpload/ProjectStorageLimitAlert.tsx
```diff
@@ -0,0 +1,83 @@
...
+        description={
+          'You do not have any remaining data within this project. In order to upload more files, you must edit your file list or upgrade your plan.'
+        }
+        primaryButtonConfig={{
+          text: 'Learn more',
+          href: SAGE_OFFERINGS_HELP_URL,
+        }}
+      ></FullWidthAlert>
+    )
+  }
```
nickgros, line 48: image

packages/synapse-react-client/src/utils/hooks/useUploadFileEntity/useTrackFileUploads.ts
```diff
@@ -155,6 +162,19 @@ export function useTrackFileUploads() {
...
+  const bytesPendingUpload = [...trackedUploadProgress].reduce(
+    (acc, [file, progress]) => {
+      if (UNTRACKED_BYTE_UPLOAD_STATES.includes(progress.status)) {
+        return acc + file.size
+      }
+
+      return acc
+    },
+    0,
+  )
```
nickgros, line 176: This sequence is possible:

1. You have a project with a 100GB limit, 0 files uploaded
2. You open the uploader and add a 90 GB file
3. Before (or very shortly after) that file finishes uploading, you try to add another 50GB file (this was not possible in the SWC uploader unless you opened two windows)

WDYT about this idea? Track the size of all files that have been uploaded or will be uploaded in the current set of files. We will add that to the reported `sumFileBytes` to decide whether to block uploads. The main risk is that files could be double-counted by both the server and the client, though I set the `staleTime` to infinity on the usage query to attempt to prevent that.

packages/synapse-react-client/src/utils/hooks/useUploadFileEntity/useTrackFileUploads.test.ts
```diff
@@ -330,6 +337,8 @@ describe('useTrackFileUploads', () => {
 
     expect(hook.current.activeUploadCount).toBe(0)
     expect(hook.current.isUploading).toBe(false)
+    // bytesPendingUpload should now only count completed uploads
+    expect(hook.current.bytesPendingUpload).toBe(completedFile.size)
```
nickgros, line 341:
```suggestion
    // only the completed upload remains
    expect(hook.current.bytesPendingUpload).toBe(completedFile.size)
```

packages/synapse-react-client/src/utils/hooks/useUploadFileEntity/useTrackFileUploads.ts
jay-hodgson: I think the logic that you have implemented is ok.  If files are double-counted, what is the UX?

jay-hodgson (approved):

packages/synapse-react-client/src/utils/hooks/useUploadFileEntity/useTrackFileUploads.ts
nickgros: The UI will erroneously block you from uploading any new files that push you over the value `limit - size of double counted files`. If we erroneously blocked the upload, you would be able to start the upload if you reloaded the page or opened a new window.

PR #1384, opened 2024-11-15
SWC-7064 - Optimize creating many dirs by using cache for entity lookup

nickgros:
Improves the performance of the case where a folder with many items is uploaded by using the react-query cache when looking up whether an entity exists, and fetching it to check if it's a folder.

For example, if you upload these files:

```
folder_1/folder_2/.../folder_m/file_1.txt
folder_1/folder_2/.../folder_m/file_2.txt
...
folder_1/folder_2/.../folder_m/file_n.txt
```

without a cache, we would call `POST entity/child` and `GET /entity/id` $m * n$ times.

With the cache, these will only be called $m$ times.

### DIFF
--- a/packages/synapse-react-client/src/synapse-queries/entity/useEntity.ts
+++ b/packages/synapse-react-client/src/synapse-queries/entity/useEntity.ts
@@ -2,12 +2,14 @@
  * Hooks to access Entity Services in Synapse
  */
 
+import { EntityLookupRequest } from '@sage-bionetworks/synapse-client'
 import { omit, pick } from 'lodash-es'
 import { useMemo } from 'react'
 import {
   InfiniteData,
   QueryClient,
   QueryKey,
+  queryOptions,
   useInfiniteQuery,
   UseInfiniteQueryOptions,
   useMutation,
@@ -19,6 +21,7 @@ import {
   useSuspenseQuery,
 } from '@tanstack/react-query'
 import SynapseClient from '../../synapse-client'
+import { allowNotFoundError } from '../../synapse-client/SynapseClientUtils'
 import { entityJsonKeys } from '../../utils/functions/EntityTypeUtils'
 import {
   createTableUpdateTransactionRequest,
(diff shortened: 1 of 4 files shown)


packages/synapse-react-client/src/utils/hooks/useUploadFileEntity/useCreateFolderPath.ts
```diff
@@ -12,32 +14,29 @@ import {
...
-            entityLookupRequest: {
-              parentId: parentId,
-              entityName: pathElement,
-            },
+        const foundEntityId = await queryClient.fetchQuery(
+          getEntityLookupQueryOptions({
+            parentId: parentId,
+            entityName: pathElement,
           }),
         )
```
nickgros, line 33: Fetch with cache to see if entity exists. 

If the entity does not exist, it is created and the cache is invalidated, so a stale cache issue should never happen

packages/synapse-react-client/src/utils/hooks/useUploadFileEntity/useCreateFolderPath.ts
```diff
@@ -12,32 +14,29 @@ import {
...
-
         if (foundEntityId) {
           // Verify that this entity is a folder
-          const entity =
-            await synapseClient.entityServicesClient.getRepoV1EntityId({
-              id: foundEntityId,
-            })
+          const entity = await queryClient.fetchQuery(
+            getEntityQueryOptions(foundEntityId),
+          )
```
nickgros, line 39: We only use this request to see if the entity is a folder or not, so it's safe to cache.

packages/synapse-react-client/src/synapse-queries/entity/useEntity.ts
```diff
@@ -46,21 +49,32 @@ import { getNextPageParamForPaginatedResults } from '../InfiniteQueryUtils'
 import { KeyFactory } from '../KeyFactory'
 import { useGetEntityBundleQueryOptions } from './useEntityBundle'
 
+export function useGetEntityQueryOptions<T extends Entity>() {
```
nickgros, line 52: Export a way to get the query options so we can pass it into `queryClient` imperatively, e.g. 

```js
const options = getEntityQueryOptions(args)
queryClient.fetchQuery(options)
```

packages/synapse-react-client/src/utils/hooks/useUploadFileEntity/useCreateFolderPath.ts
jay-hodgson: What would happen in the unlikely event that someone else renames an entity (which is not reflected in our local cache)?  In the case where the new name conflicts, would the process fail during File entity creation?  In the case where the new name no longer conflicts, I suppose the user would be prompted to add a new version for the existing entity?

nickgros: If someone else modifies a node in the tree before we can completely set it up, an error may happen, but I think we have that risk with or without a cache. Ideally we would have an API that would just let us set up the tree in one call.

In your specific scenario, if a folder is renamed before this check, we'll just end up creating a new folder. If a folder is renamed after this check, then we would probably end up uploading the file to the existing folder with the 'wrong' name. 

This set of tasks only sets up folders, and in the context of an upload, if this task fails, an error message is shown and uploads never start. The user can attempt to address the issue (e.g. delete an existing file that has the same name as one of the new folders) or just restart the upload, depending on the error (e.g. a folder existed, but was renamed or deleted when we tried to use it).

jay-hodgson (approved):

PR #1388, opened 2024-11-18
PORTALS-3195 - HMR for local monorepo packages

nickgros:
- Set up tsconfig paths to reference local monorepo projects when possible
- Add vite-tsconfig-paths plugin to achieve HMR without rebuilding projects
- Set up TypeScript project references where possible to speed up type-checking


apps/SageAccountWeb/tsconfig.json
```diff
@@ -9,7 +9,32 @@
...
+      "@sage-bionetworks/synapse-types/*": [
+        "../../../packages/synapse-types/src/*"
+      ],
+      "synapse-react-client": ["../../../packages/synapse-react-client/src"],
+      "synapse-react-client/*": [
+        "../../../packages/synapse-react-client/src/*"
+      ],
+      "vite-config": ["../../../packages/vite-config/src"],
+      "vite-config/*": ["../../../packages/vite-config/src/*"]
+    }
```
nickgros, line 32: In each Vite project, adding these tsconfig `paths` properties informs the TypeScript compiler to use the sources of the projects rather than the built files. We combine this with a popular Vite plugin to achieve HMR across modules. You can start the Vite dev server for the app, make a change in synapse-react-client, and instantly see the change in the app without rebuilding.

apps/SageAccountWeb/package.json
```diff
@@ -74,7 +74,7 @@
     "preview": "vite preview",
     "build": "vite build",
     "test": "vitest",
-    "type-check": "tsc --noEmit"
+    "type-check": "tsc --build"
```
nickgros, line 77: Build using [TypeScript project references](https://www.typescriptlang.org/docs/handbook/project-references.html), which provides incremental type checking

apps/synapse-portal-framework/src/components/index.ts
```diff
@@ -1,4 +1,4 @@
-import RouteControlWrapper from './RouteControlWrapper'
+import ExploreWrapper from './Explore/ExploreWrapper'
```
nickgros, line 1: Component & file were already renamed, this was just an outdated reference that caused build errors

apps/synapse-portal-framework/tsconfig.json
```diff
@@ -1,17 +1,39 @@
 {
   "extends": "../../shared/tsconfig.base.json",
   "compilerOptions": {
+    "composite": true,
```
nickgros, line 4: With TypeScript project references, projects that can be referenced by other projects must set `"composite": true`. This enforces a few other rules that can be seen in other changes in the tsconfig files.

apps/synapse-portal-framework/tsconfig.json
```diff
@@ -1,17 +1,39 @@
...
     "rootDir": ".",
     "baseUrl": ".",
+    "outDir": "../../node_modules/.cache/.tsc_cache/synapse-portal-framework",
+    "noEmit": false,
```
nickgros, line 8: With `"composite": true`, `noEmit` must be false, since dependent projects use the output files. For projects where we don't use the TypeScript compiler output as a build artifact, we can redirect the output to a cache folder in node_modules. Idea taken from [this doc from moonrepo](https://moonrepo.dev/docs/config/toolchain#routeoutdirtocache), which we don't use, but they have a [fantastic document](https://moonrepo.dev/docs/guides/javascript/typescript-project-refs) that thoroughly explains setting up a TypeScript project references in a monorepo.

packages/synapse-client/package.json
```diff
@@ -38,8 +38,7 @@
     "get-spec": "pnpm get-spec:production",
     "configure": "openapi-generator-cli version-manager set 7.8.0",
     "generate": "MODEL_NAME_MAPPINGS=$(npx tsx src/codegen/generateModelNameMappings.ts); pnpm configure && rimraf src/generated && openapi-generator-cli generate -i src/spec/openapispecification.json -c config.json -o src/generated -t out/ --model-name-mappings $MODEL_NAME_MAPPINGS",
-    "typecheck": "pnpm tsc",
-    "build": "rimraf ./dist && pnpm tsc -p tsconfig.build.json",
```
nickgros, line 42: We had a separate tsconfig.build.json for this project, but it was basically a subset of the one we use for type-checking, so I just removed it.

packages/synapse-react-client/src/components/CertificationQuiz/CertificationQuiz.tsx
```diff
@@ -334,7 +334,7 @@ const CertificationQuiz: React.FunctionComponent = () => {
 }
 
 const CertificationQuizSkeleton: React.FunctionComponent = () => {
-  const questions = []
+  const questions: ReactNode[] = []
```
nickgros, line 337: Fix some type errors that came up, I think it may be that we have a dependent project that has more strict rules than this project (synapse-react-client), but I didn't look into it.

packages/synapse-react-client/src/components/entity/metadata/MetadataTable.tsx
```diff
@@ -34,12 +34,10 @@ export const MetadataTable = ({
   const storageLocationId = dataFileHandle?.storageLocationId
   const { data: storageLocationUploadDestination } =
     useGetUploadDestinationForStorageLocation(parentId!, storageLocationId!, {
-      enabled:
-        parentId !== undefined &&
-        storageLocationId != null,
```
nickgros, line 39: this line change was the code formatter

packages/synapse-react-client/tsconfig.json
```diff
@@ -1,15 +1,52 @@
...
+    "paths": {
+      // HMR will work if we uncomment these, but doing so causes various build/typecheck errors
+      //
+      // In the markdown-it packages, the TypeScript compiler had an issue with the output type declaration file being
+      // generated by tsup and not TypeScript. We could probably replace tsup with the TypeScript compiler, but that
+      // might cause breaking changes that must be handled in SWC.
+      //
+      // In the synapse-client and synapse-types packages cause failures when parsed by the vite-plugin-dts plugin
+      // that we use to generate a single type declaration file in synapse-react-client
+
```
nickgros, line 23: Incremental type-checking and HMR in synapse-react-client won't work for these dependencies. We can probably deal with them, but I didn't want to deal with it right now.

IMO not a big deal, since these dependencies are pretty small and easy to rebuild. The big advantage to this PR is getting HMR for synapse-react-client components in Vite apps.

packages/synapse-types/package.json
```diff
@@ -21,14 +20,12 @@
   },
   "license": "Apache-2.0",
   "scripts": {
-    "build": "rimraf ./dist && tsup ./src/index.ts --format esm,cjs --dts",
-    "type-check": "tsc --noEmit",
+    "build": "tsc --build",
```
nickgros, line 23: Replaced tsup with tsc. Outside of the monorepo, this is used in SynapseWebClient e2e tests, where this seemed to be fine, so I think this is a safe change.

packages/vite-config/src/baseConfig.ts
```diff
@@ -17,7 +18,7 @@ const baseConfig: UserConfig = defineConfig({
       transformMixedEsModules: true,
     },
   },
-  plugins: [nodePolyfills()],
+  plugins: [nodePolyfills(), tsconfigPaths()],
```
nickgros, line 21: The vite-tsconfig-paths plugin inspects the tsconfig `paths` property and modifies the Vite config so modules are resolved in the same way TypeScript resolves types, ensuring our type-checking and bundling are consistent.

jay-hodgson (approved):

PR #1395, opened 2024-11-19
PORTALS-3195

nickgros:
- fix NF portal OrganizationDetailsPage to pass searchParams along since there are multiple lookup keys
- add hook to dynamically set document.title
- add clean jobs that remove tsbuildinfo

### DIFF
--- a/apps/portals/nf/src/pages/OrganizationDetailsPage/OrganizationDetailsPage.tsx
+++ b/apps/portals/nf/src/pages/OrganizationDetailsPage/OrganizationDetailsPage.tsx
@@ -29,7 +29,7 @@ const tabConfig: DetailsPageTabConfig[] = [
 ]
 
 function OrganizationDetailsPage() {
-  const { abbreviation } = useGetPortalComponentSearchParams()
+  const searchParams = useGetPortalComponentSearchParams()
   return (
     <>
       <CardContainerLogic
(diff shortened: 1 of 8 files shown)


apps/portals/nf/src/pages/OrganizationDetailsPage/OrganizationDetailsPage.tsx
```diff
@@ -29,7 +29,7 @@ const tabConfig: DetailsPageTabConfig[] = [
 ]
 
 function OrganizationDetailsPage() {
-  const { abbreviation } = useGetPortalComponentSearchParams()
```
nickgros, line 32: This details page also supports the searchParam `fundingAgency`, so just pass the entire object along

apps/synapse-portal-framework/src/App.tsx
```diff
@@ -7,8 +7,11 @@ import {
 import AppInitializer from './components/AppInitializer'
 import Footer from './components/Footer'
 import Navbar from './components/navbar/Navbar'
+import { useDocumentTitleFromRoutes } from './utils/useDocumentTitleFromRoutes'
 
 export default function App(props: PropsWithChildren) {
+  useDocumentTitleFromRoutes()
```
nickgros, line 13: use new hook to set document title in all portals

jay-hodgson (approved):

PR #1397, opened 2024-11-20
PORTALS-3195 - add canonical URL tags to Explore pages 

nickgros:
...and a few details pages

One hypothesis is that Google is not indexing pages in our portals because they do not identify them as unique. Adding the canonical URL to the document head could help Google dedifferentiate duplicate pages, and also identify unique pages.

### DIFF
--- a/apps/portals/nf/src/pages/DatasetDetailsPage.tsx
+++ b/apps/portals/nf/src/pages/DatasetDetailsPage.tsx
@@ -40,6 +40,7 @@ export default function DatasetDetailsPage() {
         ContainerProps={{
           maxWidth: 'xl',
         }}
+        resourcePrimaryKey={['id']}
       >
         <DetailsPageContent
           content={[
--- a/apps/synapse-portal-framework/src/components/DetailsPage/DetailsPage.tsx
+++ b/apps/synapse-portal-framework/src/components/DetailsPage/DetailsPage.tsx
@@ -19,6 +19,7 @@ import {
 import { DetailsPageProps } from '../../types/portal-util-types'
 import { useGetPortalComponentSearchParams } from '../../utils/UseGetPortalComponentSearchParams'
 import { DetailsPageContextProvider } from './DetailsPageContext'
+import { DetailsPageDocumentMetadata } from './DetailsPageDocumentMetadata'
 import { useScrollOnMount } from './utils'
 
 const goToExplorePage = () => {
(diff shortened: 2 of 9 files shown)


apps/synapse-portal-framework/src/components/DetailsPage/DetailsPage.tsx
```diff
@@ -117,6 +119,7 @@ export default function DetailsPage(props: DetailsPageProps) {
         rowData: row,
       }}
     >
+      <DetailsPageDocumentMetadata resourcePrimaryKey={resourcePrimaryKey} />
```
nickgros, line 122: For all DetailsPage instances, add document metadata (internally, nothing will happen if `resourcePrimaryKey` is undefined)

apps/synapse-portal-framework/src/utils/useSetCanonicalUrl.ts
```diff
@@ -0,0 +1,32 @@
+import { useEffect } from 'react'
+
+/**
+ * Hook to set the canonical URL of the page.
+ *
+ * The canonical URL helps search crawlers understand which URL(s) to index when there are multiple URLs that point to
+ * the same content.
+ * @param canonicalUrl
+ */
+export function useSetCanonicalUrl(canonicalUrl?: string) {
```
nickgros, line 10: One hypothesis is that Google is not indexing pages in our portals because they do not identify them as unique. Adding the canonical URL to the document head could help Google dedifferentiate duplicate pages, and also identify unique pages.

apps/portals/nf/src/pages/DatasetDetailsPage.tsx
```diff
@@ -40,6 +40,7 @@ export default function DatasetDetailsPage() {
         ContainerProps={{
           maxWidth: 'xl',
         }}
+        resourcePrimaryKey={['id']}
```
nickgros, line 43: Identifies which searchParams should be used for the canonical URL. This is necessary because some details pages use multiple keys to show a resource. Without this parameter, a canonical URL will not be added (so this must be manually curated for each DetailsPage)

apps/synapse-portal-framework/src/components/Explore/ExploreWrapper.tsx
```diff
@@ -36,15 +36,22 @@ export default function ExploreWrapper(props: ExploreWrapperProps) {
...
+      } - ${pageName}`
+      if (document.title !== newTitle) {
+        document.title = newTitle
+      }
     }
-  }
+  }, [pathname, pageName])
+
+  // The canonical URL is the explore route with no searchParams
+  useSetCanonicalUrl(new URL(pathname, window.location.origin).toString())
```
nickgros, line 54: Also set the canonical URL on explore pages.

jay-hodgson (approved):

PR #1399, opened 2024-11-20
PORTALS-3308 - What's in Portals Component (GoalsV2)

afwillia:
[PORTALS-3308](https://sagebionetworks.jira.com/browse/PORTALS-3308) This PR modifies the Goals component by creating cards for each panel, surrounded by a box.

[PORTALS-3308]: https://sagebionetworks.jira.com/browse/PORTALS-3308?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/synapse-react-client/src/components/GoalsV2/GoalsV2.tsx
```diff
@@ -0,0 +1,235 @@
...
+          // assume that we recieve assets in order of rows and there is an asset for each item
+          // can revisit if this isn't the case.
+          const asset = assets?.[index] ?? ''
+          const goalsV2DataProps: GoalsV2DataProps = {
+            countSql,
+            title,
+            summary,
+            link,
+            asset,
+          }
```
nickgros, line 211: This is a lot of logic in the returned JSX. Suggest creating an array of `goalsV2DataProps` outside of the `return` and then calling `goalsV2DataPropsArray.map` here

packages/synapse-react-client/src/style/components/_goalsv2.scss
```diff
@@ -0,0 +1,123 @@
...
+  justify-content: center;
+  align-items: center;
+}
+
+%side-padding {
+  padding-left: 25px;
+  padding-right: 25px;
+}
+
+.GoalsV2 {
```
nickgros, line 21: There is a lot of duplicate styling between GoalsV1 and GoalsV2, and it's a bit confusing because there is a lot of styling happening both here and in the component file. Can you either:
- Refactor to share styles with GoalsV1 and minimize the duplicates, and put style overrides here instead of in the component
- Remove most/all of these styles and for GoalsV2 just use MUI builtin components and sx styles

packages/synapse-react-client/src/components/GoalsV2/GoalsV2.test.tsx
```diff
@@ -0,0 +1,113 @@
...
+        </SynapseContextProvider>
+      </QueryClientProvider>,
+    )
+
+    await waitFor(() => {
+      expect(screen.getByText("What's in the Portal?")).toBeInTheDocument()
+    })
+  })
+
+  // test('displays error message on fetch failure', async () => {
```
nickgros, line 80: any reason this is commented out?

nickgros (changes requested):
had some questions and suggestions, let me know if you have any questions!

packages/synapse-react-client/src/components/GoalsV2/GoalsV2.test.tsx
afwillia: it was commented out because it wasn't passing, but will rewrite it to work.

packages/synapse-react-client/src/style/components/_goalsv2.scss
afwillia: I deleted the goalsV2 scss file and instead used the original goals styling with some MUI overrides.

packages/synapse-react-client/src/components/GoalsV2/GoalsV2.tsx
```diff
@@ -0,0 +1,235 @@
...
+              boxShadow: 'none',
+              '&:hover': {
+                backgroundColor: darken('#5BA998', 0.05),
+                boxShadow: 'none',
+              },
+            }}
+          >
+            Start Exploring Data
+          </Button>
+        </Link>
```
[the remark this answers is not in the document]
afwillia: I applied this and also made the href into a prop called dataLink.

packages/synapse-react-client/src/components/GoalsV2/GoalsV2.Mobile.tsx
```diff
@@ -0,0 +1,37 @@
...
+import { Button } from '@mui/material'
+
+export default function GoalsV2Mobile({
+  link,
+  summary,
+  countSql,
+  title,
+}: GoalsV2DataProps) {
+  const titleElement = (
+    <div className="GoalsV2__Mobile__Header">
```
nickgros, line 14: Go ahead and remove all classNames that are not being used for styling

packages/synapse-react-client/src/components/GoalsV2/GoalsV2.test.tsx
```diff
@@ -0,0 +1,128 @@
...
+  isInExperimentalMode: false,
+  utcTime: false,
+  withErrorBoundary: false,
+  downloadCartPageUrl: '/DownloadCart',
+  appId: undefined,
+}
+
+const queryClient = new QueryClient()
+
+const server = setupServer(
```
nickgros, line 23: No need to call `setupServer`, we do that already. You can import `server` from '../mocks/msw/server'

packages/synapse-react-client/src/components/GoalsV2/GoalsV2.test.tsx
```diff
@@ -0,0 +1,128 @@
...
+  rest.post(
```
nickgros, line 24: Your handlers look good, but we have a utility `generateAsyncJobHandlers` that can abstract away setting up the `async/start` and `async/get` handlers that might reduce the amount of code here. Optionally consider using that here

packages/synapse-react-client/src/components/GoalsV2/index.ts
nickgros, line 1: You can delete this file, I'm trying to work towards removing these index files (for performance) and it's no longer true that the component must be exported this way to use it in a portal.

packages/synapse-react-client/src/components/GoalsV2/GoalsV2.tsx
```diff
@@ -0,0 +1,235 @@
...
+          What's in the Portal?
+        </Typography>
+        <Link
+          href={dataLink}
+          target="_blank"
+          rel="noopener noreferrer"
+          fontFamily="Lato"
+          fontSize="18"
+          fontStyle="semi-bold"
+        >
```
nickgros, line 196: Remove this `Link` component since you're already using a `Button` with an `href`

packages/synapse-react-client/src/components/GoalsV2/GoalsV2.tsx
```diff
@@ -0,0 +1,235 @@
...
+          <Button
+            href={dataLink}
+            target="_blank"
+            rel="noopener noreferrer"
+            variant="contained"
+            sx={{
+              backgroundColor: '#5BA998',
+              border: '1px solid white',
+              boxShadow: 'none',
```
nickgros, line 205: You can just remove the box shadow from all states (hover, active) with `!important` here. Should be safe since it only applies to this specific button.

That said, I think @kianamcc is moving these button styles to the EL Home Page theme override in #1456, so you might be able to remove some of these style overrides entirely 

```suggestion
              boxShadow: 'none !important',
```

packages/synapse-react-client/src/components/GoalsV2/GoalsV2.tsx
```diff
@@ -0,0 +1,235 @@
...
+        sx={{
+          textAlign: 'center',
+          paddingBottom: '30px',
+        }}
+      >
+        <Typography
+          variant="headline1"
+          style={{
+            color: 'grey.1000',
+            fontFamily: 'Merriweather',
```
nickgros, line 176: Should be unnecessary now that this is in the HomePageThemeProvider in the EL Portal
```suggestion
```

packages/synapse-react-client/src/components/GoalsV2/GoalsV2.tsx
```diff
@@ -0,0 +1,235 @@
...
+          includePreSignedURLs: true,
+          includePreviewPreSignedURLs: false,
+          requestedFiles: fileHandleAssociationList,
+        }
+        const files = await getFiles(batchFileRequest, accessToken)
+        setError(undefined)
+        setAssets(
+          files.requestedFiles
+            .filter(el => el.preSignedURL !== undefined)
+            .map(el => el.preSignedURL!),
```
nickgros, line 100: I am slightly concerned about these presigned URLs expiring, but this hasn't been an issue with the Goals v1 component, so I think it's fine for now.

packages/synapse-react-client/src/components/GoalsV2/GoalsV2.tsx
```diff
@@ -0,0 +1,235 @@
...
+  )
+  const summaryColumnIndex = getFieldIndex(
+    ExpectedColumns.SUMMARY,
+    queryResultBundle,
+  )
+  const linkColumnIndex = getFieldIndex(ExpectedColumns.LINK, queryResultBundle)
+
+  const goalsDataArray: GoalsV2DataProps[] = []
+
+  queryResultBundle?.queryResult!.queryResults.rows.forEach((el, index) => {
```
nickgros, line 131: You could use `Array,map` instead of `forEach`, but it's basically the same

nickgros (changes requested):
Looking good! Just a few more things to fix - remove unused class names, fix setting up MSW to match our other tests. The rest of the comments are optional suggestions.

packages/synapse-react-client/src/components/GoalsV2/GoalsV2.test.tsx
```diff
@@ -0,0 +1,88 @@
...
+            dataLink="https://eliteportal.synapse.org/Explore/Data"
+          />
+        </SynapseContextProvider>
+      </QueryClientProvider>,
+    )
+
+    await waitFor(() => {
+      expect(screen.getByText("What's in the Portal?")).toBeInTheDocument()
+    })
+  })
```
nickgros, line 87: This test is identical to the first one

packages/synapse-react-client/src/components/GoalsV2/GoalsV2.test.tsx
```diff
@@ -0,0 +1,88 @@
...
+    render(
+      <QueryClientProvider client={queryClient}>
+        <SynapseContextProvider synapseContext={mockSynapseContext}>
+          <GoalsV2
+            entityId="syn22315959"
+            dataLink="https://eliteportal.synapse.org/Explore/Data"
+          />
+        </SynapseContextProvider>
+      </QueryClientProvider>,
+    )
```
nickgros, line 36: We have a test wrapper that provides a QueryClient and SynapseContext, `createWrapper` in TestingLibraryUtils.tsx. You can do `render(<MyComponent />, { wrapper: createWrapper() })`

packages/synapse-react-client/src/components/GoalsV2/GoalsV2.tsx
```diff
@@ -0,0 +1,226 @@
...
+            .filter(el => el.preSignedURL !== undefined)
+            .map(el => el.preSignedURL!),
+        )
+      } catch (e) {
+        console.error('Error on get data', e)
+        setError(e)
+      }
+    }
+    getData()
+  }, [entityId, accessToken, queryResultBundle])
```
nickgros, line 111: Is the logic identical between this in GoalsV1 and V2? If so, consider extracting a custom hook used in both GoalsV1 and V2

packages/synapse-react-client/src/components/GoalsV2/GoalsV2.tsx
```diff
@@ -0,0 +1,226 @@
...
+          textAlign: 'center',
+          paddingBottom: '30px',
+        }}
+      >
+        <Typography
+          variant="headline1"
+          style={{
+            color: 'grey.1000',
+            fontSize: '32px',
+          }}
```
nickgros, line 178: Move these to `sx`. Also, `grey.1000` won't work in `style`, only `sx`

packages/synapse-react-client/src/components/GoalsV2/GoalsV2.tsx
```diff
@@ -0,0 +1,226 @@
...
+          sx={{
+            backgroundColor: '#5BA998',
+            border: '1px solid white',
+            boxShadow: 'none !important',
+            borderRadius: '5px',
+            '&:hover': {
+              backgroundColor: darken('#5BA998', 0.05),
+              boxShadow: 'none !important',
+            },
+          }}
```
nickgros, line 203: Putting ELITE portal theme colors here will make this component non-generic. I think many of the required colors have already been added to the ELITE portal homepage theme (HomePageThemeProvider.tsx), so I think you can probably remove most/all of these inline styles. Any styles that should be applied to all buttons on the ELITE portal home page can be added to that theme file.

packages/synapse-react-client/src/components/GoalsV2/GoalsV2.Desktop.tsx
```diff
@@ -0,0 +1,51 @@
...
+            justifyContent: 'center',
+          }}
+        >
+          <Typography variant="h6" component="strong" sx={{ marginRight: 1 }}>
+            {countSql && (
+              <QueryCount parens={false} query={{ sql: countSql }} />
+            )}
+          </Typography>
+          <Typography variant="body1">{title}</Typography>
+          <IconButton sx={{ color: '#5BA998' }}>
```
nickgros, line 44: 1. Use theme colors so the component is generic for reuse in other portals
2. Maybe the color changed in Figma, but I'm seeing `#39AC97`, so I'm not sure where `#5BA998` came from. Just pointing this out because I usually search the code to see if a specific color matches an existing variable, and I couldn't find `#5BA998` 

```suggestion
          <IconButton sx={{ color: 'primary.main' }}>
```

packages/synapse-react-client/src/components/GoalsV2/GoalsV2.tsx
```diff
@@ -0,0 +1,159 @@
...
+    },
+  }
+  const { data: queryResultBundle } =
+    useGetQueryResultBundle(queryBundleRequest)
+
+  const { assets: goalAssets, error: goalError } = useGetGoalData(
+    entityId,
+    queryResultBundle,
+  )
+  console.log(goalAssets)
```
nickgros, line 57: Remove this log

```suggestion
```

packages/synapse-react-client/src/utils/hooks/useGetGoalData.test.ts
```diff
@@ -0,0 +1,143 @@
...
+            error: string | SynapseClientError | undefined
+          }
+      )[]
+      error?: Error | undefined
+    }
+    await waitFor(() => {
+      result = renderHook(() => useGetGoalData(entityId, tableQueryResult), {
+        wrapper: createWrapper(),
+      }).result
+    })
```
nickgros, line 131: Can you just do this? This `waitFor` isn't really waiting for anything--if it's just to get rid of the `act` warning, we need a different waitFor called _after_ `renderHook`
```suggestion
    const { result } = renderHook(() => useGetGoalData(entityId, tableQueryResult), {
        wrapper: createWrapper(),
    })
```

nickgros (changes requested):

nickgros (approved):
This looks great!

afwillia:
Great, thanks for all the help! I will merge now.

PR #1407, opened 2024-11-24
PORTALS-3344 - fix reference issue causing mismatched artifact usage

nickgros:
- Fix erroneous tsconfig paths that caused one package to use the build artifacts and another to use source code, leading to the app using 2 different copies of the same React context
- Fill in missing tsconfig references in portals projects

### DIFF
--- a/apps/portals/arkportal/tsconfig.json
+++ b/apps/portals/arkportal/tsconfig.json
@@ -22,5 +22,10 @@
     }
   },
   "include": ["src/**/*.tsx", "src/**/*.ts"],
-  "files": ["src/types.d.ts"]
+  "files": ["src/types.d.ts"],
+  "references": [
+    { "path": "../../synapse-portal-framework" },
+    { "path": "../../../packages/synapse-react-client" },
+    { "path": "../../../packages/vite-config" }
+  ]
 }
--- a/apps/synapse-portal-framework/tsconfig.json
+++ b/apps/synapse-portal-framework/tsconfig.json
@@ -15,18 +15,14 @@
     "moduleResolution": "bundler",
     "types": ["vite/client"],
     "paths": {
-      "@sage-bionetworks/synapse-types": [
-        "../../../packages/synapse-types/src"
-      ],
+      "@sage-bionetworks/synapse-types": ["../../packages/synapse-types/src"],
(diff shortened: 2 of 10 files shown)


apps/portals/arkportal/tsconfig.json
```diff
@@ -22,5 +22,10 @@
     }
   },
   "include": ["src/**/*.tsx", "src/**/*.ts"],
-  "files": ["src/types.d.ts"]
+  "files": ["src/types.d.ts"],
+  "references": [
+    { "path": "../../synapse-portal-framework" },
+    { "path": "../../../packages/synapse-react-client" },
+    { "path": "../../../packages/vite-config" }
+  ]
```
nickgros, line 30: Add `references` for incremental type-checking. This does not affect the bug, I only applied this optimization to one portal and forgot to apply it to the rest.

apps/synapse-portal-framework/tsconfig.json
```diff
@@ -15,18 +15,14 @@
...
-      ],
-      "synapse-react-client": ["../../../packages/synapse-react-client/src"],
-      "synapse-react-client/*": [
-        "../../../packages/synapse-react-client/src/*"
+        "../../packages/synapse-types/src/*"
       ],
-      "vite-config": ["../../../packages/vite-config/src"],
-      "vite-config/*": ["../../../packages/vite-config/src/*"]
+      "synapse-react-client": ["../../packages/synapse-react-client/src"],
+      "synapse-react-client/*": ["../../packages/synapse-react-client/src/*"],
```
nickgros, line 23: This should fix the bug; synapse-portal-framework had the wrong path for the synapse-react-client sources. This caused this package to use the build artifacts while the individual portals (using the correct path) used the source code. This led to 2 copies of global objects, such as the React.Context that we store the auth token in. Only one of the authentication contexts was initialized, so consumers of the other context would never get the auth token.

I have no idea why this didn't result in a build-time error :shrug:

jay-hodgson: Nice sleuthing, Nick!  What a difference and superfluous "../" makes!

jay-hodgson (approved):

PR #1411, opened 2024-11-26
Upgrade lint packages

nickgros:
- upgrade lint packages
- fix new lint errors
- add eslint-plugin-barrel-files, add warning for now
- run `eslint --fix` on all packages
- add `eslint --fix` to lint-staged to run on precommit


apps/portals-e2e/src/explore.spec.ts
```diff
@@ -274,7 +274,9 @@ const expectFacetCharts = async (page: Page) => {
         const availableFacetNames: string[] = []
         for (let i = 0; i < nAvailableFacets; i++) {
           const buttonText = await availableFacetButtons.nth(i).textContent()
-          buttonText && availableFacetNames.push(buttonText)
```
nickgros, line 277: New default lint rule requires being more explicit about these nullish checks. I think this improves readability, but I can turn it off if desired.

eslint.config.mjs
```diff
@@ -1,61 +1,37 @@
...
-      sourceType: 'script',
+    },
+    languageOptions: {
       parserOptions: {
-        project: [
-          './tsconfig.json',
-          './packages/**/tsconfig.json',
-          './apps/**/tsconfig.json',
-        ],
+        projectService: true,
```
nickgros, line 31: This `projectService: true` is new functionality in typescript-eslint, and is supposedly much faster.

packages/synapse-react-client/src/components/CardContainer/CardContainer.tsx
```diff
@@ -79,7 +79,7 @@ function Card(props: { propsToPass: any; type: string }) {
   }
 }
 
-function _CardContainer(props: CardContainerProps) {
+function CardContainerInternal(props: CardContainerProps) {
```
nickgros, line 82: New default rule requires that React components follow a naming convention that the leading underscore breaks.

packages/synapse-react-client/src/components/DownloadCart/RequestDownloadCard.tsx
```diff
@@ -75,9 +75,11 @@ export const RequestDownloadCard: React.FunctionComponent<
...
             variant="outlined"
             onClick={() => {
-              useReactACLEditor
-                ? setShowSharingSettings(true)
-                : onViewSharingSettingsClicked(entityId)
+              if (useReactACLEditor) {
+                setShowSharingSettings(true)
+              } else {
+                onViewSharingSettingsClicked(entityId)
+              }
```
nickgros, line 82: New default lint rule bans ternaries that contain a function call.

eslint.config.mjs
```diff
@@ -105,15 +80,10 @@ export default [
         },
       ],
       '@typescript-eslint/no-redundant-type-constituents': 'warn',
-    },
-  },
-  {
-    files: ['apps/portals-e2e/*.ts'],
-    rules: {
-      '@typescript-eslint/no-floating-promises': 'error',
+      'barrel/avoid-barrel-files': 'warn',
```
nickgros, line 83: The only lint rule change I manually added was to add this plugin to avoid barrel files. We should remove barrel files to speed up bundling (in both dev mode and production builds) ([source](https://e18e.dev/guide/speedup.html)) -- for now, just warn against it

packages/synapse-react-client/src/components/JSONArrayEditor/useParseCsv.ts
```diff
@@ -36,7 +44,7 @@ export default function useParseCsv(
           dynamicTyping: !itemsAreString,
           complete: result => {
             if (result.errors.length > 0) {
-              reject(result.errors)
+              reject(new UseParseCsvError(result.errors))
```
nickgros, line 47: New lint rule requires that `reject` is called with an object that has `Error` as a superclass.

packages/synapse-react-client/src/components/Markdown/MarkdownSynapse.tsx
```diff
@@ -583,7 +584,7 @@ const MarkdownSynapse: MarkdownSynapseComponent = class MarkdownSynapse extends
     )
   }
 
-  public async componentDidMount() {
+  public componentDidMount(): void {
```
nickgros, line 587: New lint rule bans inherited methods from returning a promise when the interface does not.

package.json
```diff
@@ -15,28 +15,34 @@
...
+    "typescript": "5.5.2",
+    "typescript-eslint": "^8.16.0"
   },
   "npmClient": "pnpm",
   "lint-staged": {
-    "*.{js,jsx,ts,tsx,json,yml,scss,css,md}": "prettier --write"
+    "*.{js,jsx,ts,tsx}": [
+      "eslint --fix",
+      "prettier --write"
+    ],
```
nickgros, line 42: run `eslint --fix` on precommit for JavaScript/TypeScript

PR #1414, opened 2024-11-26
PORTALS-3307 - getting started component

lakikowolfe:
Create a getting started component for the EL Portal

jira ticket: https://sagebionetworks.jira.com/browse/PORTALS-3307

I got this as close as I could to the figma mock up but to my eye it's still not quite right.


apps/synapse-portal-framework/src/components/elportal/ELGettingStarted.tsx
```diff
@@ -0,0 +1,162 @@
...
+            color: 'white',
+            fontStyle: 'italic',
+            fontSize: '13px',
+            lineHeight: '20px',
+          }}
+          sx={{
+            maxWidth: '100%',
+            mt: '16px',
+            mb: '16px',
+          }}
```
nickgros, line 60: Ditto, no need to use both `style` and `sx`

apps/synapse-portal-framework/src/components/elportal/ELGettingStarted.tsx
```diff
@@ -0,0 +1,162 @@
...
+            padding: '6px 24px',
+            boxShadow: 'none',
+            '&:hover': {
+              backgroundColor: darken('#5BA998', 0.05),
+              boxShadow: 'none',
+            },
+          }}
+        >
+          Visit Our Help Section
+        </Button>
```
nickgros, line 84: Probably don't need some of these styles -- the default button color is `primary.main`, and it handles hover/active states for us. 

Nice catch to remove the box shadow though, that would look weird here! Normally I would never advise to use `!important` but since it's only applied to one element, I think you could get away with it. This would handle the default, hover, and active states of the button.


```suggestion
        <Button
          variant="contained"
          href="https://help.eliteportal.org/help/"
          sx={{
            border: '1px solid white',
            padding: '6px 24px',
            boxShadow: 'none !important',
          }}
        >
          Visit Our Help Section
        </Button>
```

apps/synapse-portal-framework/src/components/elportal/ELGettingStarted.tsx
```diff
@@ -0,0 +1,162 @@
...
+      </Box>
+      <IconSquare
+        iconUrl={exploreIcon}
+        headline="Explore"
+        description="Get the latest data, publications, and news from the Exceptional
+         Longevity Research community"
+      />
```
nickgros, line 91: nice encapsulation of this component

apps/synapse-portal-framework/src/components/elportal/ELGettingStarted.tsx
```diff
@@ -0,0 +1,162 @@
...
+export function IconSquare({ iconUrl, headline, description }) {
+  return (
+    <Box
+      sx={{
+        flex: '1 1 auto',
+        width: '100%',
+        height: '100%',
+        overflow: 'hidden',
+      }}
+    >
```
nickgros, line 119: This is not in a flex container, so `flex` isn't doing anything. I also think the hidden overflow may end up being a headache if this layout changes

```suggestion
    <Box
      sx={{
        width: '100%',
        height: '100%',
      }}
    >
```

You could consider making this a grid container or flex container, but what you have now looks fine to me.

If you want the pixel-perfect alignment between sections (which is waaaay easier for Adam to do in Figma than for us to do in CSS), you could probably use [subgrid](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_grid_layout/Subgrid). We have a couple of examples I could point you to if you're interested--it's a fairly advanced (and pretty new) CSS feature

apps/synapse-portal-framework/src/components/elportal/ELGettingStarted.tsx
```diff
@@ -0,0 +1,162 @@
...
+      <Box
+        sx={{
+          flex: '0 0 auto',
```
nickgros, line 122:
```suggestion
```
Not doing anything unless you make the wrapper a flex container

apps/synapse-portal-framework/src/components/elportal/ELGettingStarted.tsx
```diff
@@ -0,0 +1,162 @@
...
+          backgroundImage: `url(${iconUrl})`,
+          backgroundRepeat: 'no-repeat',
+          backgroundSize: 'contain',
+        }}
+      ></Box>
+      <Typography
+        variant="headline3"
+        style={{
+          color: 'white',
+          fontFamily: "'Merriweather', serif",
```
nickgros, line 135: not clear to me if `headline3` should always have this fontFamily in the EL HomePageThemeProvider. If not, this this is fine

nickgros (changes requested):
I left a bunch of super nitpicky comments and optional suggestions, but as a whole what you have looks great and is pretty close to ready. I am mostly looking for code style changes and removal of the CSS properties that aren't doing anything. Please feel free to reach out if you have questions or want clarification

lakikowolfe:
hm something about merging main into this branch broke the site running locally. Looking into that now 👀

apps/synapse-portal-framework/src/components/elportal/ELGettingStarted.tsx
```diff
@@ -0,0 +1,162 @@
...
+// TODO Fix vertical column space (getting started section is slightly bigger)
+// TODO Set padding for mobile and desktop
+
+const ELGettingStarted: React.FC = () => {
+  return (
+    <Box
+      sx={{
+        backgroundColor: '#39AC97',
+        display: 'grid',
+        gridTemplateColumns: '25% 25% 25% 25%',
```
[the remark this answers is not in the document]
nickgros: I can't find a good reference in the MUI docs but when using `sx`, you can set the property to be an object with the specific breakpoints `xs`, `sm`, `md`, `lg`, `xl`, and that style will be applied at that breakpoint and sizes above it (until the next specified breakpoint).

So passing the object 
```ts
       {
          xs: '100%',
          sm: '50% 50%',
          md: '25% 25% 25% 25%',
        },
```

would set `gridTemplateColumns` to '100%' from 0 - sm (I think 600px wide), '50%-50%' from 600 (sm) -> 900 (md) and '25% 25% 25% 25%' from 900px up

apps/synapse-portal-framework/src/components/elportal/ELGettingStarted.tsx
```diff
@@ -0,0 +1,162 @@
...
+        position: 'relative',
+        pt: '80px',
+        pb: '80px',
+        pl: '80px',
+        pr: '80px',
+      }}
+    >
+      <Box
+        sx={{
+          flex: '1 1 auto',
```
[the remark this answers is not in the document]
nickgros: Correct. They are similar, in both cases you have a grid container & grid items or a flex container & flex items. Grid item properties won't do anything in flex containers and Flex item properties won't do anything in flex containers.

There are some properties that work for both, e.g. `gap` works for both flex and grid containers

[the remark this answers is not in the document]
nickgros: These are are my preferred references for flex & grid, if it's helpful:

https://css-tricks.com/snippets/css/a-guide-to-flexbox/

https://css-tricks.com/snippets/css/complete-guide-grid/

apps/synapse-portal-framework/src/components/elportal/ELGettingStarted.tsx
nickgros: Hidden overflow will hide excess content and make it impossible to see all of it as a user. If we have excess content in this component, the layout should accommodate it so it isn't hidden. In other cases (but probably not this component) you may want `overflow: scroll` or `overflow: auto` so a scroll bar appears.

Sometimes `overflow: hidden;` is appropriate, but it doesn't seem right in this case (unless I'm missing something, please let me know if I am)

apps/synapse-portal-framework/src/components/elportal/ELGettingStarted.tsx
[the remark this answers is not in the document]
lakikowolfe: That's really helpful thanks!

apps/synapse-portal-framework/src/components/elportal/ELGettingStarted.tsx
```diff
@@ -0,0 +1,139 @@
...
+          borderTop: '3px solid #ffffff88',
+          py: '20px',
+        }}
+      >
+        <Typography
+          variant="headline2"
+          sx={{
+            width: '100%',
+            color: 'white',
+            fontFamily: "'Merriweather', serif",
```
nickgros, line 36: This will be applied by the `HomePageThemeProvider`
```suggestion
```

apps/synapse-portal-framework/src/components/elportal/ELGettingStarted.tsx
```diff
@@ -0,0 +1,139 @@
...
+            fontSize: '13px',
+            lineHeight: '20px',
+          }}
+        >
+          We provide all the help you need for navigating the portal and
+          accelerating your research.
+        </Typography>
+        <Button
+          variant="contained"
+          href="https://help.eliteportal.org/help/"
```
nickgros, line 58: Since this goes to another page, it seems like it should open in a new tab
```suggestion
          href="https://help.eliteportal.org/help/"
          rel="noopener noreferrer"
          target="_blank"
```

apps/synapse-portal-framework/src/components/elportal/ELGettingStarted.tsx
```diff
@@ -0,0 +1,139 @@
...
+      <Typography
+        variant="body1"
+        sx={{
+          maxWidth: '100%',
+          color: 'white',
+          fontFamily: "'DM Sans', sans-serif",
+          fontSize: '13px',
+          lineHeight: '15px',
+          fontWeight: 400,
+        }}
```
nickgros, line 133: Remove a few of these values that match the defaults
```suggestion
        sx={{
          maxWidth: '100%',
          color: 'white',
          fontSize: '13px',
          lineHeight: '15px',
        }}
```

nickgros (approved):
This looks good, had a few small suggestions. Did you want to add this to the HomePageV2 component?

lakikowolfe:
> hm something about merging main into this branch broke the site running locally. Looking into that now 👀

solved by rebuilding repo

PR #1422, opened 2024-12-02
PORTALS-3332: add team member count, and use it in the Challenge portal

jay-hodgson:


packages/synapse-react-client/src/synapse-queries/team/useTeamMembers.ts
```diff
@@ -31,6 +32,18 @@ export function useGetTeamMembers(
...
+export function useGetTeamMemberCount(
+  teamId: string | number,
+  options?: Partial<UseQueryOptions<Count, SynapseClientError>>,
+) {
+  const { accessToken, keyFactory } = useSynapseContext()
+  return useQuery({
+    ...options,
+    queryKey: keyFactory.getTeamMemberCountQueryKey(String(teamId)),
+    queryFn: () => SynapseClient.getTeamMemberCount(accessToken, teamId),
+  })
```
nickgros, line 44: You're using the `Count` type generated from the OpenAPI definition, so why not also use the generated client:
```suggestion
  const { keyFactory, synapseClient } = useSynapseContext()
  return useQuery({
    ...options,
    queryKey: keyFactory.getTeamMemberCountQueryKey(String(teamId)),
    queryFn: () => synapseClient.teamServicesClient.getRepoV1TeamMembersCountId({id: teamId}),
  })
```

nickgros (approved):

PR #1431, opened 2024-12-03
SWC-7200: If there's an error reading an evaluation queue in the scop…

jay-hodgson:
…e, show the error (and a way to remove the evaluation queue ID)


jay-hodgson:
Screenshot 2024-12-03 at 1 42 02 PM

nickgros (approved):

PR #1436, opened 2024-12-04
PORTALS-3355

jay-hodgson:


jay-hodgson:
ELITE home page redesign implementation should be made to the HomepageV2 version.  We will eventually switch the index to point to HomepageV2 and simply delete the old Homepage

afwillia (approved):
This sounds good to me, thanks for the heads up.

PR #1447, opened 2024-12-06
PORTALS-3340 - Load Merriweather from Google Fonts

nickgros:


kianamcc (approved):

apps/portals/elportal/src/App.scss
```diff
@@ -5,3 +5,6 @@
 
 // style_overrides should be last to ensure its styles have highest priority
 @use './config/style/style_overrides';
+
+// Merriweather font is used only on the ELITE Portal
```
jay-hodgson, line 9: Could you also make a note in HomePageThemeProvider that this import is required (for the future engineer who applies this theme to other portal home pages)?

nickgros: https://github.com/Sage-Bionetworks/synapse-web-monorepo/pull/1450

PR #1454, opened 2024-12-09
Remove openapi-generator version configuration step

nickgros:
Fetching the list of versions queries search.maven.org, which can experience downtime. This request is unnecessary to run unless changing the version of the generator tool.


packages/synapse-client/package.json
```diff
@@ -37,8 +37,7 @@
     "get-spec:staging": "npx tsx src/codegen/downloadSpec.ts --stack staging && pnpm format-spec",
     "get-spec:production": "npx tsx src/codegen/downloadSpec.ts --stack production && pnpm format-spec",
     "get-spec": "pnpm get-spec:production",
-    "configure": "openapi-generator-cli version-manager set 7.8.0",
```
nickgros, line 40: note the current versions is saved in openapitools.json, so this step is unnecessary.

jay-hodgson (approved):

PR #1455, opened 2024-12-09
PORTALS-3358: add an On Synapse link

jay-hodgson:


jay-hodgson:
See https://sagebionetworks.jira.com/browse/PORTALS-3358?focusedCommentId=229181

nickgros (approved):

PR #1459, opened 2024-12-12
SWC-7166: Hide Banner on Mobile Screens

lakikowolfe:
Hide Orientation Banner on mobile screens

Jira: https://sagebionetworks.jira.com/browse/SWC-7166

### DIFF
--- a/packages/synapse-react-client/src/components/OrientationBanner/OrientationBanner.tsx
+++ b/packages/synapse-react-client/src/components/OrientationBanner/OrientationBanner.tsx
@@ -56,6 +56,7 @@ function OrientationBanner(props: OrientationBannerProps) {
 
   const theme = useTheme()
   const defaultSx = {
+    display: { xs: 'none', md: 'unset' },
     backgroundColor: '#F9FAFB',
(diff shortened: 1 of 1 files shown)


packages/synapse-react-client/src/components/OrientationBanner/OrientationBanner.tsx
```diff
@@ -82,7 +82,11 @@ function OrientationBanner(props: OrientationBannerProps) {
         }
         setShowBanner(false)
       }}
-      sx={{ ...defaultSx, ...sx }}
+      sx={{
+        ...defaultSx,
+        ...sx,
+        display: { xs: 'none', md: 'unset' },
```
nickgros, line 88: Sorry, I didn't realize that we already had an `sx` object -- can you move this `display` property to the `defaultSx` object?

lakikowolfe: oh yep! that makes more sense. On it!

nickgros (approved):

PR #1463, opened 2024-12-16
SWC-7166 - Fix OrientationBanner display & add spreadSx utility

nickgros:
- Add utility to properly spread Sx props
- Fix `display` of OrientationBanner
- Use utility in components spreading sx


packages/synapse-react-client/src/components/OrientationBanner/OrientationBanner.tsx
```diff
@@ -54,9 +56,10 @@ function OrientationBanner(props: OrientationBannerProps) {
     localStorage.getItem(storageBannerId) === null,
   )
 
-  const theme = useTheme()
-  const defaultSx = {
-    display: { xs: 'none', md: 'unset' },
```
nickgros, line 59: Setting this to `unset` ended up replacing the `flex` layout, so we had to remove `unset` somehow

packages/synapse-react-client/src/components/OrientationBanner/OrientationBanner.tsx
```diff
@@ -54,9 +56,10 @@ function OrientationBanner(props: OrientationBannerProps) {
...
+  const defaultSx: SxProps<Theme> = theme => ({
+    [theme.breakpoints.down('md')]: {
+      display: 'none',
+    },
```
nickgros, line 62: More concise/conventional way to just change the breakpoint styles is to do this. Now we aren't touching md and up.

packages/synapse-react-client/src/components/OrientationBanner/OrientationBanner.tsx
```diff
@@ -83,10 +86,7 @@ function OrientationBanner(props: OrientationBannerProps) {
         }
         setShowBanner(false)
       }}
-      sx={{
-        ...defaultSx,
-        ...sx,
-      }}
+      sx={spreadSx(defaultSx, sx)}
```
nickgros, line 89: I learned that you can pass an array of SxProps to `sx` and MUI will handle merging them, but we need to use a new utility to do so.

kianamcc (approved):

PR #1469, opened 2024-12-17
small elite portal components styling updates

kianamcc:


apps/portals/elportal/src/themes/HomePageThemeProvider.tsx
```diff
@@ -32,6 +32,11 @@ export function HomePageThemeProvider({ children }: React.PropsWithChildren) {
               // TODO: Bootstrap 3 link color applies to all a:hover elements. Remove this override when Bootstrap 3 is removed
               color: '#FFF !important',
             },
+            '&:focus': {
+              textDecoration: 'none',
+              // TODO: Bootstrap 3 link color applies to all a:hover elements. Remove this override when Bootstrap 3 is removed
+              color: '#FFF !important',
```
nickgros, line 38: Pull this up and out of the pseudo-class selectors since it's always white (and keep the !important so it still works!)

nickgros (approved):

PR #1471, opened 2024-12-18
Links to Titles on RecentPublicationsGrid Component

kianamcc:
Hover color given by Adam:

Screenshot 2024-12-17 at 4 53 32 PM

### DIFF
--- a/packages/synapse-react-client/src/components/RecentPublicationsGrid/RecentPublicationsGrid.tsx
+++ b/packages/synapse-react-client/src/components/RecentPublicationsGrid/RecentPublicationsGrid.tsx
@@ -25,7 +25,101 @@ export type RecentPublicationsGridProps = {
 
 type PublicationCardProps = {
   pub: Row
+  isLoading: boolean
+  categoryColIndex: number
+  titleColIndex: number
+  journalColIndex: number
+  publicationDateColIndex: number
+  doiColIndex: number
 }
+
+const PublicationCard = ({
+  pub,
+  isLoading,
+  categoryColIndex,
+  titleColIndex,
+  journalColIndex,
+  publicationDateColIndex,
+  doiColIndex,
+}: PublicationCardProps) => (
+  <Grid
+    key={pub.rowId}
+    height={{ xs: 'auto', sm: 'initial' }}
+    minWidth={{ xs: '280px', lg: 'initial' }}
+    maxWidth={{ xs: '450px', lg: 'initial' }}
+  >
+    <Box sx={{ height: '100%' }}>
+      {isLoading ? (
+        <Skeleton variant="rectangular" height={275} width="100%" />
+      ) : (
+        <div>
+          {pub.values[categoryColIndex] && (
+            <Typography
+              variant="overline"
+              fontSize={'14px'}
+              sx={{
+                backgroundColor: 'grey.300',
+                borderRadius: '3px',
+                padding: '4px 8px',
+                border: 'none',
+                lineHeight: 'initial',
+              }}
+            >
+              {pub.values[categoryColIndex]}
+            </Typography>
+          )}
+          <Typography
+            variant="headline2"
+            color="grey.1000"
+            fontSize={'21px'}
+            sx={{
+              padding: '20px 0px',
+            }}
+          >
+            <MuiLink
+              href={convertDoiToLink(pub.values[doiColIndex] || '')}
+              target="_blank"
+              sx={{
+                color: 'inherit',
+                textDecoration: 'none',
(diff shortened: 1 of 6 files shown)


packages/synapse-react-client/src/components/RecentPublicationsGrid/RecentPublicationsGrid.tsx
```diff
@@ -20,6 +27,14 @@ type PublicationCardProps = {
   pub: Row
 }
 
+const convertDoiToLink = (doi: string) => {
+  doi = doi.trim()
+  if (DOI_REGEX.test(doi)) {
```
kianamcc, line 32: Did we have a utility function for this?

nickgros: It looks like we have similar logic in GenericCard.tsx and CardFooter.tsx, you are welcome to refactor to create this utility.

packages/synapse-react-client/src/components/RecentPublicationsGrid/RecentPublicationsGrid.tsx
```diff
@@ -60,6 +76,9 @@ function RecentPublicationsGrid(props: RecentPublicationsGridProps) {
     ExpectedColumns.PUBLICATION_DATE,
     queryResultBundle,
   )
+  const doiColIndex = getFieldIndex(ExpectedColumns.DOI, queryResultBundle)
+
+  console.log('doi', doiColIndex)
```
nickgros, line 81: Remove log

packages/synapse-react-client/src/components/RecentPublicationsGrid/RecentPublicationsGrid.tsx
```diff
@@ -96,7 +115,21 @@ function RecentPublicationsGrid(props: RecentPublicationsGridProps) {
...
-              {pub.values[titleColIndex]}
+              <MuiLink
+                href={convertDoiToLink(pub.values[doiColIndex] || '')}
+                target="_blank"
+                sx={{
+                  color: 'inherit',
+                  textDecoration: 'none',
+                  fontWeight: 'inherit',
+                  '&:hover': {
+                    color: '#1C9F87',
```
nickgros, line 126: This is not reusable across portals, and IMO this is indistinguishable from the default hover color on the EL portal.

If we need this color, please add it to the ELITE Portal Home page theme or ELITE Portal CSS overrides.
```suggestion
```

packages/synapse-react-client/src/utils/functions/RegularExpressions.ts
```diff
@@ -5,6 +5,14 @@ import { normalizeSynPrefix } from './EntityTypeUtils'
 // note - had to add an escape character for the second and third forward slash in the regex above
 export const DOI_REGEX = /^10.\d{4,9}\/[-._;()/:a-z0-9]+$/i
 
+export function convertDoiToLink(doi: string) {
```
nickgros, line 8: Can you just add a couple of tests for this, at least these input cases, to ensure future changes to this method do not break the components using it?

* Valid DOI
* Malformed DOI
* Empty string

packages/synapse-react-client/src/components/RecentPublicationsGrid/RecentPublicationsGrid.tsx
```diff
@@ -60,6 +68,7 @@ function RecentPublicationsGrid(props: RecentPublicationsGridProps) {
     ExpectedColumns.PUBLICATION_DATE,
     queryResultBundle,
   )
+  const doiColIndex = getFieldIndex(ExpectedColumns.DOI, queryResultBundle)
 
   const PublicationCard = ({ pub }: PublicationCardProps) => (
```
nickgros, line 73: I just noticed that this component is declared inside of another component. This violates the [no-unstable-nested-component](https://github.com/jsx-eslint/eslint-plugin-react/blob/master/docs/rules/no-unstable-nested-components.md) rule. Can you move the `PublicationCard` definition outside of the `RecentPublicationsGrid` function component?

Also I recommend ensuring you have the ESLint VSCode extension enabled to highlight these issues for you.

nickgros: This has come up a few times from multiple engineers, so I created PORTALS-3369 to change the lint warning to an error here, to make it easier to catch.

nickgros (approved):
Looks good, tests and recommended change to extract internal component can occur in this PR or can be in follow-up PR(s)

PR #1475, opened 2024-12-26
SWC-7170: SRC UserCard updates on mobile

kianamcc:
Jira Ticket: https://sagebionetworks.jira.com/browse/SWC-7170

I used overrides + editing styles in Chrome to get the output below:
Screenshot 2024-12-26 at 11 26 41 AM

### DIFF
--- a/packages/synapse-react-client/src/components/UserCard/UserCardMedium.tsx
+++ b/packages/synapse-react-client/src/components/UserCard/UserCardMedium.tsx
@@ -1,4 +1,4 @@
-import { Card, IconButton, Skeleton, Tooltip } from '@mui/material'
+import { Card, IconButton, Skeleton, Tooltip, SxProps } from '@mui/material'
 import {
   MutableRefObject,
   SyntheticEvent,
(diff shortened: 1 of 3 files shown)


packages/synapse-react-client/src/components/UserCard/UserCardMedium.tsx
```diff
@@ -78,6 +78,13 @@ export function UserCardMedium({
   }
   const [showModal, setShowModal] = useState(false)
 
+  const userCardMediumBaseStyles: SxProps = {
+    height: '120px',
+    width: ['-webkit-fill-available', '-moz-available', '100%'],
+    paddingLeft: '25px',
+    minWidth: '350px',
+  }
```
nickgros, line 86: Why copy these styles and remove the `SRC-userCardMediumUp` class? I'd be ok with keeping the class if there are no changes.

If you want to remove the CSS class entirely, that would be OK, but your change should consider the other usages of this class (there are at least 2 in this file)

kianamcc: There was some override issues with scss and sx but I just increased the specificity for the sx styling so it works with SRC-userCardMediumUp now!

nickgros (approved):

PR #1476, opened 2024-12-30
SWC-7169-V2

kianamcc:
Fixing ConfirmationDialog overflow. Waiting for Adam's input: https://sagebionetworks.jira.com/browse/SWC-7169?focusedCommentId=230439&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-230439

*Update*: My Loom video to confirm things/update Adam with changes: https://www.loom.com/share/b33e639251254d4ca4a2eb2caf72ab33?replyToComment=66429327&start_and_pause=1&t=127

Adjusted padding on mobile since video.

### DIFF
--- a/packages/synapse-react-client/src/components/DialogBase.tsx
+++ b/packages/synapse-react-client/src/components/DialogBase.tsx
@@ -12,8 +12,8 @@ import {
   SxProps,
 } from '@mui/material'
 import { ReactNode } from 'react'
-
 import { HelpPopover, HelpPopoverProps } from './HelpPopover/HelpPopover'
+import { spreadSx } from '../theme/spreadSx'
 
 const EMPTY_OBJECT = {}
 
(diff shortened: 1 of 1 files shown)


packages/synapse-react-client/src/components/DialogBase.tsx
```diff
@@ -102,19 +102,22 @@ export const DialogBase = ({
...
+        [theme.breakpoints.down('sm')]: {
+          width: '100vw',
+          '.MuiDialog-container > .MuiPaper-root': {
+            [theme.breakpoints.down('sm')]: {
+              padding: '33px',
+              margin: 0,
+              width: '100%',
+              height: '100%',
+              maxHeight: 'unset',
+            },
```
nickgros, line 116: Is this separate wrapper necessary? You're already inside a `theme.breakpoints.down`
```suggestion
            padding: '33px',
            margin: 0,
            width: '100%',
            height: '100%',
            maxHeight: 'unset',
```

packages/synapse-react-client/src/components/DialogBase.tsx
```diff
@@ -102,19 +102,22 @@ export const DialogBase = ({
...
-        sx: theme => ({
-          [theme.breakpoints.down('sm')]: {
-            margin: 0,
-            width: '100%',
-            height: '100%',
-            maxHeight: 'unset',
-            padding: '16px',
+      sx={theme => ({
+        ...sx,
+        [theme.breakpoints.down('sm')]: {
```
nickgros, line 107: Can you use the [spreadSx utility](https://github.com/Sage-Bionetworks/synapse-web-monorepo/blob/84d5f03846ba335aba0a033ce0658a9f8d8a6106/packages/synapse-react-client/src/theme/spreadSx.ts#L10) instead of manually spreading the `sx` passed from props?

packages/synapse-react-client/src/components/DialogBase.tsx
kianamcc: Whoops, no, that is a mistake!

packages/synapse-react-client/src/components/DialogBase.tsx
kianamcc: Ah, right! Will be using that from now on.

nickgros (approved):

PR #1477, opened 2025-01-02
Portals 3313

afwillia:
This updates the large user card so the name links to the user profile and the header background is transparent.


packages/synapse-react-client/src/components/UserCard/UserCardMedium.tsx
```diff
@@ -251,7 +243,8 @@ export function UserCardMedium({
   return (
     <Card>
       <div
-        className={`SRC-primary-background-color SRC-userCard SRC-userCardMediumUp`}
+        className={`SRC-userCard`}
```
nickgros, line 246: Why remove the `SRC-userCardMediumUp` style? When I look at this in Storybook, it looks like that caused us to lose some crucial padding in the top + left part of the card

image

packages/synapse-react-client/src/components/UserCard/UserCardMedium.tsx
```diff
@@ -251,7 +243,8 @@ export function UserCardMedium({
...
+        style={{ backgroundColor: 'transparent' }}
```
nickgros, line 247: Is this background color set anywhere else? It seems like this override is unnecessary, but I might be missing something
```suggestion
```

packages/synapse-react-client/src/components/UserCard/UserCardMedium.tsx
```diff
@@ -176,14 +168,14 @@ export function UserCardMedium({
           )}
         </p>
         {(position || company) && (
-          <p className={`${isLarge ? 'SRC-whiteText' : ''}`}>
+          <p className={`${isLarge ? 'SRC-blackText' : ''}`}>
```
nickgros, line 171: I think the default color is fine (that's what's used in the design) for both medium and large variants, so you can remove the className prop entirely for all of these where you swapped `whiteText` for `blackText`
```suggestion
          <p>
```

nickgros (changes requested):
Couple of small suggestions and a question

packages/synapse-react-client/src/components/UserCard/UserCardMedium.tsx
afwillia: I thought this was necessary to remove the blue background, but that doesn't seem to be the case anymore

packages/synapse-react-client/src/components/UserCard/UserCardMedium.tsx
afwillia: I added this back - I agree I don't think it needed to be removed, thanks!

packages/synapse-react-client/src/components/UserCard/UserCardMedium.tsx
```diff
@@ -176,14 +168,14 @@ export function UserCardMedium({
...
-          <p className={`${isLarge ? 'SRC-whiteText' : ''}`}>
+          <p>
             {position} {position ? ' / ' : ''} {company}
           </p>
         )}
         {!hideEmail && (
           <p
             ref={copyToClipboardRef}
-            className={`${isLarge ? 'SRC-whiteText' : ''}
+            className={`${isLarge ? 'SRC-blackText' : ''}
```
nickgros, line 178: Remove these from other places -- unless it's still necessary
```suggestion
```

packages/synapse-react-client/src/components/UserCard/UserCardMedium.tsx
```diff
@@ -198,7 +190,7 @@ export function UserCardMedium({
             tabIndex={0}
           >
             <span style={{ paddingRight: '5px', paddingBottom: '2px' }}>
-              <a className={`link ${isLarge ? 'SRC-whiteText' : ''}`}>
+              <a className={`link ${isLarge ? 'SRC-blackText' : ''}`}>
```
nickgros, line 193: Remove these from other places -- unless it's still necessary
```suggestion
              <a className={`link`}>
```

packages/synapse-react-client/src/components/UserCard/UserCardMedium.tsx
```diff
@@ -212,7 +204,7 @@ export function UserCardMedium({
             rel="noopener noreferrer"
             style={{ width: 'fit-content' }}
             tabIndex={0}
-            className={isLarge ? 'SRC-whiteText' : ''}
+            className={isLarge ? 'SRC-blackText' : ''}
```
nickgros, line 207: Remove these from other places -- unless it's still necessary
```suggestion
```

nickgros (approved):
Few more cases where you can remove the unnecessary CSS class -- fix those and you're good to go!

PR #1481, opened 2025-01-03
Fix router issue in Storybook

nickgros:
Add `withRouter` parameter to Storybook to optionally configure a router per-story

Fix the ChangePassword stories by adding a memory router


jay-hodgson (approved):

jay-hodgson:
Coudl this also be used in RecentPublicationsGrid.stories.tsx (instead of the BrowserRouter)?

PR #1482, opened 2025-01-03
SWC-7207

jay-hodgson:


packages/synapse-react-client/src/components/ChangePassword/useChangePasswordFormState.tsx
```diff
@@ -98,7 +98,30 @@ export default function useChangePasswordFormState(
...
+    }
+
+    const hasSpecialChar = /[~!@#$%^&*_\-+=`|\\(){}[\]:;"'<>,.?/]/.test(
+      newPassword,
+    ) // Checks for at least one special character
+    if (!hasSpecialChar) {
+      return 'A valid password must include special characters ~!@#$%^&*_-+=`|\\(){}[]:;"\'<>,.?/'
+    }
+    return undefined
+  }, [])
```
nickgros, line 124: This function should be pulled out of the hook (ok to keep it in this file), since it doesn't rely on any hook state/effects. It could be unit tested without instantiating the entire hook (but I think the tests you have are comprehensive enough for this new functionality, and I don't think we have unit tests for this hook to begin with).

The function doesn't even need to be returned by the hook, but I recognize that there is value in co-locating it with the stateful password utilities, so that decision makes sense.

nickgros (approved):
Minor suggestion about organization, logic looks good

nickgros (approved):

PR #1485, opened 2025-01-06
PORTALS-2795 - Support virtualizing TanStack tables

nickgros:
- Add StyledVirtualTanStackTable component, a wrapper around StyledTanStackTable with support for @tanstack/react-virtual
- Add @tanstack/react-virtual
- Augment StyledTanStackTable, TableBody to support slots, trying to match the MUI design pattern. The use of slots reduces duplicated code that we would need in StyledVirtualTanStackTable


packages/synapse-react-client/package.json
```diff
@@ -59,7 +59,8 @@
     "@tanstack/query-core": "5.22.2",
     "@tanstack/react-query": "5.22.2",
     "@tanstack/react-query-devtools": "5.24.0",
-    "@tanstack/react-table": "^8.20.5",
+    "@tanstack/react-table": "^8.20.6",
+    "@tanstack/react-virtual": "^3.11.1",
```
nickgros, line 63: @tanstack/react-virtual is very small, and was easy to figure out how to get it to work with @tanstack/react-table compared to react-window, since examples exist in the react-table docs.

We're also removing another virtualization library embedded in react-base-table, so while I don't _want_ to have multiple dependencies that do the same thing, I think this is progress.

packages/synapse-react-client/src/components/TanStackTable/StyledTanStackTable.tsx
```diff
@@ -1,25 +1,59 @@
...
 
-type StyledTanStackTableProps<T = unknown> = {
-  table: Table<T>
+export type StyledTanStackTableProps<TData = unknown, TRowType = Row<TData>> = {
+  table: Table<TData>
   styledTableContainerProps?: StyledTableContainerProps
   fullWidth?: boolean
-}
+  slots?: StyledTanStackTableSlots<TData, TRowType>
+  slotProps?: StyledTanStackTableSlotProps<TData, TRowType>
```
nickgros, line 19: In an effort to reduce duplicate code, I added `slots`/`slotsProps` props to `StyledTanStackTable`. I borrowed heavily from MUI's [slots design pattern](https://mui.com/x/common-concepts/custom-components/) since it seems to be a straightforward way to override/customize components wrapped by a complex component.

packages/synapse-react-client/src/components/TanStackTable/StyledVirtualTanStackTable.tsx
```diff
@@ -0,0 +1,123 @@
...
+    />
+  )
+}
+
+/**
+ * Stylized table component that is compatible with @tanstack/react-table and @tanstack/react-virtual to display a
+ * virtualized table, i.e. a table that uses a virtualizer to only render the DOM nodes that contain data that is visible
+ * within the viewport. For cases where the table rows do not need to be virtualized, use {@link StyledTanStackTable}.
+ */
+export default function StyledVirtualTanStackTable<T = unknown>(
```
nickgros, line 53: New component that adds the styles / component overrides needed for virtualization to work

packages/synapse-react-client/src/components/TanStackTable/StyledVirtualTanStackTable.tsx
```diff
@@ -0,0 +1,123 @@
...
+            },
+          },
+          Th: {
+            style: {
+              display: 'flex',
+            },
+          },
+          Tbody: {
+            style: {
+              display: 'grid',
```
nickgros, line 114: Uses a `grid` layout, but still uses semantic table elements for a11y

packages/synapse-react-client/src/components/TanStackTable/TableBody.tsx
```diff
@@ -1,52 +1,74 @@
...
+})({})
 
-type TableBodyProps<T = unknown> = {
-  table: Table<T>
-}
+export type TableBodyProps<TData = unknown, TRowType = Row<TData>> = {
+  /** The table instance */
+  table: Table<TData>
+  slots?: TableBodySlots<TData, TRowType>
+  slotProps?: TableBodySlotProps<TData, TRowType>
```
nickgros, line 23: Add slots

packages/synapse-react-client/src/components/TanStackTable/TableCellRenderer.tsx
```diff
@@ -0,0 +1,35 @@
+import { Skeleton } from '@mui/material'
+import { Cell, flexRender } from '@tanstack/react-table'
+import ExpandableTableDataCell from '../SynapseTable/ExpandableTableDataCell'
+import { getColumnSizeCssVariable } from './TanStackTableUtils'
+
+export function TableCellRenderer<T = unknown>(cell: Cell<T, unknown>) {
```
nickgros, line 6: Moved here from from `TableBody` component, no changes

packages/synapse-react-client/src/components/TanStackTable/types.ts
```diff
@@ -0,0 +1,136 @@
+import { Cell, Row } from '@tanstack/react-table'
+import { RequireAllOrNone } from 'type-fest'
+import { TableBodyProps } from './TableBody'
+
```
nickgros, line 4: Making the slots work requires a fair amount of complex TypeScript. To keep the component files simple, I moved the types here. 

I don't think these are so complex that they are unmaintainable, but I do want to know if you think this is overengineered and if there's a better approach to take.

packages/synapse-react-client/src/components/TanStackTable/TableBody.tsx
```diff
@@ -1,52 +1,74 @@
...
+  props: TableBodyProps<TData, TRowType>,
+) {
+  const { table, slots = {}, slotProps = {} } = props
+
+  // By default, use TanStack Table Rows and the identity function as a transform.
+  // This can be overridden e.g. to accomplish row virtualization.
+  const {
+    rows = table.getRowModel().rows as TRowType[],
+    rowTransform = identity,
+  } = props
```
nickgros, line 41: For instance, when we create a virtual table, `rows` will be the virtualized items that should be rendered (e.g. a sliding window of indices), and `rowTransform` will be a function that transforms a virtualized item to the corresponding row data

packages/synapse-react-client/src/components/styled/StyledTableContainer.tsx
```diff
@@ -1,64 +1,80 @@
...
 export type StyledTableContainerProps = BoxProps & {
+  /**
+   * Controls the padding of the table cells
+   * @default 'default'
+   */
   density?: 'default' | 'compact'
+  /**
+   * Controls whether the rows of the table are striped using CSS. For virtualized tables, this should be set to false, since the
+   * row even/oddness is not guaranteed to be consistent as DOM nodes are loaded and unloaded.
+   * @default true
```
nickgros, line 13:
```suggestion
   * @default false
```

packages/synapse-react-client/src/components/styled/StyledTableContainer.tsx
```diff
@@ -1,64 +1,80 @@
...
-      userSelect: 'none',
-      touchAction: 'none',
-    },
-    '.resizer.isResizing': {
-      background: theme.palette.primary.main,
-      opacity: 1,
-    },
-    '@media (hover: hover)': {
+  styled(Box, {
+    shouldForwardProp: prop => prop !== 'density' && prop !== 'striped',
```
nickgros, line 21:
```suggestion
    shouldForwardProp: prop => prop !== 'density' && prop !== 'noStripedRows',
```

packages/synapse-react-client/src/components/styled/StyledTableContainer.tsx
```diff
@@ -1,64 +1,80 @@
...
-      userSelect: 'none',
-      touchAction: 'none',
-    },
-    '.resizer.isResizing': {
-      background: theme.palette.primary.main,
-      opacity: 1,
-    },
-    '@media (hover: hover)': {
+  styled(Box, {
+    shouldForwardProp: prop => prop !== 'density' && prop !== 'noStripedRows',
```
nickgros, line 21: Do not attach these props as DOM node attributes

packages/synapse-react-client/src/utils/slots/SlotUtils.ts
```diff
@@ -0,0 +1,60 @@
...
+    const slotProps2Fn: SlotPropsFunction<TArgs> =
+      typeof slotProps2 === 'function' ? slotProps2 : () => slotProps2
+
+    return function (...args: TArgs) {
+      return merge(slotProps1Fn(...args), slotProps2Fn(...args))
+    } as MergeReturn<TArgs, T1, T2>
+  } else {
+    return merge(slotProps1, slotProps2) as MergeReturn<TArgs, T1, T2>
+  }
+}
```
nickgros, line 60: This was a beast to get TypeScript to correctly infer types, took a fair amount of help from claude.ai

packages/synapse-react-client/src/components/styled/StyledTableContainer.tsx
```diff
@@ -1,64 +1,80 @@
...
+   * Controls the padding of the table cells
+   * @default 'default'
+   */
   density?: 'default' | 'compact'
+  /**
+   * Controls whether the rows of the table are striped using CSS. For virtualized tables, this should be set to false, since the
+   * row even/oddness is not guaranteed to be consistent as DOM nodes are loaded and unloaded.
+   * @default false
+   */
+  noStripedRows?: boolean
```
jay-hodgson, line 15: noStripedRows should default to false for virtualized tables (where the row even/odd is not guaranteed)?  I would expect false if this was "stripedRows" in this case, or true for noStripedRows.  Is my understanding flawed, or is the logic flipped in the field name?

packages/synapse-react-client/src/components/TanStackTable/types.ts
jay-hodgson: Thank you for pulling out these types into a separate file.  These make sense, but could be a cognitive burden if defined in the component files

packages/synapse-react-client/src/components/TanStackTable/StyledTanStackTable.tsx
jay-hodgson: This is my first exposure to the slots design pattern, and I can see the utility.  I think it was a good call to copy this pattern in this implementation

jay-hodgson (approved):

packages/synapse-react-client/src/components/styled/StyledTableContainer.tsx
nickgros: This is the shared `StyledTableContainer`. Most of our tables are not virtualized, so having it striped by default is the most typical case. The virtual table component can set this to true

PR #1486, opened 2025-01-07
PORTALS-2795 - Replace react-base-table usage with TanStack table

nickgros:
- Update EntityFinder's DetailsView to use StyledVirtualTanStackTable
- Update ChallengeDataDownload component to use EntityFinder components, remove custom components with slightly modified code
  - Updated DetailsView to support showing/hiding Add to Download Cart & Direct Download columns
- Update DatasetItemsEditor to use StyledVirtualTanStackTable
  - Fixed memoization issue in useSet
  - Fixed broken test by explicitly refetching the dataset on successful update

## Visual Changes

|Component|Old|New|
|---|---|---|
|Entity Finder| image | image |
|Dataset Editor|image | image |
|ChallengeDataDownload| image | image |


packages/synapse-react-client/src/components/ChallengeDataDownload/ChallengeDataDownload.tsx
```diff
@@ -70,13 +71,21 @@ export function ChallengeDataDownload({
...
+    ],
   }
   return (
     <>
       <Box>
-        <ChallengeEntityChildrenDetails
+        <EntityChildrenDetails
           parentContainerId={parentContainerId}
           {...sharedProps}
+          // TODO: replace the AddToDownloadList column with direct download column
```
nickgros, line 88:
```suggestion
```

packages/synapse-react-client/src/components/ChallengeDataDownload/ChallengeDataDownload.tsx
```diff
@@ -70,11 +71,18 @@ export function ChallengeDataDownload({
     isIdSelected,
     isSelectable: () => true,
     toggleSelection,
+    hiddenColumns: [
+      DetailsViewColumn.BADGES,
+      DetailsViewColumn.ADD_TO_DOWNLOAD_CART,
+      DetailsViewColumn.CREATED_ON,
+      DetailsViewColumn.VERSION,
+      DetailsViewColumn.MODIFIED_BY,
+    ],
```
nickgros, line 80: For the challenge portal version of this component, hide these columns (instead of maintaining tons of duplicate code!).

packages/synapse-react-client/src/components/EntityFinder/details/EntityDetailsList.tsx
```diff
@@ -55,6 +58,8 @@ export type EntityDetailsListSharedProps = {
   isSelectable: (header: EntityFinderHeader) => boolean
   toggleSelection: (entity: Reference | Reference[]) => void
   setCurrentContainer?: Dispatch<SetStateAction<EntityTreeContainer>>
+  /** Chosen columns to hide. Defaults to just the DirectDownload column. */
+  hiddenColumns?: DetailsViewColumn[]
```
nickgros, line 62: Exposed for challenge portal customization

packages/synapse-react-client/src/components/EntityFinder/details/view/DetailsView.tsx
```diff
@@ -252,119 +505,190 @@ export function DetailsView({
...
+            versionSelection !== VersionSelectionType.DISALLOWED,
+          [DetailsViewColumn.SIZE]: visibleTypes.includes(EntityType.FILE),
+          [DetailsViewColumn.MD5]: visibleTypes.includes(EntityType.FILE),
+          [DetailsViewColumn.ADD_TO_DOWNLOAD_CART]: visibleTypes.includes(
+            EntityType.FILE,
+          ),
+          [DetailsViewColumn.DIRECT_DOWNLOAD]: visibleTypes.includes(
+            EntityType.FILE,
+          ),
+          ...hideColumnOverrides,
```
nickgros, line 626: Columns passed via `hiddenColumns` prop will pass along show/hide state to `useReactTable` here

packages/synapse-react-client/src/components/EntityBadgeIcons/EntityBadgeIcons.test.tsx
```diff
@@ -1,43 +1,40 @@
-import { act, render, screen, waitFor } from '@testing-library/react'
```
nickgros, line 1: No logical changes to these tests. An unused prop was removed from the component under test. I moved this test file to the `src` folder and fixed type errors.

packages/synapse-react-client/src/components/EntityBadgeIcons/EntityBadgeIcons.tsx
```diff
@@ -68,8 +68,6 @@ export type EntityBadgeIconsProps = {
   onUnlinkError?: (error: unknown) => void
   /** Whether or not the badges (e.g. Annotations) can trigger opening a modal on click */
   canOpenModal: boolean
-  /** Whether this component should render a ReactTooltip or if an external component is managing it */
-  renderTooltipComponent: boolean
```
nickgros, line 72: Unused prop that dates back to when we used the `react-tooltip` library.

packages/synapse-react-client/src/components/EntityFinder/details/view/DetailsViewTableRenderers.tsx
nickgros, line 1: Moved all of these components to their own files, and modified them to use the TanStack table API. Minimal changes to rendered content (e.g. removed a wrapper `div` in a few places)

packages/synapse-react-client/src/components/EntityFinder/details/view/table/AddToDownloadListCell.tsx
```diff
@@ -0,0 +1,45 @@
...
+import { CellContext } from '@tanstack/react-table'
+import useGetEntityBundle from '../../../../../synapse-queries/entity/useEntityBundle'
+import { useAddFileToDownloadList } from '../../../../../synapse-queries/index'
+import { useSynapseContext } from '../../../../../utils/index'
+import { displayFilesWereAddedToDownloadListSuccess } from '../../../../download_list/DownloadConfirmationUtils'
+import IconSvg from '../../../../IconSvg/IconSvg'
+import { displayToast } from '../../../../ToastMessage/index'
+import { EntityFinderTableViewRowData } from '../DetailsView'
+
+export function AddFileToDownloadListCell(
```
nickgros, line 12: All of these cell renderers were moved into their own files to keep them simple & optimized for HMR.

They were changed to use the TanStack table API (which TypeScript will catch), but otherwise should not have many changes, so I don't think they need to be closely reviewed.

packages/synapse-react-client/src/components/SynapseTable/datasets/DatasetItemsEditor.test.tsx
```diff
@@ -136,22 +136,24 @@ async function clickRemove() {
...
-  onUnsavedChangesChange: mockOnUnsavedChangesFn,
-}
-
 async function renderComponent(wrapperProps?: SynapseContextType) {
-  const wrapper = render(<DatasetItemsEditor {...defaultProps} />, {
+  const user = userEvent.setup()
+
+  const mockOnUnsavedChangesFn = jest.fn()
+  const mockOnSaveFn = jest.fn()
+  const mockOnCloseFn = jest.fn()
```
nickgros, line 147: Fixed some flaky test issues by creating new mocks/`userEvent` instance each render. I think I had an issue where some actions in tests would fire after the test had ended and next test started, which made mock invocation counts incorrect.

packages/synapse-react-client/src/components/SynapseTable/datasets/DatasetItemsEditor.tsx
```diff
@@ -705,14 +788,18 @@ export function DatasetItemsEditor(props: DatasetItemsEditorProps) {
...
+            <Button
+              disabled={!datasetToUpdate || !hasChangedSinceLastSave}
+              variant="contained"
+              color="primary"
+              onClick={() => mutate(datasetToUpdate!)}
+            >
+              Save
+            </Button>
+          </div>
+        </Tooltip>
```
nickgros, line 802: Update 'Save' button to only enable when changes are pending. Add tooltip + div wrapper to tell the user why it's disabled.

packages/synapse-react-client/src/components/styled/StyledTableContainer.tsx
```diff
@@ -23,6 +23,9 @@ export const StyledTableContainer: StyledComponent<StyledTableContainerProps> =
     ({ theme, density = 'default', noStripedRows = false }) => ({
       overflow: 'auto',
       paddingLeft: '2px',
+      ['thead > tr']: {
+        backgroundColor: theme.palette.grey[200],
+      },
```
nickgros, line 28: In addition to the `th` cells, add a background color to the entire header row. This prevents a visual regression where the full table width is wider than the sum of all columns (react-base-table auto-sized the columns to take up full width, so this wasn't an issue before).

packages/synapse-react-client/src/components/EntityFinder/details/view/DetailsView.tsx
```diff
@@ -78,36 +109,271 @@ export type EntityFinderTableViewRowData = EntityFinderHeader & {
...
+      <span
+        role="link"
+        className="EntityFinderTableCellContainerLink"
+        onClick={e => {
+          e.stopPropagation()
+          setCurrentContainer(row.original.id)
+        }}
+      >
+        <Tooltip title={rowName}>
+          <span className="nameColumnCell">{rowName}</span>
```
jay-hodgson, line 132: Should these cell renderers be using MUI Typography?

jay-hodgson: (and if so, how much overhead does that introduce in the DOM?)

jay-hodgson: (was worried about scalability in the DOM, but remembered this is virtualized)

packages/synapse-react-client/src/components/EntityFinder/details/view/DetailsView.tsx
```diff
@@ -78,36 +109,271 @@ export type EntityFinderTableViewRowData = EntityFinderHeader & {
...
+    return null
+  }
+
+  return (
+    <Checkbox
+      inputProps={{ 'aria-label': 'Select All' }}
+      checked={checked}
+      disabled={disabled}
+      onChange={() => {
+        if (!disabled) {
```
jay-hodgson, line 162: Surprised you need to check for this, but it certainly doesn't hurt!

packages/synapse-react-client/src/components/EntityFinder/details/view/DetailsView.tsx
```diff
@@ -252,119 +504,190 @@ export function DetailsView({
...
+  const rowVirtualizer = useVirtualizer({
+    count: totalEntities ?? table.getRowModel().rows.length,
+    estimateSize: () => 40, // estimate row height for accurate scrollbar dragging
+    getScrollElement: () => tableContainerRef.current,
+    // measure dynamic row height, except in firefox because it measures table border height incorrectly
+    measureElement:
+      typeof window !== 'undefined' &&
+      navigator.userAgent.indexOf('Firefox') === -1
+        ? element => element?.getBoundingClientRect().height
+        : undefined,
```
jay-hodgson, line 641: Why must you be like this, FF? :)

packages/synapse-react-client/src/components/EntityFinder/details/view/DetailsView.tsx
```diff
@@ -380,209 +703,46 @@ export function DetailsView({
...
-          </BaseTable>
-        )}
-      </AutoResizer>
+            },
+          }}
+        />
+      )}
+      {isLoading && (
+        <div className="EntityFinderDetailsViewPlaceholder">
+          <SynapseSpinner size={30} />
```
jay-hodgson, line 738: Would a skeleton be preferred in this case?

jay-hodgson (approved):

packages/synapse-react-client/src/components/EntityFinder/details/view/DetailsView.tsx
nickgros: They could, but I don't think it affects the end result. I tend to use Typography when I need alternate variants or the content would benefit from standard margins/line height, neither of which are necessary here

packages/synapse-react-client/src/components/EntityFinder/details/view/DetailsView.tsx
nickgros: I can remove it--I think this is an artifact of some refactoring. This used to be wrapped in a div that had the handler, which had to check the `disabled` state.

packages/synapse-react-client/src/components/EntityFinder/details/view/DetailsView.tsx
nickgros: I think skeletons are always better than spinners! In this case, the editor area is already rendered with a defined height/width, so a skeleton wouldn't affect layout shift. So I don't think it's necessarily worth the effort to change it

PR #1488, opened 2025-01-08
PORTALS-3372: If LLFS_Metabolomics study key, redirect to LLFS

jay-hodgson:


apps/portals/elportal/src/pages/StudyDetailsPage.tsx
```diff
@@ -49,63 +50,71 @@ export const studyDetailsPageRoutes: RouteObject[] = [
   {
     path: studyDetailsPageTabConfig[0].path,
     element: (
-      <DetailsPageContent
```
jay-hodgson, line 52: No change to the DetailsPageContent, just wrapped in a react fragment

apps/portals/elportal/src/pages/StudyDetailsPage.tsx
```diff
@@ -49,63 +50,71 @@ export const studyDetailsPageRoutes: RouteObject[] = [
...
+                        searchParams={{ id: value }}
+                      />
+                    )
+                  }}
+                </DetailsPageContextConsumer>
+              ),
+            },
+          ]}
+        />
+        <RedirectToURL
```
jay-hodgson, line 113: Added a RedirectToURL that only does something if the search matches (LLFS_Metabolomics)

PR #1489, opened 2025-01-08
PORTALS-3305: New Featured Research ("What's New") component

kianamcc:
Jira Ticket: https://sagebionetworks.jira.com/browse/PORTALS-3305

Screenshot 2025-01-08 at 1 00 16 PM


packages/synapse-react-client/src/components/FeaturedResearch/FeaturedResearch.tsx
```diff
@@ -0,0 +1,361 @@
...
+  Typography,
+  Skeleton,
+} from '@mui/material'
+import {
+  FileHandleAssociateType,
+  FileHandleAssociation,
+  QueryBundleRequest,
+  Row,
+} from '@sage-bionetworks/synapse-types'
+import { SynapseConstants, SynapseUtilityFunctions } from '../../utils'
```
nickgros, line 15: This imports every single utility, which is unnecessary/slow. You can make these imports more specific:
```suggestion
import * as SynapseConstants from '../../utils/SynapseConstants'
import { parseEntityIdFromSqlStatement } from '../../utils/functions/SqlFunctions'
```

packages/synapse-react-client/src/components/FeaturedResearch/FeaturedResearch.tsx
```diff
@@ -0,0 +1,361 @@
...
+        </Link>
+      </Stack>
+    </div>
+  )
+}
+
+function FeaturedResearch(props: FeaturedResearchProps) {
+  const { sql } = props
+
+  const entityId = SynapseUtilityFunctions.parseEntityIdFromSqlStatement(sql)
```
nickgros, line 253: To correspond with import suggestion:
```suggestion
  const entityId = parseEntityIdFromSqlStatement(sql)
```

packages/synapse-react-client/src/components/FeaturedResearch/FeaturedResearch.tsx
```diff
@@ -0,0 +1,361 @@
...
+        }}
+      />
+    )
+  }
+  return (
+    <div>
+      <CardMedia
+        component="img"
+        image={url}
+        alt="Top Research Card Image"
```
nickgros, line 191: I like that you are looking at accessibility here! Unfortunately, this is not really useful alt text ([examples](https://accessibility.huit.harvard.edu/describe-content-images)). Is it possible to have the alt text in the table row to correspond with each image? If not, I suggest we hide it from screen readers
```suggestion
        aria-hidden="true"
```

packages/synapse-react-client/src/components/FeaturedResearch/FeaturedResearch.tsx
```diff
@@ -0,0 +1,361 @@
...
+        sx={{
+          objectFit: 'cover',
+          maxHeight: { xs: '300px', md: '600.3px' },
```
nickgros, line 194: 600.3 is oddly specific

packages/synapse-react-client/src/components/FeaturedResearch/FeaturedResearch.tsx
```diff
@@ -0,0 +1,361 @@
...
+          borderRadius: '10px',
+          marginBottom: '30px',
+        }}
+      />
+      <Stack
+        direction="row"
+        useFlexGap
```
nickgros, line 201: I didn't know about this! I just assumed that it was already using `gap`. We don't support any browsers that don't support `gap`, so if you want to enable `useFlexGap` by default in the theme, go for it

packages/synapse-react-client/src/components/FeaturedResearch/FeaturedResearch.tsx
```diff
@@ -0,0 +1,361 @@
...
+          lineHeight={'normal'}
+          sx={{
+            fontWeight: 700,
+            backgroundColor: 'primary.main',
+            color: '#FFFF',
+            padding: '3px 8px',
+            borderRadius: '3px',
+          }}
+        >
+          {parseTags(tagsColIndex, research)?.[0] || ''}
```
nickgros, line 217: Intentional to only show the first tag? Seems like we should show a chip for each one, but maybe this is intentional to keep the design simple

nickgros: Is it possible there are no tags? You might want to only render this `Typography` component if `parseTags(tagsColIndex, research)[0]` exists

packages/synapse-react-client/src/components/FeaturedResearch/FeaturedResearch.tsx
```diff
@@ -0,0 +1,361 @@
...
+  if (isLoading) {
+    return <Skeleton variant="rectangular" height={142} width="100%" />
+  }
+  return (
+    <Box
+      display={'flex'}
+      gap={'30px'}
+      borderBottom={'1px solid #EAECEE'}
+      padding={'24px 0'}
+      sx={{
```
nickgros, line 92:
```suggestion
      sx={{
        borderColor: 'grey.300',
```

packages/synapse-react-client/src/components/FeaturedResearch/FeaturedResearch.tsx
```diff
@@ -0,0 +1,361 @@
...
+  const fileId = research.values[imageColIndex] ?? ''
+  const url = useImageUrl(fileId ?? '', entityId)
+  if (isLoading) {
+    return <Skeleton variant="rectangular" height={142} width="100%" />
+  }
+  return (
+    <Box
+      display={'flex'}
+      gap={'30px'}
+      borderBottom={'1px solid #EAECEE'}
```
nickgros, line 90:
```suggestion
      borderBottom={'1px solid'}
```

packages/synapse-react-client/src/components/FeaturedResearch/FeaturedResearch.tsx
nickgros: Also seems like the `?` is unnecessary since `parseTags` always returns an array

packages/synapse-react-client/src/components/FeaturedResearch/FeaturedResearch.test.tsx
```diff
@@ -0,0 +1,192 @@
...
+    )
+  })
+
+  const renderWithRouter = (props: FeaturedResearchProps) => {
+    const router = createMemoryRouter([
+      {
+        path: '/',
+        element: <FeaturedResearch {...props} />,
+      },
+    ])
```
nickgros, line 159: Seems like the component is not using react-router, so you probably don't need to use it in tests

nickgros (commented):
Looks very good, I had some minor suggestions

packages/synapse-react-client/src/components/FeaturedResearch/FeaturedResearch.tsx
kianamcc: I took this specific number from the figma design! I went ahead and removed this.

nickgros (approved):
Looks great, nice work!

packages/synapse-react-client/src/components/FeaturedResearch/FeaturedResearch.tsx
```diff
@@ -0,0 +1,364 @@
...
+  const url = useImageUrl(fileId ?? '', entityId)
+  if (isLoading) {
+    return <Skeleton variant="rectangular" height={142} width="100%" />
+  }
+  return (
+    <Box
+      display={'flex'}
+      gap={'30px'}
+      borderBottom={'1px solid'}
+      padding={'24px 0'}
```
nickgros, line 92: FYI, I just learned these system props are deprecated in MUI v6, so you will want to get in the habit of putting these in `sx`. See https://mui.com/material-ui/migration/migrating-from-deprecated-apis/#system-props

nickgros: No need to change these in this PR, I can run a codemod to fix them in my future PR to update to MUI 6! But just an FYI

PR #1492, opened 2025-01-09
SWC-4693

nickgros:
Support drag-and-drop upload before opening modal

- Update useUploadFileEntities to indicate when required data has been fetched, so the caller knows when imperative upload can begin
- Add callback prop to EntityUpload/EntityUploadModal to indicate to SWC widget when imperative upload can begin
- EntityUploadModal changes to ensure upload tab is shown after imperative upload starts
- Use keepMounted dialog prop to ensure EntityUpload component is mounted to be ready for drag-and-drop

### DIFF
--- a/packages/synapse-react-client/src/utils/hooks/useUploadFileEntity/useUploadFileEntities.ts
+++ b/packages/synapse-react-client/src/utils/hooks/useUploadFileEntity/useUploadFileEntities.ts
@@ -96,6 +96,8 @@ export type UseUploadFileEntitiesReturn = {
   initiateUpload: (args: InitiateUploadArgs) => void
   activeUploadCount: number
   uploadProgress: FileUploadProgress[]
+  /** True when files can be uploaded. */
+  isUploadReady: boolean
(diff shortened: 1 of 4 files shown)


packages/synapse-react-client/src/utils/hooks/useUploadFileEntity/useUploadFileEntities.ts
```diff
@@ -358,10 +362,16 @@ export function useUploadFileEntities(
 
   const initiateUpload = useCallback(
     (args: InitiateUploadArgs) => {
+      if (uploadDestination == null) {
+        console.error(
+          'Upload destination was not loaded, or failed to load! Aborting upload.',
+        )
+        return
+      }
```
nickgros, line 370: Had issues where `uploadDestination` was not ready when we called `initiateUpload`. I added the callback to avoid this scenario. This should not happen now, but I added an error log just in case.

jay-hodgson (approved):

PR #1494, opened 2025-01-09
PORTALS-3350 - Upgrade react-router to v7

nickgros:
Followed the [migration guide](https://reactrouter.com/upgrading/v6)


apps/SageAccountWeb/src/App.tsx
```diff
@@ -127,15 +127,17 @@ function App() {
...
+          <Route
+            path="*"
+            element={
+              <>
+                <AuthenticatedRoutes />
+                <Footer />
+              </>
+            }
+          />
+        </Route>
```
nickgros, line 140: Had to nest `Route`s here in one of the v7 breaking changes. I think it only affects behavior if, within this route tree, we were to use a relative path in a Link component, which we do not do.

packages/synapse-react-client/package.json
```diff
@@ -249,8 +249,6 @@
     "build:umd": "vite build -c vite.config.umd.ts",
     "storybook:build": "pnpm storybook build",
     "type-check": "tsc --build",
-    "link-src": "pnpm link && cd ./node_modules/react && pnpm link && cd ../react-router && pnpm link && cd ../react-router-dom && pnpm link && cd ../../",
-    "unlink-src": "pnpm unlink && cd ./node_modules/react && pnpm unlink && cd ../react-router && pnpm unlink && cd ../react-router-dom && pnpm unlink && cd ../../",
```
nickgros, line 253: These haven't done anything since we moved this and portals to a pnpm monorepo.

Linking might still be useful for SWC, but today that would use `yarn`, not `pnpm`

apps/synapse-portal-framework/src/components/Explore/ExploreWrapperTabs.tsx
```diff
@@ -64,7 +64,9 @@ export function ExploreWrapperTabs(props: ExploreWrapperProps) {
             key={path}
             value={encodeURI(path)}
             label={displayName}
-            onClick={() => navigate(path)}
+            onClick={() => {
+              void navigate(path)
+            }}
```
nickgros, line 69: `navigate` now returns a promise, so `void` must be added to placate ESLint

jay-hodgson (approved):

PR #1498, opened 2025-01-13
SWC-7221: SRC UserCards don't stretch for 2-column grid layout on some devices

kianamcc:
Jira: https://sagebionetworks.jira.com/browse/SWC-7221

Screenshot 2025-01-13 at 9 57 10 AM


nickgros (approved):
You may also want to give the cards a bit of padding so the text/copy icon doesn't go all the way to the edge

PR #1500, opened 2025-01-14
SWC-7168

lakikowolfe:
Redesign DownloadCart on mobile

https://sagebionetworks.jira.com/jira/software/c/projects/SWC/issues/SWC-7168?jql=project%20%3D%20%22SWC%22%20AND%20assignee%20%3D%20currentUser%28%29%20ORDER%20BY%20created%20DESC

Ready for a review but I still plan to consolidate my media queries and add some comments. And I can't figure out why the buttons aren't centered.

See ticket for screenshots


nickgros (commented):
To center the buttons you can either wrap it in a flex container and do 

```css
flex-direction: column;
align-items: center;
```

or change the button style to
```css
display: flex; // it's currently inline-flex
margin: auto;
```

packages/synapse-react-client/src/style/components/_download-cart-page.scss
```diff
@@ -26,35 +35,51 @@
...
+      height: auto;
+      flex-direction: column;
+
+      .nav-item button {
+        font-size: 15px;
+      }
+    }
   }
+
   .availableForDownloadTableContainer {
```
lakikowolfe, line 83: Would you do anything to the styling here?

To me it would look nicer to center the filtering drop down

```
    @media (max-width: 768px) {
      .filterFilesContainer {
        display: flex;
        justify-content: center;
      }
    }
```

What do you think @nickgros

nickgros: Sounds like a good idea to me! I haven't looked at this too closely but it seems like we don't necessarily need the grid. If it makes sense, you may be able to simplify this if you remove the grid and replace it with  `flex` and set `justify-content: flex-end` outside of the media query, then just override the `justify-content` in your media query.

If you think it looks better, you could further enhance this so the dropdown button takes up the full width, e.g.:

```scss
.filterFilesContainer {
  justify-content: center;
  gap: 16px;
  .dropdown {
    flex: 1;
    .dropdown-toggle {
      width: 100%
    }
  }
}
```

nickgros: We also have this ticket: https://sagebionetworks.jira.com/browse/SWC-6543 - not trying to scope-creep but if you're digging deep into this layout, you could consider also fixing it here.

Instead of going to flex, you could see if it makes sense to keep the grid, and tweak the CSS it to fix this bug at wider screen sizes and get your preferred layout at narrow widths -- up to you.

lakikowolfe: I'm going to call this ticket done and open a separate PR for SWC-6543

nickgros (approved):
This looks good! Can you create a new ticket to add mobile support to the "Actions Required" list (the first tab on the download cart page) and link it to SWC-7168? I think it is important that we do that, but that work certainly doesn't block merging these improvements

lakikowolfe:
> This looks good! Can you create a new ticket to add mobile support to the "Actions Required" list (the first tab on the download cart page) and link it to [SWC-7168](https://sagebionetworks.jira.com/browse/SWC-7168)? I think it is important that we do that, but that work certainly doesn't block merging these improvements

Done! Anthony is assigned to the actions required work right now

PR #1501, opened 2025-01-15
SWC-7225

jay-hodgson:
Output.  Note the Available Filters text, the spacing and alignment of facet filter controls, as well as position of the available facets list:

image

image

### DIFF
--- a/packages/synapse-react-client/src/components/widgets/query-filter/FacetFilterControls.tsx
+++ b/packages/synapse-react-client/src/components/widgets/query-filter/FacetFilterControls.tsx
@@ -221,6 +221,25 @@ function FacetFilterControls(props: FacetFilterControlsProps) {
 
   return (
     <div className={`FacetFilterControls`}>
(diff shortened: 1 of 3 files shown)


packages/synapse-react-client/src/components/widgets/query-filter/FacetFilterControls.tsx
```diff
@@ -221,6 +221,25 @@ function FacetFilterControls(props: FacetFilterControlsProps) {
 
   return (
     <div className={`FacetFilterControls`}>
+      <div>
```
jay-hodgson, line 224: Moved up.  Only changed the header label (to Available Filters)

nickgros: also an elegant way to fix an annoying layout shift issue, where filter sections would appear/disappear as each chip was clicked!

PR #1507, opened 2025-01-21
PORTALS-3389: only update tab configuration if the count for that tab…

jay-hodgson:
… changed, and share the onQueryResultBundleChange handler function


nickgros (approved):
Not clear to me why `onQueryResultBundleChange` was called so often, but this change makes sense to reduce the number of renders

PR #1508, opened 2025-01-21
ADM-4286: Include UpsetPlot in ELITE portal homepage v2

jay-hodgson:


jay-hodgson:
ELITE portal configuration change only

PR #1533, opened 2025-01-30
PORTALS-3398: fix type issue

jay-hodgson:
Error does not show up locally (on build or in the IDE), so it's difficult to test.  This is my initial attempt to fix


jay-hodgson:
Running `pnpm type-check` I could test the issue locally.  I ended up casting it to the type that was being reported by the typecheck.  No change in runtime behavior

nickgros (approved):

PR #1537, opened 2025-01-30
PORTALS-2651 - Forums/Threads

nickgros:
- Remove react-bootstrap from forums components

### DIFF
--- a/packages/synapse-react-client/src/components/Forum/DiscussionThread.tsx
+++ b/packages/synapse-react-client/src/components/Forum/DiscussionThread.tsx
@@ -1,34 +1,33 @@
+import { Box, Button, TextField, Typography } from '@mui/material'
+import {
+  ALL_ENTITY_BUNDLE_FIELDS,
+  ObjectType,
+  SubscriptionObjectType,
+} from '@sage-bionetworks/synapse-types'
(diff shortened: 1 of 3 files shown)


packages/synapse-react-client/src/components/Forum/DiscussionThread.tsx
```diff
@@ -103,7 +102,7 @@ export function DiscussionThread(props: DiscussionThreadProps) {
   const replies = replyData?.pages.flatMap(page => page.results) ?? []
 
   return (
-    <div className="bootstrap-4-backport DiscussionThread">
+    <div className="DiscussionThread">
```
nickgros, line 105: image

packages/synapse-react-client/src/components/Forum/ForumThreadEditor.tsx
```diff
@@ -107,10 +106,11 @@ export function ForumThreadEditor(props: ForumThreadEditorProps) {
   }
 
   const editorContent = (
-    <div className="bootstrap-4-backport">
+    <div>
```
nickgros, line 109: image

jay-hodgson (approved):

PR #1541, opened 2025-01-31
SWC-7236

nickgros:
- Fix column resizing issue
- SynapseTable: set initial column width to 200 (default is 150)

### DIFF
--- a/packages/synapse-react-client/src/components/SynapseTable/SynapseTable.tsx
+++ b/packages/synapse-react-client/src/components/SynapseTable/SynapseTable.tsx
@@ -153,6 +153,7 @@ export function SynapseTable(props: SynapseTableProps) {
           id: selectColumn.name,
           enableSorting: isSortableColumn(selectColumn.columnType),
           enableResizing: true,
+          size: 200,
           header: TableDataColumnHeader,
           cell: TableDataCell,
         })
--- a/packages/synapse-react-client/src/components/TanStackTable/StyledTanStackTable.tsx
+++ b/packages/synapse-react-client/src/components/TanStackTable/StyledTanStackTable.tsx
@@ -73,10 +73,15 @@ export default function StyledTanStackTable<
         header.column.getSize()
     }
     return colSizes
+    // Intentionally limit the dependencies to only recompute when the column sizes change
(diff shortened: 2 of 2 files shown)


packages/synapse-react-client/src/components/SynapseTable/SynapseTable.tsx
```diff
@@ -153,6 +153,7 @@ export function SynapseTable(props: SynapseTableProps) {
           id: selectColumn.name,
           enableSorting: isSortableColumn(selectColumn.columnType),
           enableResizing: true,
+          size: 200,
```
nickgros, line 156: For Synapse Tables, make the initial column width a bit wider per request

packages/synapse-react-client/src/components/TanStackTable/StyledTanStackTable.tsx
```diff
@@ -73,10 +73,15 @@ export default function StyledTanStackTable<
         header.column.getSize()
     }
     return colSizes
+    // Intentionally limit the dependencies to only recompute when the column sizes change
     // eslint-disable-next-line react-hooks/exhaustive-deps
-  }, [table.getState().columnSizingInfo, table.getState().columnSizing])
+  }, [
+    table.getState().columnSizingInfo,
+    table.getState().columnSizing,
+    table.getState().columnVisibility, // If a column is added, its width should be recalculated
```
nickgros, line 81: Fixes an occasional bug where these CSS variables would be initialized before SynapseTable had configured its visible columns.

packages/synapse-react-client/src/components/TanStackTable/StyledTanStackTable.tsx
```diff
@@ -89,6 +94,7 @@ export default function StyledTanStackTable<
         {...tableSlotProps}
         style={{
           ...columnSizeVars,
+          tableLayout: 'fixed',
```
nickgros, line 97: Fixes the re-sizing issue. This seems to work well for both dynamic-width tables and fullWidth (100%) tables

jay-hodgson (approved):

PR #1547, opened 2025-02-03
PORTALS-2651

nickgros:
Remove bootstrap 4 and react-bootstrap


packages/synapse-react-client/src/components/EntityUpload/EntityUpload.tsx
```diff
@@ -201,7 +201,6 @@ export const EntityUpload = forwardRef(function EntityUpload(
             />
             {/* File input */}
             <input
-              className="form-control"
               type="file"
               id="filesToUpload"
               name="file upload"
```
nickgros, line 206: These `input`s with `form-control` are not visible, so removing the class doesn't matter

packages/synapse-react-client/src/style/base/_core.scss
```diff
@@ -721,15 +720,13 @@ main,
...
+    &:active,
+    &:link {
       @include SRC_Mixins.ignoreLink;
-      &:visited,
-      &:active,
-      &:link {
-        @include SRC_Mixins.ignoreLink;
-      }
     }
   }
```
nickgros, line 731: This selector touches a lot, and isn't bootstrap-specific, so I left it--even though I want to remove it.

packages/synapse-react-client/package.json
```diff
@@ -62,11 +62,9 @@
     "@tanstack/react-virtual": "^3.11.1",
     "@upsetjs/react": "^1.11.0",
     "animate.css": "^4.1.1",
-    "bootstrap": "^4.6.2",
     "classnames": "^2.5.1",
     "dagre": "^0.8.5",
     "dayjs": "^1.11.10",
-    "downshift": "^6.1.12",
```
nickgros, line 69: was previously used in `UserSearchBox` (v1), is no longer used

packages/synapse-react-client/package.json
```diff
@@ -88,11 +86,8 @@
     "plotly.js": "2.28.0",
     "plotly.js-basic-dist": "^2.30.0",
     "pluralize": "^8.0.0",
-    "prop-types": "^15.8.1",
```
nickgros, line 91: Not used, plus not needed since we use TypeScript

jay-hodgson (approved):

PR #1548, opened 2025-02-03
Consolidate calls to createPlotlyComponent

nickgros:
Some bundlers (like esbuild) do not produce expected output when bundling react-plotly.js/factory (I think it has unusual CJS syntax). I consolidated usage of this to one module to simplify debugging, but it's a generally-useful refactor.


packages/synapse-react-client/src/components/Plot/Plot.stories.ts
```diff
@@ -1,11 +1,11 @@
 import { Meta, StoryObj } from '@storybook/react'
-import PlotlyWrapper from '../PlotlyWrapper'
+import Plot from './Plot'
 import rawData from '../../mocks/distribution_data.json'
 import Plotly from 'plotly.js-basic-dist'
 
 const meta = {
-  title: 'Components/PlotlyWrapper',
-  component: PlotlyWrapper,
+  title: 'Components/PlotlyPlot',
```
nickgros, line 7: Update story to just use the Plot component instead of the unused PlotlyWrapper component

jay-hodgson (approved):

packages/synapse-react-client/src/umd.index.ts
```diff
@@ -143,7 +142,6 @@ const SynapseComponents = {
   displayToast,
   IconSvg,
   UserProfileLinks,
-  PlotlyWrapper,
```
jay-hodgson, line 146: I think we ran a POC with utilizing this in Agora at one point, but I think they're only using SRC for rendering Synapse Wikis now (@sagely1 or @hallieswan , could you confirm?)

nickgros: [A search in Sage-Bionetworks/sage-monorepo](https://github.com/search?q=repo%3ASage-Bionetworks%2Fsage-monorepo%20synapse-react-client&type=code) tells me that no projects there, including Agora, are currently using `synapse-react-client`.

PR #1549, opened 2025-02-03
SWC-7237

nickgros:
Fix issue where we improperly indicated that we will not force the user to re-accept the ToU after they are redirected to show the ToU.

### DIFF
--- a/apps/SageAccountWeb/src/hooks/useMaybeRedirectToSignTermsOfService.test.ts
+++ b/apps/SageAccountWeb/src/hooks/useMaybeRedirectToSignTermsOfService.test.ts
@@ -1,0 +1,214 @@
+import { TermsOfServiceState } from '@sage-bionetworks/synapse-types'
+import { renderHook, waitFor } from '@testing-library/react'
+import { useApplicationSessionContext } from 'synapse-react-client'
+import { vi } from 'vitest'
+import useMaybeRedirectToSignTermsOfService, {
+  SKIPPED_SIGNING_TOS_SESSIONSTORAGE_KEY,
+} from './useMaybeRedirectToSignTermsOfService'
+import { useLocation, useNavigate, Location } from 'react-router'
+
+vi.mock('synapse-react-client', () => {
+  return {
+    storeLastPlace: vi.fn(),
+    useApplicationSessionContext: vi.fn(),
+  }
+})
+
+vi.mock('react-router', () => {
+  return {
+    useLocation: vi.fn(),
+    useNavigate: vi.fn(),
+  }
+})
+const mockLocation: Location = {
+  key: '',
+  state: undefined,
+  pathname: '/',
+  search: '',
+  hash: '',
+}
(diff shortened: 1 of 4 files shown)


apps/SageAccountWeb/src/hooks/useMaybeRedirectToSignTermsOfService.test.ts
```diff
@@ -0,0 +1,214 @@
...
+
+    const hook = renderHook(() => useMaybeRedirectToSignTermsOfService())
+    await waitFor(() => {
+      expect(hook.result.current.mayPromptTermsOfUse).toBe(true)
+      expect(mockNavigate).toHaveBeenCalledWith(
+        '/authenticated/signUpdatedTermsOfUse',
+      )
+    })
+  })
+  it('redirects to signTermsOfUse when user ToS is MUST_AGREE_NOW and has never signed ToU', async () => {
```
nickgros, line 141:
```suggestion
  it('redirects to signTermsOfUse when user ToS is MUST_AGREE_NOW and has signed ToU', async () => {
```

apps/SageAccountWeb/src/hooks/useMaybeRedirectToSignTermsOfService.test.ts
```diff
@@ -0,0 +1,214 @@
...
+    })
+
+    const hook = renderHook(() => useMaybeRedirectToSignTermsOfService())
+    await waitFor(() => {
+      expect(hook.result.current.mayPromptTermsOfUse).toBe(true)
+      expect(mockNavigate).toHaveBeenCalledWith(
+        '/authenticated/signUpdatedTermsOfUse',
+      )
+    })
+  })
```
nickgros, line 161: This case is closest to the scenario reported by the user. The assertion `expect(hook.result.current.mayPromptTermsOfUse).toBe(true)` fails with the old logic and passes with the new logic.

apps/SageAccountWeb/src/hooks/useMaybeRedirectToSignTermsOfService.ts
```diff
@@ -44,8 +46,9 @@ export default function useMaybeRedirectToSignTermsOfService(): UseMaybeRedirect
           storeLastPlace()
           navigate('/authenticated/signUpdatedTermsOfUse')
         }
+      } else {
+        setMayPromptTermsOfUse(false)
       }
-      setMayRedirect(false)
     }
```
nickgros, line 52: Wrapping in `else` fixes the bug.

apps/SageAccountWeb/src/hooks/useMaybeRedirectToSignTermsOfService.test.ts
```diff
@@ -0,0 +1,204 @@
...
```
nickgros, line 157: This case is closest to the scenario reported by the user. The assertion `expect(hook.result.current.mayPromptTermsOfUse).toBe(true)` fails with the old logic and passes with the new logic.

jay-hodgson (approved):

PR #1552, opened 2025-02-05
PORTALS-3209: "Cite As" button for cards

kianamcc:
Jira ticket: https://sagebionetworks.jira.com/browse/PORTALS-3209

Cite As Popover:
Screenshot 2025-02-04 at 9 40 29 AM

Cite As button:
Screenshot 2025-02-04 at 12 18 30 PM

Cite As Popover on mobile:
Screenshot 2025-02-04 at 12 20 05 PM


packages/synapse-react-client/src/components/CitationPopover/CitationPopover.tsx
```diff
@@ -0,0 +1,335 @@
...
+          )
+        }
+        const citationText = await response.text()
+        setCitation(citationText)
+      } catch (err) {
+        console.log(err)
+      } finally {
+        setLoading(false)
+      }
+    }
```
nickgros, line 151: Can you encapsulate the logic to retrieve a query into a custom hook? Both for testability and reusability with different UI in the future.

Also, please use `useQuery` from `@tanstack/react-query` to fetch data asynchronously, since that will manage providing loading & error states that you can handle in your code.

packages/synapse-react-client/src/components/CitationPopover/CitationPopover.tsx
```diff
@@ -0,0 +1,325 @@
...
+  const open = Boolean(anchorEl)
+  const id = open ? 'cite-as-popover' : undefined
+
+  const useCitation = (doi: string, format: string) => {
+    return useQuery({
+      queryKey: ['citation', doi, format],
+      queryFn: () => fetchCitation(doi, format),
+      enabled: !!doi && !!format && !!open,
+    })
+  }
```
nickgros, line 153: Please put this & `fetchCitation` in their own `useCitation.ts` file and import here

packages/synapse-react-client/src/components/CitationPopover/CitationPopover.tsx
```diff
@@ -0,0 +1,325 @@
...
+        transformOrigin={{
+          vertical: 'top',
+          horizontal: 'left',
+        }}
+        slotProps={{
+          paper: {
+            sx: theme => ({
+              postion: 'relative',
+              borderRadius: 0,
+              boxShadow: '0px 8px 24px 0px rgba(53, 58, 63, 0.15)',
```
nickgros, line 200: Is there a big difference between this and the default `boxShadow`?

packages/synapse-react-client/src/components/CitationPopover/CitationPopover.tsx
```diff
@@ -0,0 +1,325 @@
...
+        anchorOrigin={{
+          vertical: 'bottom',
+          horizontal: 'left',
+        }}
+        transformOrigin={{
+          vertical: 'top',
+          horizontal: 'left',
+        }}
+        slotProps={{
+          paper: {
```
nickgros, line 196: Feels like this should have a max width at sizes >= md to prevent taking up a ton of space

packages/synapse-react-client/src/components/CitationPopover/CitationPopover.tsx
```diff
@@ -0,0 +1,325 @@
...
+                color: '#FFF !important',
+                whiteSpace: 'nowrap',
+                '&:hover': {
+                  backgroundColor: darken(theme.palette.primary.main, 0.2),
+                  textDecoration: 'none',
+                },
+              }}
+            >
+              Download Citation
+            </Button>
```
nickgros, line 303: Why can't we just use the variant='contained' prop instead of adding all of these custom colors and styles? This feels like it should just be

```suggestion
            <Button
              onClick={() => handleDownload(citation || '')}
              sx={{
                flex: 1,
                whiteSpace: 'nowrap',
              }}
            >
              Download Citation
            </Button>
```

packages/synapse-react-client/src/components/CitationPopover/CitationPopover.tsx
```diff
@@ -0,0 +1,325 @@
...
+      <path
+        d="M1.2002 6.85162H3.497C3.497 7.82963 3.09981 8.71402 2.4614 9.35362C2.45901 9.356 2.45421 9.3608 2.45179 9.36322V10.3364C3.18859 10.2116 3.84859 9.85881 4.35379 9.35361C4.99457 8.71283 5.38939 7.82961 5.38939 6.85161L5.38943 2.66241H1.20022L1.2002 6.85162Z"
+        fill="#395979"
+      />
+      <path
+        d="M6.61035 2.66242V6.85162H8.90715C8.90715 7.82963 8.50996 8.71402 7.87155 9.35362C7.86917 9.356 7.86436 9.3608 7.86194 9.36322V10.3364C8.59874 10.2116 9.25874 9.85881 9.76394 9.35361C10.4048 8.714 10.7995 7.83081 10.7995 6.85161V2.66241L6.61035 2.66242Z"
+        fill="#395979"
+      />
+    </svg>
+  )
```
nickgros, line 76: Move to its own file in `src/assets`

packages/synapse-react-client/src/components/CitationPopover/CitationPopover.tsx
```diff
@@ -0,0 +1,325 @@
...
+            borderRadius: '3px',
+            border: '1px solid',
+            borderColor: 'grey.400',
+          }}
+        >
+          <Box sx={{ display: 'flex' }}>{quotationSvg}</Box>
+          <Typography variant="label" sx={{ lineHeight: '20px' }}>
+            Cite As
+          </Typography>
+        </Button>
```
nickgros, line 179: Use props `variant='outlined'` and `startIcon={quotationSvg}` and remove extraneous styles

packages/synapse-react-client/src/components/GenericCard/GenericCard.tsx
```diff
@@ -43,6 +43,8 @@ import {
 import { useGetEntity } from '../../synapse-queries'
 import { useQueryContext } from '../QueryContext'
 import { convertDoiToLink } from '../../utils/functions/RegularExpressions'
+import { Stack } from '@mui/system'
```
nickgros, line 46:
```suggestion
import { Stack } from '@mui/material'
```

packages/synapse-react-client/src/components/CitationPopover/CitationPopover.tsx
```diff
@@ -0,0 +1,325 @@
...
+  title,
+  boilerplateText,
+}: CitationPopoverProps) {
+  const [anchorEl, setAnchorEl] = useState<HTMLButtonElement | null>(null)
+  const [citationFormat, setCitationFormat] = useState('bibtex')
+  const theme = useTheme()
+
+  const selectSx = {
+    width: '100%',
+    backgroundColor: '#F1F3F5',
```
nickgros, line 37: isn't this already the default background color?

packages/synapse-react-client/src/components/GenericCard/GenericCard.tsx
```diff
@@ -59,6 +61,8 @@ export type GenericCardSchema = {
   title: string
   subTitle?: string
   description?: string
+  includeCitation?: boolean
+  citationBoilerplateText?: string
```
nickgros, line 65: the presence / formatting of this text seems like it would just make copying and pasting the actual citation hard & confusing.

packages/synapse-react-client/src/components/CitationPopover/CitationPopover.test.tsx
```diff
@@ -0,0 +1,166 @@
...
+  Promise.resolve({
+    ok: true,
+    status: 200,
+    statusText: 'OK',
+    text: () =>
+      Promise.resolve(
+        '@misc{test2025,\n  title = {Some BibTeX Entry},\n  year = {2025}\n}',
+      ),
+  }),
+) as jest.Mock
```
nickgros, line 21: If you're mocking `useQuery` then you don't need to mock `fetch`. 

I think ultimately you will want to mock `useCitation` after you extract that and import it separately.

packages/synapse-react-client/src/components/CitationPopover/CitationPopover.test.tsx
```diff
@@ -0,0 +1,166 @@
...
+
+const openPopover = () => {
+  const button = screen.getByText('Cite As')
+  fireEvent.click(button)
```
nickgros, line 25: Please use `click` from user event and avoid using `fireEvent.click` unless absolutely necessary (across this whole test file)

[More info](https://testing-library.com/docs/user-event/intro/#differences-from-fireevent)

packages/synapse-react-client/src/components/CitationPopover/CitationPopover.tsx
kianamcc: I would say there's enough of a difference:

Screenshot 2025-02-05 at 2 11 40 PM
Screenshot 2025-02-05 at 2 11 34 PM

packages/synapse-react-client/src/components/CitationPopover/CitationPopover.tsx
kianamcc: This should have had a fixed width of 500px according to the design so I added that here.

packages/synapse-react-client/src/components/GenericCard/GenericCard.tsx
kianamcc: It was mentioned by Jineta to have this boilerplate text and it's also in the designs as well. However, I don't include this in the actual citation that users can copy/download since if I were to do that I would probably need to find out where and how to incorporate this text in each format type we're supporting. Should I just remove this prop (citationBoilerplateText) and make note of it in the jira ticket?

packages/synapse-react-client/src/components/CitationPopover/CitationPopover.test.tsx
kianamcc: whoops, forgot to remove this

packages/synapse-react-client/src/components/GenericCard/GenericCard.tsx
nickgros: You can leave it for now. My issue is with the design, not so much your implementation.

packages/synapse-react-client/src/assets/icons/DoubleQuotes.tsx
```diff
@@ -0,0 +1,20 @@
+export const DoubleQuotes = (
```
nickgros, line 1: Would prefer if this was just an svg file, since React isn't really doing anything special here.

packages/synapse-react-client/src/components/CitationPopover/CitationPopover.test.tsx
```diff
@@ -0,0 +1,194 @@
...
+import userEvent from '@testing-library/user-event'
+import { UseQueryResult } from '@tanstack/react-query'
+import CitationPopover from './CitationPopover'
+import { useCitation } from './useCitation'
+
+jest.mock('./useCitation', () => ({
+  useCitation: jest.fn(),
+}))
+
+const mockUseCitation = useCitation as jest.MockedFunction<typeof useCitation>
```
nickgros, line 11:
```suggestion
const mockUseCitation = jest.mocked(useCitation)
```

packages/synapse-react-client/src/components/CitationPopover/CitationPopover.test.tsx
```diff
@@ -0,0 +1,194 @@
...
+
+const openPopover = async () => {
+  const button = screen.getByTestId('CiteAsButton')
+  await act(async () => {
+    await userEvent.click(button)
+  })
```
nickgros, line 17: No need to wrap in `act` when using `userEvent`
```suggestion
  await userEvent.click(button)
```

packages/synapse-react-client/src/components/CitationPopover/CitationPopover.test.tsx
```diff
@@ -0,0 +1,194 @@
...
+      isLoading: false,
+      data: data,
+      error: null,
+      isError: false,
+      isSuccess: true,
+    } as UseQueryResult<string, Error>)
+  })
+  afterEach(() => {
+    cleanup()
+  })
```
nickgros, line 42: This shouldn't be necessary, I'm like 90% sure `cleanup` is already called automatically

```suggestion
```

packages/synapse-react-client/src/components/CitationPopover/CitationPopover.test.tsx
```diff
@@ -0,0 +1,194 @@
...
+
+  it('renders button', () => {
+    render(<CitationPopover {...mockProps} />)
+    const button = screen.queryByTestId('CiteAsButton')
+    expect(button).toBeInTheDocument()
```
nickgros, line 47: Will this work? 2 notes:
1. Prefer `xxxByRole` over `xxxByTestId` -- https://testing-library.com/docs/queries/about#priority
2. `findByXxx` will fail if it doesn't find the element, so it's more succinct than `queryByXxx` followed by `toBeInTheDocument`

```suggestion
    await screen.findByRole('button', { name: 'Cite As' })
```

packages/synapse-react-client/src/components/CitationPopover/CitationPopover.test.tsx
```diff
@@ -0,0 +1,194 @@
...
+        'bibtex',
+        true,
+      )
+    })
+
+    await waitFor(() => {
+      expect(
+        screen.getByText(content => content.includes('Some BibTeX Entry')),
+      ).toBeInTheDocument()
+    })
```
nickgros, line 70: More succinct

```suggestion
    await screen.findByText(content => content.includes('Some BibTeX Entry'))
```

But I have never seen this pattern. What's the difference between `getByText('foo')` and `getByText(content => content.includes('foo'))`?

packages/synapse-react-client/src/components/CitationPopover/CitationPopover.test.tsx
```diff
@@ -0,0 +1,194 @@
...
+    openPopover()
+
+    await waitFor(() => {
+      expect(screen.getByTestId('CiteAsPopover')).toBeInTheDocument()
+    })
+
+    const select = screen.getByRole('combobox')
+    await act(async () => {
+      await userEvent.click(select)
+    })
```
nickgros, line 84: Remove `act` wrapper

packages/synapse-react-client/src/components/CitationPopover/CitationPopover.test.tsx
```diff
@@ -0,0 +1,194 @@
...
+  })
+
+  it('shows menu options when select button is clicked with bibtex as default', async () => {
+    render(<CitationPopover {...mockProps} />)
+    openPopover()
+
+    await waitFor(() => {
+      expect(screen.getByTestId('CiteAsPopover')).toBeInTheDocument()
+    })
```
nickgros, line 79: `findByRole`

packages/synapse-react-client/src/components/CitationPopover/CitationPopover.test.tsx
```diff
@@ -0,0 +1,194 @@
...
+
+  it('displays boilerplate text when available', async () => {
+    render(<CitationPopover {...mockProps} />)
+    openPopover()
+
+    await waitFor(() => {
+      expect(
+        screen.getByText(content => content.includes('Some boilerplate text')),
+      ).toBeInTheDocument()
+    })
```
nickgros, line 104: As above, use `findByText`, no need to check `toBeInTheDocument`

packages/synapse-react-client/src/components/CitationPopover/CitationPopover.test.tsx
```diff
@@ -0,0 +1,194 @@
...
+    render(<CitationPopover {...mockProps} />)
+    const mockWriteText = jest.fn().mockResolvedValue('copied')
+    Object.assign(navigator, {
+      clipboard: { writeText: mockWriteText },
+    })
+    openPopover()
+
+    await waitFor(() => {
+      expect(screen.getByTestId('CiteAsPopover')).toBeInTheDocument()
+    })
```
nickgros, line 117: As above, use `findByRole`

packages/synapse-react-client/src/components/CitationPopover/CitationPopover.test.tsx
```diff
@@ -0,0 +1,194 @@
...
+
+    await waitFor(() => {
+      expect(
+        screen.getByText(content => content.includes('Some BibTeX Entry')),
+      ).toBeInTheDocument()
+    })
```
nickgros, line 123: As above, use `findByText`

packages/synapse-react-client/src/components/CitationPopover/CitationPopover.test.tsx
```diff
@@ -0,0 +1,194 @@
...
+
+    const copyButton = screen.getByTestId('CopyButton')
```
nickgros, line 125: As above, use `getByRole`

packages/synapse-react-client/src/components/CitationPopover/CitationPopover.test.tsx
```diff
@@ -0,0 +1,194 @@
...
+    await act(async () => {
+      await userEvent.click(copyButton)
+    })
```
nickgros, line 128: Remove `act` wrapper

packages/synapse-react-client/src/components/CitationPopover/CitationPopover.test.tsx
```diff
@@ -0,0 +1,194 @@
...
+
+    await waitFor(() => {
+      expect(
+        screen.getByText(content => content.includes('Some BibTeX Entry')),
+      ).toBeInTheDocument()
+    })
+
+    const createElementSpy = jest
+      .spyOn(document, 'createElement')
+      .mockReturnValue(document.createElement('a'))
```
nickgros, line 151: Do you need this `mockReturnValue` if you're just going to call the method itself?

```suggestion
    const createElementSpy = jest
      .spyOn(document, 'createElement')
```

packages/synapse-react-client/src/components/CitationPopover/useCitation.tsx
```diff
@@ -0,0 +1,35 @@
...
+    if (!response.ok) {
+      throw new Error(
+        `Failed to fetch citation: ${response.status}: ${response.statusText}`,
+      )
+    }
+    return await response.text()
+  } catch (err) {
+    console.log(err)
+    return ''
+  }
```
nickgros, line 26: Since this is an external API that we don't control, I think we need to do more than log and silently fail by returning an empty string. `useQuery` can handle catching a thrown error. I think you could remove the `try`/`catch` and let the component handle error UI
```suggestion
  const response = await fetch(
    `https://citation.doi.org/format?doi=${formattedDoi}&style=${format}&lang=en-US`,
    {
      headers: {
        Accept: `text/x-${format}`,
      },
    },
  )
  if (!response.ok) {
    throw new Error(
      `Failed to fetch citation: ${response.status}: ${response.statusText}`,
    )
  }
  return await response.text()
```

packages/synapse-react-client/src/components/CitationPopover/CitationPopover.test.tsx
kianamcc: ah I missed this where userEvents internally wrap these actions with wrap!

packages/synapse-react-client/src/components/CitationPopover/CitationPopover.test.tsx
kianamcc: Checks for any element with the substring!

packages/synapse-react-client/src/components/CitationPopover/CitationPopover.tsx
```diff
@@ -0,0 +1,267 @@
...
+
+  const open = Boolean(anchorEl)
+  const id = open ? 'cite-as-popover' : undefined
+
+  const { data: citation, isLoading } = useCitation(
+    doi || '',
+    citationFormat,
+    open,
+  )
+
```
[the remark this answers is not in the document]
kianamcc: I added an Alert inside the popover to display the error message!

packages/synapse-react-client/src/components/CitationPopover/CitationPopover.tsx
kianamcc: I did elevation: 9:
Screenshot 2025-02-06 at 3 56 15 PM

At lower levels it looks too sharp but at higher levels the "blurs" look softer but also darker.

packages/synapse-react-client/src/components/GenericCard/GenericCard.tsx
```diff
@@ -617,6 +636,22 @@ class _GenericCard extends Component<GenericCardPropsInternal> {
               </Box>
             )}
           </div>
+          {includeCitation && (
+            <Box
+              sx={{
+                display: 'flex',
```
nickgros, line 642: It doesn't seem like flex properties are being used, but you can keep this if I'm wrong
```suggestion
```

packages/synapse-react-client/src/components/CitationPopover/CitationPopover.tsx
```diff
@@ -0,0 +1,267 @@
...
+function CitationPopover({
+  doi,
+  title,
+  boilerplateText,
+}: CitationPopoverProps) {
+  const [anchorEl, setAnchorEl] = useState<HTMLButtonElement | null>(null)
+  const [citationFormat, setCitationFormat] = useState('bibtex')
+
+  const selectSx = {
```
nickgros, line 35: add type, consider pulling this object out of the function component (or using `styled` from `'@mui/material'`)
```suggestion
  const selectSx: SxProps = {
```

packages/synapse-react-client/src/components/CitationPopover/CitationPopover.tsx
```diff
@@ -0,0 +1,267 @@
...
+              sx={{
+                padding: '6px 16px',
+                borderRadius: 0,
+              }}
+            >
+              Download Citation
+            </Button>
+            <CopyToClipboardIcon
+              value={citation || ''}
+              onClick={() => handleCopy(citation || '')}
```
nickgros, line 249: do you need this prop? It seems like this is already implemented in the `CopyToClipboardIcon`, so remove it if you don't need it.

packages/synapse-react-client/src/components/CitationPopover/CitationPopover.test.tsx
```diff
@@ -0,0 +1,178 @@
...
+    await screen.findByRole('button', { name: /Cite As/i })
+  })
+
+  it('opens popover when button is clicked and fetches citation', async () => {
+    render(<CitationPopover {...mockProps} />)
+    openPopover()
+
+    await screen.findByRole('presentation', {
+      name: /Citation options popover/i,
+    })
```
nickgros, line 55: The [`presentation` role](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/presentation_role) seems to hide this from the accessibility tree, which is not what we want. Can you set the `aria-role` of this element to be `'dialog'` and query for the `'dialog'` role in these tests?

packages/synapse-react-client/src/components/CitationPopover/CitationPopover.test.tsx
nickgros: Cool, I usually use a RegEx for this but your way seems less error prone

nickgros (changes requested):

packages/synapse-react-client/src/components/CitationPopover/CitationPopover.tsx
```diff
@@ -0,0 +1,261 @@
...
+            '& .MuiButton-startIcon': { marginRight: '4px', marginLeft: 0 },
+          }}
+        >
+          <Typography variant="label" sx={{ lineHeight: '20px' }}>
+            Cite As
+          </Typography>
+        </Button>
+      )}
+      <Popover
+        aria-label="Citation options popover"
```
nickgros, line 121: This is redundant, the accessibility device (screen reader) will already see that this is a dialog
```suggestion
        aria-label="Citation options"
```

nickgros (approved):
Small nit on the accessible label, looks good otherwise

PR #1553, opened 2025-02-05
PORTALS-3253 - Remove imports against `src/index.ts`

nickgros:
synapse-react-client: Remove local imports against `src/index.ts`

We will remove index.ts in a future change (or replace it with umd.index.ts). These imports are unnecessarily broad and may slow down building the project, since they trigger importing basically the entirety of synapse-react-client.

### DIFF
--- a/packages/synapse-react-client/README.md
+++ b/packages/synapse-react-client/README.md
@@ -97,7 +97,7 @@ In this example, make sure that your `node_modules` folder is in your Dart Sass 
 
 ## Adding a new component and Publishing a new version of SRC
 
-To expose a component from the library you must export it from [index.ts](src/index.ts). Ideally, your component will get its own subfolder within `src/components/`, which will contain an `index.ts` file that exports the public API of the component (typically the component and its props type). In `src/components/index.ts`, you can then export the contents of the new subfolder.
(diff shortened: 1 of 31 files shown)


packages/synapse-react-client/README.md
```diff
@@ -97,7 +97,7 @@ In this example, make sure that your `node_modules` folder is in your Dart Sass
 
 ## Adding a new component and Publishing a new version of SRC
 
-To expose a component from the library you must export it from [index.ts](src/index.ts). Ideally, your component will get its own subfolder within `src/components/`, which will contain an `index.ts` file that exports the public API of the component (typically the component and its props type). In `src/components/index.ts`, you can then export the contents of the new subfolder.
+JavaScript packages and applications inside the monorepo can directly import modules within the `src` folder, so modules created within the `src` directory are accessible from other TypeScript packages and applications within the monorepo--they have been configured to reference the synapse-react-client source files.
```
nickgros, line 100:
```suggestion
JavaScript packages and applications inside the monorepo that have been configured to reference the synapse-react-client source files can directly import modules within the `src` folder without exporting them from any additional file.
```

PR #1554, opened 2025-02-05
PORTALS-3382

jay-hodgson:


apps/portals/cancercomplexity/src/pages/CCKPSearchPage.tsx
```diff
@@ -94,14 +96,28 @@ export function CCKPSearchPage(props: CCKPSearchPageProps) {
         searchPageTabs[tabIndex].count = newCount
         setSearchPageTabsState([...searchPageTabs])
       }
+      // PORTALS-3382: If no tab is selected and all counts have been set, then redirect to the item with the highest count
+      const allCountsSet = searchPageTabs.every(tab => tab.count !== undefined)
+      if (selectedTabIndex == undefined && allCountsSet) {
+        // navigate to the tab that has the highest count
+        const maxCountTab = searchPageTabs.reduce((max, tab) =>
+          max === null || tab.count! > max.count! ? tab : max,
```
jay-hodgson, line 104: Is this too dense?

jay-hodgson:
Adds a loading UI (while the initial queries are executed), and navigates to the tab with the max count once all counts are set.
image

apps/portals/cancercomplexity/src/pages/CCKPSearchPage.tsx
nickgros: I think it's ok. I might suggest splitting the null case from the comparison with the accumulator. The tricky thing about this statement is trying to remember order of operations between a ternary, `||`, `>`, and `===`.

That said, I don't know that you need a null check here. The initial value will be `searchPageTabs[0]`, so `max` should never be `null`.

So I'd either separate the null check into its own statement or remove it altogether

apps/portals/cancercomplexity/src/pages/CCKPSearchPage.tsx
```diff
@@ -94,14 +96,28 @@ export function CCKPSearchPage(props: CCKPSearchPageProps) {
...
   )
   // on search field value update, update the special search parameter FTS_SEARCH_TERM, which the QueryWrapperPlotNav will load as the search term
   return (
     <Box sx={{ p: { xs: '10px', lg: '50px' } }}>
-      <PortalFullTextSearchField />
-      <PortalSearchTabs tabConfig={searchPageTabsState} />
+      <PortalFullTextSearchField disabled={selectedTabIndex == undefined} />
+      {selectedTabIndex != undefined && (
+        <PortalSearchTabs tabConfig={searchPageTabsState} />
+      )}
```
nickgros, line 120: Ah, so you effectively won't even show the tabs until you have a query. This order of interactions is making more sense to me

nickgros (approved):

PR #1555, opened 2025-02-06
Fix lint precommit issue

nickgros:
Fix issue where vite config files were excluded from tsconfig, which would cause them to break on precommit when trying to lint them.

This issue still persists in synapse-react-client, where a handful of files are intentionally not included in the tsconfig because they have type errors


apps/SageAccountWeb/tsconfig.json
```diff
@@ -1,8 +1,7 @@
 {
   "extends": "../../shared/tsconfig.base.json",
   "compilerOptions": {
-    "rootDir": ".",
-    "baseUrl": "src",
```
nickgros, line 5: Match existing projects

eslint.config.mjs
```diff
@@ -7,6 +7,20 @@ import barrel from 'eslint-plugin-barrel-files'
...
+    .filter(dirent => dirent.isDirectory())
+    .map(dirent => dirent.name)
+
+// TODO: Use Nx API to retrieve project directories
+const appDirs = getDirectories('apps').map(app => `apps/${app}`)
+const portalDirs = getDirectories('apps/portals').map(
+  app => `apps/portals/${app}`,
+)
+const packageDirs = getDirectories('packages').map(pkg => `packages/${pkg}`)
+const allProjectDirs = [...appDirs, ...portalDirs, ...packageDirs]
```
nickgros, line 23: Nx has a really friendly command-line API but an unintuitive JavaScript API for building things that are not plugins. Using Nx would be the right way to do this but I couldn't figure it out very quickly.

eslint.config.mjs
```diff
@@ -28,7 +42,18 @@ export default tseslint.config(
     },
     languageOptions: {
       parserOptions: {
-        projectService: true,
+        projectService: {
+          // We must enumerate each file that we want to lint that is not explicitly captured by a project-specific tsconfig
+          // allowDefaultProject does not allow `**` globs.
+          // https://github.com/typescript-eslint/typescript-eslint/issues/9739
+          allowDefaultProject: allProjectDirs.map(dir => `${dir}/*.ts`)
+        },
```
nickgros, line 50: This fixes the issue -- any top-level `.ts` file will be linted against the top-level tsconfig.json

packages/synapse-react-client/src/components/Markdown/MarkdownSynapse.test.tsx
nickgros, line 1: fix type + lint errors and move inside of `synapse-react-client/src`

packages/synapse-react-client/src/components/Markdown/MarkdownSynapse.test.tsx
```diff
@@ -1,22 +1,25 @@
 import { render, screen, waitFor, within } from '@testing-library/react'
```
nickgros, line 1: Changed a bunch of tests and stories -- for all of them, I fixed type + lint errors and moved them inside of `synapse-react-client/src`.

Now we have no more non-linted/non type-safe tests!

eslint.config.mjs
```diff
@@ -7,6 +7,24 @@ import barrel from 'eslint-plugin-barrel-files'
...
+const appDirs = getDirectories(`${import.meta.dirname}/apps`).map(
+  app => `apps/${app}`,
+)
+const portalDirs = getDirectories(`${import.meta.dirname}/apps/portals`).map(
+  app => `apps/portals/${app}`,
+)
+const packageDirs = getDirectories(`${import.meta.dirname}/packages`).map(
+  pkg => `packages/${pkg}`,
+)
+const allProjectDirs = [...appDirs, ...portalDirs, ...packageDirs]
```
nickgros, line 27: Nx has a really friendly command-line API but an unintuitive JavaScript API for building things that are not plugins. Using Nx would be the right way to do this but I couldn't figure it out.

apps/SageAccountWeb/tsconfig.json
```diff
@@ -1,8 +1,7 @@
...
+    "rootDir": "src",
     "lib": ["dom", "dom.iterable", "esnext"],
```
nickgros, line 5: Update rootDir to `src` to match other projects

packages/synapse-react-client/jest.config.cjs
```diff
@@ -19,17 +19,17 @@ module.exports = {
...
         `node_modules/(?!(?:.pnpm/)?(${esModules.join('|')}))`,
       ],
-      setupFilesAfterEnv: ['<rootDir>/test/setupTests.ts'],
+      setupFilesAfterEnv: ['<rootDir>/src/testutils/setupTests.ts'],
       resetMocks: false,
     },
   ],
   reporters: [
-    '<rootDir>/test/testutils/LogForFailedTestsOnlyReporter.ts',
+    '<rootDir>/src/testutils/LogForFailedTestsOnlyReporter.cjs',
```
nickgros, line 32: This was always a `cjs` file (it had no type annotations). Jest has not been configured to transform this file, so the extension was updated to denote that it is in fact a CommonJS file.

packages/synapse-react-client/src/components/EntityForm/EntityForm.test.tsx
```diff
@@ -0,0 +1,39 @@
+import { render } from '@testing-library/react'
```
nickgros, line 1: Moved file and fixed type/lint issues, test logic itself is not new

eslint.config.mjs
```diff
@@ -28,7 +46,22 @@ export default tseslint.config(
     },
     languageOptions: {
       parserOptions: {
-        projectService: true,
+        projectService: {
+          // We must enumerate each file that we want to lint that is not explicitly captured by a project-specific tsconfig
+          // allowDefaultProject does not allow `**` globs.
+          // https://github.com/typescript-eslint/typescript-eslint/issues/9739
+          allowDefaultProject: [
+            ...allProjectDirs.map(dir => `${dir}/*.ts`),
```
jay-hodgson, line 54: You don't need this entry?
...allProjectDirs.map(dir => `${dir}/*.tsx`)

jay-hodgson (approved):
Contingent on a successful build

eslint.config.mjs
nickgros: I don't think we have any top-level `tsx` files. So we shouldn't need it for now

PR #1562, opened 2025-02-11
SWC-6915

lakikowolfe:
Update date range picker design
- Control flow from desktop to mobile
- Style button to match designs in figma
- Update to use MUI components

Update enum range slider design
- Style apply button to match figma designs

https://sagebionetworks.jira.com/browse/SWC-6915


lakikowolfe:
@nickgros still working on making sure the mobile flow is handled correctly. Would love to get your eyes on this early to make sure that I haven't lost any functionality in the refactor from base `input` to MUI `DatePicker`

packages/synapse-react-client/src/components/widgets/RangeSlider/RangeSlider.tsx
```diff
@@ -78,8 +78,14 @@ function RangeSlider(props: RangeSliderProps) {
...
               size="small"
-              variant="contained"
+              variant="outlined"
               onClick={() => onApplyClicked({ min: values[0], max: values[1] })}
+              sx={{
+                borderRadius: 0,
+                border: '1px solid #395979',
+                p: '6px 8px',
+                fontSize: 14,
+              }}
```
nickgros, line 88: Just like the changes in the other file, I think this would just become
```suggestion
              sx={{
                fontSize: '16px',
              }}
```

nickgros (commented):
I think you could simplify the layout quite a bit, so I tried to communicate one potential path forward in these comments--it's a lot, so please reach out if you have questions!

nickgros:
Oh and it seems like the functionality is fine just from trying it in Storybook, but I think existing tests should verify this too!

lakikowolfe:
@nickgros really appreciate this in depth review! It's making sense after I read through it a couple times. I'll ping you for rereview once I've implemented.

packages/synapse-react-client/src/components/widgets/Range.tsx
```diff
@@ -72,39 +74,79 @@ export function Range(props: RangeProps) {
...
+        display: 'flex',
+        flexDirection: 'row',
+      }}
+    >
+      <Box
+        sx={{
+          display: 'flex',
+          marginRight: '10px',
+          gap: '10px',
+          color: '#71767F',
```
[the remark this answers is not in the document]
lakikowolfe: The default color is `#4a5056` so I'll implement the change you suggested here.

lakikowolfe:
@nickgros I'm a little stuck on how to make an update to the storybook entry. Here's what I've figured out so far:

Storybook entry: https://github.com/Sage-Bionetworks/synapse-web-monorepo/blob/df26c7fba9aac78f92f6410558193d70c4da64a0/packages/synapse-react-client/src/components/widgets/query-filter/RangeFacetFilter.stories.tsx#L89 

Runs RangeFaceFilterUI: https://github.com/Sage-Bionetworks/synapse-web-monorepo/blob/main/packages/synapse-react-client/src/components/widgets/query-filter/RangeFacetFilterUI.tsx#L48 

The Storybook entry doesn’t really seem appropriate for the DateRangePicker?

export const SelectedDate: Story = {
  args: {
    label: 'foo',
    columnType: 'DATE',
    facetResult: {
      columnMin: '0',
      columnMax: '100',
      selectedMin: '5',
      selectedMax: '95',
    },
  },
}

I’ve been playing around with these values via the storybook UI and they don’t update the values in the DateRangePicker.

I've updated the values in my local instance but it's taking forever to boot up locally and I need to log off for the day. @nickgros can you point me towards where the default storybook dates are coming from?

nickgros:
@lakikowolfe I think you're in the right place. If you change the values in that story to dates (e.g. `'2025-02-13'`), then I think it will work. You might not see the change in storybook without a full refresh to the page

packages/synapse-react-client/src/components/widgets/Range.tsx
```diff
@@ -72,39 +74,85 @@ export function Range(props: RangeProps) {
   }
 
   return (
-    <div className={className} style={{ display: 'flex', flexWrap: 'wrap' }}>
-      <div style={{ marginRight: '10px' }}>
-        <input
-          aria-label="min"
-          key="range_min"
-          type={props.type}
```
jay-hodgson, line 80: The old range picker supports different types.  We cannot (should not) use a DatePicker for all types, only dates.  In other cases, should probably use a MUI TextField for styling (with the type property set)

lakikowolfe: @jay-hodgson just pushed an update to address this problem. Can you take a look?

lakikowolfe: Screenshot 2025-02-26 at 3 07 40 PM

lakikowolfe:
After some tweaking this is what the date picker and num picker look like:

Screenshot 2025-02-27 at 10 45 28 AM

lakikowolfe:
Tests are still failing. It looks like MUI expects/returns a different date format than the previous component.

```
    ● Range input test › date range › should render with correct checked state
  
      expect(received).toBe(expected) // Object.is equality
  
      Expected: "2019-01-04"
      Received: "01/04/2019"
  
        80 |     it('should render with correct checked state', () => {
        81 |       renderComponent({ type: 'date', initialValues: initialValues })
      > 82 |       expect(getInput('min').value).toBe('2019-01-04')
           |                                     ^
        83 |       expect(getInput('max').value).toBe('2019-10-03')
        84 |     })
        85 |
  
        at Object.toBe (src/components/widgets/Range.test.tsx:82:37)
  
    ● Range input test › date range › should call callbackFn correctly
  
      expect(jest.fn()).toHaveBeenCalledWith(...expected)
  
      - Expected
      + Received
  
        Object {
      -   "max": "2019-08-03",
      -   "min": "2019-08-01",
      +   "max": "Invalid Date",
      +   "min": "Invalid Date",
        },
```

PR #1563, opened 2025-02-11
SWC-7241

nickgros:
SynapseTrendingProjects should display PROJECT_DESCRIPTION when available

image


packages/synapse-react-client/src/components/SynapseHomepageV2/SynapseTrendingProjects.tsx
```diff
@@ -33,14 +34,15 @@ export function SynapseTrendingProjects({
 
   const rowSet = past30DaysDownloadData?.responseBody?.queryResult?.queryResults
   const headers = rowSet?.headers
-  const entityIdColIndex = headers?.findIndex(
-    selectColumn => selectColumn.name == 'PROJECT_ID',
+  const entityIdColIndex = getColumnIndex('PROJECT_ID', headers)!
```
nickgros, line 37: Use existing utility

packages/synapse-react-client/src/components/SynapseHomepageV2/SynapseTrendingProjects.tsx
```diff
@@ -33,14 +34,15 @@ export function SynapseTrendingProjects({
...
   )!
-  const nUniqueUsersColIndex = headers?.findIndex(
-    selectColumn => selectColumn.name == 'N_UNIQUE_USERS',
-  )!
-  const egressSizeColIndex = headers?.findIndex(
-    selectColumn => selectColumn.name == 'ESTIMATED_PROJECT_SIZE_IN_GIB',
+  const projectDescriptionColumnIndex = getColumnIndex(
+    'PROJECT_DESCRIPTION',
+    headers,
   )!
```
nickgros, line 46: New column PROJECT_DESCRIPTION

packages/synapse-react-client/src/mocks/query/mockHomepageQueryResultData.ts
```diff
@@ -38,120 +38,135 @@ export const mockHomepageTrendingQueryResultBundle: QueryResultBundle = {
           id: '215953',
         },
         {
```
nickgros, line 40: I updated this mock data, but I don't think it's currently used in tests, since the homepage snapshot tests did not update.

It would be a bit of a pain to add tests, since we would need to add handlers for each entity in the list, so I decided against it for now.

packages/synapse-react-client/src/components/SynapseHomepageV2/SynapseTrendingProjectsItem.tsx
```diff
@@ -107,7 +112,7 @@ export function SynapseTrendingProjectItem({
           {rank}
         </Typography>
         <Box>
-          <Typography variant="body1">{entityHeader?.name}</Typography>
+          <Typography variant="body1">{displayedName}</Typography>
```
kianamcc, line 115: If the projectDescription is displayed and is lengthy, how do we handle overflow (assuming we want the displayedName to be in a single line)? Or can we expect the displayName to not be too long?

packages/synapse-react-client/src/components/SynapseHomepageV2/SynapseTrendingProjectsItem.tsx
```diff
@@ -63,6 +67,7 @@ export function SynapseTrendingProjectItem({
       borderRadius = '0px 0px 12px 12px'
       break
   }
+  const displayedName = projectDescription ?? entityHeader?.name
```
kianamcc, line 70: Should there be a fallback for this? (entityHeader)

kianamcc (approved):

packages/synapse-react-client/src/components/SynapseHomepageV2/SynapseTrendingProjectsItem.tsx
nickgros: Today it wraps, which I assume is fine: 

image

packages/synapse-react-client/src/components/SynapseHomepageV2/SynapseTrendingProjectsItem.tsx
nickgros: This is a good point--I think this list of projects will always be public, so the name should always be available.

 It might be possible for the entity to be changed to private after these statistics are surfaced, but in that case we might want to hide the row altogether.

PR #1564, opened 2025-02-12
PORTALS-3405: Consistent styling and mobile responsiveness review

kianamcc:
Jira: https://sagebionetworks.jira.com/browse/PORTALS-3405

The majority of the new components used a similar section header so I created a reusable component out of it.

Screenshot 2025-02-11 at 8 59 19 PM
Screenshot 2025-02-11 at 8 28 44 PM

### DIFF
--- a/apps/synapse-portal-framework/src/components/elportal/ELContributeYourData.tsx
+++ b/apps/synapse-portal-framework/src/components/elportal/ELContributeYourData.tsx
@@ -31,7 +31,6 @@ function ELContributeYourData() {
         sx={{
           flex: '1 1 auto',
           zIndex: 1,
-          textAlign: 'center',
           padding: { xs: '100px 40px', md: '100px 0' },
         }}
       >
@@ -45,13 +44,17 @@ function ELContributeYourData() {
           buttonText="Start Contributing"
           link="https://sagebionetworks.jira.com/servicedesk/customer/portal/12"
           sx={{
-            '*': { color: '#FFF', borderTop: 'none' },
             h2: {
+              color: 'primary.contrastText',
               fontSize: '24px',
-              borderTop: '4px solid rgba(255, 255, 255, 0.40)',
+              borderColor: 'rgba(255, 255, 255, 0.40)',
             },
-            a: { border: '1px solid white' },
-            '& p': { fontSize: '16px', paddingBottom: '10px' },
+            '.MuiButton-root': { border: '1px solid white' },
+            '& p': {
+              fontSize: '16px',
+              paddingBottom: '10px',
+              color: 'primary.contrastText',
+            },
           }}
         />
       </Box>
--- a/apps/synapse-portal-framework/src/components/elportal/ELContributeYourData.tsx
+++ b/apps/synapse-portal-framework/src/components/elportal/ELContributeYourData.tsx
@@ -1,6 +1,7 @@
-import { Box, Link, Button, Typography } from '@mui/material'
+import { Box } from '@mui/material'
 import backgroundSpotsLeft from './assets/dot_blob_top_left.png'
 import backgroundSpotsRight from './assets/dot_blob_bottom_right.png'
+import { PortalSectionHeader } from 'synapse-react-client'
 
 function ELContributeYourData() {
   return (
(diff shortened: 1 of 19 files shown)


apps/synapse-portal-framework/src/components/elportal/ELContributeYourData.tsx
```diff
@@ -34,47 +35,25 @@ function ELContributeYourData() {
...
+        our data curation team through our service desk."
+          buttonText="Start Contributing"
+          link="https://sagebionetworks.jira.com/servicedesk/customer/portal/12"
           sx={{
-            pt: 4,
-            mb: 2,
-            mx: 'auto',
-            width: 'max-content',
-            borderTop: '6px solid #ffffff88',
+            '*': { color: '#FFF', borderTop: 'none' },
```
nickgros, line 48: I suggest making the `color` `'primary.contrastText'`, which will make it more clear that you're setting the color to contrast with the background (which is the primary color). And I think it might make more sense to move this `color` style to the same `Box` where the `backgroundColor` is set to primary.main.

I am also not a big fan of using the `*` selector here either, but it's not obvious to me if there's a better way.

apps/synapse-portal-framework/src/components/elportal/ELGettingStarted.tsx
```diff
@@ -54,53 +55,20 @@ const ELGettingStarted = () => {
...
+        reverseButtonAndText
+        title="Getting Started"
+        summaryText="We provide all the help you need for navigating the portal and
+        accelerating your research."
+        buttonText="Visit Our Help Section"
+        link="https://help.eliteportal.org/help/"
         sx={{
-          borderTop: '3px solid #ffffff88',
-          py: '20px',
+          '*': { color: '#FFF', borderColor: 'rgba(255, 255, 255, 0.40)' },
```
nickgros, line 66: Ditto on lifting the `color` style up and using `'primary.contrastText`. 

If you could make the `borderColor` style more specific to the headline element, that would be great too (not all borders in this component should be this color)

apps/synapse-portal-framework/src/components/elportal/ELSupportedByNIH.tsx
```diff
@@ -65,19 +37,23 @@ const ELSupportedByNIH: React.FC = () => {
...
         >
           The National Institute on Aging (NIA) is a leading research
           organization dedicated understanding the nature of aging, and extend
           the healthy, active years of life. As part of the National Institutes
           of Health (NIH), NIA supports cutting-edge research on aging and
           age-related diseases, including Alzheimer's disease and other forms of
-          dementia. With a mission to improve the health and well-being of older
+          dementia.
+          <br />
+          <br />
```
nickgros, line 55: Probably just use two `Typography` elements instead of adding `br`, since these will correspond to `p` tags, which would be appropriate for these two paragraphs.

packages/synapse-react-client/src/components/PortalSectionHeader/PortalSectionHeader.tsx
```diff
@@ -0,0 +1,121 @@
...
+const PortalSectionHeader = ({
+  title,
+  buttonText,
+  summaryText,
+  link,
+  sx,
+  centered = false,
+  reverseButtonAndText = false,
+}: PortalSectionHeaderProps) => {
+  const iisExternalLink =
```
nickgros, line 32: Typo
```suggestion
  const isExternalLink =
```

packages/synapse-react-client/src/components/PortalSectionHeader/PortalSectionHeader.tsx
```diff
@@ -0,0 +1,121 @@
...
+        }}
+      >
+        <Typography
+          variant="headline2"
+          paddingTop="30px"
+          color="grey.1000"
+          fontSize={{ xs: '24px', md: '32px' }}
+        >
+          {title}
+        </Typography>
```
nickgros, line 59: As mentioned elsewhere, I suggest moving the border to the h2, and in components, make h2 100% width when the border should take up the full width of the section
```suggestion
      <Stack
        sx={{
          gap: '16px',
          alignItems: centered ? 'center' : 'flex-start',
        }}
      >
        <Typography
          variant="headline2"
          paddingTop="30px"
          color="grey.1000"
          fontSize={{ xs: '24px', md: '32px' }}
          sx={{
            borderTop: '4px solid',
            borderColor: 'grey.400',
          }}
        >
          {title}
        </Typography>
```

apps/synapse-portal-framework/src/components/elportal/ELContributeYourData.tsx
nickgros: Another change to consider is if the `borderTop` should be on the `h2` or the container surrounding `h2` (how you have it now).  If you had it on the `h2`, then you wouldn't need all these custom border styles here, and `PortalSectionHeader` component would just handle it.

In other cases, you would probably need to override `h2` to be `width: 100%` so the border takes up the full width of the container. I'm not 100% sure if this idea is an improvement, what do you think of it?

packages/synapse-react-client/src/components/PortalSectionHeader/PortalSectionHeader.tsx
```diff
@@ -0,0 +1,121 @@
...
+  title: string
+  buttonText?: string
+  summaryText?: React.ReactNode
+  link?: string
+  sx?: SxProps<Theme>
+  centered?: boolean
+  reverseButtonAndText?: boolean
+}
+
+const PortalSectionHeader = ({
```
nickgros, line 23: Optionally consider adding stories

nickgros (approved):
This is a great refactor -- I had some suggestions for improvement

PR #1565, opened 2025-02-12
SWC-7242

nickgros:
Update our bundle config so that we build and distribute the modules used by SWC. In https://github.com/Sage-Bionetworks/SynapseWebClient/pull/5630 we are adding a bundler to SWC that can handle more modern ESM & CJS code, so we can update synapse-react-client to emit ESM/CJS versions of the same bundle here and later remove the UMD bundle.

- Update synapse-react-client ESM/CJS bundles to use 'UMD' entry file. SynapseWebClient will use this new bundle in a future change. In the future, we will also replace `src/index.ts` with the contents of `src/umd.index.ts` and delete the UMD-specific build processes.
- Include @sage-bionetworks/synapse-types in the synapse-react-client bundled output since it is prone to drift from the version on NPM, since we do not typically release every @sage-bionetworks/synapse-types change to NPM.

While this is a breaking change, no known applications currently use the synapse-react-client ESM/CJS bundles, so it should be safe.

### DIFF
--- a/packages/synapse-react-client/vite.config.ts
+++ b/packages/synapse-react-client/vite.config.ts
@@ -6,8 +6,13 @@ import { ConfigBuilder } from 'vite-config'
  */
 const config = new ConfigBuilder()
   .setIncludeReactConfig(true)
-  .setIncludeLibraryConfig(true)
-  .setBuildLibEntry(resolve(__dirname, 'src/index.ts'))
+  .setIncludeLibraryConfig(true, {
+    except: [
+      // Include @sage-bionetworks/synapse-types because the local version may drift from the version released on NPM
+      '@sage-bionetworks/synapse-types',
+    ],
+  })
+  .setBuildLibEntry(resolve(__dirname, 'src/umd.index.ts'))
   .setConfigOverrides({
     root: '.',
     build: {
--- a/packages/vite-config/src/ConfigBuilder.ts
+++ b/packages/vite-config/src/ConfigBuilder.ts
@@ -5,7 +5,6 @@ import viteLibraryConfig from './vite-library-config.js'
 import vitestConfig from './vitest-config.js'
 
 export class ConfigBuilder {
-  private includeReactConfig = false
   private includeLibraryConfig = false
   private buildLibEntry: string | string[] | undefined = undefined
   private includeVitestConfig = false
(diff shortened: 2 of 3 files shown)


packages/vite-config/src/ConfigBuilder.ts
```diff
@@ -5,15 +5,14 @@ import viteLibraryConfig from './vite-library-config.js'
 import vitestConfig from './vitest-config.js'
 
 export class ConfigBuilder {
-  private includeReactConfig = false
```
nickgros, line 8: This can be captured in the existing `pluginConfigOptions` instance variable, so I removed it.

packages/vite-config/src/ConfigBuilder.ts
```diff
@@ -27,15 +26,14 @@ export class ConfigBuilder {
     return this
   }
 
-  setIncludeLibraryConfig(includeLibraryConfig: boolean): ConfigBuilder {
-    this.includeLibraryConfig = includeLibraryConfig
-    return this
-  }
-
-  setPluginConfigOptions(
```
nickgros, line 35: This method was unused.

packages/synapse-react-client/vite.config.ts
```diff
@@ -6,8 +6,13 @@ import { ConfigBuilder } from 'vite-config'
...
   .setIncludeReactConfig(true)
-  .setIncludeLibraryConfig(true)
-  .setBuildLibEntry(resolve(__dirname, 'src/index.ts'))
+  .setIncludeLibraryConfig(true, {
+    except: [
+      // Include @sage-bionetworks/synapse-types because the local version may drift from the version released on NPM
+      '@sage-bionetworks/synapse-types',
+    ],
+  })
+  .setBuildLibEntry(resolve(__dirname, 'src/umd.index.ts'))
```
nickgros, line 15: Update this config to use the `umd.index.ts` file, which is really "the set of synapse-react-client modules used in SWC", and has nothing to do with the UMD module format.

In the near future, I plan to make this more clear. I will update file names and remove the UMD-specific bundle scripts so we just have one build that creates ESM/CJS packages for SynapseWebClient.

For now, I'm just trying to make the changes to get this in SWC as small as possible.

packages/vite-config/src/pluginConfig.ts
```diff
@@ -34,14 +35,18 @@ const REACT_PLUGINS: PluginOption[] = [
...
-  dts({
-    rollupTypes: true,
-  }),
-]
+function getLibraryPlugins(
+  externalizeDepsOptions?: Parameters<typeof externalizeDeps>[0],
+): PluginOption[] {
+  return [
+    // Do not bundle any dependencies; the consumer's bundler will resolve and link them.
+    externalizeDeps(externalizeDepsOptions),
```
nickgros, line 43: allow passing arg

kianamcc (approved):

PR #1575, opened 2025-02-17
Add deployment to staging for portals

xschildw:
This PR adds GH workflows deployments for all the portals (AdKnowledge and Challenges already existed).
Don't merge. Need to disable jobs on Jenkins.


nickgros (approved):
LGTM

PR #1581, opened 2025-02-18
PORTALS-3390: Fix missing radio buttons

afwillia:
PORTALS-3390: Add left padding to radio buttons to fix them rendering outside of the download popup.
Screenshot 2025-02-18 at 11 10 06 AM


packages/synapse-react-client/src/template_style/_form.scss
```diff
@@ -24,7 +24,8 @@
 
 .checkbox label,
 .radio label {
-  padding-left: 0px;
+  // PORTALS-3390: Radio button renders outside of box without padding
+  padding-left: 20px;
```
afwillia, line 28: Take the same padding from[ SWC buttons](https://github.com/Sage-Bionetworks/synapse-web-monorepo/blob/a3ddd78ec9b6ddeaced8a7580df9059a656563fb/apps/SageAccountWeb/src/style/_core.scss#L6), which display properly.

nickgros: 2 thoughts:


1. I think this matches the default Bootstrap 3 style, which is already loaded--would you get the same result if you get rid of this entire block?
2. I think this component may be the only place where we use Bootstrap checkbox/radio in SRC/portals, since
  - At some point we replaced all raw checkbox/radios with MUI versions
  - Usages of `'@rjsf/core'` could pull in Bootstrap 3 checkboxes and radios, but the only other usages are in EntityForm.tsx (which I think is used in SWC only, if it's used at all) and SynapseForm, which is using custom radio checkbox

So maybe you could add the following props to the `Form` component used in ModalDownload.tsx, to match the usage in SynapseForm.


```tsx
              widgets={{
                CheckboxWidget: SynapseFormCheckboxWidget,
                CheckboxesWidget: SynapseFormCheckboxesWidget,
                RadioWidget: SynapseFormRadioWidget,
              }}
```

Then possibly delete this entire file?

nickgros (commented):
See comment, I think you can either get rid of the entire style block that you changed, or with a little more work, delete the file

packages/synapse-react-client/src/template_style/_form.scss
afwillia: Removing the entire style block also fixes the issue. I like the sound of updating ModalDownload to use the MUI buttons, so I'll update this PR.

afwillia:
The _form.scss file was still being imported but not used in a few places, so I removed the ones necessary to complete the build without failing.

nickgros (approved):
Thanks! This gets us a bit closer to dropping Bootstrap 3 entirely!

PR #1585, opened 2025-02-19
SWC-7242 - Update endpoint override var

nickgros:
Change endpoint override global variable to not be inside of SRC. When synapse-react-client is imported as SRC using a namespace import (`import * as SRC from synapse-react-client`), `SRC` becomes an immutable object, making it impossible to set an override.

Do not merge, a corresponding PR in SWC must also be created and merged in coordination with this PR.


kianamcc (approved):
Checked, updates all 18 instances of endpoint override variable.

PR #1592, opened 2025-02-24
PORTALS-3411: Initial HeaderCardV2 implementation from Sigfried Gold

jay-hodgson:


packages/synapse-react-client/src/components/HeaderCard/HeaderCardV2.tsx
```diff
@@ -0,0 +1,319 @@
...
+  descriptionConfig,
+  href,
+  target,
+  icon,
+  backgroundImage,
+  forceStackedLayout = false,
+  ctaButtons,
+}: HeaderCardV2Props) {
+  const theme = useTheme()
+  // const isMobile = useMediaQuery(theme.breakpoints.down('md'))
```
kianamcc, line 137: Is this needed?

kianamcc:
Was the "DICOM" title in dark grey intentional? Just wondering because it's difficult to see on my screen!

kianamcc:
Looks really nice! Very detailed.

kianamcc (approved):

packages/synapse-react-client/src/components/HeaderCard/HeaderCardV2.tsx
jay-hodgson: I guess not!

jay-hodgson:
> Was the "DICOM" title in dark grey intentional? Just wondering because it's difficult to see on my screen!

Nice catch.  I emailed Sigfried asking for a fix a couple of hours ago.  I don't think this blocks merge, but it's something to keep an eye on (and needs to be fixed before rollout)

PR #1594, opened 2025-02-25
SWC-7257

jay-hodgson:
Use a memoized version of the callback that only changes if onQueryResultBundleChange changes.


kianamcc (approved):
Nice! That should help keep onQueryResultBundleChange more stable

PR #1597, opened 2025-02-25
IT-3849: show a warning iff the source table is not found on the stac…

jay-hodgson:
…k (allow accepting).  Should only occur in stacks where source table does not exist

Screenshot 2025-02-25 at 12 39 25 PM


packages/synapse-react-client/src/components/TermsAndConditions/TermsAndConditions.tsx
```diff
@@ -130,6 +137,22 @@ function TermsAndConditions({
           </Link>
           ), including:
         </label>
+        {isTermsAndConditionsTableNotFound && (
+          <Alert severity="warning">
```
kianamcc, line 141: Helpful UI addition for users!

kianamcc (approved):

xschildw (approved):

PR #1610, opened 2025-03-03
PORTALS-3436:  Update copy only

jay-hodgson:


jay-hodgson:
Milan reviewed change to the copy and (after another update) approved.  Merging

PR #1613, opened 2025-03-03
move SonarQube config into project directories instead of monorepo root

afwillia:
for SonarQube monorepo scans, each project can be analyzed separately by providing individual configs and a project-specific key.


afwillia:
This workflow will fail when initiated from a fork because it cannot access the SONAR_TOKEN secret. It should pass when running on merging into main. After confirming the workflow passes in main, I will set up a github environment to allow authorized forks to run an updated version of this workflow.

jay-hodgson (approved):

PR #1615, opened 2025-03-04
PORTALS-3459

jay-hodgson:
Rely on vite build environment to set the endpoint


xschildw (approved):
Same as we did for OneSage...

PR #1620, opened 2025-03-05
PORTALS-3385: [CCKP Homepage_V1] Add icons to CCKP cards for data types

kianamcc:
Jira: https://sagebionetworks.jira.com/browse/PORTALS-3385?atlOrigin=eyJpIjoiNDVlMmNiODdkZmU0NGVkZGFiYmMwNmFiNzhiOTY1YjgiLCJwIjoiaiJ9

IconList:
Screenshot 2025-03-05 at 2 30 27 PM

Created a handleIconClick function to implement this functionality.

Will create a PORTALS-3385-V2 branch to add this feature to CCKP when data is available and add a link to that PR in the ticket.

### DIFF
--- a/packages/synapse-react-client/src/components/IconList.tsx
+++ b/packages/synapse-react-client/src/components/IconList.tsx
@@ -1,6 +1,8 @@
 import { useMemo } from 'react'
 import IconSvg, { IconSvgProps } from './IconSvg/IconSvg'
 import { merge } from 'lodash-es'
+import { useQueryContext } from './QueryContext'
+import { UniqueFacetIdentifier } from '../utils'
 
 type IconConfigs = {
   [index: string]: IconSvgProps // if the icon option has the "label" set, it will show tooltip in IconSvg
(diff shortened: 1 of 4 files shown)


packages/synapse-react-client/src/components/IconList.tsx
```diff
@@ -29,6 +31,15 @@ function IconList(props: IconListProps) {
   let noMatch: boolean = false
   const css = useTheme ? 'icon-list themed' : 'icon-list'
   const componentCss = useBackground ? `${css} bg-circle` : css
+  const queryContext = useQueryContext()
+  const { addValueToSelectedFacet } = queryContext
+
+  const handleIconClick = (dataType: 'string') => {
```
jay-hodgson, line 37: Function is recreated on rerender (I think).  Is this a good place for a useCallback?  (It's unclear to me if IconSvg is going to be re-rendered when this is recreated)

packages/synapse-react-client/src/components/IconList.tsx
```diff
@@ -29,6 +31,15 @@ function IconList(props: IconListProps) {
...
+    const facet: UniqueFacetIdentifier = {
+      columnName: 'dataType',
+    }
+    addValueToSelectedFacet(facet, dataType)
```
jay-hodgson, line 41: I verified that addValueToSelectedFacet in TableQueryReducerActions handles the case when the caller attempts to add the same facet value more than once (duplicate value is not pushed).  So this should be fine!

jay-hodgson (approved):

PR #1621, opened 2025-03-06
PORTALS-3388: [CCKP Homepage_V1]: Hard coded conditional secondary label value

kianamcc:
Jira: https://sagebionetworks.jira.com/browse/PORTALS-3388?atlOrigin=eyJpIjoiNGM5M2NiN2MzZjE2NDk3NmE2ZThhZDE5MzZiNDM2NDYiLCJwIjoiaiJ9

I originally had it set up differently but I forgot about the comments left on the Jira ticket on how this should be set up. @jay-hodgson, let me know if this is what you and Nick were thinking of?

Screenshot 2025-03-06 at 11 02 01 AM


packages/synapse-react-client/src/components/GenericCard/GenericCard.tsx
```diff
@@ -67,6 +67,15 @@ export type GenericCardSchema = {
   imageFileHandleColumnName?: string
   thumbnailRequiresPadding?: boolean
   secondaryLabels?: string[]
+  customSecondaryLabelConfig?: {
```
jay-hodgson, line 70: This makes sense as an optional input parameter

packages/synapse-react-client/src/components/GenericCard/GenericCard.tsx
```diff
@@ -416,6 +425,20 @@ class _GenericCard extends Component<GenericCardPropsInternal> {
     )
     const values: string[][] = []
     const { secondaryLabels = [] } = genericCardSchemaDefined
+    genericCardSchemaDefined.customSecondaryLabelConfig = {
```
jay-hodgson, line 428: I would expect customSecondaryLabelConfig to be set in the portal config, not in the Generic Card itself (so anyone using a Generic Card can add a custom hardcoded value, whose visibility logic is custom)

packages/synapse-react-client/src/components/GenericCard/GenericCard.tsx
jay-hodgson: I would expect to see a columnName parameter as well

packages/synapse-react-client/src/components/GenericCard/GenericCard.tsx
```diff
@@ -416,6 +425,20 @@ class _GenericCard extends Component<GenericCardPropsInternal> {
...
+              genericCardSchemaDefined.customSecondaryLabelConfig?.key ?? '',
+              genericCardSchemaDefined.customSecondaryLabelConfig?.value ?? '',
+              'externalLink',
+            ]
+          : undefined
+      },
+    }
+
     const isView = table && !isTableEntity(table)
     for (let i = 0; i < secondaryLabels.length; i += 1) {
```
jay-hodgson, line 443: just before this for loop block, I think it should try to check for the customSecondaryLabelConfig and add the entry iff exist.  Something like:
```
        if (customSecondaryLabelConfig?.isVisible(
              schema,
              data,
            )) {
        const {key, value, columnName} = customSecondaryLabelConfig
        const keyValue = [key, value, columnName]
        values.push(keyValue)
   }
   ```

Comment (kianamcc) [reply packages/synapse-react-client/src/components/GenericCard/GenericCard.tsx:443]: I had my original logic inside the for loop so that the new label would be right above the external link row. But since we're making it more flexible, would it be ok if the "how to download" label is the first label? (This would be the case using the logic placement above).

Comment (kianamcc) [reply packages/synapse-react-client/src/components/GenericCard/GenericCard.tsx:440]: It would look like this: 
Screenshot 2025-03-11 at 5 20 37 PM

Unsure about this placement.

Comment (jay-hodgson) [reply packages/synapse-react-client/src/components/GenericCard/GenericCard.tsx:440]: Yes, I think this placement(the first item) matches the design

File: apps/portals/cancercomplexity/src/config/synapseConfigs/datasets.ts [hunk 4]
```diff
@@ -14,6 +14,19 @@ export const datasetSchema: GenericCardSchema = {
   title: 'datasetName',
   description: 'description',
   includeCitation: true,
+  customSecondaryLabelConfig: {
+    key: 'HOW TO DOWNLOAD',
+    value: 'This file is hosted externally, follow the External Link, below',
```
jay-hodgson, line 19: Nitpick - could extract 2 constants and reuse

jay-hodgson: I guess I'm confused why each one is defined twice.  Could isVsible() return a Boolean and you use the key and value?  Or remove the key and value, and rename isVisible() to something like getCustomLabel()?

jay-hodgson (approved):

PR #1622, opened 2025-03-06
PORTALS-3422: Fix dropdown menu

afwillia:
Set the dropdown menu anchor to the top and add a top margin so it doesn't cover the button when opened.


kianamcc (approved):
Looks much better!

PR #1625, opened 2025-03-07
Hook up real standards MV and a dummy Standard Details page

jay-hodgson:


jay-hodgson:
Changes are for an unreleased portal only (configuring the standards explorer)

PR #1627, opened 2025-03-07
PORTALS-3404

lakikowolfe:
Consolidate various card components into a single multi-use component called CardDeck with variants that match styling and functionality required for B2AI Standards Explorer, CCKP Resources Available, and the existing Programs component

https://sagebionetworks.jira.com/browse/PORTALS-3404


lakikowolfe:
Looking for reveiw on the path forward I chose to handle custom css styling. My gut feeling is that this is messier and more tedious that using overrides. On the other hand, having a few preset css styles doesn't prevent further css styling using overrides 🤷 its a tough call

packages/synapse-react-client/src/components/CardDeck/CardDeck.Mobile.tsx
```diff
@@ -0,0 +1,53 @@
...
+          <a
+            className="highlight-link"
+            href={link}
+            target="_blank"
+            rel="noreferrer"
+          >
+            Visit website
+          </a>
+        </p>
+      )} */}
```
nickgros, line 40: Why is this commented out? Planning to add it back later?

packages/synapse-react-client/src/components/CardDeck/CardDeck.Desktop.tsx
```diff
@@ -0,0 +1,108 @@
...
+              </a>
+            </p>
+          )} */}
+        </div>
+      </div>
+      <div className="CardDeck__Card__summary">
+        <p> {description} </p>
+        <LargeButton
+          color="secondary"
+          variant={cardDeckType === 'ccpk' ? 'outlined' : 'contained'} // Update variant based on cardDeckType
```
nickgros, line 82:
```suggestion
          variant={cardDeckType === 'cckp' ? 'outlined' : 'contained'} // Update variant based on cardDeckType
```

nickgros (commented):
In general I like the approach. Would you want to merge this (maybe after others provide feedback as well) and then continue iterating to fix the outstanding issues in new PR(s)?

nickgros:
CCKP desktop: looks like the heights aren't the same as in the design:

image

you might need constrain the heights or use something like CSS subgrid (which might be a lot of work) to align card content

Programs desktop looks weird in Storybook, might just need a bit of tweaking:

image

Programs mobile looks off since all of the cards are different heights despite only having one line of text, looks like you might need to constrain the icon size or something else (I think this is an existing issue with the component)

image


b2ai desktop has some alignment/padding & button coloring issues:

image


Mobile also has some minor alignment issues:
image

lakikowolfe:
> In general I like the approach. Would you want to merge this (maybe after others provide feedback as well) and then continue iterating to fix the outstanding issues in new PR(s)?

@nickgros I think this is a solid path forward.

packages/synapse-react-client/src/components/CardDeck/CardDeck.Desktop.tsx
```diff
@@ -0,0 +1,108 @@
...
+          href={ctaButtonURL}
+          sx={{
```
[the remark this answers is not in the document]
lakikowolfe: Yeah there was some minimal mui sx styling already present and I ran with it. It's also not my favorite 😂

packages/synapse-react-client/src/components/CardDeck/CardDeck.Desktop.tsx
```diff
@@ -0,0 +1,108 @@
...
+              backgroundColor: 'white',
+              boxShadow: 'none',
+              color: '#395979',
+            }),
+            ...(cardDeckType === 'ccpk' && {
+              width: '170px',
+              fontSize: '14px',
+              boxShadow: 'none',
+              borderColor: '#314C83',
+              color: '#314C83',
```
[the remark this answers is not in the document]
lakikowolfe: Yeah this is also a primary color

packages/synapse-react-client/src/components/CardDeck/CardDeck.Mobile.tsx
lakikowolfe: This is commented out because I moved my in progress code to a different local branch while I worked on the css styling. This code corresponds to the header website link that is present in the original Programs component. I realized I needed to learn more about how the links are configured using the LinkConfig before implementing that.

Tldr: I will add this functionality back in. Just need to figure out how to do it.

lakikowolfe:
Split remaining work out to individual tickets:
- https://sagebionetworks.jira.com/browse/PORTALS-3489
- https://sagebionetworks.jira.com/browse/PORTALS-3490
- https://sagebionetworks.jira.com/browse/PORTALS-3491
- https://sagebionetworks.jira.com/browse/PORTALS-3492

packages/synapse-react-client/src/components/CardDeck/CardDeck.Desktop.tsx
```diff
@@ -89,14 +88,14 @@ export default function CardDeckDesktop({
...
             }),
-            ...(cardDeckType === 'ccpk' && {
+            ...(cardDeckType === 'cckp' && {
               width: '170px',
               fontSize: '14px',
               boxShadow: 'none',
-              borderColor: '#314C83',
-              color: '#314C83',
+              borderColor: '#SRC.$primary-action-color',
+              color: '#SRC.$primary-action-color',
```
nickgros, line 98: This won't work in `sx`. When we compile scss files (which won't look at this code at all), these Sass variables get transformed to the actual variable value, so this string has no meaning at runtime.

You can just use the `primary.main` MUI theme variable.

lakikowolfe: that makes a lot of sense. Thanks for explaining this!

PR #1628, opened 2025-03-07
PORTALS-3432: [CCKP Homepage_V1] Select a role dropdown for new portal home page search component

kianamcc:
Jira: https://sagebionetworks.jira.com/browse/PORTALS-3432

Screenshot 2025-03-07 at 2 26 52 PM

### DIFF
--- a/apps/portals/cancercomplexity/src/pages/CCKPSearchPage.tsx
+++ b/apps/portals/cancercomplexity/src/pages/CCKPSearchPage.tsx
@@ -5,7 +5,13 @@ import {
 import { PortalFullTextSearchField } from '@sage-bionetworks/synapse-portal-framework/components/PortalSearch/PortalFullTextSearchField'
 import { SearchParamAwareStandaloneQueryWrapper } from '@sage-bionetworks/synapse-portal-framework/components/PortalSearch/SearchParamAwareStandaloneQueryWrapper'
 import { Box } from '@mui/material'
-import { Outlet, RouteObject, useLocation, useNavigate } from 'react-router'
+import {
+  Outlet,
+  RouteObject,
+  useLocation,
+  useNavigate,
+  useSearchParams,
+} from 'react-router'
 import cckpConfigs from 'src/config/synapseConfigs'
 import { QueryResultBundle } from '@sage-bionetworks/synapse-types'
 import { useCallback, useState } from 'react'
@@ -80,6 +86,14 @@ function getQueryCount(queryResultBundleJSON: string) {
   return queryCount
 }
 
+const roleMapping: { [key: string]: string } = {
+  researcher: 'Datasets',
+  principalInvestigator: 'Grants',
+  funder: 'Grants',
+  student: 'Educational Resources',
+  patientAdvocate: 'Educational Resources',
+}
+
 export function CCKPSearchPage(props: CCKPSearchPageProps) {
   const { selectedTabIndex } = props
   const { datasets, education, grants, people, publications, tools } =
@@ -88,6 +102,8 @@ export function CCKPSearchPage(props: CCKPSearchPageProps) {
     useState<PortalSearchTabConfig[]>(searchPageTabs)
   const navigate = useNavigate()
   const location = useLocation()
+  const [searchParams] = useSearchParams()
+  const role = searchParams.get('role')
 
   const onQueryResultBundleChange = useCallback(
     (tabIndex: number, newQueryResultBundleJSON: string) => {
@@ -101,14 +117,26 @@ export function CCKPSearchPage(props: CCKPSearchPageProps) {
       if (selectedTabIndex == undefined && allCountsSet) {
         // Navigate to the tab that has the highest count.
         // Explicitly initialize the accumulator ("max") to the first element (searchPageTabs[0]). "max" will never be null
-        const maxCountTab = searchPageTabs.reduce(
-          (max, tab) => (tab.count! > max.count! ? tab : max),
-          searchPageTabs[0],
-        )
-        navigate({
-          pathname: `/Search/${maxCountTab.path}`,
-          search: location.search,
-        })
+        if (role && roleMapping[role]) {
+          const roleTab = searchPageTabs.find(
+            tab => tab.title === roleMapping[role],
+          )
+          if (roleTab) {
+            navigate({
+              pathname: `/Search/${roleTab.path}`,
+              search: location.search,
+            })
+          }
+        } else {
+          const maxCountTab = searchPageTabs.reduce(
+            (max, tab) => (tab.count! > max.count! ? tab : max),
+            searchPageTabs[0],
+          )
+          navigate({
+            pathname: `/Search/${maxCountTab.path}`,
+            search: location.search,
+          })
+        }
       }
     },
     [searchPageTabsState, selectedTabIndex],
--- a/apps/portals/cancercomplexity/src/pages/CCKPSearchPage.tsx
+++ b/apps/portals/cancercomplexity/src/pages/CCKPSearchPage.tsx
@@ -16,7 +16,7 @@ import cckpConfigs from 'src/config/synapseConfigs'
 import { QueryResultBundle } from '@sage-bionetworks/synapse-types'
 import { useCallback, useState } from 'react'
 import { SynapseSpinner } from 'synapse-react-client/components/LoadingScreen/LoadingScreen'
-export const searchPageTabs: PortalSearchTabConfig[] = [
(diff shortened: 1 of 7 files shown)


apps/portals/cancercomplexity/src/pages/CCKPSearchPage.tsx
```diff
@@ -80,6 +86,14 @@ function getQueryCount(queryResultBundleJSON: string) {
   return queryCount
 }
 
+const roleMapping: { [key: string]: string } = {
+  researcher: 'Datasets',
+  principalInvestigator: 'Grants',
+  funder: 'Grants',
+  student: 'Educational Resources',
+  patientAdvocate: 'Educational Resources',
+}
```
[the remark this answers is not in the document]
kianamcc: This is great! I didn't know about this.

apps/synapse-portal-framework/src/components/HeaderSearchBox.tsx
```diff
@@ -6,35 +6,48 @@ import {
...
+  roles,
 }: HeaderSearchBoxProps) => {
+  const [role, setRole] = useState('')
   const [, setSearchParams] = useSearchParams()
   const theme = useTheme()
 
   const handleTermClick = (term: string) => {
     const trimmedTerm = term.trim()
-    setSearchParams({ FTS_SEARCH_TERM: trimmedTerm })
+    setSearchParams({ FTS_SEARCH_TERM: trimmedTerm, role: role || '' })
```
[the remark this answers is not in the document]
kianamcc: Yes, the role param is still added to the URL even if it's value is an empty string. I created a check to make sure role is defined before adding it to the searchParams.

apps/portals/cancercomplexity/src/pages/CCKPSearchPage.tsx
```diff
@@ -101,14 +117,26 @@ export function CCKPSearchPage(props: CCKPSearchPageProps) {
       if (selectedTabIndex == undefined && allCountsSet) {
         // Navigate to the tab that has the highest count.
         // Explicitly initialize the accumulator ("max") to the first element (searchPageTabs[0]). "max" will never be null
```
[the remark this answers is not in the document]
kianamcc: Confirmed with Adam that this should default back to the max results if the prioritized tab is empty.

apps/synapse-portal-framework/src/components/HeaderSearchBox.tsx
```diff
@@ -33,12 +37,20 @@ const HeaderSearchBox = ({
...
     const trimmedTerm = term.trim()
-    setSearchParams({ FTS_SEARCH_TERM: trimmedTerm, role: role || '' })
+    const newSearchParams = new URLSearchParams(searchParams)
+    newSearchParams.set(FTS_SEARCH_TERM, trimmedTerm)
+
+    if (role) {
+      newSearchParams.set(FTS_SEARCH_ROLE, role)
+    }
+
+    setSearchParams(newSearchParams)
```
nickgros, line 52: Pass a callback to the state setter to ensure consistency

```suggestion
    setSearchParams(prev => {
  	    prev.set(FTS_SEARCH_TERM, trimmedTerm)
  	    if (role) {
	      prev.set(FTS_SEARCH_ROLE, role)
    	}
    	return prev
  	})
```

apps/portals/cancercomplexity/src/pages/CCKPSearchPage.tsx
```diff
@@ -41,7 +41,7 @@ export const searchPageTabs: PortalSearchTabConfig[] = [
     title: 'Educational Resources',
     path: 'EducationalResources',
   },
-]
+] satisfies PortalSearchTabConfig[]
```
nickgros, line 44: Sorry, this won't do anything unless you
1. remove `: PortalSearchTabConfig[]` above on L19
2. Here, use `as const` (i.e. `] as const satisfies PortalSearchTabConfig[]`)

But if you do that, then the `count` property cannot be dynamically changed.

So if you want to get this type-safety, then you would need to separate the count into a different object (maybe a `Map<(typeof searchPageTabs)[number]['title'], number>`?). Up to you if you want to do that work or just remove the `satisfies` here.

```suggestion
]
```

apps/synapse-portal-framework/src/components/PortalSearch/PortalSearchPage.tsx
```diff
@@ -11,10 +11,14 @@ import {
   StandaloneQueryWrapperProps,
 } from 'synapse-react-client'
 
-export type PortalSearchPageProps = {
+export type PortalSearchPageProps<
+  RoleType extends string = string,
+  TabTitleType extends string = string,
+> = {
```
kianamcc, line 17: I was thinking of using generic types to make sure we don't lose the type safety we did for roleMapping and searchPageTabs. What do you think? @nickgros

nickgros: It sounds like a good idea, but I don't see how these generic types are being used currently. It looks like you are always falling back to the default types unless you also add type parameters to the component.

You may also want to parameterize the type of `searchPageTabs` and use that to determine your default TabTitleType. IME that can be tricky to do and might not be worth the effort.

kianamcc: I created a new type to create a union of titles from the SearchTabs array. The roleMapping object uses the inferred TabTitleType, with valid values restricted to those found in the title field of the SearchTabs array. However, I'm unsure how TypeScript may behave in more complex cases but I'm hoping that if set up correctly in the specific portal using this feature (in this case like the CCKPSearchPage.tsx), it should be ok. How might I improve this?

nickgros: I spent some time looking at this to try to get a better understanding. I thought it might be possible to use [`const` type parameters](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-5-0.html#const-type-parameters) to get the TypeScript compiler to use type literals when possible when checking these types.

Basically, I found that this won't work unless the caller always uses `as const` or provides the configuration inline. I don't think we can enforce that, so I don't think it's possible to always check these types so strictly.

So I updated the props to look like:

```ts
type ExtractTabTitles<Tabs extends PortalSearchTabConfig[]> =
  Tabs[number]['title']

export type PortalSearchPageProps<
  SearchTabs extends PortalSearchTabConfig[],
  RoleType extends string,
> = {
  selectedTabIndex?: number
  configs: QueryWrapperPlotNavProps[]
  searchPageTabs: SearchTabs
  roleMapping?: Record<RoleType, ExtractTabTitles<SearchTabs>>
}
```

and the function definition to look like:

```tsx
export function PortalSearchPage<
  const SearchTabs extends PortalSearchTabConfig[],
  const RoleType extends string,
>(props: PortalSearchPageProps<SearchTabs, RoleType>) {
```

I then wrote a pretty simple test to validate these types. Note that I don't think I'm using `assertType` correctly, but I added comments noting what's happening in each.

```tsx
test('PortalSearchPage type enforces roleMapping to match the searchPageTabs', () => {
  const portalSearchPageConfigs = []
  const searchPageTabs = [{ title: 'foo', path: 'foo' }]
  const validRoleMapping = { role1: 'foo' }

  assertType( // This passes because the literal types are not used
    <PortalSearchPage
      configs={portalSearchPageConfigs}
      searchPageTabs={searchPageTabs}
      roleMapping={validRoleMapping}
    />,
  )

  assertType( // This passes because the literal types match
    <PortalSearchPage
      configs={portalSearchPageConfigs}
      searchPageTabs={[{ title: 'foo', path: 'foo' }]}
      roleMapping={{ role1: 'foo' }}
    />,
  )


  const invalidRoleMapping = { role1: 'bar' }
  assertType( // We do NOT get the expected TS error because literal types are not used
    <PortalSearchPage
      configs={portalSearchPageConfigs}
      searchPageTabs={searchPageTabs}
      // @ts-expect-error invalidRoleMapping is not a valid roleMapping
      roleMapping={invalidRoleMapping}
    />,
  )

  assertType( // We do get the expected error because literal types are used
    <PortalSearchPage
      configs={portalSearchPageConfigs}
      searchPageTabs={[{ title: 'foo', path: 'foo' }]}
      // @ts-expect-error invalidRoleMapping is not a valid roleMapping
      roleMapping={{ role1: 'bar' }}
    />,
  )
})
```

nickgros (approved):

PR #1631, opened 2025-03-11
PORTALS-3471: use DomPurify for all portal html sanitization

jay-hodgson:


packages/synapse-react-client/src/components/Markdown/MarkdownSynapse.test.tsx
```diff
@@ -305,7 +305,8 @@ describe('MarkdownSynapse tests', () => {
     expect(spyOnRender).toHaveBeenCalled()
   })
 
-  it('renders a synapse reference', async () => {
+  // PORTALS-3471:  The migration to DomPurify broke the markdown reference/bookmark functionality
+  it.skip('renders a synapse reference', async () => {
```
nickgros, line 309: Do you know if this is a limitation in DomPurify? Or does this require additional investigation (a new ticket)? This is used in at least one portals page: https://nf.synapse.org/Explore/Hackathon/DetailsPage/Background%24Results?id=syn24304262

packages/synapse-react-client/src/utils/functions/SanitizeHtmlUtils.test.ts
```diff
@@ -9,31 +8,31 @@ function createHTML(innerHTML: string) {
 
 describe('HTML Santization', () => {
   test('Prevents XSS through onload', () => {
-    const sanitized = xss("<span onload='alert('XSS')'>foo</span>", xssOptions)
+    const sanitized = sanitize("<span onload='alert('XSS')'>foo</span>")
```
nickgros, line 11: Consider running these tests against both libraries while we still use xss in SWC.

packages/synapse-react-client/src/components/Markdown/MarkdownSynapse.test.tsx
jay-hodgson: Fixed!  The order of the attributes changed, so the regex to find these elements also needed to be updated (to no longer assume the order)

packages/synapse-react-client/src/utils/functions/SanitizeHtmlUtils.test.ts
jay-hodgson: Good idea

nickgros (approved):

PR #1632, opened 2025-03-11
SWC-7269

jay-hodgson:
Test first development - test that verifies the number of times the redirect is called fails before the code change:
Screenshot 2025-03-11 at 1 04 29 PM

### DIFF
--- a/apps/synapse-oauth-signin/src/OAuth2Form.tsx
+++ b/apps/synapse-oauth-signin/src/OAuth2Form.tsx
@@ -43,10 +43,6 @@ const sendGTagEvent = (event: string) => {
   }
 }
 
-function redirectToURL(redirectURL: string) {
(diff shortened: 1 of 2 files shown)


apps/synapse-oauth-signin/src/OAuth2Form.tsx
```diff
@@ -312,7 +313,7 @@ export function OAuth2Form() {
       }}
     >
       <div style={{ marginTop: '50px' }}>
-        {showPendingRedirectUI && (
+        {pendingRedirectURL !== undefined && (
```
nickgros, line 316: Suggest still creating a variable derived from the new state for clarity, e.g.

`const showPendingRedirectUi = pendingRedirectURL !== undefined`

nickgros (approved):

PR #1638, opened 2025-03-14
PORTALS-3475

jay-hodgson:


jay-hodgson:
Changed configuration for component only exposed in Experimental Mode

PR #1639, opened 2025-03-14
PORTALS 3386 - Add Dataset to Download Cart

afwillia:
[PORTALS 3386](https://sagebionetworks.jira.com/browse/PORTALS-3386) - on dataset cards, create a button that adds the dataset to a user's download cart. The button should only appear if `synapseLink` is populated in the card container backend table. This uses `getEntity` to determine the entity type of the dataset then `<TableQueryDownloadConfirmation />` or `<FolderDownloadConfirmation />` to handle the download cart.


packages/synapse-react-client/src/components/AddToDownloadCart/AddToDownloadCart.tsx
```diff
@@ -0,0 +1,96 @@
...
+        )
+        setEntityType(entityType)
+      } catch (e) {
+        console.error('Error on getting entity information = ', e)
+        // should this error out or just default to folder?
+        //setEntityType('org.sagebionetworks.repo.model.Folder');
+      }
+    }
+    getEntity()
+  }, [folderId, accessToken])
```
nickgros, line 41:
```suggestion
  const { data: entity, isLoading } = useGetEntity(entityId)
  const entityConcreteType = entity?.concreteType
```

packages/synapse-react-client/src/components/AddToDownloadCart/AddToDownloadCart.tsx
```diff
@@ -0,0 +1,96 @@
...
+      <QueryWrapper initQueryRequest={initQueryRequest} key={queryWrapperKey}>
+        <QueryWrapperErrorBoundary>
+          <TableQueryDownloadConfirmation />
+        </QueryWrapperErrorBoundary>
+      </QueryWrapper>
+    )
+  }
+
+  return (
+    <div>
```
nickgros, line 80: also add a loading state when `useGetEntity`'s `isLoading` is true.

packages/synapse-react-client/src/components/AddToDownloadCart/AddToDownloadCart.tsx
```diff
@@ -0,0 +1,96 @@
...
+
+  function tableTsx() {
+    return (
+      <QueryWrapper initQueryRequest={initQueryRequest} key={queryWrapperKey}>
+        <QueryWrapperErrorBoundary>
+          <TableQueryDownloadConfirmation />
+        </QueryWrapperErrorBoundary>
+      </QueryWrapper>
+    )
+  }
```
nickgros, line 77: 3 issues with this pattern:
1. These are function components, so they should be TitleCase
2. Function components [should always be rendered using JSX](https://react.dev/reference/rules/react-calls-components-and-hooks#never-call-component-functions-directly) `<MyComponent />`, not function calls.
3. Once you make these changes, you'll be breaking the [no-unstable-nested-components](https://github.com/jsx-eslint/eslint-plugin-react/blob/master/docs/rules/no-unstable-nested-components.md) rule, so you'll need to refactor them out of this component (in the same file is fine)

packages/synapse-react-client/src/components/row_renderers/Dataset_V2.tsx
```diff
@@ -0,0 +1,132 @@
...
+  data?: any
+  schema?: any
+  secondaryLabelLimit?: number
+  selectColumns?: SelectColumn[]
+  columnModels?: ColumnModel[]
+  genericCardSchema?: GenericCardSchema
+  queryVisualizationContext: QueryVisualizationContextType
+}
+
+class Dataset_V2 extends Component<Dataset_V2_Props, never> {
```
nickgros, line 22: Why a class component? Please convert to a function component--it should be pretty straightforward since this component is stateless.

nickgros: Also nit but remove underscores for consistency
```suggestion
class DatasetV2 extends Component<DatasetV2Props, never> {
```

packages/synapse-react-client/src/components/AddToDownloadCart/AddToDownloadCart.tsx
```diff
@@ -0,0 +1,96 @@
+import { useState, useEffect } from 'react'
```
nickgros, line 1: Suggest renaming the component to `AddToDownloadCartButton` for clarity

packages/synapse-react-client/src/components/row_renderers/Dataset_V2.tsx
```diff
@@ -0,0 +1,132 @@
...
+              >
+                {datasetName}
+              </a>
+            </h3>
+          </div>
+          <div className="SRC-description-dataset">
+            <p className="SRC-description-text">{summary}</p>
+            <div className="SRC-cardAction">
+              {synapseLink && <AddToDownloadCart folderId={synapseLinkId} />}
+            </div>
```
nickgros, line 117: I think the CCKP Datasets feature actually uses GenericCard, so I think you want to integrate this new feature into that component, not create a new DatasetV2 card component. Or am I misunderstanding?

packages/synapse-react-client/src/components/AddToDownloadCart/AddToDownloadCart.tsx
```diff
@@ -0,0 +1,96 @@
...
+import { DEFAULT_PAGE_SIZE } from '../../utils/SynapseConstants'
+import { SynapseConstants } from '../../utils'
+import { QueryBundleRequest } from '@sage-bionetworks/synapse-types'
+import { QueryWrapper } from '../QueryWrapper'
+import { QueryWrapperErrorBoundary } from '../QueryWrapperErrorBoundary'
+import SynapseClient from '../../synapse-client'
+import { GetAppTwoTone } from '@mui/icons-material'
+
+export type AddToDownloadCartProps = {
+  folderId: string
```
nickgros, line 15: Rename to entityId since it may not be a folder

packages/synapse-react-client/src/components/AddToDownloadCart/AddToDownloadCart.tsx
```diff
@@ -0,0 +1,96 @@
...
+    entityId: folderId,
+    query: {
+      sql: `select * from ${folderId}`,
+      limit: DEFAULT_PAGE_SIZE,
+      sort: undefined,
+      additionalFilters: undefined,
+    },
+    partMask: SynapseConstants.BUNDLE_MASK_QUERY_RESULTS,
+  }
+  const queryWrapperKey = JSON.stringify(initQueryRequest)
```
nickgros, line 61: You probably won't need this key after refactoring to use JSX to render

packages/synapse-react-client/src/components/AddToDownloadCart/AddToDownloadCart.tsx
```diff
@@ -0,0 +1,96 @@
...
+        onClick={onAddClick}
+        variant="contained"
+        startIcon={<GetAppTwoTone />}
+      >
+        Add to Download Cart
+      </Button>
+      {showConfirmation &&
+        (entityType === 'org.sagebionetworks.repo.model.Folder'
+          ? folderTsx()
+          : tableTsx())}
```
nickgros, line 91: How does this look in the card if it's shown in the same `div` as the button? Feels like this might look weird

nickgros (changes requested):

afwillia:
@nickgros thanks for the feedback. I made removed Dataset_V2 and added the button component to the Generic Card. Let me know what you think!

packages/synapse-react-client/src/components/GenericCard/GenericCard.tsx
```diff
@@ -420,6 +421,8 @@ class _GenericCard extends Component<GenericCardPropsInternal> {
       schema,
       rowId,
     )
+    const synapseLinkId: string =
+      data[schema.synapseLink]?.match(SYNAPSE_ENTITY_ID_REGEX)?.[1] || ''
```
nickgros, line 425: is `schema.synapseLink` the expected way to instantiate this? I would have expected something like

1) Add a property to the type `GenericCardSchema`, like (name of prop that I came up with might not be great)

```ts
  downloadButtonSynId?: string
```

2) Get the ID from the row data (maybe around the same place `doiColumnIndex` is defined)

```tsx
    let downloadButtonSynIdColumnIndex: number | undefined
    let downloadButtonSynIdValue: string | undefined
    if (genericCardSchemaDefined.downloadButtonSynId) {
      downloadButtonSynIdColumnIndex = getColumnIndex(genericCardSchemaDefined.downloadButtonSynId, selectColumns, columnModels)
    downloadButtonSynIdValue =
      downloadButtonSynIdColumnIndex !== undefined ? data[downloadButtonSynIdColumnIndex] : undefined
    }
```

packages/synapse-react-client/src/components/GenericCard/GenericCard.tsx
```diff
@@ -45,6 +45,7 @@ import { useQueryContext } from '../QueryContext'
 import { convertDoiToLink } from '../../utils/functions/RegularExpressions'
```
nickgros, line 45: if it's not too hard (I think the current GenericCard test setup might be kind of a mess), adding a test to GenericCard to make sure the button appears when it should would be great too

afwillia:
There are still a couple issues here. One is that the download button renders in a nice place along the right edge of the card, but once clicked, moves to the left side.
Screenshot 2025-03-31 at 3 11 00 PM
Screenshot 2025-03-31 at 3 11 21 PM

Another is that clicking `cancel` does nothing when the entity is a table, thought it works fine for a folder.

afwillia:
@nickgros I am having issues adding a test for this to the generic card tests, but I think I've resolved your other comments. Another outstanding problem is that the download cart banner for tables does not respond to clicks. Do you have any ideas why either the AddToDownloadCartButton or [TableQueryDownloadConfirmation](https://github.com/Sage-Bionetworks/synapse-web-monorepo/blob/main/packages/synapse-react-client/src/components/download_list/TableQueryDownloadConfirmation.tsx) components are not responsive? This isn't an issue for folder entities using `FolderDownloadConfirmation`.

packages/synapse-react-client/src/components/GenericCard/GenericCard.tsx
```diff
@@ -584,6 +587,12 @@ class _GenericCard extends Component<GenericCardPropsInternal> {
               }}
             >
               <div className="SRC-type">{type}</div>
+              {/* PORTALS-3386 Use synapseLink in schema to add entity to download cart */}
+              {synapseLinkId && (
+                <div style={{ marginLeft: 'auto' }}>
+                  <AddToDownloadCartButton entityId={synapseLinkId} />
+                </div>
+              )}
```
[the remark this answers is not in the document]
afwillia: Screenshot 2025-04-01 at 11 34 47 AM
here's what it looks like in the citation popover div. I feel like the lack of padding looks worse than its original location - what are your thoughts?

nickgros:
> Do you have any ideas why either the AddToDownloadCartButton or [TableQueryDownloadConfirmation](https://github.com/Sage-Bionetworks/synapse-web-monorepo/blob/main/packages/synapse-react-client/src/components/download_list/TableQueryDownloadConfirmation.tsx?rgh-link-date=2025-04-01T17%3A57%3A00Z) components are not responsive? This isn't an issue for folder entities using FolderDownloadConfirmation.

This might be because `TableQueryDownloadConfirmation` uses `useQueryVisualizationContext` to handle hiding the confirmation. We may want to convert that into a callback prop passed to TableQueryDownloadConfirmation.

Also, I think we need to wrap the `QueryWrapper` in a React `Suspense` component

nickgros:
We may also want to separate the button and the ...DownloadConfirmation components to put them in different places on the card.

nickgros:
As I look at this locally, it also seems like TableQueryDownloadConfirmation isn't using the correct instance of the QueryContext, but I haven't been able to figure out why.

packages/synapse-react-client/src/components/AddToDownloadCart/AddToDownloadCartButton.tsx
```diff
@@ -0,0 +1,96 @@
...
+import { TableQueryDownloadConfirmation } from '../download_list/TableQueryDownloadConfirmation'
+import { QueryWrapper } from '../QueryWrapper'
+import { QueryWrapperErrorBoundary } from '../QueryWrapperErrorBoundary'
+
+/**
+ * Props for the AddToDownloadCartButton component.
+ *
+ * @property entityId - The ID of the entity to be added to the download cart.
+ */
+export type AddToDownloadCartButtonProps = {
```
nickgros, line 16:
```suggestion
/**
 * Props for the EntityDownloadConfirmation component.
 *
 * @property entityId - The ID of the entity to be added to the download cart.
 */
export type EntityDownloadConfirmationProps = {
```

packages/synapse-react-client/src/components/AddToDownloadCart/index.tsx
```diff
@@ -0,0 +1 @@
+export * from './AddToDownloadCartButton'
```
nickgros, line 1:
```suggestion
export * from './EntityDownloadConfirmation'
```

packages/synapse-react-client/src/components/GenericCard/GenericCard.tsx
```diff
@@ -654,22 +697,42 @@ class _GenericCard extends Component<GenericCardPropsInternal> {
...
-                title={title}
-                doi={doiValue}
-                boilerplateText={citationBoilerplateText}
-                defaultCitationFormat={defaultCitationFormat}
-              />
-            </Box>
-          )}
+
+          {includeCitation ||
+            (downloadCartSynIdValue && (
```
nickgros, line 702: These parentheses are wrong, it should be

```suggestion
          {(includeCitation || downloadCartSynIdValue) && (
```

packages/synapse-react-client/src/components/GenericCard/GenericCard.tsx
```diff
@@ -77,6 +80,7 @@ export type GenericCardSchema = {
 
   link?: string
   dataTypeIconNames?: string
+  downloadCartSynId?: string
```
nickgros, line 83:
```suggestion
  /** The column name whose data contains a synId that can be used to show a button to add the corresponding entity to the download cart. */
  downloadCartSynId?: string
```

nickgros (changes requested):
Few small changes but this is very close

afwillia:
@nickgros thanks for the feedback and revisions - I just applied them. Also renamed to EntityDownloadConfirmation

nickgros (approved):
This set of changes looks good, but there are layout issues to fix (in another PR is fine) before the work is done.

nickgros:
I am working to refactor the generic card a bit, so I'm merging this to avoid a mess of conflicts later.

PR #1643, opened 2025-03-17
ADM-4286: update UpsetPlot sql on ELITE portal homepage (v2), and app…

jay-hodgson:
…lied facet value target column


jay-hodgson:
@melissarklein reviewed/approved.  Merging

PR #1648, opened 2025-03-18
Workflow to run sonarcloud scan on PR

afwillia:
Add a workflow to run sonarcloud scans from a PR using the main repo's sonar-pr environment to get the sonar token. This is necessary to securely run workflows from forks.

### DIFF
--- a/.github/workflows/sonarqube-scan-pr.yml
+++ b/.github/workflows/sonarqube-scan-pr.yml
@@ -18,7 +18,7 @@ jobs:
           # Disabling shallow clones is recommended for improving the relevancy of reporting
           fetch-depth: 0
       - name: SonarQube Scan
-        uses: SonarSource/sonarqube-scan-action@v5.0.0
+        uses: SonarSource/sonarqube-scan-action@v5
         env:
           SONAR_TOKEN: ${{ secrets.SONAR_TOKEN }}
         with:
--- a/.github/workflows/sonarqube-scan-pr.yml
+++ b/.github/workflows/sonarqube-scan-pr.yml
@@ -2,7 +2,7 @@ on:
   # Warning: using the pull_request_target event without the cautionary measures may allow
   # unauthorized GitHub users to open a “pwn request” and exfiltrate secrets.
   pull_request_target:
-    types: [opened, synchronize, reopened, labeled]
+    types: [opened, synchronize, reopened]
 
 name: SonarCloud Code Scan
 jobs:
--- a/.github/workflows/sonarqube-scan-pr.yml
+++ b/.github/workflows/sonarqube-scan-pr.yml
@@ -1,0 +1,25 @@
+on:
+  # Warning: using the pull_request_target event without the cautionary measures may allow
+  # unauthorized GitHub users to open a “pwn request” and exfiltrate secrets.
+  pull_request_target:
+    types: [opened, synchronize, reopened, labeled]
+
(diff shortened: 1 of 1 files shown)


afwillia:
I think this needs to be merged into this repo, not a fork, before the workflow will start running on PR.

.github/workflows/sonarqube-scan-pr.yml
```diff
@@ -0,0 +1,25 @@
+on:
+  # Warning: using the pull_request_target event without the cautionary measures may allow
+  # unauthorized GitHub users to open a “pwn request” and exfiltrate secrets.
+  pull_request_target:
+    types: [opened, synchronize, reopened, labeled]
```
nickgros, line 5: Any reason we should have `labeled` here? Seems like the other triggers are appropriate, but I don't think labeling would affect the scan results
```suggestion
    types: [opened, synchronize, reopened]
```

.github/workflows/sonarqube-scan-pr.yml
```diff
@@ -0,0 +1,25 @@
...
+
+name: SonarCloud Code Scan
+jobs:
+  # For monorepos, create one job per project and specify the directory
+  sonarqube-src:
+    environment: ${{ github.event_name == 'pull_request_target' &&
+      github.event.pull_request.head.repo.full_name != github.repository && 'sonar-pr' || 'none' }}
```
nickgros, line 12: Can you help me understand this environment config? The event will always be pull_request_target, so I think the first condition can be removed (unless pull_request events are a subset of pull_request_target events--but I don't think that's true).

So if it's in Sage-Bionetworks, it will be 'sonar-pr' and otherwise it will be 'none'? What does this do?

.github/workflows/sonarqube-scan-pr.yml
```diff
@@ -0,0 +1,25 @@
...
+    name: scan-synapse-react-client
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          # Disabling shallow clones is recommended for improving the relevancy of reporting
+          fetch-depth: 0
+      - name: SonarQube Scan
+        uses: SonarSource/sonarqube-scan-action@v5.0.0
```
nickgros, line 21: Do we want to pin to `v5` to get minor/patch version upgrades automatically, or should we stick with `v5.0.0` for consistency/security?

.github/workflows/sonarqube-scan-pr.yml
afwillia: Agreed, I'm ok with removing this.

.github/workflows/sonarqube-scan-pr.yml
afwillia: This is to ensure the workflow runs on PRs initiated from a fork, not within the main repo. If so, it will run in the `sonar-pr` environment, which contains the `SONAR_TOKEN` variable and is configured so our team members can approve workflows initiated by forks. Otherwise, if no environment is selected, this workflow should fail because forks can't access secrets.

.github/workflows/sonarqube-scan-pr.yml
afwillia: That's a good idea. I took this from SonarCloud docs, but it looks like they now just use `v5`

nickgros (approved):

PR #1649, opened 2025-03-18
PORTALS-3432: Standards Explorer: Standalone component to surface key/value data, similar to CardFooter

kianamcc:
Jira: https://sagebionetworks.jira.com/browse/PORTALS-3409?atlOrigin=eyJpIjoiMzkzMDJlZDQyNGIzNDgyMThmZmE5NTZmNmJiMGY4MDYiLCJwIjoiaiJ9

Screenshot 2025-03-18 at 10 37 04 AM


packages/synapse-react-client/src/components/CardMetadataDisplay/CardMetadataDisplay.tsx
```diff
@@ -0,0 +1,159 @@
...
+import { getFieldIndex } from '../../utils/functions/queryUtils'
+
+export type CardMetadataDisplayProps = {
+  sql: string
+  sqlOperator?: SQLOperator
+  searchParams?: Record<string, string>
+  columnLinks?: LabelLinkConfig
+  labels?: string[]
+  columnAliases?: Record<string, string>
+} & Pick<QueryVisualizationWrapperProps, 'additionalFiltersSessionStorageKey'>
```
nickgros, line 31: I didn't expect you to know this, but the `sqlOperator` & `searchParams` props are a legacy pattern from when these components directly read/accepted the URLSearchParams to add the filter. Since I refactored how data is passed in DetailsPages, these prop names don't really make sense, so I don't want to keep propagating this pattern that is now (IMO) poorly named and an unintuitive API.

Instead of `sql`, `sqlOperator` and `searchParams`, can we refactor so we just have

```ts
/** The query used to retrieve data */
query: Query
```

Then in the details page, you can then pass something like
```tsx
const query: Query = { 
  sql: dataSql, 
  additionalFilters: [{
  concreteType: 'org.sagebionetworks.repo.model.table.ColumnSingleValueQueryFilter',
  columnName: 'Name',
  operator: ColumnSingleValueFilterOperator.EQUAL,
  values: [value],
  ]]
} 

<CardMetadataDisplay query={query} /* ...other props */ />
```

Honestly, we could even keep using `getAdditionalFilters` here:
```tsx
const query: Query = { 
  sql: dataSql, 
  additionalFilters: getAdditionalFilters(undefined, { Name: value! }, ColumnSingleValueFilterOperator.EQUAL)
} 
```

You can also remove the `Pick<QueryVisualizationWrapperProps, 'additionalFiltersSessionStorageKey'>` from the props signature since this component won't need to save/load dynamic filters from sessionStorage.

Thinking bigger picture, I think calls to `getAdditionalFilters` should (eventually, not in this PR) be refactored to be applied before a query gets passed into a component instead of within the component.

Curious to know what @jay-hodgson thinks of migrating towards this pattern

packages/synapse-react-client/src/components/CardMetadataDisplay/CardMetadataDisplay.tsx
```diff
@@ -0,0 +1,159 @@
...
+
+const CardMetadataDisplay = (props: CardMetadataDisplayProps) => {
```
nickgros, line 33: This isn't (necessarily) shown on a card, whether or not it's "Metadata" is use-case specific, and "Display" is very generic...how about renaming it to `RowDataTable`? It only shows one Row of data, in a table UI.

Naming things is hard :slightly_smiling_face:

In another comment I point out that I think we already have the Row in the DetailsPage component, but I don't want to get rid the work you did to also orchestrate a query, so I think we could refactor to have two components

1. RowDataTable takes rowData and the other props used for display and displays the table
2. QueryForRowDataTable (this name could be improved) which queries for a row and passes the `row: Row` and `headers: SelectColumn[]` to RowDataTable.

packages/synapse-react-client/src/components/CardMetadataDisplay/CardMetadataDisplay.tsx
```diff
@@ -0,0 +1,159 @@
...
+} from '@mui/material'
+import { getFieldIndex } from '../../utils/functions/queryUtils'
+
+export type CardMetadataDisplayProps = {
+  sql: string
+  sqlOperator?: SQLOperator
+  searchParams?: Record<string, string>
+  columnLinks?: LabelLinkConfig
+  labels?: string[]
+  columnAliases?: Record<string, string>
```
nickgros, line 30: Not sure how I feel about `labels` as a prop name but I like that it's succinct--I think that it would be more clear if you added a docstring for every prop. It would help understand what's happening, especially now that external folks are increasingly developing in this package

```suggestion
  /** The column names referring to the ordered keys in the key-value pairs that should be shown */
  labels?: string[]
  /** Optional display value overrides for column names */
  columnAliases?: Record<string, string>
```

packages/synapse-react-client/src/components/CardMetadataDisplay/CardMetadataDisplay.tsx
```diff
@@ -0,0 +1,159 @@
...
+
+  if (isLoading)
+    return (
+      <Typography
+        sx={{
+          fontSize: '14px',
+        }}
+      >
+        Loading...
+      </Typography>
```
nickgros, line 89: Prefer to use `SkeletonTable` component so it takes up roughly the same size/shape

apps/portals/standards/src/pages/StandardsDetailsPage.tsx
```diff
@@ -28,7 +36,26 @@ export const standardDetailsPageContent: DetailsPageContentType = [
...
+              'Acronym',
+              'Is_Open',
+              'Registration',
+            ]}
+            columnAliases={dataColumnAliases}
+            searchParams={{
+              Name: value!,
+            }}
+          />
+        )}
```
nickgros, line 57: You actually already have the RowSet and Row in the DetailsPage context. In another comment I suggest splitting the component so you don't always have to issue a new table query. Here, you would use the component that doesn't create a new query

```suggestion
        {({ value, context }) => (
          <RowDataTable
            rowData={context.rowData}
            headers={context.rowSet.headers}
            labels={[
              'Name',
              'Collections',
              'Organizations',
              'Data_Topic',
              'Acronym',
              'Is_Open',
              'Registration',
            ]}
            columnAliases={dataColumnAliases}
          />
        )}
```

packages/synapse-react-client/src/components/CardMetadataDisplay/CardMetadataDisplay.tsx
```diff
@@ -0,0 +1,159 @@
...
+
+  if (labels) {
+    for (let i = 0; i < labels.length; i += 1) {
+      const columnName = labels[i]
+      const columnIndex = getFieldIndex(columnName, queryResultBundle)
+      const value = rowData[columnIndex]
+      // value can be 'null' if column type is of type STRING or '[null]' if column type is of STRING_LIST
+      if (value && value !== '[null]' && value !== 'null') {
+        labelValueArray.push([columnName, value])
+      }
```
nickgros, line 77: I am surprised we don't already have logic for seeing if a cell is empty, but I didn't find an exact fit in SynapseCardLabel or SynapseTableCell

Is it true that we should hide keys that have `null` values? What about an empty array?

packages/synapse-react-client/src/components/CardMetadataDisplay/CardMetadataDisplay.tsx
```diff
@@ -0,0 +1,159 @@
...
+        sx={{
+          fontSize: '14px',
+        }}
+      >
+        Loading...
+      </Typography>
+    )
+
+  return (
+    <div>
```
nickgros, line 93: Consider using `StyledTableContainer` to apply styles--you may be able to remove some of your custom styles. You might also have to use html elements instead of MUI components to use the StyledTableContainer styles

apps/portals/standards/src/pages/StandardsDetailsPage.tsx
kianamcc: Where exactly does rowData and rowSet come from? I see we can pass in a columnName to filter by but how do we know exactly what row to return?

packages/synapse-react-client/src/components/CardMetadataDisplay/CardMetadataDisplay.tsx
kianamcc: I am ok with removing the query component! Can we assume we will always have context.rowSet and context.rowData?

packages/synapse-react-client/src/components/CardMetadataDisplay/CardMetadataDisplay.tsx
kianamcc: I assume we should hide rows with a null value like in the card component. I don't know when we would have an empty array. If there is no value for a cell, does it always show as null or [null] or is it possible that it shows as []?

apps/portals/standards/src/pages/StandardsDetailsPage.tsx
```diff
@@ -28,7 +42,58 @@ export const standardDetailsPageContent: DetailsPageContentType = [
...
+                  'Collections',
+                  'Organizations',
+                  'Data_Topic',
+                  'Acronym',
+                  'Is_Open',
+                  'Registration',
+                ]}
+                columnAliases={dataColumnAliases}
+              />
+            )
```
nickgros, line 93: We should always (eventually) have `rowData` and `rowSet`, so in this case we should show a loading component like a skeleton.

apps/portals/standards/src/pages/StandardsDetailsPage.tsx
```diff
@@ -28,7 +42,58 @@ export const standardDetailsPageContent: DetailsPageContentType = [
...
+              ColumnSingleValueFilterOperator.EQUAL,
+            ),
+          }
+          if (context.rowData && context.rowSet) {
+            return (
+              <RowDataTable
+                rowData={
+                  context.rowData.values?.filter(
+                    (value): value is string => value !== null,
+                  ) ?? []
```
nickgros, line 61: Would prefer that the filter is applied in the component (couldn't it also mess up the mapping of column -> data?). Remove the filter here and update the type in the component prop

packages/synapse-react-client/src/components/RowDataTable/RowDataTable.tsx
```diff
@@ -0,0 +1,93 @@
+import { SelectColumn } from '@sage-bionetworks/synapse-types'
+import { LabelLinkConfig } from '../CardContainerLogic'
+import { StyledTableContainer } from '../styled/StyledTableContainer'
+import { SynapseCardLabel } from '../GenericCard'
+
+export type RowDataTableProps = {
+  /** Array of string values corresponding to the data for a row */
+  rowData: string[]
```
nickgros, line 8: Allow passing `null` since Synapse Table row data may be null. If you need to do filtering, it's easy/centralized if you do it inside this component, then callers don't need to think about it/do it every time

```suggestion
  /** Array of string values corresponding to the data for a row */
  rowData: Array<string | null>
```

packages/synapse-react-client/src/components/RowDataTable/RowDataTable.tsx
```diff
@@ -0,0 +1,93 @@
...
+  labels: string[]
+  /** Optional configuration for linking columns */
+  columnLinks?: LabelLinkConfig
+  /** key-value pair mapping to display in rows */
+  labelValueArray?: [string, string][]
+  /** Optional display value overrides for column names */
+  columnAliases?: Record<string, string>
+}
+
+const RowDataTable = (props: RowDataTableProps) => {
```
nickgros, line 21:
```suggestion
/**
 * Displays a table of column names and Synapse Table row data represented as key/value pairs
 */
const RowDataTable = (props: RowDataTableProps) => {
```

packages/synapse-react-client/src/components/RowDataTable/RowDataTableWithQuery.tsx
```diff
@@ -0,0 +1,76 @@
...
+  headers: SelectColumn[]
+  /** List of column names to display as labels */
+  labels: string[]
+  /** Optional display value overrides for column names */
+  columnAliases?: Record<string, string>
+  /** Optional configuration for linking columns  */
+  columnLinks?: LabelLinkConfig
+}
+
+const RowDataTableWithQuery = (props: RowDataTableWithQueryProps) => {
```
nickgros, line 33:
```suggestion
/**
 * For the first row returned by the query, displays a table of column names and Synapse Table row data represented as key/value pairs
 */
const RowDataTableWithQuery = (props: RowDataTableWithQueryProps) => {
```

packages/synapse-react-client/src/components/RowDataTable/RowDataTableWithQuery.tsx
```diff
@@ -0,0 +1,76 @@
...
+import { SkeletonTable } from '../Skeleton'
+import { getFieldIndex } from '../../utils/functions/queryUtils'
+
+export type RowDataTableWithQueryProps = {
+  /** The query used to retrieve data */
+  query: Query
+  /** The specific row data to be rendered */
+  row: Row
+  /** The headers for the columns in the table */
+  headers: SelectColumn[]
```
nickgros, line 24: For this component, the query will retrieve the row and columns, so those should not be passed

```suggestion
```

packages/synapse-react-client/src/components/RowDataTable/RowDataTableWithQuery.tsx
```diff
@@ -0,0 +1,76 @@
...
+    const columnName = headers[i].name
+    const columnIndex = getFieldIndex(columnName, queryResultBundle)
+    const value = row.values[columnIndex]
+    // value can be 'null' if column type is of type STRING or '[null]' if column type is of STRING_LIST
+    if (value && value !== '[null]' && value !== 'null') {
+      labelValueArray.push([columnName, value])
+    }
+  }
+
+  const rowData = row.values.filter((value): value is string => value !== null)
```
nickgros, line 60: Just pass along for the inner component to handle, so probably remove this

packages/synapse-react-client/src/components/CardMetadataDisplay/CardMetadataDisplay.tsx
nickgros: At the moment for the current use-case, we can, but I don't know if that will always be the case, which is why the query version of the component might be useful to have

packages/synapse-react-client/src/components/RowDataTable/RowDataTable.tsx
```diff
@@ -0,0 +1,93 @@
...
+  /** Array of string values corresponding to the data for a row */
+  rowData: string[]
+  /** The headers for the columns in the table */
+  headers: SelectColumn[]
+  /** List of column names to display as labels */
+  labels: string[]
+  /** Optional configuration for linking columns */
+  columnLinks?: LabelLinkConfig
+  /** key-value pair mapping to display in rows */
+  labelValueArray?: [string, string][]
```
nickgros, line 16: I don't think you will ever want to pass this prop, you will always create this data structure in the component
```suggestion
```

packages/synapse-react-client/src/components/RowDataTable/RowDataTable.tsx
```diff
@@ -0,0 +1,93 @@
...
+import { StyledTableContainer } from '../styled/StyledTableContainer'
+import { SynapseCardLabel } from '../GenericCard'
+
+export type RowDataTableProps = {
+  /** Array of string values corresponding to the data for a row */
+  rowData: string[]
+  /** The headers for the columns in the table */
+  headers: SelectColumn[]
+  /** List of column names to display as labels */
+  labels: string[]
```
nickgros, line 12: How about renaming this and slightly modifying logic to something like...
```suggestion
  /** Optional list of column names whose data should be displayed in the table. If not provided, all columns with data will be shown. */
  displayedColumns?: string[]
```

packages/synapse-react-client/src/components/RowDataTable/RowDataTable.tsx
```diff
@@ -0,0 +1,93 @@
...
+    <StyledTableContainer sx={{ width: '100%' }}>
+      <table style={{ width: '100%' }}>
+        <tbody>
+          {(labelValueArray ?? localLabelValueArray)?.map(
+            ([columnName, value], index) => (
+              <tr
+                key={columnName}
+                style={{
+                  backgroundColor: index % 2 === 0 ? '#f5f5f5' : 'transparent',
+                }}
```
nickgros, line 58: StyledTableContainer should already do this, is it possible for you to remove it?
```suggestion
```

packages/synapse-react-client/src/components/RowDataTable/RowDataTable.tsx
```diff
@@ -0,0 +1,93 @@
...
+                    style={{
+                      fontSize: '14px',
+                      lineHeight: '20px',
+                      color: '#333',
+                    }}
+                  >
+                    {columnAliases?.[columnName] ?? columnName}
+                  </span>
+                </td>
+                <td style={{ padding: '7px 20px 7px 10px' }}>
```
nickgros, line 71: Does the default padding provided by `StyledTableContainer` work for this case?

packages/synapse-react-client/src/components/RowDataTable/RowDataTable.tsx
```diff
@@ -0,0 +1,93 @@
...
+      const columnName = labels[i]
+      const columnIndex = headers.findIndex(
+        header => header.name === columnName,
+      )
+      const value = columnIndex !== -1 ? rowData[columnIndex] : undefined
+
+      // value can be 'null' if column type is of type STRING or '[null]' if column type is of STRING_LIST
+      if (value && value !== '[null]' && value !== 'null') {
+        localLabelValueArray.push([columnName, value])
+      }
```
nickgros, line 44: This might need to be more sophisticated to handle lists and non-lists separately, like 
```ts
const column = headers[columnIndex]
const isNullValue = value === "null"
const isListType = column.columnType.endsWith("_LIST")
const isEmptyListOrListOfNull = value === "[]" || value === "[null]"
if (!(isNullValue || (isListType && isEmptyListOrListOfNull))) {
  labelValueArray.push([columnName, value])
}
```

apps/portals/standards/src/pages/StandardsDetailsPage.tsx
```diff
@@ -28,7 +42,58 @@ export const standardDetailsPageContent: DetailsPageContentType = [
...
+      <DetailsPageContextConsumer columnName={'Name'}>
+        {({ value, context }) => {
+          const query: Query = {
+            sql: dataSql,
+            additionalFilters: getAdditionalFilters(
+              undefined,
+              { Name: value! },
+              ColumnSingleValueFilterOperator.EQUAL,
+            ),
+          }
```
nickgros, line 54: This is exactly how I think a query should be created in these cases, I think this looks great, though I don't think you'll need it in this specific case (based on other comments in this review). Thanks for switching to use this new pattern. I think we should look to update our other components to do this

nickgros (approved):

PR #1650, opened 2025-03-18
PORTALS-3420 - Add wrappers to improve testability

nickgros:
Refactor to create simple hook wrappers to get row selection Jotai atom values and the ReactQuery result for queryMetadata. This refactor improve testability by simplifying the process to create mocks. Without the wrappers, we would have to mock `useQuery` or `useAtomValue` and the mock implementation would have to add logic to handle the different atoms or query options.

With the wrapper, we can simply mock the wrapper without writing complex mock logic for external dependencies.

### DIFF
--- a/packages/synapse-react-client/src/components/CardContainer/CardContainer.tsx
+++ b/packages/synapse-react-client/src/components/CardContainer/CardContainer.tsx
@@ -1,3 +1,10 @@
+import { Box } from '@mui/material'
+import {
+  ColumnTypeEnum,
+  EntityHeader,
+  Row,
+  RowSet,
+} from '@sage-bionetworks/synapse-types'
 import { Suspense } from 'react'
 import useGetInfoFromIds from '../../utils/hooks/useGetInfoFromIds'
 import {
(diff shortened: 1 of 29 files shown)


packages/synapse-react-client/src/components/CardContainer/CardContainer.tsx
```diff
@@ -92,8 +92,7 @@ function CardContainerInternal(props: CardContainerProps) {
   } = props
   const { NoContentPlaceholder } = useQueryVisualizationContext()
   const queryContext = useQueryContext()
-  const { queryMetadataQueryOptions } = queryContext
-  const { data: queryMetadata } = useSuspenseQuery(queryMetadataQueryOptions)
+  const { data: queryMetadata } = useSuspenseGetQueryMetadata()
```
nickgros, line 95: in tests, we could then just mock `useSuspenseGetQueryMetadata` instead of either trying to mock react-query or creating a mock queryFn to pass to react-query.

packages/synapse-react-client/src/components/QueryWrapper/QueryWrapper.tsx
```diff
@@ -74,7 +74,7 @@ function QueryWrapperInternal(props: QueryWrapperProps) {
     fileVersionColumnName,
   } = props
 
-  const hasSelectedRows = useAtomValue(hasSelectedRowsAtom)
+  const hasSelectedRows = useHasSelectedRowsAtomValue()
```
nickgros, line 77: ditto for simplfying mocking -- instead of trying to mock `useAtomValue` or faking `hasSelectedRowsAtom`, we can just mock `useHasSelectedRowsAtomValue`

kianamcc:
I haven't heard of Jotai atoms before, but I'm interested in trying out this wrapper when I write tests that require useQuery especially!

kianamcc (approved):

PR #1655, opened 2025-03-20
PORTALS-3493

jay-hodgson:


jay-hodgson:
Do not encode '&' before error_description

apps/synapse-oauth-signin/src/mocks/handlers.ts
```diff
@@ -7,7 +7,10 @@ export function resetConsentedInMockService(newValue: boolean) {
   hasConsented = newValue
 }
 
-export const ACCESS_CODE_PROVIDED_BY_SERVER = 'someAccessCode'
+export const ACCESS_CODE_PROVIDED_BY_SERVER = 'some Access Code' //contains value that will be encoded in the redirect to test for double-encoding
```
jay-hodgson, line 10: In reality, I don't think the access codes have ever contained values that require encoding (which is why we have never seen the double-encoding bug in the wild), but I'd like to fix it anyway

apps/synapse-oauth-signin/src/URLUtils.tsx
```diff
@@ -15,22 +15,24 @@ export const handleErrorRedirect = (
...
+          error['error_description'],
+        )
+      }
+      let redirectUrl = `${redirectUri}?${redirectURLSearchParams.toString()}`
       const state = searchParams.get('state')
       if (state) {
-        redirectURLSearchParams.set('state', encodeURIComponent(state))
+        // SWC-7272 - append raw state to the end of the URL
+        // We do not use URLSearchParams for state because it encodes the value, and it should be passed untouched.
+        redirectUrl += `&state=${state}`
```
jay-hodgson, line 33: Same fix made for happy case in https://github.com/Sage-Bionetworks/synapse-web-monorepo/pull/1636/files

jay-hodgson:
only [merge-e2e-reports](https://github.com/Sage-Bionetworks/synapse-web-monorepo/actions/runs/13975926361/job/39130592063?pr=1655#logs) failed

PR #1657, opened 2025-03-21
PORTALS-3478

jay-hodgson:


jay-hodgson:
Screenshot 2025-03-21 at 11 18 25 AM

jay-hodgson:
Note that the search panel will not function properly until we enable FTS on the data table (which will also be changing based on Sigfried's work)

apps/synapse-portal-framework/src/components/standards/StandardsHeader.tsx
```diff
@@ -0,0 +1,121 @@
...
+    'Integration',
+    'Acquisition',
+    'File formats',
+    'Ontologies',
+    'Lab automation',
+  ]
+
+  const isFeatureEnabled = useGetFeatureFlag(
+    FeatureFlagEnum.PORTAL_SEARCH_HEADER,
+  )
```
nickgros, line 26: Is this feature flag check necessary for this portal?

apps/synapse-portal-framework/src/components/standards/StandardsHeader.tsx
```diff
@@ -0,0 +1,121 @@
...
+            searchExampleTerms={searchExampleTerms}
+            searchPlaceholder={searchPlaceholder}
+            callback={searchString => {
+              const filter: TextMatchesQueryFilter = {
+                concreteType:
+                  'org.sagebionetworks.repo.model.table.TextMatchesQueryFilter',
+                searchExpression: searchString,
+              }
+              const query: Query = {
+                sql: 'select * from syn64960277',
```
nickgros, line 99: Add to resources.ts (or add a TODO if you're waiting on Sigfried to add it)

```suggestion
                sql: 'select * from syn64960277',
```

apps/synapse-portal-framework/src/components/HeaderSearchBox.tsx
```diff
@@ -42,17 +45,19 @@ const HeaderSearchBox = ({
...
-
     if (path) {
+      setSearchParams(prev => {
+        prev.set(FTS_SEARCH_TERM, trimmedTerm)
+        if (role) {
+          prev.set(FTS_SEARCH_ROLE, role)
+        }
+        return prev
+      })
       window.location.pathname = `${path}`
```
nickgros, line 56: I must have missed this in a previous code review, could you update this to use a react-router hook instead of updating `window` (or create a ticket to follow-up)?

apps/synapse-portal-framework/src/components/PortalSearch/PortalFullTextSearchField.tsx
```diff
@@ -42,17 +44,19 @@ export function PortalFullTextSearchField({
...
-
           if (path) {
+            setSearchParams(prev => {
+              prev.set(FTS_SEARCH_TERM, trimmedInput)
+              if (role) {
+                prev.set(FTS_SEARCH_ROLE, role)
+              }
+              return prev
+            })
             window.location.pathname = `${path}`
```
nickgros, line 55: ditto for this change to `window`

nickgros (approved):

apps/synapse-portal-framework/src/components/standards/StandardsHeader.tsx
jay-hodgson: Probably not!

apps/synapse-portal-framework/src/components/standards/StandardsHeader.tsx
jay-hodgson: Can I reference the portal resources.ts from the synapse-portal-framework component?

jay-hodgson: I'll just add a prop to the component and set it in the standards portal config (maybe this is what you were suggesting)

apps/synapse-portal-framework/src/components/HeaderSearchBox.tsx
```diff
@@ -40,20 +40,21 @@ const HeaderSearchBox = ({
   roles,
 }: HeaderSearchBoxProps) => {
   const [role, setRole] = useState('')
-  const [, setSearchParams] = useSearchParams()
   const theme = useTheme()
+  const navigate = useNavigate()
```
jay-hodgson, line 44: Using navigate to change the route, and set the search params

PR #1661, opened 2025-03-24
SWC-7287: ensure consentToRequest is called once (max)

jay-hodgson:


apps/synapse-oauth-signin/src/OAuth2Form.tsx
```diff
@@ -222,8 +224,12 @@ export function OAuth2Form() {
...
   const onConsent = useCallback(() => {
-    sendGTagEvent('UserConsented')
-    if (oidcAuthorizationRequestFromSearchParams) {
+    if (
+      oidcAuthorizationRequestFromSearchParams &&
+      !hasCalledOnConsent.current
+    ) {
+      hasCalledOnConsent.current = true
+      sendGTagEvent('UserConsented')
       consentToRequest(oidcAuthorizationRequestFromSearchParams)
```
nickgros, line 233: Would it be possible to reproduce this in a test and verify that this is only called once?

nickgros (approved):

apps/synapse-oauth-signin/src/OAuth2Form.tsx
jay-hodgson: Good idea

PR #1670, opened 2025-03-27
PLFM-3499: Add env var for distribution id

xschildw:
This PR adds an environment variable to specify the distribution ID for each portal (staging and prod).
Another PR will modify run.sh to create an invalidation for the distribution.


jay-hodgson:
Note, this was for https://sagebionetworks.jira.com/browse/PORTALS-3499 (not PLFM-3499)

PR #1674, opened 2025-03-27
PORTALS-3435: use  instead of  to assign -Dsonar.pullrequest.key

afwillia:
@tschaffter the original workflow failed because I assigned the env variable incorrectly using `:` instead of `=`


afwillia:
merging this without review since it's a small change to a previously approved workflow file.

PR #1678, opened 2025-03-28
PORTALS-3435: pass target and source branch to sonar scan to pick up changed files

afwillia:
While the sonar scan is now running on PR and commenting its analysis, it has yet to pick up code changes. Perhaps passing branch information in addition to the PR number will give the scan the information it needs to compare code.


afwillia:
Merging this w/out review since I'm able to and these are small change to a non-critical workflow file that has already been reviewed.

PR #1682, opened 2025-03-29
Use '@/' path alias for local imports

nickgros:
Should improve VS Code DX since it otherwise tries to import from 'src/*', which does not work in synapse-react-client

We had this in synapse-react-client a long time ago but removed it since it was a pain to configure with Webpack via the 'create-react-app' API. These aliases are already supported by our current Vite setup (thanks to the vite-tsconfig-paths plugin). The only 'extra' configuration change that was required was in our synapse-react-client Jest config.


jay-hodgson:
There are many SRC test failures (unexpected token?), are these related to the changes in this PR?  
Checked a few files to understand the change, and it looks good to me.

PR #1683, opened 2025-03-29
Add code-workspace file

nickgros:
Create a VS Code workspace file to suggest extensions, configure sensible default settings, and preconfigure test runner extensions. This should improve DX by allowing devs to interactively run and debug tests in the IDE.

### DIFF
--- a/synapse-web-monorepo.code-workspace
+++ b/synapse-web-monorepo.code-workspace
@@ -1,0 +1,31 @@
+{
+  // Shared VS Code settings for synapse-web-monorepo
+  // This file is used to configure the workspace settings and suggested extensions used by all devs
+  // Your preferred overrides can go in .vscode/settings.json
+
+  "folders": [
+    {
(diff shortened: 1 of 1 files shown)


synapse-web-monorepo.code-workspace
```diff
@@ -0,0 +1,31 @@
...
+    // General settings
+    "typescript.tsdk": "node_modules/typescript/lib",
+    "editor.defaultFormatter": "esbenp.prettier-vscode",
+
+    // Jest extension settings
+    "jest.runMode": "on-demand",
+    "jest.rootPath": "packages/synapse-react-client", // synapse-react-client is the only Jest project
+
+    // Vitest extension settings
+    "vitest.maximumConfigs": 99
```
nickgros, line 21: Load all vitest projects. All of our vitest suites are small, so I am not super worried about this causing performance issues. If it does, we can decrease it

PR #1695, opened 2025-04-01
PORTALS-3444: initialize AMP-ALS knowledge portal

jay-hodgson:


apps/portals/ampals/src/config/style/header.svg
nickgros, line 1: Is this AI-generated? Do we plan to replace this before final release? If yes to both, is it too much to ask that we add a visible "non-final" watermark to make it obvious that this will be replaced before the portal is pushed to production?

apps/portals/ampals/src/config/scripts/exportS3ProductionBucketName.sh
```diff
@@ -0,0 +1,2 @@
+export S3_PRODUCTION_BUCK_LOCATION=s3://prod.ampals.synapse.org
+export CF_PRODUCTION_DIST_ID=????????
```
nickgros, line 2: Hope this doesn't break things :) Has this been set up?

apps/portals/ampals/src/config/style/header.svg
jay-hodgson: I think it's a stock photo, not sure if it's going to be reworked before release

apps/portals/ampals/src/config/scripts/exportS3ProductionBucketName.sh
jay-hodgson: Chicken and the egg problem.  I think Xa is waiting for this configuration file to update to the correct value (and set up the dns)

nickgros (approved):

apps/portals/ampals/src/config/scripts/exportS3ProductionBucketName.sh
xschildw: prod: E1J1L5FUOCXZHU
staging: E27YTHLWH817NN

xschildw: Wait... Was it not amp-als.org?

apps/portals/ampals/src/config/style/header.svg
adamhindman: This is a stock photo, we have a license for it. The asset is attached—happy to resize it (including getting a larger version)
image

jay-hodgson:
bucket names are wrong, so this PR should not be merged until this is fixed

xschildw (approved):

PR #1699, opened 2025-04-02
fix: limit columns for data table to display to user

katiestahl:


katiestahl:
closes this issue in b2ai-standards-registry repo https://github.com/bridge2ai/b2ai-standards-registry/issues/252

katiestahl:
could possibly have a prettier/eslint verification failure here - for some reason the script isn't getting found, even when I manually add it. Still working on figuring that out

apps/portals/standards/src/config/synapseConfigs/data.ts
```diff
@@ -12,8 +12,8 @@ export const dataColumnLinks: LabelLinkConfig = [
   {
     isMarkdown: false,
     baseURL: 'Explore/Standard/DetailsPage',
-    URLColumnName: 'id',
-    matchColumnName: 'id',
+    URLColumnName: 'acronym',
```
jay-hodgson, line 15: Are we going to use acronym (instead of id) for the fk relationship in other tables on the details page?

katiestahl:
going to close this - I think this has already been addressed in another PR.

PR #1709, opened 2025-04-04
PORTALS-3532: use provided file name column in query context

jay-hodgson:


apps/portals/eliteportal/src/config/synapseConfigs/cohortbuilder.ts
```diff
@@ -120,7 +120,7 @@ export const filesViewQueryWrapperPlotNavProps: QueryWrapperPlotNavProps = {
   name: 'Data',
   enabledExternalAnalysisPlatforms: enabledAnalysisPlatforms,
   fileIdColumnName: 'id',
-  fileNameColumnName: 'name',
+  fileNameColumnName: 'fileName',
```
jay-hodgson, line 123: Use String column that contains the actual file name (the "name" column is an Entity type column that contains the Synapse ID)

packages/synapse-react-client/src/components/QueryWrapper/QueryWrapper.tsx
```diff
@@ -191,7 +192,7 @@ function QueryWrapperInternal(props: QueryWrapperProps) {
     lockedColumn,
     fileIdColumnName,
     fileVersionColumnName,
-    fileNameColumnName: fileIdColumnName,
+    fileNameColumnName,
```
jay-hodgson, line 195: Pass file name column through (do not overwrite with file ID column)

packages/synapse-react-client/src/synapse-queries/entity/useExportToCavatica.ts
```diff
@@ -24,9 +24,10 @@ export function useExportToCavatica(
     try {
       // add drs_uri to select
       const selectColumnsList = selectColumns
-        ?.map(col => `"${col.name}"`)
+        ?.filter(col => col.name != 'name')
```
jay-hodgson, line 27: Filter out any existing select column called 'name', since we will be adding the 'name' column in our sql to generate the manifest.

nickgros: may want to add a test for the filter

nickgros (approved):

PR #1711, opened 2025-04-07
SWC-7290 - fix bundling issue

nickgros:
- synapse-react-client: Bundle @sage-bionetworks/synapse-client to prevent future drift
- Release new versions of synapse-client and synapse-types to fix existing drift. These have not changed since the release of synapse-react-client v3.4.8, so they are compatible
- Remove references to UMD bundle since we no longer use it. Rename umd.index.ts to SWC.index.ts


packages/synapse-react-client/CONTRIBUTING.md
```diff
@@ -4,37 +4,35 @@
 
 ### Setup
```
nickgros, line 5: This doc is not well maintained. I saw some references to UMD that I wanted to remove, since we don't use UMD anymore. I made some other easy changes.

packages/synapse-react-client/src/SWC.index.ts
```diff
@@ -1,3 +1,10 @@
+/**
+ * SWC.index.ts represents the entry point for the Synapse React Client library as it is published to NPM and
+ * used by SynapseWebClient.
+ *
+ * If you wish to export a component or other object to be used by SynapseWebClient, you should add it to this file.
+ */
```
nickgros, line 6: Renamed umd.index.ts to SWC.index.ts and added this to clarify what this file is.

packages/synapse-react-client/vite.config.ts
```diff
@@ -8,11 +8,12 @@ const config = new ConfigBuilder()
   .setIncludeReactConfig(true)
   .setIncludeLibraryConfig(true, {
     except: [
-      // Include @sage-bionetworks/synapse-types because the local version may drift from the version released on NPM
+      // Include certain monorepo projects because the local versions may drift from the versions released on NPM
       '@sage-bionetworks/synapse-types',
+      '@sage-bionetworks/synapse-client',
```
nickgros, line 13: This will fix the bug

packages/synapse-react-client/vite.config.umd.ts
nickgros, line 1: File is not used since we removed our UMD build

packages/synapse-types/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@sage-bionetworks/synapse-types",
-  "version": "0.0.4",
+  "version": "0.0.5",
```
nickgros, line 3: also release new version of synapse-types just for consistency

packages/synapse-client/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@sage-bionetworks/synapse-client",
-  "version": "0.0.1",
+  "version": "0.0.2",
```
nickgros, line 3: release latest changes for @sage-bionetworks/synapse-client. One way to fix the bug in release 537 would be to update synapse-react-client to use this version once it's released to npm.

PR #1716, opened 2025-04-08
feat: add cards linking to related standards and training resources

katiestahl:
There's a lot of repeated code here that I would be happy to refactor, but I know @Sigfried is working on this file as well, so I'll hold off on that for now


katiestahl:
@jay-hodgson or @nickgros any idea how I can fix the failing check? It seems unrelated to my changes (I think)
`Run gh run cancel 14338069470
failed to determine base repo: failed to run git: fatal: not a git repository (or any of the parent directories): .git`

nickgros (approved):

PR #1720, opened 2025-04-08
SWC-7218: group ARs by type, then sort by approval status

jay-hodgson:


packages/synapse-react-client/src/components/AccessRequirementList/AccessRequirementList.stories.tsx
```diff
@@ -138,36 +139,81 @@ export const HasMetRequirements: Story = {
...
 }
 
-export const HasUnmetRequirements: Story = {
+const expandedMockAccessRequirements = [
+  ...mockAccessRequirements,
+  { ...mockManagedACTAccessRequirement, id: 21 },
+  { ...mockToUAccessRequirement, id: 22 },
+
+  { ...mockSelfSignAccessRequirement, id: 23 },
+  { ...mockACTAccessRequirement, id: 24 },
```
jay-hodgson, line 148: The mock access requirements are a mixed bag of types, where the first set have approvals and the second set do not.  The output is now sorted by AR type, and within each type they are sorted by approval status. 
We will never see this kind of mess outside of a testing environment (all types of ARs, where each type has both met and unmet represented).

packages/synapse-react-client/src/components/AccessRequirementList/AccessRequirementList.tsx
```diff
@@ -217,6 +220,21 @@ export default function AccessRequirementList(
...
+        // init if we have not seen this concrete type before
+        if (!acc[item.concreteType]) {
+          acc[item.concreteType] = []
+        }
+        acc[item.concreteType].push(item)
+        return acc
+      }, {} as Record<AccessRequirement['concreteType'], AccessRequirement[]>)
+    }
+    return undefined
+  }, [accessRequirements])
```
nickgros, line 236: Can you use lodash `groupBy`?

```suggestion
  const groupedAccessRequirementsByType: Record<AccessRequirement['concreteType'], AccessRequirement[]> = useMemo(() => {
    if (accessRequirements) {
      return groupBy(accessRequirements, 'concreteType')
    }
    return undefined
  }, [accessRequirements])
```

packages/synapse-react-client/src/components/AccessRequirementList/AccessRequirementList.tsx
```diff
@@ -232,6 +250,25 @@ export default function AccessRequirementList(
       },
     )
 
+  const sortedGroupedAcessRequirementsByType = useMemo(() => {
+    if (groupedAccessRequirementsByType && sortedAccessRequirementIds) {
```
nickgros, line 254: If the grouping `useMemo` turns into effectively a one-liner, I think you could just move the logic here.

```suggestion
    if (accessRequirements && sortedAccessRequirementIds) {
      const groupedAccessRequirementsByType = groupBy(accessRequirements, 'concreteType')
```

nickgros (approved):
LGTM, couple of suggestions

nickgros (approved):

PR #1726, opened 2025-04-09
PORTALS-3533

jay-hodgson:


packages/synapse-react-client/src/components/DownloadCart/ShowDownloadV2.tsx
```diff
@@ -67,4 +69,12 @@ export function ShowDownloadV2({ to, className = '' }: ShowDownloadV2Props) {
...
-export default ShowDownloadV2
+export default function ShowDownloadV2WithSuspense(props: ShowDownloadV2Props) {
+  return (
+    <ErrorBoundary FallbackComponent={EmptyFallbackComponent}>
+      <Suspense fallback={<SkeletonInlineBlock width={100} />}>
+        <ShowDownloadV2 {...props} />
+      </Suspense>
+    </ErrorBoundary>
+  )
+}
```
nickgros, line 80: Suggest renaming the inner component to ShowDownloadV2Internal and removing the export, then do

```suggestion
export function ShowDownloadV2(props: ShowDownloadV2Props) {
  return (
    <ErrorBoundary FallbackComponent={EmptyFallbackComponent}>
      <Suspense fallback={<SkeletonInlineBlock width={100} />}>
        <ShowDownloadV2Internal {...props} />
      </Suspense>
    </ErrorBoundary>
  )
}

export default function ShowDownloadV2

```

Having two exported components with different logic could be very confusing--I would not expect different behavior from the default export and the ShowDownloadV2 named export

nickgros:
This could happen in any other component that we forget to wrap in an error boundary. Should we add an error boundary to the top level of all portals?

nickgros (approved):

PR #1727, opened 2025-04-09
Add CreateOrUpdateDoiModal

nickgros:
- Reimplementation of DOI metadata editor in React


packages/synapse-client/src/spec/openapispecification.json
nickgros, line 1: Updated to latest version of the Synapse OpenAPI spec

packages/synapse-react-client/src/components/JsonSchemaForm/templates/BaseInputTemplate.tsx
```diff
@@ -34,7 +34,8 @@ export default function BaseInputTemplate<
     disabled,
     type,
     label,
-    // hideLabel,
+    hideLabel: _hideLabel,
+    hideError: _hideError,
```
nickgros, line 38: Prevent these from being included in the `...textFieldProps` spread below. These were passed to a component that didn't recognize them, causing a warning to be thrown in dev/test

packages/synapse-react-client/src/components/doi/CreateOrUpdateDoiModal.tsx
```diff
@@ -0,0 +1,324 @@
...
+/**
+ * Returns true if the DOI has fields that are not exposed in the UI, so we can let the user know that any changes
+ * may be destructive to hidden fields.
+ */
+function getWasModifiedViaAPI(doi: V2Doi) {
+  if (doi.creators?.some(creator => !isEmpty(creator.nameIdentifiers))) {
+    return true
+  }
+  return false
+}
```
nickgros, line 81: e.g. you can add an ORCiD to your DOI programmatically, but we don't show that in the UI

packages/synapse-react-client/src/components/doi/CreateOrUpdateDoiModal.tsx
```diff
@@ -0,0 +1,324 @@
...
+      <Typography variant={'body1'} gutterBottom>
+        Minting a DOI allows you to credit contributors and makes it easy to
+        cite your work.
+      </Typography>
+      <Typography variant={'body1'} gutterBottom>
+        Note that while the DOI and its associated metadata will be publicly
+        visible outside of Synapse, your data will still adhere to its existing
+        access conditions. If your data is private, it will remain restricted,
+        but the DOI will still be listed.
+        {/* Once a DOI is minted, the resource will be added to the Synapse Data Catalog. */}
```
nickgros, line 212: This was in the design, but is it true?

packages/synapse-react-client/src/testutils/ReactQueryMockUtils.ts
```diff
@@ -121,3 +121,28 @@ export function getUseMutationMock<TData = any, TError = any, TVariables = any>(
     submittedAt: 0,
   } satisfies UseMutationResult<TData, TError, TVariables>
 }
+
+export function getUseMutationPendingMock<
```
nickgros, line 125: utility to simplify mocking a mutation in the `'pending'` state

packages/synapse-react-client/src/components/doi/DoiFormSchemas.ts
```diff
@@ -0,0 +1,78 @@
...
+        'PhysicalObject',
+        'Service',
+        'Software',
+        'Sound',
+        'Text',
+        'Workflow',
+        'Other',
+      ],
+    },
+  },
```
nickgros, line 26: Definition for DoiResourceTypeGeneral was copied from the OpenAPI schema

packages/synapse-react-client/src/components/doi/DoiFormSchemas.ts
```diff
@@ -0,0 +1,78 @@
...
+      $ref: '#/definitions/org.sagebionetworks.repo.model.doi.v2.DoiResourceTypeGeneral',
+      title: 'General Resource Type',
+      description: 'The general type of content that best matches this work.',
+    },
+    publicationYear: {
+      type: 'integer',
+      title: 'Publication Year',
+      description:
+        'The year your work became or will become public. May be at most one year in the future.',
+    },
```
nickgros, line 56: The rest of the schema is purpose-built for the form. Using the raw Synapse OpenAPI schema creates a form that is much more complex than the UI should be.

packages/synapse-client/src/spec/openapispecification.json
nickgros, line 1: Updated to latest version of the Synapse OpenAPI spec

packages/synapse-react-client/src/synapse-queries/doi/useDOI.ts
```diff
@@ -44,3 +54,32 @@ export function useGetDOI(
...
+    ...options,
+    mutationFn: async request => {
+      const asyncJobId =
+        await synapseClient.doiServicesClient.postRepoV1DoiAsyncStart({
+          doiRequest: request,
+        })
+
+      return waitForAsyncResult(() =>
+        synapseClient.doiServicesClient.getRepoV1DoiAsyncGetAsyncToken({
+          asyncToken: asyncJobId.token!,
```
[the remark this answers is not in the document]
nickgros: This should never happen. The Synapse OpenAPI spec should indicate that `token` is non-null.

packages/synapse-client/src/util/waitForAsyncResult.ts
```diff
@@ -0,0 +1,29 @@
...
+ * @throws SynapseClientError if the job fails
+ */
+export async function waitForAsyncResult<
+  T extends AsynchronousResponseBody | AsynchronousJobStatus,
+>(
+  getAsyncResult: () => Promise<T>,
+  options?: { onCurrentStatusRetrieved: (response: T) => void },
+): Promise<T> {
+  let response = await getAsyncResult()
+  while ('jobState' in response && response.jobState === 'PROCESSING') {
```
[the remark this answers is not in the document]
nickgros: Not necessary for this case--as long as the backend tells us it is PROCESSING, we should keep polling.

packages/synapse-client/src/spec/openapispecification.json
nickgros, line 1: Updated Synapse OpenAPI spec to latest version

packages/synapse-client/src/util/waitForAsyncResult.ts
```diff
@@ -0,0 +1,29 @@
...
+    await delay(500)
```
jay-hodgson, line 22: We could make this delay logic more sophisticated, say by starting with a delay of 150ms and increasing until we hit some ceiling (like 500ms or a second)

packages/synapse-react-client/src/components/doi/CreateOrUpdateDoiModal.tsx
```diff
@@ -0,0 +1,324 @@
...
+
+  const [formData, setFormData] = useState<DoiFormData>({
+    titles: [''],
+    creators: [''],
+  })
+  const { data: currentUser, isLoading: isLoadingCurrentUser } =
+    useGetCurrentUserProfile()
+  const { data: entity, isLoading: isLoadingEntity } = useGetEntity(
+    objectId,
+    defaultVersionNumber,
```
jay-hodgson, line 108: any reason to use the selectedVersionNumber for the entity data instead (should they select a different version)?

packages/synapse-react-client/src/components/doi/CreateOrUpdateDoiModal.tsx
```diff
@@ -0,0 +1,324 @@
...
+  const entityTypeDisplay = entityTypeToFriendlyName(entityType)
+  let versionHelpMarkdown = `A DOI can be associated with a specific version of this ${entityTypeDisplay}.
+  
+  Versioned DOIs will link to the specified version of the ${entityTypeDisplay}.
+  
+  A DOI without a version will always link to the newest version of this ${entityTypeDisplay},
+     so the data that someone retrieves using the DOI may change over time.`
+  if (entity && isTable(entity)) {
+    versionHelpMarkdown += `\n\nTo create a DOI that will always link to the current set of data in the ${entityTypeDisplay},
+     create a new version and mint a DOI for that version.`
```
jay-hodgson, line 148: If it's a Table, should this say "create a new snapshot" instead of "create a new version"?  Version works for me, but I know we have other UI that refers to creating Snapshots

packages/synapse-react-client/src/components/doi/CreateOrUpdateDoiModal.tsx
jay-hodgson: @thomasyu888 ?  When a DOI is minted for an entity (let's assume the entity is made public), is there a process that will eventually discover and add a row to [the Data Catalog table](https://www.synapse.org/Synapse:syn61609402)?

packages/synapse-react-client/src/components/doi/CreateOrUpdateDoiModal.tsx
```diff
@@ -0,0 +1,324 @@
...
+                  version.versionLabel !== String(version.versionNumber)
+                    ? ` / ${version.versionLabel}`
+                    : ''}
+                </MenuItem>
+              ))}
+          </Select>
+        </StyledFormControl>
+      )}
+      <div className="JsonSchemaFormContainer">
+        <JsonSchemaForm
```
jay-hodgson, line 263: Smart use of a RJSFSchema for this case

jay-hodgson (approved):

packages/synapse-react-client/src/components/doi/CreateOrUpdateDoiModal.tsx
thomasyu888: This is _not_ true right now, but I think that was Alberto's intent for this to be true

packages/synapse-react-client/src/components/doi/CreateOrUpdateDoiModal.tsx
nickgros: I think that's what the original implementation does -- it would be useful because the title of the entity may have changed between versions, which we could use to populate the form. I will update this.

packages/synapse-react-client/src/components/doi/CreateOrUpdateDoiModal.tsx
nickgros: This is the same text we have in the GWT implementation. We have some utilities to handle the different cases (e.g. Tables have "snapshots" but Datasets have "versions"). I will see if it's easy to integrate those utilities

jay-hodgson (approved):

PR #1729, opened 2025-04-09
PORTALS-3448: AMP-ALS - More Resources component

kianamcc:
Jira: https://sagebionetworks.jira.com/browse/PORTALS-3448?atlOrigin=eyJpIjoiM2IyZGQ5MTMxZjIyNDQyZThhZmZlZGQ3ZTdkZDQxNWMiLCJwIjoiaiJ9

Screenshot 2025-04-09 at 11 41 33 AM


packages/synapse-react-client/.storybook/main.mts
```diff
@@ -1,4 +1,4 @@
-import { StorybookConfig } from '@storybook/react-vite'
+import type { StorybookConfig } from '@storybook/react-vite'
```
kianamcc, line 1: https://github.com/storybookjs/storybook/blob/next/MIGRATION.md#typescript-storybookconfig-type-moved

packages/synapse-react-client/src/components/MoreResources/MoreResources.tsx
```diff
@@ -0,0 +1,93 @@
...
+  resources: ResourceConfig[]
+}
+
+const MoreResources = ({ resources }: MoreResourcesProps) => {
+  const theme = useTheme()
+  return (
+    <Box
+      sx={{
+        padding: { xs: '40px', lg: '60px 80px 80px 80px' },
+        background: `linear-gradient(0deg, rgba(255, 255, 255, 0.90) 0%, rgba(255, 255, 255, 0.90) 100%), ${theme.palette.primary.main}`,
```
nickgros, line 20: maybe I'm going crazy but I feel like this is just a flat color when I look at it in Figma or Storybook. If it is, can we just use the flat color?

packages/synapse-react-client/src/components/MoreResources/MoreResources.tsx
```diff
@@ -0,0 +1,93 @@
...
+          fontSize: '24px',
+          fontWeight: 600,
+          lineHeight: 'normal',
+          marginBottom: '40px',
+          textAlign: 'center',
+          color: 'grey.1000',
+        }}
+      >
+        More Resources
+      </Typography>
```
nickgros, line 34: This wrapper & title probably should not be in the component. I think in the portal, this should be handled using the `SectionLayout` component just like other home pages. As a bonus, the `SectionLayout` will automatically handle hash routing as needed.

packages/synapse-react-client/src/components/MoreResources/MoreResources.tsx
```diff
@@ -0,0 +1,93 @@
...
+                  lineHeight: '150%',
+                  color: 'primary.main',
+                }}
+              >
+                More Resources
+              </Typography>
+            </Link>
+          </Stack>
+        ))}
+      </Box>
```
nickgros, line 88: Let's try to use what MUI & our theme variants already provide us? It would help us get rid of a lot of code.

So for example, use the builtin `Card` instead of custom box shadow/background, use the Typography `variant`s, the Link is automatically the primary color.

I took a crack at it and this is what I came up with

```suggestion
      <Box
        sx={{
          display: 'flex',
          gap: 2,
          width: '100%',
          flexDirection: { xs: 'column', md: 'row' },
        }}
      >
        {resources.map((resource, index) => (
          <Card
            key={index}
            sx={{
              padding: '24px',
              flex: '1',
            }}
          >
            <Stack sx={{ height: '100%', gap: 2 }}>
              <Typography variant={'headline3'}>{resource.title}</Typography>
              <Typography
                variant={'body1'}
                sx={{ fontSize: '14px', flex: '1' }}
              >
                {resource.description}
              </Typography>
              <Typography
                sx={{
                  fontSize: '14px',
                  marginTop: 'auto',
                }}
              >
                <Link
                  component={RouterLink}
                  to={encodeURI(`${resource.link}`)}
                  sx={{ textDecoration: 'none' }}
                >
                  More Resources
                </Link>
              </Typography>
            </Stack>
          </Card>
        ))}
      </Box>

```

packages/synapse-react-client/src/components/MoreResources/MoreResources.tsx
```diff
@@ -0,0 +1,93 @@
...
+  title: string
+  description: string
+  link: string
+}
+
+export type MoreResourcesProps = {
+  resources: ResourceConfig[]
+}
+
+const MoreResources = ({ resources }: MoreResourcesProps) => {
```
nickgros, line 14: If we extract the title, is this still a good name for this component? These are just basic cards with descriptions and links, they don't have to represent whatever 'resource' means to the ALS portal

apps/portals/ampals/src/pages/HomePage.tsx
```diff
@@ -4,11 +4,31 @@ import AMPALSHeader from '@sage-bionetworks/synapse-portal-framework/components/
 // import { FeaturedDataTabs } from 'synapse-react-client'
 // import columnAliases from '../config/columnAliases'
 import headerSvg from '../config/style/header.svg?url'
-import { FullWidthAlert } from 'synapse-react-client'
+import { FullWidthAlert, MoreResources } from 'synapse-react-client'
 import { OrientationBanner } from 'synapse-react-client'
 
 //TODO
 export default function HomePage() {
+  const resources = [
```
nickgros, line 12: Please rename to `moreResourcesContent` to avoid confusion with the code in resources.ts

packages/synapse-react-client/src/components/MoreResources/MoreResources.tsx
```diff
@@ -0,0 +1,93 @@
...
+              }}
+            >
+              {resource.title}
+            </Typography>
+            <Typography sx={{ fontSize: '14px', lineHeight: 'normal' }}>
+              {resource.description}
+            </Typography>
+            <Link
+              component={RouterLink}
+              to={encodeURI(`${resource.link}`)}
```
nickgros, line 72: The URI should be encoded before it is passed in. You may end up double-encoding it if you do it here

kianamcc:
I changed the name of the component to CardGridWithLinks to make it more generic.

apps/portals/ampals/src/pages/HomePage.tsx
```diff
@@ -61,6 +83,25 @@ export default function HomePage() {
             ]}
           /> */}
         </SectionLayout>
+        <SectionLayout
+          title={'More Resources'}
+          centerTitle
+          ContainerProps={{
+            sx: {
+              padding: { xs: '40px', lg: '60px 80px 80px 80px' },
```
nickgros, line 91: this padding (or some parts of it) may be unnecessary since the `Container` component already constrains the width

apps/portals/ampals/src/pages/HomePage.tsx
```diff
@@ -61,6 +83,25 @@ export default function HomePage() {
...
+              backgroundColor: alpha(theme.palette.primary.main, 0.05),
```
nickgros, line 92: I think the other portal homepages have the background in a separate div that wraps the `SectionLayout`. That way the background element is always full width, and the SectionLayout can constrain the width of the contents without worrying about the background style

packages/synapse-react-client/src/components/CardGridWithLinks/CardGridWithLinks.tsx
```diff
@@ -0,0 +1,67 @@
...
+    <Box
+      sx={{
+        display: 'flex',
+        gap: 2,
+        width: '100%',
+        flexDirection: { xs: 'column', md: 'row' },
+      }}
+    >
+      {cards.map((card, index) => {
+        const encodedURI = encodeURI(card.link)
```
nickgros, line 26: Sorry, I was unclear. I meant: don't call `encodeURI` at all in this component. If the URI needs to be encoded, it should be encoded before passing it to `CardGridWithLinks`

packages/synapse-react-client/src/components/CardGridWithLinks/CardGridWithLinks.tsx
```diff
@@ -0,0 +1,67 @@
...
+  description: string
+  link: string
+}
+
+export type CardGridWithLinksProps = {
+  linkText: string
+  cards: CardGridWithLinksConfig[]
+}
+
+const CardGridWithLinks = ({ cards, linkText }: CardGridWithLinksProps) => {
```
nickgros, line 15: New name is great

apps/portals/ampals/src/pages/HomePage.tsx
kianamcc: Looks like I can use the home-bg-dark class for the background color instead!

packages/synapse-react-client/src/components/CardGridWithLinks/CardGridWithLinks.stories.tsx
```diff
@@ -0,0 +1,42 @@
...
+        link: 'https://example-1.com',
+      },
+      {
+        title: 'Title 2',
+        description: 'Description text 2.',
+        link: 'https://example-2.com',
+      },
+      {
+        title: 'Title 2',
+        description: 'Description text 3.',
```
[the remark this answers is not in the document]
kianamcc: Verifying that cards remain the same size regardless of content length: Screenshot 2025-04-10 at 10 06 00 AM

apps/portals/ampals/src/pages/HomePage.tsx
kianamcc: Could we actually remove the maxWidth from SectionLayout so that this component has padding consistent with the header component?:

Screenshot 2025-04-10 at 10 33 21 AM

nickgros (approved):

PR #1731, opened 2025-04-10
PORTALS-3539: use raw file content, where url is in a table

jay-hodgson:


jay-hodgson:
The DatasetJsonLdScript.tsx component is currently only included in the NF portal Dataset Details page, but it could be extended to other portal dataset pages.
I need to be logged in (since the table is not yet anonymously downloadable), but it works!
Screenshot 2025-04-10 at 3 34 52 PM

PR #1737, opened 2025-04-11
Portals 3386-1 - use datasetAlias instead of synapseLink for dataset card

afwillia:


afwillia:
Most records in the CCKP dataset table do not have a synapse ID in the `datasetAlias` field. So take make testing easier, I filter to `Research Theme = Computational Resource` which atm has 14 entries, one of which has a synapse ID.

This needs a bit more work because the download and citation buttons stack vertically instead of horizontally. 
image

jay-hodgson (approved):

PR #1743, opened 2025-04-14
Create Form for Contributing a New Standard

jennifer-bowser:
Creates a form to allow users to submit requests for new standards. The form has no submit logic yet; this will be implemented in a future ticket. Relatedly, the new "Contribute A Standard" page containing this form has also not been hooked up to the UI yet.

### DIFF
--- a/packages/synapse-react-client/src/components/Form/Form.tsx
+++ b/packages/synapse-react-client/src/components/Form/Form.tsx
@@ -17,14 +17,14 @@ type FieldConfig = {
   componentType: FieldType
   label: string
   helperText: string | React.ReactElement<'span'>
-  placeholder?: string
+  placeholder: string
   isRequired: boolean
   selections?: {
     [id: string]: string
-  }
-  additionalOptions?: {
+  } // Dropdown options for 'Select' componentTypes
+  additionalProperties?: {
     [id: string]: any
-  }
+  } // intentionally flexible
 }
 
 export type FieldSchema = {
@@ -92,7 +92,6 @@ export default function Form({ fields, onSubmit }: FormProps) {
 
   const createField = (id: string, config: FieldConfig) => {
     if (config.componentType === Select) {
-      const placeholderText = 'Select a Category'
       return (
         <FormControl
           fullWidth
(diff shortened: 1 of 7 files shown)


packages/synapse-react-client/src/components/Form/Form.tsx
```diff
@@ -0,0 +1,188 @@
...
+export type FormProps = {
+  fields: FieldSchema
+  onSubmit: (data: FormData) => boolean // return True on succesful submission; else False
+}
+
+type ValidFields = {
+  [id: string]: boolean
+}
+
+export default function Form({ fields, onSubmit }: FormProps) {
```
nickgros, line 47: In some components [1], [2], we have been using [react-jsonschema-form](https://rjsf-team.github.io/react-jsonschema-form/docs/), which is really similar to what you have built here. You could just construct the schema (and maybe a uiSchema) and pass it in to the [JsonSchemaForm component](https://github.com/Sage-Bionetworks/synapse-web-monorepo/blob/1e4d55f1f79e16f5f17de5f9550c945c24b833a6/packages/synapse-react-client/src/components/JsonSchemaForm/JsonSchemaForm.tsx#L25) (it was recently refactored so you might have to synchronize with `main` to use the component).

Do you think using this existing pattern might help us accomplish the same/similar UI with less code? It is possible that your solution handles some trivial case that is not easy/possible in RJSF.

I'm sorry that we didn't identify this and point out this option to you earlier!!

[1] [CreateOrUpdateDoiModal](https://github.com/Sage-Bionetworks/synapse-web-monorepo/blob/a02aa154dcb9575f6c11a7e5a93800af7db3952a/packages/synapse-react-client/src/components/doi/CreateOrUpdateDoiModal.tsx#L101)
image

[2] [SchemaDrivenAnnotationEditor](https://github.com/Sage-Bionetworks/synapse-web-monorepo/blob/b777462c0945446f81f51da95abfe37a2af17717/packages/synapse-react-client/src/components/SchemaDrivenAnnotationEditor/SchemaDrivenAnnotationEditor.tsx#L113)
image

jennifer-bowser: Thanks, I'll take a look and see if I can use the JsonSchemaForm instead!

jennifer-bowser: @nickgros Update: I see what you mean, JsonSchemaForm was a great fit for this! I've finished the refactor and am ready for re-review.

nickgros (approved):

PR #1745, opened 2025-04-14
IT-4359: change standards to b2ai.standards

jay-hodgson:


nickgros (approved):

nickgros (approved):
A lot could go wrong, or maybe it will all work. Worth a try

xschildw (approved):
Let's try...

PR #1752, opened 2025-04-15
Portals 3386 f/u - Fix download cart button

afwillia:
This PR fixes bad logic that prevented the Download button to trigger download cart logic. Also, it spaces card buttons horizontally instead of vertically

### DIFF
--- a/packages/synapse-react-client/src/components/EntityDownloadConfirmation/EntityDownloadConfirmation.tsx
+++ b/packages/synapse-react-client/src/components/EntityDownloadConfirmation/EntityDownloadConfirmation.tsx
@@ -68,7 +68,7 @@ export function EntityDownloadConfirmation({
     onIsLoadingChange(isLoading)
   }, [isLoading, onIsLoadingChange])
 
-  if (isLoading || (!isLoading && entity)) {
+  if (isLoading) {
     if (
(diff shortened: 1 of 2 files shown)


afwillia:
Screenshot 2025-04-15 at 11 01 51 AM

packages/synapse-react-client/src/components/EntityDownloadConfirmation/EntityDownloadConfirmation.tsx
```diff
@@ -68,7 +68,7 @@ export function EntityDownloadConfirmation({
     onIsLoadingChange(isLoading)
   }, [isLoading, onIsLoadingChange])
 
-  if (isLoading || (!isLoading && entity)) {
+  if (isLoading) {
```
jay-hodgson, line 71: What happens if the entity ID points to a FileEntity or a Docker repo (for example)?
Does the QueryWrapperTableDownloadConfirmation handle this case properly?

afwillia: Screenshot 2025-04-15 at 12 02 32 PM
you get an error message like this if the entity type isn't compatible

jay-hodgson (approved):

PR #1756, opened 2025-04-16
PORTALS-3552

nickgros:
Fix bug where MarkdownSynapse would not re-process markdown when the `markdown` prop changed.

### DIFF
--- a/packages/synapse-react-client/src/components/Markdown/MarkdownSynapse.test.tsx
+++ b/packages/synapse-react-client/src/components/Markdown/MarkdownSynapse.test.tsx
@@ -46,18 +46,17 @@ jest.mock('../ProvenanceGraph/ProvenanceGraph', () => {
   }
 })
 
+function getComponent(props: MarkdownSynapseProps) {
+  return (
+    <MarkdownSynapse ownerId="mock_owner_id" wikiId="mock_wiki_id" {...props} />
+  )
+}
+
 const renderComponent = (
   props: MarkdownSynapseProps,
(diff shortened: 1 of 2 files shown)


packages/synapse-react-client/src/components/Markdown/MarkdownSynapse.test.tsx
```diff
@@ -114,6 +113,18 @@ describe('MarkdownSynapse tests', () => {
...
+      const firstVal = 'loremipsum....'
+      const secondVal = 'dolorsit....'
+      const { rerender } = renderComponent({
+        markdown: firstVal,
+      })
+      screen.getByText(firstVal)
+
+      rerender(getComponent({ markdown: secondVal }))
+      screen.getByText(secondVal)
+    })
```
nickgros, line 126: Verified this test fails without the code change

packages/synapse-react-client/src/components/Markdown/MarkdownSynapse.tsx
```diff
@@ -615,14 +615,24 @@ const MarkdownSynapse: MarkdownSynapseComponent = class MarkdownSynapse extends
 
   // on component update find and re-render the math/widget items accordingly
   public componentDidUpdate(prevProps: MarkdownSynapseProps) {
```
nickgros, line 617: Resisting every urge to rewrite this with hooks :melting_face:

jay-hodgson (approved):

PR #1759, opened 2025-04-16
PORTALS-3451: AMP ALS: Access column in Explore Files table

kianamcc:
Jira: https://sagebionetworks.jira.com/browse/PORTALS-3451

Screenshot 2025-04-16 at 1 01 49 PM


packages/synapse-react-client/src/components/SynapseTable/SynapseTableRenderers.tsx
```diff
@@ -180,9 +181,16 @@ const getEntityOrRowVersion = (
...
   return (
     <div data-testid={'AccessCell'}>
-      <HasAccessV2 key={entityId} entityId={entityId} showButtonText={false} />
+      {!isExternalFile && (
+        <HasAccessV2
+          key={entityId}
+          entityId={entityId}
+          showButtonText={false}
+        />
+      )}
```
nickgros, line 193: `HasAccessV2` already looks at data to make a decision about what icon to show. Is there a compelling reason to put this logic in the `SynapseTableRenderer` and not inside of `HasAccessV2`?

packages/synapse-react-client/src/components/SynapseTable/SynapseTableRenderers.tsx
```diff
@@ -180,9 +181,16 @@ const getEntityOrRowVersion = (
 
 function AccessCell(props: CellContext<Row, unknown>) {
   const entityId = getEntityOrRowId(props)!
+  const isExternalFile = useIsExternalFileEntity({ entityId })
```
nickgros, line 184: Note that today, nearly all of the data we need to render a table is retrieved in `usePrefetchTableData` to reduce the number of HTTP requests we must make (that work was tracked in PORTALS-2970). Adding this hook as implemented would add 2 HTTP requests per row (typically 50 more requests per page of data)

Unfortunately, none of the existing requests in `usePrefetchTableData` give you the file handle handle data unless the view has a `dataFileHandleId` column, which is not guaranteed (and I'm not sure if our react-query cache would even reuse that data today).

Further, there is no Synapse API that makes it possible to batch-retrieve a set of dataFileHandleIds given a set of FileEntity IDs. Ideally we would have PLFM-8447 to do this, but unfortunately that has not been implemented.

So I guess the question to figure out is: how can we do this efficiently without adding a ton of network overhead to all portals?

packages/synapse-react-client/src/utils/hooks/useIsExternalFileEntity.ts
```diff
@@ -0,0 +1,55 @@
...
+          ],
+          includeFileHandles: true,
+          includePreSignedURLs: false,
+          includePreviewPreSignedURLs: false,
+        }
+        const file = await SynapseClient.getFiles(batchFileRequest, accessToken)
+        const externalFileHandle = file.requestedFiles[0].fileHandle
+        if (externalFileHandle && isExternalFileHandle(externalFileHandle)) {
+          setIsExternal(true)
+        }
```
nickgros, line 46: Should use existing react-query hooks (ok to compose them in a larger hook), but hold off on changing until we resolve the fundamental issues

packages/synapse-react-client/src/components/SynapseTable/SynapseTableRenderers.tsx
nickgros: Couple of options:
* We could make this check opt-in for the AMP-ALS portal only (fast/easy to build, but AMP-ALS gets the performance issue and it's basically technical debt)
* Ask the backend to prioritize batch entity bundle (or adding file handle info to some other call)
* @kianamcc could work on the required backend work :wink: - would require coordination with the backend engineers

are there better options I am overlooking?

packages/synapse-react-client/src/utils/hooks/useIsExternalFileEntity.ts
nickgros: You have two options, you could either use 
- `useGetEntity` and `useGetFileBatch` OR
- `useGetEntityBundle` <- probably this one since it's only one call

packages/synapse-react-client/src/components/SynapseTable/SynapseTableRenderers.tsx
jay-hodgson: Are we at risk of getting being throttled with this kind of activity 50 additional requests for a page of data)?  To guard against this, we may want to only render if it has ever entered the viewport (useInView from react-intersection-observer).

packages/synapse-react-client/src/components/SynapseTable/SynapseTableRenderers.tsx
kianamcc: logic inside HasAccessV2 now

packages/synapse-react-client/src/components/SynapseTable/SynapseTableRenderers.tsx
nickgros: @jay-hodgson There the main throttle filter that this would affect is the [UserConcurrentConnectionThrottler](https://github.com/Sage-Bionetworks/Synapse-Repository-Services/blob/43c91e917c60c11b83bc89ae658df76c834bd357/services/repository/src/main/java/org/sagebionetworks/repo/web/filter/throttle/UserConcurrentConnectionThrottler.java#L20), which limits users to 3 concurrent connections per machines. There are 6 machines, so we can issue up to 18 requests at a time on a user's behalf before getting throttled (maybe fewer on average depending on our load balancing algorithm).

packages/synapse-react-client/src/components/HasAccess/HasAccessV2.tsx
```diff
@@ -126,6 +132,46 @@ export function useGetRestrictionUiType(
...
+  }
+
+  const hasExternalFileHandle =
+    fileHandle.concreteType ===
+    'org.sagebionetworks.repo.model.file.ExternalFileHandle'
+
+  return hasExternalFileHandle
+}
+
+const InternalOnlyAccessWrapper = ({
```
kianamcc, line 163: I can't call useIsExternalFileEnitity custom hook conditionally so I created another helper function.

packages/synapse-react-client/src/components/SynapseTable/SynapseTableRenderers.tsx
```diff
@@ -179,10 +180,30 @@ const getEntityOrRowVersion = (
 }
 
 function AccessCell(props: CellContext<Row, unknown>) {
+  const { ref, inView } = useInView({ triggerOnce: true })
```
kianamcc, line 183: When to render logic

packages/synapse-react-client/src/components/HasAccess/HasAccessV2.tsx
```diff
@@ -126,6 +142,40 @@ export function useGetRestrictionUiType(
...
+        const fileHandles = bundle.fileHandles as FileHandle[]
+        const fileHandle: FileHandle | undefined = fileHandles?.find(
+          fileHandle => fileHandle.id === fileEntity.dataFileHandleId,
+        )
+        return fileHandle && implementsExternalFileHandleInterface(fileHandle)
+      },
+    },
+  )
+
+  return { isExternalFileHandle: entityBundle, isLoading }
```
nickgros, line 176: Just return the entire `UseQueryResult`
```suggestion
  return useGetEntityBundle(
    entityId,
    undefined,
    undefined,
    {
      ...options,
      select: bundle => {
        if (!isFileEntity(bundle.entity)) {
          return false
        }
        const fileEntity = bundle.entity as FileEntity
        const fileHandles = bundle.fileHandles as FileHandle[]
        const fileHandle: FileHandle | undefined = fileHandles?.find(
          fileHandle => fileHandle.id === fileEntity.dataFileHandleId,
        )
        return fileHandle && implementsExternalFileHandleInterface(fileHandle)
      },
    },
  )
```

packages/synapse-react-client/src/components/HasAccess/HasAccessV2.tsx
```diff
@@ -137,10 +187,18 @@ export function HasAccessV2(props: HasAccessProps) {
     AccessRequirement[]
   >([])
 
-  const { entityId, showButtonText = true } = props
-
+  const {
+    entityId,
+    showButtonText = true,
+    hideAccessIconForExternalFileHandle,
```
nickgros, line 193: Set the default value here so you don't have to do `??` or anything like that
```suggestion
    hideAccessIconForExternalFileHandle = false,
```

packages/synapse-react-client/src/components/HasAccess/HasAccessV2.tsx
```diff
@@ -137,10 +187,18 @@ export function HasAccessV2(props: HasAccessProps) {
...
+  } = props
   const restrictionUiTypeValue = useGetRestrictionUiType(entityId)
 
+  const { isExternalFileHandle, isLoading } = useGetIsExternalFileHandle(
+    entityId,
+    { enabled: hideAccessIconForExternalFileHandle ?? false },
+  )
```
nickgros, line 200: Simplify & rename `isLoading` for clarity
```suggestion
  const { isExternalFileHandle, isLoading: isLoadingIsExternalFileHandle } = useGetIsExternalFileHandle(
    entityId,
    { enabled: hideAccessIconForExternalFileHandle },
  )
```

packages/synapse-react-client/src/components/HasAccess/HasAccessV2.tsx
```diff
@@ -257,11 +315,15 @@ export function HasAccessV2(props: HasAccessProps) {
...
 
-  if (!restrictionUiTypeValue) {
+  if (!restrictionUiTypeValue || isLoading) {
     // loading
     return <></>
   }
 
+  if (hideAccessIconForExternalFileHandle && isExternalFileHandle) {
+    return null
+  }
```
nickgros, line 325: I feel like this 'no icon' design is a stopgap solution where ideally we would be able to just give the user better information. It looks like this might be the case in the design ([jira comment asking for clarification](https://sagebionetworks.jira.com/browse/PORTALS-3451?focusedCommentId=246618)).

With that in mind, I wonder if we should consider this case to be a new `RestrictionUiType` instead of just returning `null`? Then, in the future we could update this as desired to show different text or an icon to better explain this case to users.

To make this change, I think you would call your new hook in `useGetRestrictionUiType` instead of here and add logic to return a new `RestrictionUiType.AccessibleExternalFileHandle` that shows no icon (or the new one in the design) and text that might say 'Access Controlled by Another System' (or something better?)

packages/synapse-react-client/src/components/SynapseTable/SynapseTableRenderers.tsx
```diff
@@ -179,10 +180,30 @@ const getEntityOrRowVersion = (
...
+  // If hideAccessIconForExternalFileHandle is true, defer rendering until the cell is in view to avoid unnecessary API calls
+  const canRenderAccessIcon =
+    !hideAccessIconForExternalFileHandle ||
+    (hideAccessIconForExternalFileHandle && inView)
+
   return (
-    <div data-testid={'AccessCell'}>
-      <HasAccessV2 key={entityId} entityId={entityId} showButtonText={false} />
+    <div
+      ref={hideAccessIconForExternalFileHandle ? ref : undefined}
```
nickgros, line 194: I think it's ok to always attach the ref, since you have `canRenderAccessIcon`
```suggestion
      ref={ref }
```

packages/synapse-react-client/src/types/tanstack-table-augmentation.d.ts
```diff
@@ -12,6 +12,7 @@ type ColumnFilterEnumValue<TValue = unknown> = {
 declare module '@tanstack/react-table' {
   interface TableMeta<TData extends RowData> {
     rowSet?: RowSet
+    hideAccessIconForExternalFileHandle?: boolean
```
nickgros, line 15: Instead of here, can you put this property in `SynapseTableContext`, which is specific to `SynapseTable`? This `TableMeta` would ideally be general to other tables in the UI (we have clearly not been great at this, so putting it here was a reasonable choice)

apps/portals/ampals/src/config/synapseConfigs/data.ts
```diff
@@ -25,6 +25,8 @@ export const dataQueryWrapperPlotNavProps: QueryWrapperPlotNavProps = {
   defaultShowSearchBox: true,
   tableConfiguration: {
     // columnLinks: dataColumnLinks,
+    showAccessColumn: true,
+    showExternalAccessIcon: true,
```
kianamcc, line 29: How does this new name sound?

packages/synapse-react-client/src/components/HasAccess/HasAccessV2.tsx
```diff
@@ -14,12 +19,17 @@ import AccessRequirementList, {
...
+  implementsExternalFileHandleInterface,
+  isFileEntity,
+} from '@/utils/types/IsType'
 
 export type HasAccessProps = {
   onHide?: () => void
   entityId: string
   className?: string
   showButtonText?: boolean
+  showExternalAccessIcon?: boolean
```
nickgros, line 32: Also add the doc here

```suggestion
  /**
   * If true, the component will show enhanced UI for the case where
   * - the entity is a FileEntity, AND
   * - the caller has permission to fetch the dataFileHandle, AND
   * - the dataFileHandle is an instance of ExternalFileHandleInterface (i.e. the file is not controlled by Synapse)
   * Note that this requires an additional API call that cannot be batched, so it should be avoided in bulk  contexts if possible.
   * @default false
   */
  showExternalAccessIcon?: boolean
```

packages/synapse-react-client/src/components/HasAccess/HasAccessV2.tsx
```diff
@@ -51,18 +62,21 @@ const iconConfiguration: Record<
...
-
   [RestrictionUiType.Accessible]: {
     icon: 'accessOpen',
     color: theme => theme.palette.success.main,
     tooltipText: '',
   },
+  [RestrictionUiType.AccessibleExternalFileHandle]: {
+    icon: 'linkOff',
+    color: theme => theme.palette.grey[700],
+    tooltipText: 'Access controlled by an external system.',
```
nickgros, line 78: Small tweak to make this more generalizable, not sure if the design prescribed this text. Generally, we don't know if there are additional access controls.

```suggestion
    tooltipText: 'Access may be controlled by an external system.',
```

packages/synapse-react-client/src/components/HasAccess/HasAccessV2.tsx
```diff
@@ -95,6 +109,7 @@ function AccessIcon(props: { restrictionUiType: RestrictionUiType }) {
  */
 export function useGetRestrictionUiType(
   entityId: string,
+  options: { enabled: boolean },
```
nickgros, line 112: Make the parameter more specific, since it doesn't apply to the other query hooks called here. Also, ideally use the full type if you can. It would probably look something like this:
```suggestion
  useGetIsExternalFileHandleOptions: Partial<UseQueryOptions<EntityBundle, SynapseClientError, boolean>>
```

packages/synapse-react-client/src/components/HasAccess/HasAccessV2.tsx
```diff
@@ -104,10 +119,24 @@ export function useGetRestrictionUiType(
...
   }
 
+  if (options.enabled) {
+    if (isLoadingExternalFile) {
+      return undefined
+    }
+    if (isExternalFileHandle) {
+      return RestrictionUiType.AccessibleExternalFileHandle
+    }
+  }
```
nickgros, line 138: Unnecessary check, if `enabled` is false, `isLoadingExternalFile` will always be false.

I think it's ok to have the `isExternalFileHandle` check here; if this is true, then we definitely have met all ARs and we do have `canDownload`.

```suggestion
  if (isLoadingExternalFile) {
    return undefined
  }
  
  if (isExternalFileHandle) {
    return RestrictionUiType.AccessibleExternalFileHandle
  }
```

packages/synapse-react-client/src/components/HasAccess/HasAccessV2.tsx
```diff
@@ -126,6 +155,33 @@ export function useGetRestrictionUiType(
...
+/**
+ * This hook determines if
+ * - the entity is a FileEntity, AND
+ * - the caller has permission to fetch the dataFileHandle, AND
+ * - the dataFileHandle is an instance of ExternalFileHandleInterface (i.e. the file is not controlled by Synapse)
+ * Note that this requires an additional API call that cannot be batched, so it should be avoided in bulk contexts if possible.
+ */
+function useGetIsExternalFileHandle(
+  entityId: string,
+  options: { enabled: boolean },
```
nickgros, line 167: ideally use the full type if you can
```suggestion
  options: Partial<UseQueryOptions<EntityBundle, SynapseClientError, boolean>>,
```

packages/synapse-react-client/src/components/SynapseTable/SynapseTable.tsx
```diff
@@ -57,6 +58,7 @@ export type SynapseTableProps = {
 
   /** If true and entity is a view or dataset, renders a column that represents if the caller has permission to download the entity represented by the row */
   showAccessColumn?: boolean
+  showExternalAccessIcon?: boolean
```
nickgros, line 61: Add doc

packages/synapse-react-client/src/components/SynapseTable/SynapseTableContext.ts
```diff
@@ -3,6 +3,12 @@ import { LabelLinkConfig } from '../CardContainerLogic'
 
 export type SynapseTableContextType = {
   columnLinks?: LabelLinkConfig
+  /**
+   * If true, the icon will not be shown if the entity is a FileEntity and the dataFileHandle is an ExternalFileHandle (caller must have permission to retrieve the dataFileHandle)
+   * NOTE: This check requires an additional API call to retrieve the file handle
```
nickgros, line 8: update doc to match new behavior

packages/synapse-react-client/src/components/SynapseTable/SynapseTableRenderers.tsx
```diff
@@ -179,10 +180,24 @@ const getEntityOrRowVersion = (
 }
 
 function AccessCell(props: CellContext<Row, unknown>) {
+  const { showExternalAccessIcon } = useSynapseTableContext()
+  const { ref, inView } = useInView({ triggerOnce: true })
   const entityId = getEntityOrRowId(props)!
+
+  // If showExternalAccessIcon is true, defer rendering until the cell is in view to avoid unnecessary API calls
```
nickgros, line 187:
```suggestion
  // If showExternalAccessIcon is true, defer rendering until the cell is in view to limit simultaneous API calls
```

apps/portals/ampals/src/config/synapseConfigs/data.ts
nickgros: Works for me!

packages/synapse-react-client/src/components/HasAccess/HasAccessV2.tsx
kianamcc: The design didn't give me a tooltipText

nickgros (approved):

PR #1760, opened 2025-04-16
SWC-6895 - Use ROR for suggestions for UserProfile.company

nickgros:
image

image


packages/synapse-react-client/src/components/RORInstitutionField/RORInstitutionField.tsx
```diff
@@ -0,0 +1,104 @@
...
+
+export type RORLinkedInstitutionFieldProps = {
+  value?: string
+  onChange?: (value: string | undefined, rorIdentifier?: string) => void
+}
+
+function getOrganizationDisplayName(organization: ROROrganization) {
+  return organization.names.filter(name =>
+    name.types.includes('ror_display'),
+  )[0]
```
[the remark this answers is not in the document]
nickgros: `find` is more concise, but a fallback doesn't make sense here -- let's just assert that this is non-null

```suggestion
  return organization.names.find(name =>
    name.types.includes('ror_display'),
  )!
```

packages/synapse-react-client/src/testutils/ReactQueryMockUtils.ts
```diff
@@ -70,7 +70,7 @@ export function getUseQueryLoadingMock<TData>(): QueryObserverLoadingResult<
 
 export function getUseQueryErrorMock<TError>(
   error: TError,
-): UseQueryResult<null, TError> {
+): UseQueryResult<never, TError> {
```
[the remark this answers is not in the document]
nickgros: The generic type is the type of `data` -- since this is mock data for when `useQuery` returns an error, we don't care about the type of `data`. `never` is appropriate.

apps/SageAccountWeb/src/components/ProfileValidation/ProfileFieldsEditor.tsx
```diff
@@ -83,14 +92,9 @@ export const ProfileFieldsEditor = (props: ProfileFieldsEditorProps) => {
...
-          <TextField
-            fullWidth
-            label={'Current Affiliation'}
-            id="company"
-            name="company"
-            onChange={handleChange}
+          <RORInstitutionField
+            onChange={value => handleCompanyChange(value || '')}
             value={values.company || ''}
-            error={!!errors.company}
```
jay-hodgson, line 93: Do we need the RORInstitutionField to handle an error?

packages/synapse-react-client/src/components/RORInstitutionField/RORInstitutionField.test.tsx
```diff
@@ -0,0 +1,129 @@
...
+
+  it('shows search results which can be selected', async () => {
+    const mockOnChange = jest.fn()
+
+    render(<RORInstitutionField onChange={mockOnChange} />)
+
+    const input = screen.getByRole('combobox')
+    await userEvent.type(input, 'sage bionetworks')
+
+    const option = await screen.findByRole('option', {
```
jay-hodgson, line 99: This is a question for my own understanding.  Does the debounced effect impact the test time, or is the 500ms debounce ignored during testing?

packages/synapse-react-client/src/components/RORInstitutionField/RORInstitutionField.tsx
```diff
@@ -0,0 +1,104 @@
...
+ * ROR API.
+ * @param props
+ * @constructor
+ */
+export default function RORInstitutionField(
+  props: RORLinkedInstitutionFieldProps,
+) {
+  const { value: valueFromProps, onChange = noop } = props
+
+  const isControlled = valueFromProps !== undefined
```
jay-hodgson, line 30: Are there cases where this component is being used independently, not controlled (by the parent defining the value prop)?

packages/synapse-react-client/src/components/RORInstitutionField/RORInstitutionField.tsx
```diff
@@ -0,0 +1,104 @@
...
+              <Typography variant={'body1'}>{displayName}</Typography>
+              <Typography
+                variant={'smallText1'}
+                sx={{
+                  color: 'grey.700',
+                  fontStyle: 'italic',
+                  fontSize: '11px',
+                }}
+              >
+                {otherNames.map(name => name.value).join(', ')}
```
jay-hodgson, line 94: Had to go back to your screenshots to see how these otherNames are output, and it looks great!

packages/synapse-react-client/src/ror-client/index.ts
```diff
@@ -0,0 +1,16 @@
+import { fetchWithExponentialTimeout } from '@sage-bionetworks/synapse-client/util/fetchWithExponentialTimeout'
+import { ROROrganization, RORSearchResult } from './types/ROROrganization'
+
+export async function searchRegistry(query: string): Promise<RORSearchResult> {
+  return fetchWithExponentialTimeout(
+    `https://api.ror.org/v2/organizations?query=${encodeURIComponent(query)}`,
```
jay-hodgson, line 6: This is cool.  A call to an external service, with exponential timeout, and the response is strongly typed.  👍

packages/synapse-react-client/src/synapse-queries/ror/index.ts
```diff
@@ -0,0 +1,46 @@
...
+  UseQueryOptions,
+} from '@tanstack/react-query'
+import { SynapseClientError } from '../../utils'
+import {
+  ROROrganization,
+  RORSearchResult,
+} from '../../ror-client/types/ROROrganization'
+import { getOrganization, searchRegistry } from '../../ror-client'
+
+export function useGetOrganization(
```
jay-hodgson, line 13: And we're utilizing the react-query for caching for these calls.  Again, 👍

jay-hodgson (approved):

apps/SageAccountWeb/src/components/ProfileValidation/ProfileFieldsEditor.tsx
nickgros: I think in this specific case, it doesn't actually matter.
1. The validation runs after you submit, and makes sure that the `company` field is not empty
2. We do not let you submit unless the `company` field is not empty.

However I'll add it back just in case this validation changes.

packages/synapse-react-client/src/components/RORInstitutionField/RORInstitutionField.test.tsx
nickgros: It does, I should have done `jest.useFakeTimers`.

PR #1765, opened 2025-04-18
PORTALS-3554

jay-hodgson:


packages/synapse-react-client/src/components/ChallengeSubmission/SubmissionDirectoryList.tsx
```diff
@@ -458,10 +458,7 @@ function SubmissionDirectoryList({
             To learn more about how to create and submit the Docker containers
             using command line, see our{' '}
             <Link
-              to={{
```
jay-hodgson, line 461: Since this is an external link, we don’t need to use an object at all - just pass the full URL as a string.

nickgros (approved):

PR #1766, opened 2025-04-18
PORTALS-3547: suggested text edit from Jineta

jay-hodgson:


jay-hodgson:
See https://sagebionetworks.jira.com/browse/PORTALS-3547?focusedCommentId=246735

PR #1770, opened 2025-04-19
Add Vitest config to synapse-react-client, migrate one test

nickgros:
Proposal to gradually migrate Vitest in synapse-react-client

As in other projects, we should adopt Vitest in synapse-react-client since it uses Vite to bundle and run tests, which more closely matches our production environment. This will let us overcome limitations in Jest that make testing certain functionality hard or impossible (for example, dynamically imported ES modules).

Since this would be a large project, we can migrate our tests gradually by renaming them to `.vitest.tsx`. I migrated one such test. In an attempt to improve test clarity (as well as fix failures), I also modified the test to use direct mocks instead of `spyOn` and usage of Mock Service Worker.


packages/synapse-react-client/src/components/Markdown/MarkdownSynapse.vitest.tsx
```diff
@@ -59,47 +54,21 @@ const renderComponent = (
...
-      },
-    ),
-  )
-}
+const spyOnMath = vi.spyOn(MarkdownSynapse.prototype, 'processMath')
 
-describe('MarkdownSynapse tests', () => {
-  beforeAll(() => server.listen())
-  afterEach(() => server.restoreHandlers())
-  afterAll(() => server.close())
```
nickgros, line 82: Dropped mock service worker in this particular test. In this case, it seemed to increase the complexity of mocks more than we benefit from behavior resembling a real server (though I don't know if I would made the same argument for certain other tests).

packages/synapse-react-client/src/components/Markdown/__snapshots__/MarkdownSynapse.vitest.tsx.snap
```diff
@@ -1,6 +1,6 @@
-// Jest Snapshot v1, https://goo.gl/fbAQLP
+// Vitest Snapshot v1, https://vitest.dev/guide/snapshot.html
```
nickgros, line 1: Promising to see that the snapshot data did not change at all

packages/synapse-react-client/package.json
```diff
@@ -227,15 +230,17 @@
...
   },
   "scripts": {
     "clean": "rimraf dist storybook-static coverage *.tsbuildinfo",
     "prepublishOnly": "pnpm install && pnpm nx run synapse-react-client:build",
     "start": "storybook dev -p 6060",
-    "test": "jest",
-    "test:coverage": "jest --coverage",
+    "test": "pnpm run \"/^test:.*/\"",
+    "test:vitest": "vitest",
+    "test:jest": "jest",
```
nickgros, line 243: Running both Vitest and Jest separately may bust our total coverage metrics, but it lets us gradually migrate

packages/synapse-react-client/src/mocks/MockSynapseContext.tsx
```diff
@@ -25,12 +25,10 @@ export const MOCK_CONTEXT = createContext(MOCK_CONTEXT_VALUE)
  *
  * If using @testing-library/react, see {@link TestingLibraryUtils#createWrapper}
  */
-export const SynapseTestContext = jest
```
nickgros, line 28: This may be used in Jest or Vitest, so we cannot make it a Jest Mocked Function. That's ok, it doesn't need to be a mock at all.

packages/synapse-react-client/src/testutils/vitest.setup.ts
```diff
@@ -0,0 +1,2 @@
+// import jest-dom for Testing Library matchers (compatible with Vitest)
+import '@testing-library/jest-dom/vitest'
```
nickgros, line 2: Provides useful matchers, [e.g. `.toBeInTheDocument()`](https://github.com/testing-library/jest-dom). They provide a Vitest implementation, so use it here.

packages/synapse-react-client/src/components/Markdown/MarkdownSynapse.vitest.tsx
```diff
@@ -59,47 +54,21 @@ const renderComponent = (
...
-
-  const mockGetWikiAttachments = jest.spyOn(
-    SynapseClient,
-    'getWikiAttachmentsFromEntity',
-  )
+mockSynapseClient.getWikiAttachmentsFromEntity.mockResolvedValue({ list: [] })
 
+describe('MarkdownSynapse tests', () => {
```
nickgros, line 61: One of the reasons I picked this test is because (in a future PR) I want to dynamically import the MarkdownWidget component implementations to break a dependency cycle and enable code-splitting in an experiment I am working on. Vite lets you do it, but [Jest's support is experimental](https://jestjs.io/docs/next/ecmascript-modules) and is not trivial to configure. We need to move to Vitest anyways, so this is a good place to start.

packages/synapse-react-client/vite.config.ts
```diff
@@ -20,6 +21,11 @@ const config = new ConfigBuilder()
       // Do not clean the output directory before building, since we build ESM/CJS and UMD separately.
       emptyOutDir: false,
     },
+    test: {
+      // require vitest tests to be .vitest.ts(x) while we migrate from Jest
```
jay-hodgson, line 25: Great approach to moving to vitest over time, and ambitious to start with the old MarkdownSynapse component tests!

jay-hodgson (approved):

packages/synapse-react-client/src/components/Markdown/MarkdownSynapse.vitest.tsx
jay-hodgson: Like to hear more about the identified dependency cycle that is blocking code-splitting.  Nice work

nickgros: One example is
MarkdownSynapse -> MarkdownWidget -> QueryWrapperPlotNav -> ... -> MarkdownSynapse

I think there are a few places this can happen, such as in cards

PR #1771, opened 2025-04-21
SWC-7289

nickgros:
Fix bug where schemas with conditional logic based on entity properties did not work because we did not pass the entity properties to RJSF.

 To fix this, we can pass the entity data to RJSF and hide the fields.


packages/synapse-react-client/package.json
```diff
@@ -69,6 +69,7 @@
     "immutable": "4.1.0",
     "jotai": "^2.7.0",
     "json-rules-engine": "^4.1.0",
+    "jsonpath": "^1.1.1",
```
nickgros, line 72: We need to get all of the possible `properties` out of a JSON schema. The easiest way to do this is with JSONPath. The alternative is writing an algorithm to recursively walk through a JSON object.

packages/synapse-react-client/src/components/SchemaDrivenAnnotationEditor/AnnotationEditorUtils.ts
```diff
@@ -1,9 +1,66 @@
...
+    // Spread the validation schema to ensure definitions are included
+    // The definitions get lost if we simply put the entire schema in `allOf`
+    ...validationSchema,
+    allOf: [
+      ...(validationSchema.allOf || []),
+      // Add in the entity properties
+      {
+        type: 'object',
+        properties: entitySchemaBaseProperties,
+      },
```
nickgros, line 34: Adding this to the schema allows us to pass the entity properties into RJSF

packages/synapse-react-client/src/components/SchemaDrivenAnnotationEditor/AnnotationEditorUtils.ts
```diff
@@ -1,9 +1,66 @@
...
+ * Generates the UiSchema for the annotations editor.
+ * @param entitySchemaBaseProperties
+ */
+export function getUiSchemaForForm(
+  entitySchemaBaseProperties: JSONSchema7['properties'] = {},
+) {
+  // Each 'base' entity field (id, name, concreteType, etc.) should be hidden
+  const uiSchemaWithHiddenEntityProperties = Object.keys(
+    entitySchemaBaseProperties,
+  ).reduce((prev, curr) => ({ ...prev, [curr]: { 'ui:widget': 'hidden' } }), {})
```
nickgros, line 51: Now that the entity fields are part of the schema/formData, we have to hide them. Easy to do with the UiSchema.

packages/synapse-react-client/src/components/SchemaDrivenAnnotationEditor/SchemaDrivenAnnotationEditor.stories.ts
```diff
@@ -124,3 +124,39 @@ export const DirectlyProvidedSchema: Story = {
...
+          },
+        },
+
+        required: ['fileType'],
+      },
+    },
+
+    entityId: 'syn55166058',
+  },
+}
```
nickgros, line 162:
```suggestion
```

packages/synapse-react-client/src/components/SchemaDrivenAnnotationEditor/field/AdditionalPropertiesSchemaField.tsx
```diff
@@ -206,11 +206,16 @@ export function AdditionalPropertiesSchemaField<
...
+    // See https://github.com/rjsf-team/react-jsonschema-form/issues/3367
+    if (!Array.isArray(formData)) {
+      onChange(convertToArray(formData) as unknown as T)
+    }
+  }, [formData, onChange])
+
   if (!Array.isArray(formData)) {
-    onChange(convertToArray(formData) as unknown as T)
     return <></>
   }
```
nickgros, line 220: Do not call `onChange` in the `render` -- call it in `useEffect`, and use the same condition to render nothing as needed.

packages/synapse-react-client/src/components/SchemaDrivenAnnotationEditor/field/SynapseAnnotationsRJSFObjectField.tsx
```diff
@@ -30,49 +33,48 @@ export function SynapseAnnotationsRJSFObjectField<
    * @param props
    */
   useEffect(() => {
-    const { schema, formData, onChange } = props
     const newFormData: Record<string, any> = { ...formData }
-    if (schema[PROPERTIES_KEY]) {
-      Object.entries(schema[PROPERTIES_KEY]).forEach(
-        ([key, propertySchema]) => {
-          const data = newFormData[key]
+    const allProperties = getAllPropertiesInFlatObjectSchema(schema)
```
nickgros, line 37: Only changed to use new helper method which will find properties that are not in `schema.properties` (may be in `schema.definitions`, referenced by `allOf`

packages/synapse-react-client/src/mocks/msw/handlers/schemaHandlers.ts
```diff
@@ -0,0 +1,53 @@
...
+import { JSONSchema7 } from 'json-schema'
+
+const validationSchemas: JSONSchema7[] = [
+  mockValidationSchema,
+  mockConditionalOnConcreteTypeSchema,
+  mockFileEntityValidationSchema,
+  mockProjectValidationSchema,
+]
+
+export function getValidationSchemaHandlers(backendOrigin?: string) {
```
nickgros, line 22: Add a basic handler so we can register more than one schema. This is a required change as our component now gets 2 validation schemas -- one for the bound schema and one for the base entity schema

packages/synapse-react-client/src/components/SchemaDrivenAnnotationEditor/SchemaDrivenAnnotationsEditor.test.tsx
```diff
@@ -640,38 +607,6 @@ describe('SchemaDrivenAnnotationEditor tests', () => {
     )
   })
 
-  it('Disallows keys that collide with the Entity JSON definition and throws a custom error message', async () => {
```
nickgros, line 643: This test is no longer needed

jay-hodgson (approved):

PR #1776, opened 2025-04-21
PORTALS-3534

jay-hodgson:
- Change SearchParamAwareStandaloneQueryWrapper to pass in a modified Query instead of relying on the search parameters (SqlFunctions.getAdditionalFilters())
- Add an optional FTSConfig to the searchConfiguration, where you set the FTS search mode, the distance (for Boolean mode only), and optional FTS searchHelpURL


nickgros (approved):
How does the BOOLEAN MODE filter appear in the `SelectionCriteriaPills` component?

jay-hodgson:
> How does the BOOLEAN MODE filter appear in the `SelectionCriteriaPills` component?

Poorly.  It's on my list to fix.  The other thing on the list is to lock this down with tests

nickgros (approved):

apps/synapse-portal-framework/src/components/cancercomplexity/CancerComplexityHeader.tsx
```diff
@@ -1,8 +1,5 @@
 import { Box, Typography, Link } from '@mui/material'
 import HeaderSearchBox from '../HeaderSearchBox'
-import Header from '../Header'
```
jay-hodgson, line 3: Also release the new CCKP header from experimental mode in this release (per Jineta)

nickgros (approved):

PR #1783, opened 2025-04-23
PORTALS-3346 - Portal DOI redirector

nickgros:
- Generic route to redirect a DOI to a resource
- Add stub for ADKP


apps/portals/adknowledgeportal/src/config/doiRedirectConfig.ts
```diff
@@ -0,0 +1,17 @@
...
+) => {
+  switch (resourceType) {
+    case 'STUDY':
+      return `/Explore/Studies/DetailsPage/StudyDetails?Study=${resourceId}`
+    default:
+      throw new Error(
+        `Resource type ${resourceType} not supported in this portal.`,
+      )
+  }
+}
```
nickgros, line 17: Each portal must define a mapping from `(resourceType, resourceId)  -> path`. I started one for the ADKP here.

apps/synapse-portal-framework/src/shared-config/DoiRedirectRoute.tsx
```diff
@@ -0,0 +1,55 @@
...
+
+  if (!doiId) {
+    console.error(
+      'Could not redirect to DOI: no id provided in the URLSearchParams.',
+    )
+    return <Navigate to={'/'} replace={true} />
+  }
+
+  // Only split on the first `.` -- the ID may contain `.` itself
+  const parts = doiId.split('.', 2)
```
nickgros, line 30: Are we OK with this ID format? I think with the new way synapse handles DOIs, this could be changed in the future (I think the Portal DOI URI is an opaque string), but it might be a huge pain to migrate all of them.

apps/portals/adknowledgeportal/src/config/doiRedirectConfig.ts
nickgros: I need to reach out to the ADKP stakeholders to make sure `Study` is the correct column to use for a permanent identifier, so the implementation is subject to change.

apps/portals/adknowledgeportal/src/config/doiRedirectConfig.ts
```diff
@@ -0,0 +1,17 @@
+import { DoiRedirectConfig } from '@sage-bionetworks/synapse-portal-framework/shared-config/DoiRedirectRoute'
+
+type ADKPResourceType = 'STUDY'
+
+export const doiRedirectConfig: DoiRedirectConfig<ADKPResourceType> = (
```
jay-hodgson, line 5: Typed for a single resource only (ADKPResourceType = 'STUDY')?

apps/synapse-portal-framework/src/shared-config/DoiRedirectRoute.tsx
jay-hodgson: If we need to add another piece of information to this ID, how easy would it be?
I think on one extreme, we could use json.  On the other, a simple dot (or colon) separator for the tokens.
Should we lean into the URL encoded nature of this ID?  That is, what if the ID was something like "type=STUDY&id=syn123"?

apps/portals/adknowledgeportal/src/config/routesConfig.tsx
```diff
@@ -40,6 +42,7 @@ const routes: RouteObject[] = [
     ),
     children: [
       ...sharedRoutes,
+      getDoiRedirectRoute(doiRedirectConfig),
```
jay-hodgson, line 45: I see, so we would have a new route for every resource type (and a new doiRedirectConfig)?

apps/synapse-portal-framework/src/shared-config/DoiRedirectRoute.tsx
nickgros: Another consideration is that our chosen format has to consistently serialize to equal strings. So if we use JSON/URLSearchParams, we have to serialize the keys in a particular order.

nickgros: I think if we let Portal owners mint a DOI for a resource, then we must have a contract that the primary key for that resource can no longer change.

apps/portals/adknowledgeportal/src/config/routesConfig.tsx
jay-hodgson: What if a single portal wants to support DOIs for more than one resource, say Study and Dataset?

nickgros: the resource is encoded into the ID. So the backend will send a request to

`https://theportal.synapse.org/doi?id=STUDY.syn123`
`https://theportal.synapse.org/doi?id=DATASET.2d4c899`

(we discussed elsewhere that this format of IDs is probably not appropriate, but it illustrates the idea)

apps/portals/adknowledgeportal/src/config/doiRedirector.ts
```diff
@@ -0,0 +1,48 @@
+import { PortalResourceRedirector } from '@sage-bionetworks/synapse-portal-framework/shared-config/DoiRedirectRoute'
+import PortalsDoiIdSerializer from '@sage-bionetworks/synapse-portal-framework/utils/PortalsDoiIdSerializer'
+
+/**
+ * Configuration for mapping a DOI to a portal resource URL.
+ *
+ * To add a new resource type:
+ *  1. Add the resource type to the string literal
+ *  2. Add a case to the switch statement in the doiRedirectConfig function.
```
nickgros, line 9:
```suggestion
 *  2. Define the ordered key set that defines this resource in `RESOURCE_TYPE_KEY_CONFIGURATION`
 *  3. Add a case to the switch statement in the doiRedirectConfig function.
```

apps/synapse-portal-framework/src/utils/PortalsDoiIdSerializer.ts
```diff
@@ -0,0 +1,70 @@
...
+      [RESERVED_PORTAL_RESOURCE_TYPE_KEY]: type,
+      ...resourceKeyAttributes,
+    }
+
+    // Stringify as JSON. The 2nd parameter ensures we only stringify the keys we want, in the specified order
+    return JSON.stringify(fullObject, [
+      RESERVED_PORTAL_RESOURCE_TYPE_KEY,
+      ...this.config.resourceTypeOrderedRequiredAttributes[type],
+    ])
+  }
```
nickgros, line 54: Between JSON and URLSearchParams, I decided to use JSON since it was easier to implement, but still open to changing it, I don't feel strongly about either.

Because the ID will be passed as a URLSearchParam, our URLs will include an encoded serialized string. If we use a URLSearchParams format, we end up with a URL part that has to be URL-encoded twice. If we use JSON, we end up with a longer URL.

```json
{
  "__type": "someResourceType",
  "key1": "some&value"
}
```

| Case | as URL Search Params | as JSON | 
|--------|--------|-------|
| DOI ID (as URLSearchParams) | `__type=someResourceType&key1=some%26value` |`{"__type":"someResourceType","key1":"some&value"}`|
| Path to resolve DOI | `/doi?id=__type%3DsomeResourceType%26key1%3Dsome%2526value` | `%7B%22__type%22%3A%22someResourceType%22%2C%22key1%22%3A%22some%26value%22%7D` |

If we use JSON, we do still have to double-encode it (as JSON, then as a URLSearchParam), but

apps/synapse-portal-framework/src/utils/PortalsDoiSerializer.test.ts
```diff
@@ -0,0 +1,79 @@
+import PortalsDoiIdSerializer from '@/utils/PortalsDoiIdSerializer'
```
jay-hodgson, line 1: does this not need to be a .vitest.ts file?

apps/synapse-portal-framework/src/utils/PortalsDoiSerializer.test.ts
```diff
@@ -0,0 +1,79 @@
...
+    it('throws an error if a key is missing', () => {
+      expect(() => {
+        serializer.serialize('STUDY', { someOtherAttribute: 'syn123' })
+      }).toThrow(
+        'Missing key(s) studyId in resource attributes for STUDY. Passed attributes: {"someOtherAttribute":"syn123"}',
+      )
+    })
+
+    it('properly escapes special characters', () => {
+      const expected = '{"__type":"STUDY","studyId":"syn&123"}'
```
jay-hodgson, line 56: This represents a properly escaped special character?

jay-hodgson (approved):

apps/synapse-portal-framework/src/utils/PortalsDoiSerializer.test.ts
nickgros: We only need to do that in synapse-react-client

apps/synapse-portal-framework/src/utils/PortalsDoiSerializer.test.ts
nickgros: Well, it did when I tried URLSearchParams, but now it doesn't -- let me switch that to  a quote!

PR #1791, opened 2025-04-23
PORTALS-3534: updated boolean mode distances on CCKP

jay-hodgson:
See https://sagebionetworks.jira.com/browse/PLFM-8897?focusedCommentId=247471


kianamcc (approved):
Updated distances match Jineta's tested values!

PR #1792, opened 2025-04-24
PORTALS-3449

lakikowolfe:
This ticket contains a couple different pieces:
- Add frame to AMPALS Resources pages
- Check and document that video and drop downs are supported (this shouldn't block the ticket)

Jira ticket : https://sagebionetworks.jira.com/browse/PORTALS-3449


lakikowolfe:
https://github.com/user-attachments/assets/84fb50be-397b-4a94-a2d9-4448a00474eb

@nickgros I'm noticing that when I click to each page that the content loads at different speeds. Is this something worth looking into (or is there an easy fix for this)?

nickgros:
@lakikowolfe Depends on what's causing the delay. I think what's happening here is, when you go from `Resources > For Researchers` to `Resources > For Contributors`,  the rendered `MarkdownSynapse` node is reused by React because the tree structure is the same.

I think there is a bug in the `MarkdownSynapse` component where, when the props change, the component should immediately display a loading state, but does not.

It would be nice to fix, but it's not a big deal and not a blocker for this work.

lakikowolfe:
https://github.com/user-attachments/assets/2ea91b81-d42c-4595-8c4c-42bfa70a83d1

apps/portals/ampals/src/config/routesConfig.tsx
```diff
@@ -7,7 +7,17 @@ import { MarkdownSynapse } from 'synapse-react-client'
...
 import ExploreWrapper from '@sage-bionetworks/synapse-portal-framework/components/Explore/ExploreWrapper'
+import AMPALSResearchPageLayout from '@sage-bionetworks/synapse-portal-framework/components/ampals/AMPALSResearchPageLayout'
 import { searchPageChildRoutes } from '@/pages/AMPALSSearchPage'
+import { Typography, Link } from '@mui/material'
+//import header images for Resources pages
+import ResearcherResourcesHeaderSlat from '@sage-bionetworks/synapse-portal-framework/components/ampals/assets/researcher_resources_slat.png'
+import ContributorResourcesHeaderSlat from '@sage-bionetworks/synapse-portal-framework/components/ampals/assets/data_contributor_resources_slat.png'
+import PWLEResourcesHeaderSlat from '@sage-bionetworks/synapse-portal-framework/components/ampals/assets/pwles_resources_slat.png'
+import PublishingRequirementsSlat from '@sage-bionetworks/synapse-portal-framework/components/ampals/assets/publishing_requirements.png'
+import ContributeDataSlat from '@sage-bionetworks/synapse-portal-framework/components/ampals/assets/contribute_data_slat.png'
```
nickgros, line 18: suggestion: Since these are not React components, use camelCase
```suggestion
import researcherResourcesHeaderSlat from '@sage-bionetworks/synapse-portal-framework/components/ampals/assets/researcher_resources_slat.png'
import contributorResourcesHeaderSlat from '@sage-bionetworks/synapse-portal-framework/components/ampals/assets/data_contributor_resources_slat.png'
import pwleResourcesHeaderSlat from '@sage-bionetworks/synapse-portal-framework/components/ampals/assets/pwles_resources_slat.png'
import publishingRequirementsSlat from '@sage-bionetworks/synapse-portal-framework/components/ampals/assets/publishing_requirements.png'
import contributeDataSlat from '@sage-bionetworks/synapse-portal-framework/components/ampals/assets/contribute_data_slat.png'
```

Also, if some of these are AMP-ALS specific, consider moving those to `portals/ampals/src/assets` (might need to create this folder)

apps/synapse-portal-framework/src/components/ampals/AMPALSResearchPageLayout.tsx
```diff
@@ -0,0 +1,105 @@
...
+            flexDirection: 'column',
+            width: '360px',
+            gap: '20px',
+          }}
+        >
+          <Typography
+            sx={{
+              fontWeight: 700,
+              fontSize: '24px',
+              color: '#353A3F',
```
nickgros, line 75:
```suggestion
              color: 'grey-900',
```

apps/synapse-portal-framework/src/components/ampals/AMPALSResearchPageLayout.tsx
```diff
@@ -0,0 +1,105 @@
...
+            }}
+          >
+            {sidebarTitle}
+          </Typography>
+          {sidebarContent}
+        </Box>
+        <Box
+          sx={{
+            marginTop: '0px',
+            width: '720px',
```
nickgros, line 85: Not sure how this would look with narrow screen sizes with this hard-coded width. What's the purpose of this? If just to constrain the max width, then I suggest

```suggestion
            maxWidth: '720px',
```

but it might be appropriate instead to use a MUI Container here -- not totally sure about that though

nickgros (commented):
Could you update the base branch to be the new `ampals-staging` branch?

apps/portals/ampals/src/config/routesConfig.tsx
lakikowolfe: @nickgros how do you make the determination to keep things under `portals/` vs within `synapse-portal-framework/`? I thought I was following the norm for the repo based on what I saw when working on some [Elite specific components](https://github.com/Sage-Bionetworks/synapse-web-monorepo/tree/main/apps/synapse-portal-framework/src/components/eliteportal/assets) previously.

apps/synapse-portal-framework/src/components/ampals/AMPALSResearchPageLayout.tsx
lakikowolfe: What makes you think that a container might be more appropriate?

apps/portals/ampals/src/config/routesConfig.tsx
lakikowolfe: Thanks this is really helpful. I'll move these assets to `/portals/`

apps/synapse-portal-framework/src/components/ampals/AMPALSResearchPageLayout.tsx
lakikowolfe: By my eye the content is not centered horizontally. So I'm going to move forward with `maxWidth`.

Screenshot 2025-05-02 at 9 17 15 AM

nickgros (approved):

PR #1794, opened 2025-04-24
PORTALS-3549 - Add DOI to portal cards

nickgros:
When DOIs have been configured for a portal resource:
* Show the DOI if it exists on the portal card
* Show a button to create a DOI if the user has permission to do so.

| Case | Screenshot |
|--------|--------|
| Details page, viewer | image |
| Details page, edit | image |
| Details page, create | image | 
| Explore page, viewer | image |

### DIFF
--- a/packages/synapse-react-client/src/components/GenericCard/CardUtils.tsx
+++ b/packages/synapse-react-client/src/components/GenericCard/CardUtils.tsx
@@ -1,0 +1,169 @@
+import { CardLink } from '@/components/CardContainer/CardLink'
+import { isTableEntity } from '@/utils/functions/EntityTypeUtils'
+import { PRODUCTION_ENDPOINT_CONFIG } from '@/utils/functions/getEndpoint'
+import {
+  convertDoiToLink,
+  SYNAPSE_ENTITY_ID_REGEX,
+} from '@/utils/functions/RegularExpressions'
+import {
+  ColumnModel,
+  ColumnType,
+  Entity,
+  FileHandleAssociateType,
+  FileHandleAssociation,
+  SelectColumn,
+} from '@sage-bionetworks/synapse-types'
+import { TargetEnum } from '../CardContainerLogic/CardContainerLogic'
+
+export const getFileHandleAssociation = (
+  table?: Entity,
+  fileHandleId?: string,
+  rowSynapseId?: string, // only applicable if this is an EntityView
+) => {
+  let fileHandleAssociation: FileHandleAssociation | undefined = undefined
+  if (table && fileHandleId) {
+    if (isTableEntity(table)) {
+      // The file handle is in the table
+      fileHandleAssociation = {
+        fileHandleId,
+        associateObjectId: table?.id ?? '',
+        associateObjectType: FileHandleAssociateType.TableEntity,
+      }
+    } else if (rowSynapseId) {
+      // We're looking at a view, so the FileEntity (whose ID matches the row ID) gives permission to download the file handle
+      fileHandleAssociation = {
+        fileHandleId,
+        associateObjectId: rowSynapseId,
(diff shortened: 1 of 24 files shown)


packages/synapse-react-client/src/components/GenericCard/CardUtils.tsx
```diff
@@ -0,0 +1,169 @@
...
+  ColumnModel,
+  ColumnType,
+  Entity,
+  FileHandleAssociateType,
+  FileHandleAssociation,
+  SelectColumn,
+} from '@sage-bionetworks/synapse-types'
+import { TargetEnum } from '../CardContainerLogic/CardContainerLogic'
+
+export const getFileHandleAssociation = (
```
nickgros, line 18: moved some of these function out of the `TableRowGenericCard` file to improve HMR

packages/synapse-react-client/src/components/GenericCard/TableRowGenericCard.tsx
```diff
@@ -195,89 +139,6 @@ export function VersionLabel(props: { synapseId: string; version: string }) {
   )
 }
 
-type ValueOrMultiValue = {
```
nickgros, line 198: Moved without changes

packages/synapse-react-client/src/components/GenericCard/TableRowGenericCard.tsx
```diff
@@ -106,80 +122,8 @@ export type TableRowGenericCardProps = {
...
-        e,
-      )
-    }
-  }
-  return { str: value, columnModelType: selectedColumnOrUndefined?.columnType }
-}
-
 // SWC-6115: special rendering of the version column (for Views)
-export function VersionLabel(props: { synapseId: string; version: string }) {
+function VersionLabel(props: { synapseId: string; version: string }) {
```
nickgros, line 126: not used outside of this component, so remove export to improve HMR

packages/synapse-react-client/src/testutils/ReactQueryMockUtils.ts
```diff
@@ -27,7 +28,7 @@ export function getUseQuerySuccessMock<TData>(
     isRefetching: false,
     isStale: false,
     isSuccess: true,
-    refetch: jest.fn(),
+    refetch: noop as any,
```
nickgros, line 31: Cannot use these helpers that contain `jest.fn()` in a Vitest module. Might need to find a way to pass `jest` or `vitest` as a parameter, but for now, just use Lodash noop

packages/synapse-react-client/src/components/GenericCard/PortalDOI/PortalDOI.tsx
```diff
@@ -0,0 +1,75 @@
...
+}
+
+/**
+ * Component used to view/create/edit a DOI for a portal resource.
+ */
+function PortalDOI(props: PortalDOIProps) {
+  const [showEditModal, setShowEditModal] = useState(false)
+  const { portalId, resourceId } = props
+
+  const { data: canMintDoi } = useGetUserPortalPermissions(portalId, {
```
jay-hodgson, line 25: Since this is shown for every card on the page, would it make sense to use react-intersection-observer (to only load when it is close to the viewport)?

jay-hodgson (approved):

packages/synapse-react-client/src/components/GenericCard/PortalDOI/PortalDOI.tsx
nickgros: Great idea!

nickgros: I think I will add the check to the logic that adds the label rather than this one

PR #1802, opened 2025-04-25
PORTALS-3567

jay-hodgson:


packages/synapse-react-client/src/components/SynapseHomepageV2/SynapseHotDropsBackground.tsx
```diff
@@ -0,0 +1,157 @@
+import { Box } from '@mui/material'
+
+const SynapseHotdropsBackground = () => {
```
nickgros, line 3: unintentionally committed?

packages/synapse-react-client/src/components/SynapseHomepageV2/SynapseHotDropsBackground.tsx
```diff
@@ -0,0 +1,157 @@
...
+          '0%, 100%': { transform: 'translateY(0px) scale(1)' },
+          '50%': { transform: 'translateY(13px) scale(1.05)' },
+        },
+        '@keyframes float5': {
+          '0%, 100%': { transform: 'translateY(0px) scale(1)' },
+          '50%': { transform: 'translateY(15px) scale(1.06)' },
+        },
+      }}
+    >
+      <svg
```
nickgros, line 41: if this SVG isn't using any JavaScript, I'd prefer it is put in a different file that can be previewed by IDEs (and might handled differently by a bundler)

jay-hodgson:
Test failure is unrelated:
DownloadListTable tests › Copy all Synapse IDs › should call clipboard.writeText with the expected Synapse IDs

PR #1803, opened 2025-04-25
Decouple card UI for case where we don't want to use a table

nickgros:
- Refactor CardDeck components to separate table query logic from UI logic
- Add simple example to b2ai for hard-coded cards

Partially completed port: 
image


apps/portals/b2ai.standards/src/components/ChallengesCardDeck.tsx
```diff
@@ -0,0 +1,63 @@
...
+  },
+  {
+    title: 'Clinical Care Challenge',
+    description:
+      'Establishing the most diverse, high-resolution, ethically sourced, AI-ready data set to explore questions around improving recovery from acute illness',
+    ctaButtonText: 'Explore Standards',
+    ctaButtonURL: createExplorePageLink(CLINICAL_CARE_CHALLENGE_FILTERED_QUERY),
+    headerImage: <img src={healthSvg} />,
+  },
+]
```
nickgros, line 57: @Sigfried I did 2/4 cards here, does it make sense how you would construct the other 2?

packages/synapse-react-client/src/components/CardDeck/CardDeck.tsx
```diff
@@ -1,88 +1,20 @@
...
-  titleIconFileHandleAssociation?: FileHandleAssociation /* title icon file handle association */
-  headerImageFileHandleAssociation?: FileHandleAssociation /* header image file handle association */
+  cards: CardDeckCardProps[]
   cardDeckType?: 'cckp' | 'b2ai' /* The type of card deck (cckp or b2ai)*/
 }
 
+/**
+ * UI for a 'deck' of cards.
+ */
 export function CardDeck(props: CardDeckProps) {
```
nickgros, line 15: Removed table query translation logic, moved to `TableQueryCardDeck`

packages/synapse-react-client/src/components/CardDeck/CardDeckCardProps.ts
```diff
@@ -0,0 +1,11 @@
+import { ReactNode } from 'react'
+
+export type CardDeckCardProps = {
```
nickgros, line 3: Extracted props for a CardDeckCard

packages/synapse-react-client/src/components/CardDeck/CardDeckCardProps.ts
```diff
@@ -0,0 +1,11 @@
...
+  title: string /* title of card */
+  description: string /* description of card */
+  ctaButtonText: string /* call to action button text */
+  ctaButtonURL: string /* call to action button URL */
+  titleIcon?: ReactNode /* icon to be displayed on the card */
+  headerImage?: ReactNode /* header image to be displayed on the card */
```
nickgros, line 9: Only real change was to change this from a fileHandleAssociation to a ReactNode (i.e. directly pass the image instead of a Synapse-specific pointer to the image)

packages/synapse-react-client/src/components/CardDeck/TableQueryCardDeck.tsx
```diff
@@ -0,0 +1,137 @@
...
+  titleIconFileHandleColumnName?: string /* The column name for the title icon image file */
+  headerImageFileHandleColumnName?: string /* The column name for the header image file */
+  cardDeckType?: 'cckp' | 'b2ai' /* The type of card deck (cckp or b2ai)*/
+  linkConfig: CardLink /* The link configuration */
+}
+
+/**
+ * Transforms a Synapse table into a card deck.
+ */
+export function TableQueryCardDeck(props: TableQueryCardDeckProps) {
```
nickgros, line 27: Identical logic/props to the old `CardDeck`. In CCKP (and other portals where we want a Table to populate cards), we will use this

packages/synapse-react-client/src/components/CardDeck/TableQueryCardDeck.tsx
```diff
@@ -0,0 +1,137 @@
...
+        entity.data,
+        headerImageFileHandleIdValue,
+        `syn${el.rowId}`,
+      )
+
+      const headerImage = headerImageFileHandleAssociation ? (
+        <ImageFileHandle
+          fileHandleAssociation={headerImageFileHandleAssociation}
+        />
+      ) : undefined
```
nickgros, line 120: Create the `ImageFileHandle` elements here so we can just pass them into the cards as arbitrary `ReactNode`s

apps/portals/b2ai.standards/src/components/ChallengesCardDeck.tsx
```diff
@@ -0,0 +1,66 @@
+import healthSvg from '@/assets/health.svg?url'
+import soundSvg from '@/assets/sound.svg?url'
```
lakikowolfe, line 2: @nickgros what is `?url` doing here?

packages/synapse-react-client/src/components/CardDeck/CardDeck.Desktop.tsx
lakikowolfe, line 1: this looks so much cleaner 🤩

lakikowolfe (approved):
Looks good!

apps/portals/b2ai.standards/src/components/ChallengesCardDeck.tsx
nickgros: It's an ["Explicit URL Import"](https://vite.dev/guide/assets#explicit-url-imports) in Vite, but looking at this, I'm not sure if it's strictly necessary how I used it.

apps/portals/b2ai.standards/src/pages/HomePage.tsx
```diff
@@ -18,7 +24,8 @@ export default function HomePage() {
           Standards Related to Bridge2AI Challenges
         </h2>
       </div>
-      <CardDeck
+      <ChallengesCardDeck />
```
jay-hodgson, line 27: Looks like the ChallengesCardDeck and TableQueryCardDeck produce duplicate output (ChallengesCardDeck is currently a subset, but can be extended).  Is that correct?  And if so, should this determination be behind a feature flag, or ChallengesCardDeck be commented out?

jay-hodgson (approved):

apps/portals/b2ai.standards/src/components/ChallengesCardDeck.tsx
Sigfried: Oh, but this has the facet name hard coded in the code rather than in the Synapse table, which is a definite improvement. But once I get the DataSet stuff imported to Synapse, I'll want to generate these cards from there. 

So I'd want to do something like

```javascript
const queryTemplate = {
  sql: dataSql,
  limit: 25,
  selectedFacets: [
    {
      concreteType:
        'org.sagebionetworks.repo.model.table.FacetColumnValuesRequest',
      columnName: DST_TABLE_COLUMN_NAMES.RELEVANT_ORG_NAMES,
    },
  ],
}

// somehow, but I don't know how I would run sql queries like this:
const gcOrgIds = `SELECT produced_by FROM DataSet` // that will be the org ids 
const orgIdList = ... // quoted, comma delimited list
const gcOrgNames = `SELECT name FROM Organization WHERE id in (${orgIdList})`


const queries = gcOrgNames.map(name => {
   let t = deepCopy(queryTemplate)
   t.selectedFacets[0].facetValues = [name]
   return t
```

Sigfried: Can I do something like this?

```typescript
const dataSetEntity = useGetEntity(datasetEntityId)
const orgEntity = useGetEntity(orgEntityId)
```

Does that give me the data from those tables?

Sigfried (commented):
This is great, @nickgros!!

Hopefully my questions make sense.

apps/portals/b2ai.standards/src/components/ChallengesCardDeck.tsx
nickgros: We have a hook [`useGetQueryResultBundle`](https://github.com/Sage-Bionetworks/synapse-web-monorepo/blob/9ef4ebe59df6aaad67aaf802ddcc068962a4d30b/packages/synapse-react-client/src/synapse-queries/entity/useGetQueryResultBundle.ts#L40) that you can use to get the data (don't mind that it's deprecated, it's still used in most of our portal components like this one).

apps/portals/b2ai.standards/src/pages/HomePage.tsx
nickgros: This was an example, I'll remove the new one to not 'break' staging, but @Sigfried can add it back when they get it working how they want.

PR #1804, opened 2025-04-25
PORTALS-3567

jay-hodgson:
Updated logic to calculate distance based on splitting search string using non-word characters (and filtering out empty strings).  This works for real-world cases, such as "Gene expression data, somatic mutation calls (VCF files from Mutect2 plus Platypus), copy-number calls (Sequenza and QDNAseq), the fraction of mutated microsatellites (MSIsensor), ATAC-seq insertion counts and allele counts of somatic SNVs in all sample types."

### DIFF
--- a/packages/synapse-react-client/src/components/FullTextSearch/FullTextSearchUtils.ts
+++ b/packages/synapse-react-client/src/components/FullTextSearch/FullTextSearchUtils.ts
@@ -41,7 +41,7 @@ export function getTextMatchesQueryFilter(
   let searchExpression = searchText
   if (textMatchesMode == 'BOOLEAN') {
     const searchTextWordLength = searchText.trim().split(/\s+/).length
-    const distanceToUse = Math.max(distance, searchTextWordLength)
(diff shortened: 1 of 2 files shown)


packages/synapse-react-client/src/components/FullTextSearch/FullTextSearchUtils.ts
```diff
@@ -40,7 +40,11 @@ export function getTextMatchesQueryFilter(
   const { textMatchesMode, distance = 0 } = ftsConfig
   let searchExpression = searchText
   if (textMatchesMode == 'BOOLEAN') {
-    const searchTextWordLength = searchText.trim().split(/\s+/).length
+    //split by non-word character and ignore empty strings
+    const searchTextWordLength = searchText
+      .trim()
+      .split(/\W+/)
+      .filter(word => word.length > 0).length
```
nickgros, line 47: Probably reuse the `getWordCount` in `TextFieldWithWordLimit.tsx` (maybe move that to a generic utils module)

nickgros (approved):

PR #1811, opened 2025-04-26
PORTALS-3571

jay-hodgson:


packages/synapse-react-client/src/utils/functions/StringUtils.ts
```diff
@@ -75,3 +75,25 @@ export function validatePassword(newPassword: string) {
   }
   return undefined
 }
+
+export function copyStringToClipboard(
```
jay-hodgson, line 79: Is this where this util function should live?

packages/synapse-react-client/src/components/EntityFinder/details/view/table/FileEntityMD5Cell.tsx
```diff
@@ -14,6 +14,7 @@ export function FileEntityMD5Cell(
     row.original.entityId,
     row.original.versionNumber,
   )
+  // TODO: Use copyStringToClipboard?  Is this mutation necessary?
   const { mutate: copyMd5ToClipboard } = useMutation({
     mutationFn: (md5: string) => navigator.clipboard.writeText(md5),
     onSuccess: () => displayToast('MD5 copied to the clipboard', 'success'),
```
jay-hodgson, line 20: Mutation needed here?  Can we just replace with a call to copyStringToClipboard()?

packages/synapse-react-client/src/utils/functions/StringUtils.ts
nickgros: in general we could use a better system for organizing utils

nickgros: If you did this, would you have to change the API?

```ts
export function copyStringToClipboard(value: string): Promise<void> {
  return new Promise((resolve, reject) => {
    // PORTALS-3571: setTimeout necessary on Safari
    setTimeout(() => {
      // use the Clipboard API
      // https://caniuse.com/mdn-api_clipboard_writetext
      navigator.clipboard.writeText(value).then(resolve).catch(reject)
    }, 0)
  })
}
```

packages/synapse-react-client/src/components/SynapseTable/TopLevelControls/TopLevelControls.tsx
```diff
@@ -289,11 +290,12 @@ const TopLevelControls = (props: TopLevelControlsProps) => {
...
-                setHasRecentlyCopiedToClipboard(true)
-                setTimeout(() => {
-                  setHasRecentlyCopiedToClipboard(false)
-                }, 3000)
+                copyStringToClipboard(window.location.href).then(() => {
+                  setHasRecentlyCopiedToClipboard(true)
+                  setTimeout(() => {
+                    setHasRecentlyCopiedToClipboard(false)
+                  }, 3000)
+                })
```
nickgros, line 298: thanks for fixing this case

nickgros (approved):

PR #1814, opened 2025-04-28
PORTALS-3572: set to valid utf-8 charset

jay-hodgson:
https://www.rfc-editor.org/rfc/rfc9110#field.content-type
utf8 is not a valid alias (but seems to be accepted!)

### DIFF
--- a/packages/synapse-react-client/src/synapse-client/HttpClient.ts
+++ b/packages/synapse-react-client/src/synapse-client/HttpClient.ts
@@ -26,7 +26,7 @@ export const doPost = <T>(
     headers: {
       Accept: '*/*',
       'Access-Control-Request-Headers': 'authorization',
-      'Content-Type': 'application/json; charset=utf8',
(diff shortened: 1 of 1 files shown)


packages/synapse-react-client/src/synapse-client/HttpClient.ts
```diff
@@ -26,7 +26,7 @@ export const doPost = <T>(
     headers: {
       Accept: '*/*',
       'Access-Control-Request-Headers': 'authorization',
-      'Content-Type': 'application/json; charset=utf8',
+      'Content-Type': 'application/json; charset=UTF-8',
```
jay-hodgson, line 29: Note that this is case insensitive.  See https://www.rfc-editor.org/rfc/rfc2046#section-4.1.2

> Unlike some other parameter values, the values of the charset parameter are NOT case sensitive.

nickgros (approved):

PR #1816, opened 2025-04-29
SWC-6990

nickgros:
- Lazy load pages in ADKP
- Replace imports to barrel files in ADKP, portals-framework
- Lazy load plotly.js-basic-dist because it is ~1MB
- Remove a couple of targeted barrel imports that caused problems (MarkdownSynapse, especially)
- Lazy load MarkdownSynapse widgets to avoid cyclic imports that prevented lazy loading


nickgros:
Done in #1833

PR #1823, opened 2025-04-30
SWC-7315: Fix missing 'moderator' tag on Discussion Board

kianamcc:
Jira: https://sagebionetworks.jira.com/browse/SWC-7315?focusedCommentId=248102&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-248102

Screenshot 2025-04-29 at 4 59 07 PM

### DIFF
--- a/packages/synapse-react-client/src/components/Forum/DiscussionReply.tsx
+++ b/packages/synapse-react-client/src/components/Forum/DiscussionReply.tsx
@@ -18,7 +18,7 @@ import { Box } from '@mui/material'
 
 export type DiscussionReplyProps = {
   reply: DiscussionReplyBundle
-  isModerator?: boolean
+  moderatorList?: string[]
   onClickLink?: () => void
 }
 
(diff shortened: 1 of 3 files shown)


packages/synapse-react-client/src/components/Forum/DiscussionReply.tsx
```diff
@@ -40,6 +41,7 @@ export function DiscussionReply(props: DiscussionReplyProps) {
   })
 
   const isCurrentUserAuthor = reply.createdBy == currentUserProfile?.ownerId
+  const isReplyAuthorModerator = moderatorList?.includes(reply.createdBy)
```
nickgros, line 44: Why not just call `useGetModerators` here? 

Alternatively, if you are going to pass the precomputed `moderatorList`, you could just precompute the value of `isReplyAuthorModerator`. Kind of unusual to pass some data but still have to compute it here.

packages/synapse-react-client/src/components/UserCard/UserBadge.tsx
```diff
@@ -30,6 +30,8 @@ export type UserBadgeProps = {
   showFullName?: boolean
 
   className?: string
+
+  isModerator?: boolean
```
nickgros, line 34: nit: Be a bit more explicit about what this prop does
```suggestion
  showModeratorBadge?: boolean
```

nickgros (approved):
couple of picky comments but looks good!

PR #1825, opened 2025-04-30
PORTALS-3538: Elite portal links to external site

kianamcc:
Jira: https://sagebionetworks.jira.com/browse/PORTALS-3538

Screenshot 2025-04-30 at 4 35 43 PM

Screenshot 2025-04-30 at 4 35 33 PM


packages/synapse-react-client/src/components/Markdown/widget/MarkdownButton.tsx
```diff
@@ -22,12 +23,25 @@ export default function MarkdownButton(
   }
   const buttonIsCenterAligned = alignLowerCase === 'center'
   const buttonColor = highlight ? 'secondary' : 'neutral'
+  const isExternalLink =
+    widgetParamsMapped.url?.startsWith('http://') ||
+    widgetParamsMapped.url?.startsWith('https://')
```
nickgros, line 28: We use this component in a lot of places where React Router is not used (i.e. in SWC)--this will break if we happen to render this in those cases.

Can you also check the value of [`useInRouterContext`](https://api.reactrouter.com/v7/functions/react_router.useInRouterContext.html) and only use the RouterLink if that is also true?

packages/synapse-react-client/src/components/Markdown/MarkdownSynapse.vitest.tsx
```diff
@@ -346,10 +347,15 @@ describe('MarkdownSynapse tests', () => {
...
     it('works with two inline widgets', () => {
-      const { container } = renderComponent({
+      const component = getComponent({
         markdown:
           '${buttonlink?text=sometext&url=#/Help/How%20It%20Works&highlight=true}${buttonlink?text=APPLY&url=#/Apply&highlight=true} ',
       })
+
+      const { container } = render(<MemoryRouter>{component}</MemoryRouter>, {
+        wrapper: createWrapper(),
+      })
```
nickgros, line 357: Can you keep the current test as is, and add a new one with a Router? This new test will make sense in the context of the other comment.

packages/synapse-react-client/src/components/Markdown/widget/MarkdownButton.tsx
```diff
@@ -22,12 +23,40 @@ export default function MarkdownButton(
...
+      component: RouterLink,
+    }
+  } else {
+    linkProps = {
+      href: widgetParamsMapped.url,
+      component: 'a',
+    }
+  }
+
+  console.log('is component router', useInRouterContext())
```
nickgros, line 53:
```suggestion
```

PR #1826, opened 2025-05-01
Improvements to TypeScript configuration

nickgros:
- All packages now emit to a cached folder (if not already emitting consumed JS)
- Reference all packages in top level tsconfig, use tsc --build instead of Nx to build full project.
- Removed redundant properties defined in base tsconfig

### DIFF
--- a/packages/markdown-it-container/test/cjs.cjs
+++ b/packages/markdown-it-container/test/cjs.cjs
@@ -2,7 +2,7 @@
 /* eslint-env mocha */
 
 const assert = require('node:assert')
-const fn = require('../')
+const fn = require('../index.cjs')
 
 describe('CJS', () => {
   it('require', () => {
--- a/shared/tsconfig.base.json
+++ b/shared/tsconfig.base.json
@@ -1,9 +1,8 @@
 {
   "compilerOptions": {
-    "outDir": "${configDir}/build",
-    "noEmit": true,
+    "outDir": "${configDir}/node_modules/.cache/.tsc_cache/build",
     "strict": true,
-    "composite": false,
+    "composite": true,
     "declaration": true,
     "resolveJsonModule": true,
     "esModuleInterop": true,
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -4,8 +4,31 @@
   "include": [],
   "compilerOptions": {
     "rootDir": ".",
-    "outDir": "build",
-    "allowJs": true,
-    "composite": false
-  }
+    "allowJs": true
+  },
+  "references": [
+    { "path": "./apps/portals/adknowledgeportal" },
+    { "path": "./apps/portals/ampals" },
+    { "path": "./apps/portals/arkportal" },
+    { "path": "./apps/portals/b2ai.standards" },
+    { "path": "./apps/portals/bsmn" },
+    { "path": "./apps/portals/cancercomplexity" },
+    { "path": "./apps/portals/challenges" },
+    { "path": "./apps/portals/digitalhealth" },
+    { "path": "./apps/portals/eliteportal" },
+    { "path": "./apps/portals/genie" },
+    { "path": "./apps/portals/nf" },
+    { "path": "./apps/portals/stopadportal" },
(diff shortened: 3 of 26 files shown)


packages/markdown-it-container/test/cjs.cjs
```diff
@@ -2,7 +2,7 @@
 /* eslint-env mocha */
 
 const assert = require('node:assert')
-const fn = require('../')
+const fn = require('../index.cjs')
```
nickgros, line 5: fix type error

tsconfig.json
```diff
@@ -4,8 +4,31 @@
...
+    { "path": "./apps/synapse-oauth-signin" },
+    { "path": "./apps/synapse-portal-framework" },
+    { "path": "./packages/markdown-it-container" },
+    { "path": "./packages/markdown-it-synapse" },
+    { "path": "./packages/markdown-it-synapse-table" },
+    { "path": "./packages/synapse-client" },
+    { "path": "./packages/synapse-react-client" },
+    { "path": "./packages/synapse-types" },
+    { "path": "./packages/vite-config" }
+  ]
```
nickgros, line 33: Reference all projects. When you run `tsc --build` from the top level directory, it will build everything.

You can run `tsc --build --watch` from the top level to get live type-checking on every file in the project

shared/tsconfig.base.json
```diff
@@ -1,9 +1,8 @@
 {
   "compilerOptions": {
-    "outDir": "${configDir}/build",
-    "noEmit": true,
+    "outDir": "${configDir}/node_modules/.cache/.tsc_cache/build",
```
nickgros, line 3: All projects now emit to a cache directory, which the TypeScript compiler will use (see [Project References](https://www.typescriptlang.org/docs/handbook/project-references.html)). We had previously only set this up for a subset of projects.

Our projects that do not use a bundler still emit to `./dist`.

shared/tsconfig.base.json
```diff
@@ -1,9 +1,8 @@
...
     "strict": true,
-    "composite": false,
+    "composite": true,
```
nickgros, line 5: Now set `composite` to `true` for all projects. This is required for project references to work.

shared/tsconfig.base.json
```diff
@@ -1,9 +1,8 @@
 {
   "compilerOptions": {
-    "outDir": "${configDir}/build",
-    "noEmit": true,
```
nickgros, line 4: All projects must emit the results of `tsc` to use project references. The files are emitted to this `node_modules/.cache/.tsc_cache` folder, so you won't notice it typically during development. Most of our projects were already doing this.

apps/synapse-portal-framework/src/components/crc-researcher/ParticipantsBarPlot.tsx
```diff
@@ -1,18 +1,18 @@
-import { CSSProperties, useEffect, useState } from 'react'
+import {
+  QueryBundleRequest,
+  QueryResultBundle,
+  RowSet,
+} from '@sage-bionetworks/synapse-types'
+import cloneDeep from 'lodash-es/cloneDeep'
 import Plotly from 'plotly.js-basic-dist'
-import _ from 'lodash-es'
```
nickgros, line 3: Not sure why upgrading TypeScript cause using the default export from lodash-es to break, but it's a good thing, because this import is hard to tree-shake

packages/markdown-it-container/test/cjs.cjs
nickgros, line 1: This test had a type error that was _not_ caused by these changes -- we never caught this one because the project does not have a `type-check` script.

The type error is because we're accessing a file generated by TypeScript in the 'dist' folder, which is obviously not allowed. This test basically verified that the bundler (tsc in this case) works, so IMO it's fine to remove.

tsconfig.json
jay-hodgson: We should probably note this in our docs related to adding a new portal

jay-hodgson (approved):

PR #1835, opened 2025-05-02
PORTALS-3393

jay-hodgson:


nickgros (approved):
Seems like all of our deps are using cross-spawn@7, so this should be ok

PR #1840, opened 2025-05-06
PORTALS-3092

lakikowolfe:
update e2e tests so that they don't run unless a change has been made to the portal


lakikowolfe:
Can see that even though I just made a change to the ampals portal e2e tests are run for all portals

lakikowolfe:
Failing because I did not see that I needed to set up GH actions to use `nx affected`

lakikowolfe:
@nickgros The `Install Playwrite Browsers` step is taking about a min each time and contributing to the bulk of the E2E testing run time when there are no changes detected.  If no changes are detected in the `Build Portal` it seems like it might be worth while to figure out how to skip `Install Playwrite Browsers` as well? What do you think?

nickgros:
@lakikowolfe It sounds like that would be ideal! Is there a way to see which projects are `affected` before we run that step?

nickgros:
Or even finding a way to cache the playwright install (if it's not already cached) could make a huge difference (and might be easier to implement)

lakikowolfe:
> Or even finding a way to cache the playwright install (if it's not already cached) could make a huge difference (and might be easier to implement)

My plan of action was to use the output of the `Build Portal` step to determine which of the following steps would be required.

I'll take a minute to look into the playwrite caching idea though. The output of `Build Portal` isn't the most straightforward to parse.

lakikowolfe:
@nickgros I was able to use the `if` param to bypass steps once it's determined that no change was made to the portal. The E2E action completed in 23 mins when no portals were changed. You can see that [here](https://github.com/Sage-Bionetworks/synapse-web-monorepo/actions/runs/14893751797/job/41831811411).

I noticed that most of the runtime comes from the step that checks out the repository from github and it got me thinking that we could fully bypass the e2e testing job by adding a job that checks if any changes had been made and controlling which jobs run from there. I've been tinkering for a bit with no success and wanted to check back with you. Is 23 min sufficient? Or is it worth while to keep tinkering? Maybe worth pushing what I had working and splitting further work to a new ticket?

Let me know what you think!

lakikowolfe:
I have moved the work exploring adding another job locally, going to push changes that revert back to what was previously working so I can fully test it

.github/workflows/end-to-end-test-portals.yml
```diff
@@ -36,19 +36,33 @@ jobs:
...
         with:
           fetch-depth: 0
+      - name: Derive appropriate SHAs for base and head for `nx affected` commands
+        uses: nrwl/nx-set-shas@v4
+
       - uses: ./.github/actions/pnpm-setup-action
+      - name: Save affected projects to environment variable
+        run: |
+          pnpm nx show projects --affected
+          echo "AFFECTED_PROJECTS=$(pnpm nx show projects --affected)" >> "$GITHUB_ENV"
```
lakikowolfe, line 46: Code fails when affected projects contains more than one project. I believe this is because pnpm will then output an array which cannot be passed to `GITHUB_ENV`

lakikowolfe: Rather than passing the array of affected projects I should just pass a true/false boolean that can be used directly in the `if` parameters of the following steps.

nickgros:
@lakikowolfe I think this reduction is great! One question: I see that the cancercomplexity portal has changed when comparing your PR to `main`, but it doesn't look like that e2e suite ran any tests. Is that expected?

lakikowolfe:
> @lakikowolfe I think this reduction is great! One question: I see that the cancercomplexity portal has changed when comparing your PR to `main`, but it doesn't look like that e2e suite ran any tests. Is that expected?

Hm I see what you're saying. That is not expected. Looking into it 👀

nickgros:
@lakikowolfe Awesome!! Are the AMP-ALS changes in this PR intentional?

nickgros (approved):

PR #1851, opened 2025-05-08
PORTALS-3578

nickgros:
Lazy-load routes in AMP-ALS portal


apps/portals/ampals/src/config/routesConfig.tsx
```diff
@@ -27,176 +14,116 @@ const routes: RouteObject[] = [
...
+            index: true,
+            lazy: () =>
+              import('@/pages/Search').then(convertModuleToRouteObject),
+          },
+          {
+            path: ':resourceType',
+            lazy: () =>
+              import('@/pages/Search').then(convertModuleToRouteObject),
+          },
+        ],
```
nickgros, line 126: instead of one route per search tab, treat the tab like a slug

apps/portals/ampals/src/pages/Search.tsx
```diff
@@ -0,0 +1,22 @@
...
+    tab => tab.path === resourceType,
+  )
+
+  return (
+    <PortalSearchPage
+      selectedTabIndex={selectedTabIndex === -1 ? undefined : selectedTabIndex}
+      configs={portalSearchPageConfigs}
+      searchPageTabs={searchPageTabs}
+    />
+  )
```
nickgros, line 19: use the slug to identify the tab (if possible)

apps/portals/ampals/src/config/routesConfig.tsx
kianamcc: These path parameters are captured by React Router's useParams() hook.

There seems to be a bit of variation in the origin story for the term "slug" but a popular answer seems to be that it comes from journalism/publishing where "slug" is a label used to identify an article. Maybe a shorthand title that is easy to remember. (https://stackoverflow.com/questions/4230846/what-is-the-etymology-of-slug-in-a-url)

apps/portals/ampals/src/pages/Search.tsx
kianamcc: Refactoring Search from using hardcoded routes to route parameters. Looks so much cleaner!

kianamcc (approved):

PR #1856, opened 2025-05-09
PORTALS-3581

nickgros:
Lazy-load routes in BSMN Portal


apps/portals/bsmn/src/pages/StudyDetailsPage.tsx
```diff
@@ -62,3 +62,5 @@ export default function StudyDetailsPage() {
     </>
   )
 }
+
+export default StudyDetailsPage
```
kianamcc, line 66: Need export default for React Router lazy loading.

apps/portals/bsmn/src/pages/Explore/Data.tsx
```diff
@@ -0,0 +1,8 @@
+import { studiesQueryWrapperPlotNavProps } from '@/config/synapseConfigs/studies'
```
kianamcc, line 1: Split explore page routes into separate files for lazy loading.

kianamcc (approved):

PR #1859, opened 2025-05-09
PORTALS-3583 - Lazy load routes in Challenge Portal

nickgros:
Lazy-load routes for Challenge Portal


apps/portals/challenges/src/config/routesConfig.tsx
```diff
@@ -21,65 +18,56 @@ const routes: RouteObject[] = [
...
+              ).then(convertModuleToRouteObject),
+          },
+          {
+            // Task tab last so we attempt to match the other paths first before using the slug
+            path: ':taskId',
+            lazy: () =>
+              import('@/pages/ChallengeDetailsPageTabContent/TaskTab').then(
+                convertModuleToRouteObject,
+              ),
+          },
```
nickgros, line 69: Before, we hard-coded `Task1`, `Task2`, etc. as individual routes. Instead, use one route with a path parameter.

kianamcc (approved):

PR #1863, opened 2025-05-13
SWC-7329 - Fix import issue in dev mode

nickgros:
SWC-7329

Use workaround for browser-bundling + server-side ESM quirk in esbuild.

See https://github.com/evanw/esbuild/issues/2480#issuecomment-1970337003 for more info & workaround reference

### DIFF
--- a/packages/synapse-react-client/src/components/Plot/safe-react-plotly-factory.ts
+++ b/packages/synapse-react-client/src/components/Plot/safe-react-plotly-factory.ts
@@ -1,0 +1,15 @@
+import _createPlotlyComponent from 'react-plotly.js/factory'
+
+// We want to bundle browser-compatible code, but use ESM (`"type": "module"` in package.json) when we use Node (e.g. tests).
+// esbuild (used by Vite in development mode) has a quirk where this will cause packages react-plotly.js/factory to be
+// imported in 'node mode', which breaks imports in the browser.
+// For further explanation and source of workaround, see: https://github.com/evanw/esbuild/issues/2480#issuecomment-1970337003
+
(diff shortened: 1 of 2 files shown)


packages/synapse-react-client/src/components/Plot/safe-react-plotly-factory.ts
```diff
@@ -0,0 +1,15 @@
+import _createPlotlyComponent from 'react-plotly.js/factory'
+
+// We want to bundle browser-compatible code, but use ESM (`"type": "module"` in package.json) when we use Node (e.g. tests).
+// esbuild (used by Vite in development mode) has a quirk where this will cause packages react-plotly.js/factory to be
+// imported in 'node mode', which breaks imports in the browser.
+// For further explanation and source of workaround, see: https://github.com/evanw/esbuild/issues/2480#issuecomment-1970337003
+
+const safeESModule = <T>(a: T | { default: T }): T => {
```
jay-hodgson, line 8: We may never need to, but safeESModule seems like a generic utility and not specific to plotly.  Consider refactoring to prepare for reuse (which we may never need until we use another module-based library)

jay-hodgson (approved):

packages/synapse-react-client/src/components/Plot/safe-react-plotly-factory.ts
nickgros: Agree with the idea. I think this is a very narrow edge case with how react-plotly.js/factory is built, so I'm going to hold off on the refactor until this happens twice. IMO this is a bug in the bundler, so I'm going to opt to not make this so easily reusable until we need to.

PR #1864, opened 2025-05-13
SWC-7327: Fix 'link to thread' button in Discussions

kianamcc:
Jira: https://sagebionetworks.jira.com/browse/SWC-7327?atlOrigin=eyJpIjoiNzRhODI1MzI4ZmRjNDM3NmFkYjRlMjUzYWRmODk4ZGUiLCJwIjoiaiJ9

SRC implementation of discussion thread was missing "link to thread reply" functionality.

Before:
Screenshot 2025-05-12 at 1 03 57 PM
Screenshot 2025-05-12 at 1 04 44 PM
Screenshot 2025-05-12 at 1 04 31 PM
Screenshot 2025-05-12 at 1 04 06 PM

After: 
Screenshot 2025-05-13 at 11 26 38 AM
Screenshot 2025-05-13 at 11 20 13 AM
Screenshot 2025-05-13 at 11 19 48 AM
Screenshot 2025-05-13 at 11 19 54 AM


packages/synapse-react-client/src/components/Forum/DiscussionReply.tsx
```diff
@@ -15,27 +15,29 @@ import { displayToast } from '../ToastMessage/ToastMessage'
...
   onClickLink?: () => void
 }
 
-const DEFAULT_ON_CLICK_LINK = () =>
-  alert('This functionality has not been implemented yet')
+const handleCopyLink = (id: string) => {
+  const baseUrl = `${window.location.origin}${window.location.pathname}`
+  const url = `${baseUrl}?replyid=${id}`
+  copyStringToClipboard(url)
+  displayToast('Reply link copied to clipboard', 'info')
```
nickgros, line 30: `copyStringToClipboard` is async
```suggestion
  copyStringToClipboard(url).then(() => displayToast('Reply link copied to clipboard', 'info'))
```

packages/synapse-react-client/src/components/Forum/DiscussionReply.tsx
```diff
@@ -15,27 +15,29 @@ import { displayToast } from '../ToastMessage/ToastMessage'
...
   reply: DiscussionReplyBundle
   isReplyAuthorModerator?: boolean
   onClickLink?: () => void
 }
 
-const DEFAULT_ON_CLICK_LINK = () =>
-  alert('This functionality has not been implemented yet')
+const handleCopyLink = (id: string) => {
+  const baseUrl = `${window.location.origin}${window.location.pathname}`
+  const url = `${baseUrl}?replyid=${id}`
```
nickgros, line 28: Make replyid a constant
```suggestion
  const url = `${baseUrl}?${REPLY_ID_PARAM_KEY}=${id}`
```

packages/synapse-react-client/src/components/Forum/DiscussionReply.tsx
```diff
@@ -46,8 +48,22 @@ export function DiscussionReply(props: DiscussionReplyProps) {
 
   const isCurrentUserAuthor = reply.createdBy == currentUserProfile?.ownerId
 
+  useEffect(() => {
+    const params = new URLSearchParams(window.location.search)
+    const replyId = params.get('replyid')
```
nickgros, line 53:
```suggestion
    const replyId = params.get('REPLY_ID_PARAM_KEY')
```

packages/synapse-react-client/src/components/Forum/DiscussionReply.tsx
```diff
@@ -46,8 +48,22 @@ export function DiscussionReply(props: DiscussionReplyProps) {
...
+        if (replyRef.current) {
+          replyRef.current.style.backgroundColor = 'transparent'
+        }
+      }, 2000)
+    }
+  }, [reply.id])
+
   return (
-    <div className="reply-container">
+    <div className="reply-container" id={reply.id} ref={replyRef}>
```
nickgros, line 66: Consider making the ID something like `reply_1234` so it's not just some number that could collide with something else on the page (which I concede is very unlikely)
```suggestion
    <div className="reply-container" id={`reply_${reply.id}`} ref={replyRef}>
```

packages/synapse-react-client/src/components/Forum/DiscussionReply.tsx
```diff
@@ -46,8 +48,22 @@ export function DiscussionReply(props: DiscussionReplyProps) {
...
+    const replyId = params.get('replyid')
+    if (replyId === reply.id && replyRef.current) {
+      replyRef.current.style.transition = 'background-color 1s ease'
+      replyRef.current.style.backgroundColor = '#fbf4e0'
+      setTimeout(() => {
+        if (replyRef.current) {
+          replyRef.current.style.backgroundColor = 'transparent'
+        }
+      }, 2000)
+    }
```
nickgros, line 62: Cleanup side effect in case this component unmounts before `setTimeout` runs
```suggestion
    let timerId
    if (replyId === reply.id && replyRef.current) {
      replyRef.current.style.transition = 'background-color 1s ease'
      replyRef.current.style.backgroundColor = '#fbf4e0'
      timerId = setTimeout(() => {
        if (replyRef.current) {
          replyRef.current.style.backgroundColor = 'transparent'
        }
      }, 2000)
    }
    return () => {
      if (timerId) {
        clearTimeout(timerId)
      }
    }
```

packages/synapse-react-client/src/components/Forum/DiscussionThread.tsx
```diff
@@ -98,14 +100,36 @@ export function DiscussionThread(props: DiscussionThreadProps) {
...
-  const replies = replyData?.pages.flatMap(page => page.results) ?? []
 
   const { data: moderatorList } = useGetModerators(threadData?.forumId ?? '')
 
   const isAuthorModerator = moderatorList?.results.includes(
     threadData?.createdBy ?? '',
   )
 
+  const params = new URLSearchParams(window.location.search)
+  const replyId = params.get('replyid')
```
nickgros, line 111:
```suggestion
  const replyId = params.get(REPLY_ID_PARAM_KEY)
```

packages/synapse-react-client/src/components/Forum/DiscussionThread.tsx
```diff
@@ -49,6 +50,7 @@ export function DiscussionThread(props: DiscussionThreadProps) {
   const [showSignInModal, setShowSignInModal] = useState(false)
   const [showRestoreModal, setShowRestoreModal] = useState(false)
   const [showSubscriberModal, setShowSubscriberModal] = useState(false)
+  const [replies, setReplies] = useState<DiscussionReplyBundle[]>([])
```
nickgros, line 53: Make the presence of the search param the source of truth for this 'state' and compute everything instead of using `useState` + `useEffect`

```suggestion
```

Suggestion continued below

packages/synapse-react-client/src/components/Forum/DiscussionThread.tsx
```diff
@@ -98,14 +100,36 @@ export function DiscussionThread(props: DiscussionThreadProps) {
...
+
+  const handleShowAllRepliesButton = () => {
+    const allReplies = replyData?.pages.flatMap(page => page.results) ?? []
+    setReplies(allReplies)
+
+    const params = new URLSearchParams(window.location.search)
+    params.delete('replyid')
+    const newUrl = `${window.location.pathname}?${params.toString()}`
+    window.history.replaceState({}, '', newUrl)
+  }
```
nickgros, line 131: Continued from above, make the URL param the single point of control over filtering logic. Just worried that the component doesn't reload and never catches the update to `window.location.search`, but hopefully it works.

Also suggest using `pushState` instead of `replaceState` so the user can go back to the reply-only if desired.

```suggestion
  const replies = useMemo(() => {
    const allReplies = replyData?.pages.flatMap(page => page.results) ?? []
    if (replyId) {
      return allReplies.filter(reply => reply.id === replyId)
    }
    return allReplies
  }, [replyId, replyData])

  const handleShowAllRepliesButton = () => {
    const params = new URLSearchParams(window.location.search)
    params.delete('replyid')
    const newUrl = `${window.location.pathname}?${params.toString()}`
    window.history.pushState({}, '', newUrl)
  }
```

kianamcc: This doesn't seem to work for me. When the URL is modified the component doesn't re-render. Should window.location.search be in the useMemo hook?

nickgros: React won't know when `window.location.search` updates because it's not tracked internally. Can we add a `useNativeSearchParams` hook that is similar to the implementation here? https://www.30secondsofcode.org/react/s/use-hash-search-param/

Then use that to get `replyid`.

packages/synapse-react-client/src/components/Forum/DiscussionThread.tsx
```diff
@@ -98,14 +122,34 @@ export function DiscussionThread(props: DiscussionThreadProps) {
...
+      return allReplies
+    }
+  }, [replyData, replyId])
+
+  const handleShowAllRepliesButton = () => {
+    const params = new URLSearchParams(window.location.search)
+    params.delete(REPLY_ID_PARAM_KEY)
+    const newUrl = `${window.location.pathname}?${params.toString()}`
+    history.pushState({}, '', newUrl)
+    window.dispatchEvent(new Event('pushstate'))
```
kianamcc, line 150: Had to manually trigger the state.

packages/synapse-react-client/src/utils/hooks/useNativeSearchParams.tsx
```diff
@@ -0,0 +1,24 @@
+import { useCallback, useState, useEffect } from 'react'
+
+export const useNativeSearchParams = (param: string) => {
```
nickgros, line 3:
```suggestion
/**
 * React hook that can be used to access the current state of window.location.search.
 * Only use this when you cannot use an equivalent hook provided by a routing library (e.g. react-router)
 */
export const useNativeSearchParams = (param: string) => {
```

packages/synapse-react-client/src/components/Forum/DiscussionThread.tsx
```diff
@@ -98,14 +99,34 @@ export function DiscussionThread(props: DiscussionThreadProps) {
...
   } = useGetRepliesInfinite(threadId, orderByDatePosted, limit)
-  const replies = replyData?.pages.flatMap(page => page.results) ?? []
 
   const { data: moderatorList } = useGetModerators(threadData?.forumId ?? '')
 
   const isAuthorModerator = moderatorList?.results.includes(
     threadData?.createdBy ?? '',
   )
 
+  const REPLY_ID_PARAM_KEY = 'replyid'
```
nickgros, line 109: use the same constant variable in both files

packages/synapse-react-client/src/components/Forum/DiscussionThread.tsx
```diff
@@ -263,6 +284,29 @@ export function DiscussionThread(props: DiscussionThreadProps) {
...
+            width: 'fit-content',
+            padding: '2px 8px',
+            marginBottom: '12px',
+            gap: '4px',
+            cursor: 'pointer',
+          }}
+        >
+          <IconSvg
+            icon="arrowBack"
+            label="Restore deleted thread"
```
nickgros, line 304: This label should not be here, right?
```suggestion
```

packages/synapse-react-client/src/components/Forum/DiscussionThread.tsx
```diff
@@ -263,6 +284,29 @@ export function DiscussionThread(props: DiscussionThreadProps) {
           />
         )}
       </Box>
+      {replyId && (
+        <Box
```
nickgros, line 288: Any reason to not just make this a MUI `Button` with a `startIcon`? You could probably get rid of a bunch of the styles

nickgros (changes requested):
Couple more comments, any opportunity to add test(s)?

packages/synapse-react-client/src/components/Forum/DiscussionThread.tsx
kianamcc: Different from what I originally had but I changed it to this MUI button:

Screenshot 2025-05-19 at 2 16 06 PM

PR #1867, opened 2025-05-15
PORTALS-3585: Lazy load routes for ELITE Portal

kianamcc:
Jira: https://sagebionetworks.jira.com/browse/PORTALS-3585


apps/portals/eliteportal/src/config/routesConfig.tsx
```diff
@@ -1,206 +1,159 @@
...
-          <SectionLayout title={'Omics Profiles in Humans'}>
-            <Markdown
-              ownerId="syn27229419"
-              wikiId="630616"
-              loadingSkeletonRowCount={15}
-            />
-          </SectionLayout>
-        ),
+        lazy: () =>
+          import('@/pages/OmnicsProfilesInHumans').then(
```
nickgros, line 134:
```suggestion
          import('@/pages/OmicsProfilesInHumans').then(
```

apps/portals/eliteportal/src/config/synapseConfigs/handleFilesToParticipants.tsx
```diff
@@ -4,10 +4,10 @@ import {
 } from '@sage-bionetworks/synapse-types'
 import {
   SynapseClient,
-  SynapseConstants,
   SynapseUtilityFunctions,
```
nickgros, line 7: Just a note: all of the non-type imports from 'synapse-react-client' trigger importing the entire bundle. So just separating SynapseConstants doesn't change anything when you still have this.

No need to fix in this PR, we can eventually do a separate pass once all of the portals have been migrated to this pattern.

apps/portals/eliteportal/src/pages/HomePage.tsx
```diff
@@ -132,7 +133,7 @@ export default function HomePage() {
     </span>
   )
   return (
-    <>
+    <HomePageThemeProvider>
```
nickgros, line 136: Note that `theme` object on line 38 will now get the outer theme and not the home page theme. You could fix this by doing something like

```tsx
function HomePageInternal() {
  const theme = useTheme()
  // ...
}

function HomePage() {
  return (
    <HomePageThemeProvider>
      <HomePageInternal />
    </HomePageThemeProvider>
  )
}
```

nickgros (changes requested):
Looks good, couple of small issues

nickgros (approved):

PR #1868, opened 2025-05-15
PLFM-8986 - don't error on entity type schema failure

nickgros:


nickgros:
The error I was trying to fix is thrown by RJSF, not returned by the Synapse API, so this approach would not work.

PR #1872, opened 2025-05-16
PORTALS-3600

jay-hodgson:


packages/synapse-react-client/src/components/GenericCard/Linkify.test.tsx
```diff
@@ -121,6 +121,18 @@ describe('Linkify tests', () => {
     )
   })
 
+  test('auto-links doi with Synapse ID', () => {
```
jay-hodgson, line 124: Adding a new test, but there are existing tests for Synapse IDs.  See 'auto-link only the Synapse ID' and 'auto-links a Synapse ID'

kianamcc (approved):
Didn't occur to me that this was in the Synapse Web Monorepo 😅 Looks like a good fix for handling embedded Synapse IDs in URLs!

PR #1880, opened 2025-05-19
PORTALS-3609

nickgros:
PORTALS-3609

Allow configuring the portal card download component with rowId/versionNumber and do so in the NF datasets props


apps/portals/nf/src/config/synapseConfigs/datasets.ts
```diff
@@ -46,7 +46,7 @@ export const datasetCardConfiguration: CardConfiguration = {
       'visualizeDataOn',
     ],
     dataTypeIconNames: 'dataType',
-    downloadCartSynId: 'id',
+    useRowIdAndVersionForDownloadCart: true,
```
nickgros, line 49: Just use the rowId/version for adding to download cart, since it's a dataset collection.

packages/synapse-react-client/src/components/GenericCard/TableRowGenericCard.tsx
```diff
@@ -84,7 +84,10 @@ export type TableToGenericCardMapping = {
   link?: string
   /** Column name of the STRING_LIST column that includes icon names that represent icons that should be displayed on the card */
   dataTypeIconNames?: string
-  /** The column name whose data contains a synId that can be used to show a button to add the corresponding entity to the download cart. */
+  /** If true, uses the rowId and row versionNumber to display a button to add the card item to the download cart */
+  useRowIdAndVersionForDownloadCart?: boolean
```
nickgros, line 88: Added new config option to use rowId/version for the Download Cart feature

afwillia (approved):
LGTM - can confirm the dataset in the ticket now downloads without error

PR #1892, opened 2025-05-21
PORTALS-3188 - Evaluation Round Info tooltip

nickgros:
PORTALS-3188

Add column to the Evaluation Queue picker with round info

image

### DIFF
--- a/packages/synapse-react-client/src/components/ChallengeSubmission/EvaluationQueueCurrentRoundInfo.tsx
+++ b/packages/synapse-react-client/src/components/ChallengeSubmission/EvaluationQueueCurrentRoundInfo.tsx
@@ -14,8 +14,7 @@ type EvaluationQueueInfoProps = {
 }
 
 /**
- * Displays an information tooltip about the current round of an Evaluation Queue.
- * If there is no current round, then
+ * Displays an information tooltip about the current (or recent) round of an Evaluation Queue.
  * @param props
  * @constructor
  */
--- a/packages/synapse-react-client/src/components/ChallengeSubmission/EvaluationQueueCurrentRoundInfo.tsx
+++ b/packages/synapse-react-client/src/components/ChallengeSubmission/EvaluationQueueCurrentRoundInfo.tsx
@@ -1,0 +1,84 @@
+import {
+  useGetEvaluation,
+  useGetEvaluationRoundsInfinite,
+} from '@/synapse-queries/evaluation/useEvaluation'
+import { getCurrentOrNextOrLatestRound } from '@/utils/challenge/evaluation/EvaluationUtils'
+import { formatDate } from '@/utils/functions/DateFormatter'
+import { InfoTwoTone } from '@mui/icons-material'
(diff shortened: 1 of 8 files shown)


packages/synapse-react-client/src/components/ChallengeSubmission/EvaluationQueueCurrentRoundInfo.tsx
```diff
@@ -0,0 +1,84 @@
...
+import dayjs from 'dayjs'
+import { useEffect, useMemo } from 'react'
+
+type EvaluationQueueInfoProps = {
+  evaluationId: string
+}
+
+/**
+ * Displays an information tooltip about the current round of an Evaluation Queue.
+ * If there is no current round, then
```
nickgros, line 18:
```suggestion
 * Displays an information tooltip about the current (or recent) round of an Evaluation Queue.
```

packages/synapse-react-client/src/components/Evaluation/EvaluationCard.tsx
```diff
@@ -191,19 +191,37 @@ function EvaluationCardDropdown({
         slotProps={{ paper: { sx: { minWidth: '120px' } } }}
       >
         {permissions.canEdit && (
```
nickgros, line 193: Fix a bug in the EvaluationCard (SWC challenge editor)  where menu options did not close after click

packages/synapse-react-client/src/components/StorybookComponentWrapper.tsx
```diff
@@ -54,6 +55,10 @@ function overrideEndpoint(stack: SynapseStack) {
   ;(window as any)['SRC_OVERRIDE_ENDPOINT_CONFIG'] = endpointConfig
 }
 
+// Store auth token in global atom to avoid race conditions where the Story component
+// may not get the login token right away
+const authTokenAtom = atom<string | undefined>(undefined)
```
nickgros, line 60: I think Storybook may have changed something that caused this wrapper component to be completely re-created instead of reused when opening a new story, so `accessToken` would always be `undefined` until it was retrieved. This would cause stories requiring authentication to tell you to log in even if you already did. This should fix that.

packages/synapse-react-client/src/utils/challenge/evaluation/EvaluationUtils.ts
```diff
@@ -0,0 +1,47 @@
...
+import dayjs from 'dayjs'
+
+/**
+ * Given a list of evaluation rounds:
+ *  - If there is a current round, return it
+ *  - Else if there is one or more upcoming rounds, return the soonest
+ *  - Else return the most recent round in the past
+ * @param rounds
+ */
+export function getCurrentOrNextOrLatestRound(
```
nickgros, line 11: We have to decide which round to pick when we show the start/end/limits. I think this algorithm aligns with what users might expect.

jay-hodgson (approved):

PR #1893, opened 2025-05-21
SWC-7334: Do not report user cannot edit sharing permissions if inher…

jay-hodgson:
…ited (user may be able to modify if they go to the benefactor)


jay-hodgson:
Cannot edit local ACL:
Screenshot 2025-05-21 at 10 41 12 AM

Cannot edit here due to inherited ACL:
Screenshot 2025-05-21 at 10 41 18 AM

Can edit:
Screenshot 2025-05-21 at 10 41 31 AM

nickgros (approved):
Oops, thanks for this fix!

PR #1897, opened 2025-05-22
SWC-7089: change type from DataCatalog to Dataset

jay-hodgson:


jay-hodgson:
DataCatalog is correct in this context, it's just not supported by the Google Rich Results tester

PR #1912, opened 2025-05-29
PORTALS-3450

jay-hodgson:
Per https://sagebionetworks.jira.com/browse/AMPALS-41?focusedCommentId=251638
There will not be any datasets hosted on Synapse for v0.1, so file level data will not be shown.
Jessica will create a description on the dataset wiki for each dataset.  Change config to show the dataset wiki on the Dataset Details Page.


apps/portals/ampals/src/pages/HomePage.tsx
```diff
@@ -87,63 +87,24 @@ export default function HomePage() {
       <ALLALSSlat />
       <div className={'home-bg-dark'}>
         <SectionLayout
-          title={'Featured Datasets'}
+          title={'More Resources'}
           centerTitle
-          ContainerProps={{ className: 'home-spacer' }}
+          ContainerProps={{
+            sx: {
+              maxWidth: '100% !important',
```
nickgros, line 94: Since this is applied to a [MUI Container](https://mui.com/material-ui/api/container/), you can use the `maxWidth` prop directly
```suggestion
            maxWidth: false,
            sx: {
```

nickgros (approved):

PR #1914, opened 2025-05-30
PORTALS-3613

nickgros:
PORTALS-3613

For https://github.com/bridge2ai/b2ai-standards-registry/issues/300

- Add MapValueToReactComponentConfig to SynapseCardLabel, which provides a way to generically map a table value to a React Component implementation
- Add IsMatureIconMap component and integrate into Standards Portal as a MapValueToReactComponentConfig

image

### DIFF
--- a/packages/synapse-react-client/src/components/CardContainerLogic/CardContainerLogic.tsx
+++ b/packages/synapse-react-client/src/components/CardContainerLogic/CardContainerLogic.tsx
@@ -10,9 +10,10 @@ import { DEFAULT_PAGE_SIZE } from '@/utils/SynapseConstants'
 import {
   Query,
   QueryBundleRequest,
+  SelectColumn,
   SortDirection,
 } from '@sage-bionetworks/synapse-types'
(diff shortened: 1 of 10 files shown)


packages/synapse-react-client/src/components/CardContainerLogic/CardContainerLogic.tsx
```diff
@@ -78,12 +79,25 @@ export type DescriptionConfig = {
...
+  // The column to apply the mapping to
+  matchColumnName: string
+  Component: React.ComponentType<{
+    /* The value of the table cell. If the column is a _LIST type, then the LIST will be parsed and this will be a string array. */
+    value: string | string[] | null
+    selectColumn: SelectColumn
+  }>
+  // discriminator
+  isMapValueToReactNodeConfig: true
+}
```
nickgros, line 92: Super generic interface. We could make _n_ reusable components to get passed into this interface instead of having _n_ different configurations

jay-hodgson (approved):

PR #1917, opened 2025-05-30
PORTALS-3351

nickgros:
- Upgrade React, React DOM to v19
- ~Fix tests broken by upgrade (mostly assertions based on an assumption about how mock function components are called)~ #1969
- ~Fix React 19-specific issue with react-transition-group by passing a nodeRef of the element to animate to CSSTransition~ #1947 
- ~Remove usage of `defaultProps`~ #1948
- ~Replace `react-dom/test-utils` import of `act` with `@testing-library/react`~ #1949 
- Upgrade various dependencies for compatibility with React 19 (our production code is not impacted by any breaking changes in these libraries)
	- `@react-hookz/web` (23.1.0 -> 25.1.1)
	- `@tanstack/react-query` (5.22.2 -> 5.80.10) (will need to update SWC to stay in sync)
	- `jotai-devtools` (0.6.3 -> 0.12.0)


packages/synapse-react-client/jest.config.cjs
```diff
@@ -6,6 +6,7 @@ const esModules = [
   'mui-one-time-password-input',
   'p-limit',
   'yocto-queue',
+  '@ver0',
```
nickgros, line 9: `@react-hookz/web` upgrade pulls in some new transitive dependencies that are ESM only. Update the Jest config to properly handle these deps.

apps/SageAccountWeb/src/components/ProfileValidation/ReturnToAppButton.tsx
```diff
@@ -11,32 +11,40 @@ import {
...
   const theme = useTheme()
   const [open, setOpen] = useState(false)
   const onClose = () => setOpen(false)
   const redirectFn = () => {
     window.location.assign('/authenticated/myaccount')
   }
-  const element = !children ? (
+  const element = children ? (
+    children(() => setOpen(true))
+  ) : (
```
nickgros, line 30: Instead of passing a new prop via `cloneElement`, use the render props pattern and pass a callback that can be consumed by `children`

apps/portals/adknowledgeportal/src/config/routesConfig.tsx
```diff
@@ -122,11 +122,12 @@ const routes: RouteObject[] = [
...
-        lazy: () =>
-          import('@/pages/Contribute/Contribute-FormSubmission').then(
-            convertModuleToRouteObject,
-          ),
+        // The FormSubmission page has been replaced with a link to Service Desk
+        // path: 'Contribute/FormSubmission',
+        // lazy: () =>
+        //   import('@/pages/Contribute/Contribute-FormSubmission').then(
+        //     convertModuleToRouteObject,
+        //   ),
```
nickgros, line 130: I thought I had changed this in an earlier PR, but somehow it ended up in this one...

This page is not exposed anywhere in the portal and I checked with Ann C that it's ok to hide this.

packages/synapse-react-client/src/components/GenericCard/CollapsibleDescription.tsx
```diff
@@ -34,8 +34,11 @@ export function LongDescription(props: {
     descriptionSubTitle,
     descriptionConfig,
   } = props
-  let content: JSX.Element | string = description
-  if (descriptionConfig?.isMarkdown) {
+  let content: ReactNode = description
+  if (
+    descriptionConfig?.isMarkdown &&
+    (content == null || typeof content === 'string')
```
nickgros, line 40: Add typeguard, does not change behavior (this would have crashed if `content` was not null or a string)

packages/synapse-react-client/src/synapse-queries/QueryFilterUtils.test.ts
```diff
@@ -41,7 +40,7 @@ describe('QueryFilterUtils', () => {
       expect(
         suppliedPredicate(
           new Query({
-            cache: new QueryCache(),
+            client: new QueryClient(),
```
nickgros, line 43: Internal changes in react-query

packages/synapse-react-client/src/utils/hooks/useFocus.ts
nickgros, line 1: Unused

apps/SageAccountWeb/src/components/LeftRightPanel.tsx
```diff
@@ -6,8 +7,8 @@ export function LeftRightPanel({
   rightContent,
   className,
 }: {
-  leftContent: JSX.Element
-  rightContent: JSX.Element
```
nickgros, line 10: The ambient `JSX` type was removed. We could `import { JSX } from 'react'` but in all of these cases, `ReactNode` is a better type to use.

packages/synapse-react-client/src/components/DownloadCart/RequestDownloadCard.test.tsx
```diff
@@ -13,14 +13,21 @@ import {
...
   REQUEST_DOWNLOAD_TITLE,
   RequestDownloadCard,
   RequestDownloadCardProps,
 } from './RequestDownloadCard'
 
+vi.mock('../EntityAclEditor/EntityAclEditorModal', () => ({
+  default: vi.fn(() => null),
+}))
+
+const mockEntityAclEditorModal = vi.mocked(EntityAclEditorModal)
```
nickgros, line 29: Test was flaky, easier to use a mock for the component where we were getting flaky behavior

packages/synapse-react-client/src/synapse-queries/chat/useChat.ts
```diff
@@ -124,7 +124,6 @@ export function useGetAgentChatSessionHistoryInfinite<
       SessionHistoryResponse,
       SynapseClientError,
       TData,
-      SessionHistoryResponse,
```
nickgros, line 127: Types changed in react-query, so this generic argument is no longer necesssary

packages/synapse-react-client/src/testutils/ComponentToBePassedPropsCustomMatcher.ts
```diff
@@ -21,10 +21,7 @@ expect.extend({
   },
   toHaveBeenRenderedWithProps(mockComponent: Mock, expectedProps: unknown) {
     try {
-      expect(mockComponent).toHaveBeenCalledWith(
-        expectedProps,
-        expect.anything(),
-      )
+      expect(mockComponent).toHaveBeenCalledWith(expectedProps, undefined)
```
nickgros, line 24: In React 19, a function component is called with `undefined` 2nd argument instead of an object.

packages/synapse-react-client/src/utils/hooks/index.ts
```diff
@@ -1,7 +1,5 @@
-export * from './useCompare'
```
nickgros, line 1: Removed some hooks that were unusued.

kianamcc (approved):

PR #1921, opened 2025-06-02
PORTALS-3621 - Fix Sass deprecations

nickgros:
PORTALS-3621

Fix Sass issues causing deprecation warnings


apps/portals/stopadportal/src/config/style/_style_overrides.scss
```diff
@@ -30,7 +31,7 @@ footer {
 }
 
 #home-link {
-  @media (max-width: map-get(SrcVariables.$breakpoints, 'medium')) {
+  @media (max-width: map.get(SrcVariables.$breakpoints, 'medium')) {
```
nickgros, line 34: Have to replace builtin functions with module versions: https://css-tricks.com/introducing-sass-modules/#aa-import-built-in-sass-modules

packages/synapse-react-client/src/style/abstracts/_mixins.scss
```diff
@@ -185,3 +186,19 @@ $link-underline-color-default: adjust-color(
...
+
+@mixin side-padding($padding: 25px) {
+  padding-left: $padding;
+  padding-right: $padding;
+}
+
+@mixin space-between() {
+  display: flex;
+  justify-content: space-between;
+}
```
nickgros, line 204: These new mixins already existed as placeholders (`%hv-center` syntax), but sharing placeholders doesn't seem to be possible with the module system. I don't fully understand why/how this changed, but moving them here to be shared mixins was very easy and solved the issue.

packages/synapse-react-client/src/style/components/entity_finder/_entity-finder.scss
```diff
@@ -3,19 +3,19 @@
...
-$-breadcrumb-height: 40px;
-$-selected-background-color: map.get(SRC.$secondary-color-palette, 200);
-$-selected-hover-background-color: color.adjust(
-  $-selected-background-color,
+$border: $-border-size solid SRC.$border-color-gray;
+$finder-height: 500px;
+$breadcrumb-height: 40px;
+$selected-background-color: map.get(SRC.$secondary-color-palette, 200);
+$selected-hover-background-color: color.adjust(
+  $selected-background-color,
```
nickgros, line 11: The sass-migrator codemod decided to change all of these. I think this is because these variables are exported and used in other scss files, so they should not have the `-` prefix, which indicates locality.

kianamcc: Seems so!: https://sass-lang.com/documentation/at-rules/use/ (Private Members section)

packages/synapse-react-client/src/style/components/synapse_form_wrapper/_synapse-form-wrapper.scss
```diff
@@ -52,8 +53,8 @@
   .btn-action:hover,
   .btn-action:focus {
     background-color: SrcVariables.$primary-action-color;
-    @include SrcMixins.box-shadow(3px, 3px, 5px, rgba(0, 0, 0, 0.5));
     color: #fff;
+    @include SrcMixins.box-shadow(3px, 3px, 5px, rgba(0, 0, 0, 0.5));
```
kianamcc, line 57: Do we know why this was moved below color? Does it make any difference?

kianamcc (approved):

PR #1926, opened 2025-06-04
Allow pnpm v10

afwillia:
Update the engines field in package.json to allow pnpm v10. Fixes the following error installing new packages
`Expected version: >=9 <10
Got: 10.11.1`


afwillia:
@nickgros ok thanks, that sounds good!

PR #1946, opened 2025-06-09
Remove redundant hook dependency

nickgros:
Replace `@react-hook/resize-observer` with `@react-hookz/web` implementation


packages/synapse-react-client/src/components/SynapseTable/ExpandableTableDataCell.tsx
```diff
@@ -1,5 +1,5 @@
 import { useRef, useState } from 'react'
-import useResizeObserver from '@react-hook/resize-observer'
+import { useResizeObserver } from '@react-hookz/web'
```
nickgros, line 2: We are using `@react-hookz/web` for other utilities, so we might as well use this one too

jay-hodgson (approved):

PR #1959, opened 2025-06-11
PORTALS-3624

jay-hodgson:


packages/synapse-react-client/src/components/GenericCard/BioregistryRules.ts
```diff
@@ -0,0 +1,869 @@
+/* eslint-disable no-useless-escape */
+const onMatch = (value: string) => `https://bioregistry.io/${value}`
+
+export const bioregistryRules = [
+  {
```
nickgros, line 5: Would it make sense to create a library that programmatically generates these regular expressions using the latest [registry](https://github.com/biopragmatics/bioregistry/blob/main/src/bioregistry/data/bioregistry.json)?

Then to update this we would just re-run the script?

jay-hodgson: Yes!  I believe that was the intent from Robert's work, that results in this file: https://github.com/Sage-Bionetworks/bioregistry-collection/blob/main/bioregistry.ts
I think we're very close to using this auto-generated output, but [my comments in the ticket](https://sagebionetworks.jira.com/browse/PORTALS-3624?focusedCommentId=255547) describe where I needed to manually change the regexes.   I added a comment to the file

PR #1967, opened 2025-06-13
SWC-7358: Add Cite As button in Synapse Entity Page

kianamcc:
Jira: https://sagebionetworks.jira.com/browse/SWC-7358?atlOrigin=eyJpIjoiMWJlNTg0OTJiNTFkNGI5ZmJmOGRkYmM4NGNiZDBmYjUiLCJwIjoiaiJ9

- LinkWithIcon component
- EntityCitation component
- Refactor CitationPopover component

Project only:
Screenshot 2025-06-13 at 1 39 20 PM

Project + Entity:
Screenshot 2025-06-13 at 1 39 28 PM
Screenshot 2025-06-13 at 1 41 20 PM

Only Entity:
Screenshot 2025-06-13 at 1 42 23 PM


packages/synapse-react-client/src/assets/icons/DoubleQuotes.svg
kianamcc, line 1: Needed to change fill-color to not be hardcoded value since these quotation marks are white in synapse.org

packages/synapse-react-client/src/components/CitationPopover/CitationPopover.tsx
```diff
@@ -117,144 +41,13 @@ function CitationPopover({
           </Typography>
         </GenericCardActionButton>
       )}
-      <Popover
-        aria-label="Citation options"
-        role="dialog"
-        id={id}
-        elevation={9}
-        open={open}
+      <CitationPopoverContent
```
kianamcc, line 44: Reusing this in EntityCitation

packages/synapse-react-client/src/components/PortalSectionHeader/PortalSectionHeader.tsx
```diff
@@ -29,8 +30,8 @@ const PortalSectionHeader = ({
   centered = false,
   reverseButtonAndText = false,
 }: PortalSectionHeaderProps) => {
-  const isExternalLink =
-    link?.startsWith('http://') || link?.startsWith('https://')
+  const isExternal = IsExternalLink(link ?? '')
```
kianamcc, line 33: utilize utility function

packages/synapse-react-client/src/components/LinkWithIcon/LinkWithIcon.tsx
kianamcc, line 1: Quick mockup from Adam in a Slack discussion called for this change: image

packages/synapse-react-client/src/components/EntityCitation/EntityCitation.tsx
```diff
@@ -0,0 +1,205 @@
...
+  entityId?: string
+  versionNumber?: number
+}
+
+const EntityCitation = ({
+  projectId,
+  entityId,
+  versionNumber,
+}: EntityCitationProps) => {
+  const { data: bundle } = useGetEntityBundle(entityId!)
```
nickgros, line 22: Should pass the version number to the `useGetEntityBundle` call

packages/synapse-react-client/src/components/EntityCitation/EntityCitationUtils.ts
```diff
@@ -0,0 +1,49 @@
+import { useGetDOIAssociation } from '@/synapse-queries/doi/useDOI'
+import { isVersionableEntity } from '@/utils/functions/EntityTypeUtils'
+import { DoiObjectType } from '@sage-bionetworks/synapse-client'
+import { EntityBundle } from '@sage-bionetworks/synapse-types'
+
+export function useGetEntityDoiAssociation(
+  entityId?: string,
+  versionNumber?: number,
+  projectId?: string,
+  bundle?: EntityBundle,
```
nickgros, line 10: why not just call `useGetEntityBundle` in this hook?

packages/synapse-react-client/src/components/EntityCitation/EntityCitation.tsx
```diff
@@ -0,0 +1,205 @@
...
+
+  const open = Boolean(anchorEl)
+  const id = open ? 'cite-as-popover' : undefined
+
+  useEffect(() => {
+    // Clear selection when switching entities or tabs
+    setSelectedDoi(undefined)
+    setCitationAnchorEl(null)
+    setAnchorEl(null)
+  }, [projectId, entityId, versionNumber])
```
nickgros, line 37: This makes sense, but would this scenario be better handled by generic behavior to close the popover on [Click Away](https://mui.com/material-ui/react-click-away-listener/)? Or is there a scenario where this happens outside of a 'click away' event?

packages/synapse-react-client/src/components/EntityCitation/EntityCitation.tsx
```diff
@@ -0,0 +1,205 @@
...
+                >
+                  <ExpandMoreIcon
+                    className="user-expand"
+                    width={18}
+                    height={18}
+                  />
+                </Box>
+              )
+            }
+          >
```
nickgros, line 189: Maybe instead of negative margin icon styling, could we use the [`DropdownSelect`](https://github.com/Sage-Bionetworks/synapse-web-monorepo/blob/48ba269795ecd5875af9e1cc816c6f771d405ac0/packages/synapse-react-client/src/components/DropdownSelect/DropdownSelect.tsx#L31) component when we have both DOIs?

packages/synapse-react-client/src/components/EntityCitation/EntityCitation.tsx
```diff
@@ -0,0 +1,205 @@
...
+      {(projectDoiAssociation || entityDoiAssociation) && (
+        <>
+          <Button
+            onClick={handleButtonClick}
+            sx={{
+              '& .MuiButton-endIcon': {
+                margin: 0,
+              },
+              borderColor: '#FFF !important',
+              color: '#FFF !important',
```
nickgros, line 158: This shouldn't be necessary--could this be removed completely? Or at least something like

```suggestion
              borderColor: 'currentColor',
              color: 'currentColor',
```

packages/synapse-react-client/src/components/EntityCitation/EntityCitation.stories.tsx
```diff
@@ -0,0 +1,61 @@
...
+import { useTheme, Box } from '@mui/material'
+
+const RenderComponent = (StoryComponent: StoryFn) => {
+  const theme = useTheme()
+  return (
+    <Box sx={{ backgroundColor: theme.palette.primary.main, padding: '20px' }}>
+      <StoryComponent />
+    </Box>
+  )
+}
```
nickgros, line 12: Better name for this decorator, and another style for accuracy

```suggestion
const ProjectHeaderStyleDecorator = (StoryComponent: StoryFn) => {
  const theme = useTheme()
  return (
    <Box sx={{ backgroundColor: theme.palette.primary.main, color: 'white', padding: '20px' }}>
      <StoryComponent />
    </Box>
  )
}
```

packages/synapse-react-client/src/components/EntityCitation/EntityCitation.test.tsx
```diff
@@ -0,0 +1,290 @@
...
+      { wrapper: createWrapper() },
+    )
+
+    act(() => {
+      setMockUseGetEntityBundleSuccess(fileWithAnnotations)
+      setEntityDOISuccess(doiEntitySuccess)
+      setProjectDOISuccess(null)
+    })
+
+    openPopover('Cite page')
```
nickgros, line 145: You should be getting a lint error here?
```suggestion
    await openPopover('Cite page')
```

packages/synapse-react-client/src/components/EntityCitation/EntityCitation.test.tsx
```diff
@@ -0,0 +1,290 @@
...
+    )
+
+    // Set mock responses after hooks have initialized
+    act(() => {
+      setMockUseGetEntityBundleSuccess(fileWithAnnotations)
+      setEntityDOISuccess(null)
+      setProjectDOISuccess(doiProjectSuccess)
+    })
+
+    openPopover('Cite project')
```
nickgros, line 181:
```suggestion
    await openPopover('Cite project')
```

packages/synapse-react-client/src/components/EntityCitation/EntityCitation.test.tsx
```diff
@@ -0,0 +1,290 @@
...
+      { wrapper: createWrapper() },
+    )
+
+    act(() => {
+      setMockUseGetEntityBundleSuccess(fileWithAnnotations)
+      setEntityDOISuccess(doiEntitySuccess)
+      setProjectDOISuccess(doiProjectSuccess)
+    })
+
+    openPopover('Cite as...')
```
nickgros, line 204:
```suggestion
    await openPopover('Cite as...')
```

packages/synapse-react-client/src/components/EntityCitation/EntityCitation.test.tsx
```diff
@@ -0,0 +1,290 @@
...
+      { wrapper: createWrapper() },
+    )
+
+    act(() => {
+      setMockUseGetEntityBundleSuccess(fileWithAnnotations)
+      setEntityDOISuccess(versionedDoiEntitySuccess)
+      setProjectDOISuccess(null)
+    })
+
+    openPopover('Cite page')
```
nickgros, line 250:
```suggestion
    await openPopover('Cite page')
```

packages/synapse-react-client/src/components/EntityCitation/EntityCitation.test.tsx
```diff
@@ -0,0 +1,290 @@
...
+      if (request.id === 'syn61841662' && request.version === 1) {
+        return useGetEntityDOIMockImpl()
+      } else if (
+        request.id === 'syn60582629' &&
+        request.version === undefined
+      ) {
+        return useGetProjectDOIMockImpl()
+      }
+      return useGetEntityDOIMockImpl()
+    })
```
nickgros, line 128: This is fine. If you prefer the more declarative style (like Mockito in Java), we have the `vitest-when` package where you could do:

```suggestion
    when(mockUseGetDOIAssociation)
      .calledWith(expect.objectMatching({ id: fileId, version: fileVersion })
      .thenReturn(useGetEntityDOIWithVersionMockImpl())
    when(mockUseGetDOIAssociation)
      .calledWith(expect.objectMatching({ id: fileId', version: undefined })
      .thenReturn(useGetEntityDOIMockImpl())
    when(mockUseGetDOIAssociation)
      .calledWith(expect.objectMatching({ id: projectId })
      .thenReturn(useGetProjectDOIMockImpl())    
```

But it's totally up to you!

packages/synapse-react-client/src/components/LinkWithIcon/LinkWithIcon.tsx
```diff
@@ -0,0 +1,52 @@
...
+import IconSvg, { IconName } from '../IconSvg'
+
+export type LinkWithIconProps = {
+  text: string
+  icon?: IconName
+  sx?: SxProps
+  href: string
+}
+
+const LinkWithIcon = (props: LinkWithIconProps) => {
```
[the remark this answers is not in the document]
kianamcc: Based on the styling in the design and for semantic purposes, I thought using a link would be more appropriate. Let me know if you would prefer that I change this into a button instead!

packages/synapse-react-client/src/components/EntityCitation/EntityCitation.tsx
kianamcc: Ah, this might actually be redundant.. It works fine without this useEffect.

packages/synapse-react-client/src/components/EntityCitation/EntityCitation.test.tsx
```diff
@@ -0,0 +1,290 @@
...
+  setSuccess: setMockUseGetEntityBundleSuccess,
+} = getUseQueryMock<Partial<EntityBundle>, SynapseClientError>()
+
+const { mock: useGetEntityDOIMockImpl, setSuccess: setEntityDOISuccess } =
+  getUseQueryMock<DoiAssociation | null, SynapseClientError>()
+
+const { mock: useGetProjectDOIMockImpl, setSuccess: setProjectDOISuccess } =
+  getUseQueryMock<DoiAssociation | null, SynapseClientError>()
+
+const fileWithAnnotations: Partial<EntityBundle> = {
```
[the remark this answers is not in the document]
kianamcc: This actually wasn’t copied from another test file. I had brain lag and for some reason was focused on annotations here, haha.

packages/synapse-react-client/src/components/EntityCitation/EntityCitation.tsx
kianamcc: Removed

packages/synapse-react-client/src/components/DropdownSelect/DropdownSelect.tsx
```diff
@@ -18,8 +26,12 @@ export type DropdownSelectProps = ButtonGroupProps & {
   selectedIndex?: number
   setSelectedIndex?: (selectedIndex: number) => void
   onButtonClick?: (selectedIndex: number) => void
+  buttonText?: string
   variant?: ButtonProps['variant']
   buttonGroupAriaLabel?: string
+  sx?: SxProps
+  anchorRef?: RefObject<HTMLElement>
```
nickgros, line 33: I think this would avoid your need to cast further down
```suggestion
  anchorRef?: RefObject<HTMLDivElement>
```

packages/synapse-react-client/src/components/DropdownSelect/DropdownSelect.tsx
```diff
@@ -88,7 +107,16 @@ export default function DropdownSelect(props: DropdownSelectProps) {
...
           }}
         >
-          {options[selectedIndex ?? selectedIndexLocal]}
+          {buttonText ? (
+            <Typography
+              variant="label"
+              sx={{ lineHeight: '20px', fontSize: '16px' }}
+            >
+              {buttonText}
+            </Typography>
```
nickgros, line 116: What's wrong with the default button text style? Why would we only want to change that in this one place?

```suggestion
            buttonText
```

packages/synapse-react-client/src/components/EntityCitation/EntityCitation.tsx
```diff
@@ -0,0 +1,159 @@
...
+              options={options}
+              anchorRef={citationButtonRef}
+              paperProps={{ square: true, elevation: 2 }}
+              sx={{
+                fontSize: '16px',
+                lineHeight: '20px',
+                '.MuiButtonBase-root': {
+                  padding: '4px 16px',
+                },
+                width: { xs: '100%', sm: 'initial' },
```
nickgros, line 115: Are these styles specific to this button or are they related to the button's placement in the SWC Project Header menu? Would it make more sense for SWC to add some of these style overrides?

packages/synapse-react-client/src/components/EntityCitation/EntityCitation.test.tsx
```diff
@@ -0,0 +1,279 @@
...
+} = getUseQueryMock<Partial<EntityBundle>, SynapseClientError>()
+
+const { mock: useGetEntityDOIMockImpl, setSuccess: setEntityDOISuccess } =
+  getUseQueryMock<DoiAssociation | null, SynapseClientError>()
+
+const { mock: useGetProjectDOIMockImpl, setSuccess: setProjectDOISuccess } =
+  getUseQueryMock<DoiAssociation | null, SynapseClientError>()
+
+const doiEntitySuccess: DoiAssociation = {
+  objectId: 'syn61841662',
```
nickgros, line 47: It would be much more clear if we replaced these IDs with a variable describing what they represent

const mockEntityWithUnversionedDoiId = 'syn61841662'
```suggestion

const doiEntitySuccess: DoiAssociation = {
  objectId: mockEntityWithUnversionedDoiId,
```

packages/synapse-react-client/src/components/EntityCitation/EntityCitation.test.tsx
```diff
@@ -0,0 +1,279 @@
...
+    await openPopover('Cite page')
+
+    await waitFor(() => {
+      screen.findByRole('dialog', {
+        name: /Citation options/i,
+      })
+      expect(
+        screen.getByText(content => content.includes('version=1')),
+      ).toBeInTheDocument()
+    })
```
nickgros, line 248: Either use `await screen.findBy..` outside of a `waitFor`, or use `getBy` in the `waitFor` (`findBy` is just a convenience wrapper over `await waitFor(() => getBy...)`)

In this case, `findBy` returns a promise that might not reject until `waitFor` has finished.

```suggestion
    await waitFor(() => {
      screen.getByRole('dialog', {
        name: /Citation options/i,
      })
      screen.getByText(content => content.includes('version=1'))
    })
```

packages/synapse-react-client/src/components/EntityCitation/EntityCitationUtils.ts
kianamcc: Ah I see, good to know!

packages/synapse-react-client/src/components/DropdownSelect/DropdownSelect.tsx
```diff
@@ -75,8 +93,9 @@ export default function DropdownSelect(props: DropdownSelectProps) {
     <>
       <ButtonGroup
         variant={variant}
-        ref={anchorRef}
+        ref={anchorRef as RefObject<HTMLDivElement>}
```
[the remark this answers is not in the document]
kianamcc: I think I set up the prop wrong. The prop type should be RefObject<HTMLDivElement> but I had it as RefObject<HTMLElement> which button group doesn't accept as it's ref.

packages/synapse-react-client/src/components/DropdownSelect/DropdownSelect.tsx
kianamcc: Changing the style in the popover component where it's used instead.

packages/synapse-react-client/src/components/EntityCitation/EntityCitation.tsx
```diff
@@ -0,0 +1,159 @@
...
+
+  return (
+    <>
+      {(projectDoiAssociation || entityDoiAssociation) && (
+        <>
+          {isProjectAndEntityDoi ? (
+            <DropdownSelect
+              options={options}
+              anchorRef={citationButtonRef}
+              paperProps={{ square: true, elevation: 2 }}
```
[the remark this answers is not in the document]
kianamcc: I wanted it to look closer to the design: Screenshot 2025-06-17 at 11 30 11 AM

I will remove this though!

packages/synapse-react-client/src/components/EntityCitation/EntityCitation.tsx
kianamcc: I moved these styles to SWC. I left 'width' alone as that is just for mobile responsiveness.

packages/synapse-react-client/src/components/EntityCitation/EntityCitation.tsx
```diff
@@ -0,0 +1,170 @@
...
+          }}
+          variant="outlined"
+          startIcon={<DoubleQuotes width={18} height={18} />}
+        >
+          <Typography
+            variant="label"
+            sx={{ lineHeight: '20px', fontSize: '16px' }}
+          >
+            {buttonText}
+          </Typography>
```
nickgros, line 162: Sorry, to clarify, I don't think we need this special button styling at all--I think design would prefer that the text on this button has the same style as all of the other buttons--any specific styles that are different in the Figma for this feature were probably not intentional (but let me know if I am wrong!)

```suggestion
           {buttonText}
```

packages/synapse-react-client/src/components/EntityCitation/EntityCitation.test.tsx
```diff
@@ -0,0 +1,259 @@
...
+        request.id === mockProjectWithDoiId &&
+        request.version === undefined
+      ) {
+        return useGetProjectDOIMockImpl()
+      } else if (
+        request.id === mockProjectWithNoDoiId &&
+        request.version === undefined
+      ) {
+        return useGetProjectDOIMockImpl()
+      }
```
nickgros, line 133: Since you're setting the mock response body in each test, isn't this the same thing, but more simple?
```suggestion
      if (request.id === mockEntityWithVersionedDoiId) {
        return useGetEntityDOIMockImpl()
      } else if (request.id === mockProjectWithDoiId) {
        return useGetProjectDOIMockImpl()
      }
```

packages/synapse-react-client/src/components/LinkWithIcon/LinkWithIcon.tsx
```diff
@@ -0,0 +1,59 @@
...
+      }}
+    >
+      <Typography
+        variant="buttonLink"
+        sx={{
+          fontWeight: 700,
+        }}
+      >
+        {text}
+      </Typography>
```
nickgros, line 54: This should already get the correct style by being in a `Button`, no?
```suggestion
      {text}
```

packages/synapse-react-client/src/components/LinkWithIcon/LinkWithIcon.tsx
```diff
@@ -0,0 +1,59 @@
...
+import IconSvg, { IconName } from '../IconSvg'
+
+export type LinkWithIconProps = {
+  text: string
+  icon?: IconName
+  sx?: SxProps
+  href: string
+}
+
+const LinkWithIcon = (props: LinkWithIconProps) => {
```
nickgros, line 12: Do we need this for SWC? thought we were planning to remove it and just use the builtin `Button`

packages/synapse-react-client/src/components/EntityCitation/EntityCitation.tsx
kianamcc: So we don't want to use MUI Typography? To match the styling of the dropdown from the ActionMenu on synapse.org where this dropdown is placed next to, I should just do the styling in _core.scss in SWC?

packages/synapse-react-client/src/components/LinkWithIcon/LinkWithIcon.tsx
kianamcc: removed component

packages/synapse-react-client/src/components/EntityCitation/EntityCitation.tsx
kianamcc: Removed styling, will add styling in synapse.org as needed.

nickgros: Shouldn't need to use Typography inside of a button. If the goal is style consistency with other items in a menu, then that styling should be co-located with the menu styles, not individual buttons.

PR #1969, opened 2025-06-16
PORTALS-3351 - Custom matcher for React component mocks/spies

nickgros:
Create a [custom matcher](https://vitest.dev/guide/extending-matchers) for checking whether React components have rendered.

The motivation for this change is that the details of how/when components are rendered is abstracted by React and subject to change with React upgrades. For instance, these assertions all break on upgrade to React 19. Doing this refactor here will limit the scope of required changes upon React 19 upgrade (essentially making it a 2-line change instead of changing every test file touched in this PR).

This custom matcher is also more expressive, so our test code can more clearly indicate when an assertion is checked against a component.

### DIFF
--- a/packages/synapse-react-client/src/testutils/ComponentToBePassedPropsCustomMatcher.ts
+++ b/packages/synapse-react-client/src/testutils/ComponentToBePassedPropsCustomMatcher.ts
@@ -1,0 +1,60 @@
+import { Mock } from 'vitest'
+
+/**
+ * Custom matchers for testing whether mock/spy React components have been rendered with specific props.
+ */
+
+expect.extend({
+  toHaveBeenRendered(mockComponent: Mock) {
+    try {
+      expect(mockComponent).toHaveBeenCalled()
+    } catch (error) {
+      return {
+        pass: false,
+        message: () => `Component was not rendered:\n  ${error.message}`,
+      }
+    }
+    return {
+      pass: true,
+      message: () => 'Component was rendered',
+    }
+  },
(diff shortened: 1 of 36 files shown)


packages/synapse-react-client/src/testutils/ComponentToBePassedPropsCustomMatcher.ts
```diff
@@ -0,0 +1,60 @@
...
+    return {
+      pass: true,
+      message: () => 'Component was rendered',
+    }
+  },
+  toHaveBeenRenderedWithProps(mockComponent: Mock, expectedProps: unknown) {
+    try {
+      expect(mockComponent).toHaveBeenCalledWith(
+        expectedProps,
+        expect.anything(),
```
nickgros, line 26: In React 19, this second argument will always be `undefined`. On upgrade, we will just have to change 2 lines in this file, instead of every single test file touched in this PR.

packages/synapse-react-client/src/testutils/vitest.d.ts
```diff
@@ -0,0 +1,12 @@
+/* eslint-disable @typescript-eslint/no-empty-object-type,@typescript-eslint/no-empty-interface */
+import 'vitest'
+
+interface CustomMatchers<R = unknown> {
+  toHaveBeenRendered: () => R
+  toHaveBeenRenderedWithProps: (props: unknown) => R
+  toHaveBeenLastRenderedWithProps: (props: unknown) => R
```
nickgros, line 7: Type augmentation so TypeScript is aware of our custom matchers.

jay-hodgson (approved):

PR #1972, opened 2025-06-17
Skip flaky SynapseTable tests

nickgros:


kianamcc (approved):
SynapseTable test timing out

PR #1984, opened 2025-06-19
PORTALS-3588 - Lazy load routes in STOP-AD Portal

nickgros:


apps/portals/stopadportal/src/config/routesConfig.tsx
```diff
@@ -15,75 +9,38 @@ const routes: RouteObject[] = [
...
-              formTitle="Your Submission"
-              formClass="drug-upload-tool"
-            />
-          </SectionLayout>
-        ),
+        lazy: () =>
+          import('@/pages/FormSubmissionPage').then(convertModuleToRouteObject),
       },
       {
         path: 'Help/How It Works',
```
kianamcc, line 27: I feel like this would be easier to read and maintain if we group these 'Help' routes under a parent 'Help' route. We can avoid the repeating prefix and keep related routes together. What do you think?

apps/portals/stopadportal/src/config/routesConfig.tsx
```diff
@@ -1,37 +1,57 @@
...
 import { convertModuleToRouteObject } from '@sage-bionetworks/synapse-portal-framework/utils/convertModuleToRouteObject'
-import { RouteObject } from 'react-router'
+import { Navigate, RouteObject } from 'react-router'
 
 const routes: RouteObject[] = [
   {
     path: '/',
     element: <App />,
     children: [
+      ...sharedRoutes,
```
nickgros, line 11: Noticed this was missing

kianamcc (approved):

PR #1990, opened 2025-06-23
SWC-7381

jay-hodgson:


packages/synapse-react-client/src/components/UserCard/UserCardMedium.tsx
```diff
@@ -101,9 +102,9 @@ export function UserCardMedium({
   let name = ''
   const linkLocation = link
     ? link
-    : `${PRODUCTION_ENDPOINT_CONFIG.PORTAL}Profile:${userProfile.ownerId}`
+    : `${PRODUCTION_ENDPOINT_CONFIG.PORTAL}Profile:${ownerId}`
   // linkLocation is overriden by custom click handler
-  const email = `${userName}@synapse.org`
+  const email = userName ? `${userName}@synapse.org` : undefined
```
nickgros, line 107: Could we update the type definition, so we can catch all of the other places where we may have a similar error?

https://github.com/Sage-Bionetworks/synapse-web-monorepo/blob/fdcfcc8909d7f5625e55bc0997700b7e168c9a6e/packages/synapse-types/src/UserProfile.ts#L27

packages/synapse-react-client/src/components/UserCard/Avatar.tsx
```diff
@@ -91,9 +93,7 @@ export function Avatar({
   let content: JSX.Element | string = <></>
 
   if (!hasProfileImage) {
-    content = userProfile.firstName
-      ? userProfile.firstName[0]
-      : userProfile.userName[0]
+    content = firstName ? firstName[0] : userName[0]
```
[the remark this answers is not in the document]
jay-hodgson: Yeah, a number is odd.  Better than an error, but still not ideal

packages/synapse-react-client/src/components/UserCard/UserCardMedium.tsx
jay-hodgson: Absolutely!  I thought I made this change, but I accidentally made the change to the output UserProfile.d.ts file instead

PR #1991, opened 2025-06-23
PORTALS-3660

jay-hodgson:


jay-hodgson:
fyi @Sigfried

jay-hodgson:
This is to temporarily reference the "draft" (latest) value of the source table.  Before releasing the staging site to production, we should lock this down to a versioned snapshot

PR #1995, opened 2025-06-24
SWC-6990 - Optimize bundle size

nickgros:
- Lazy-load each React component, so Vite will code-split and reduce the initial bundle size
- Remove unused React components from SWC bundle

### DIFF
--- a/packages/synapse-react-client/src/SWC.index.ts
+++ b/packages/synapse-react-client/src/SWC.index.ts
@@ -5,68 +5,16 @@
  * If you wish to export a component or other object to be used by SynapseWebClient, you should add it to this file.
  */
 
-import UserAccessRequestHistoryPlace from '@/components/dataaccess/UserAccessRequestHistory/UserAccessRequestHistoryPlace'
-import { EntityUploadModal } from './components/EntityUpload/EntityUploadModal'
-import { SkeletonButton } from './components/Skeleton/SkeletonButton'
-import { AccountLevelBadges } from './components/AccountLevelBadges/AccountLevelBadges'
-import ChangePassword from './components/ChangePassword/ChangePassword'
-import { ReviewerDashboard } from './components/dataaccess/ReviewerDashboard'
-import { FolderDownloadConfirmation } from './components/download_list/FolderDownloadConfirmation'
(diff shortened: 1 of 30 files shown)


packages/synapse-react-client/src/SWC.index.ts
```diff
@@ -127,88 +47,166 @@ const SynapseContext = {
...
-  EntityAclEditorModal,
-  SynapseChat,
-  RejectProfileValidationRequestModal,
-  GovernanceMarkdownGithub,
-  MarkdownGithubLatestTag,
-  ProjectDataAvailability,
-  EntityUploadModal,
-  CreateOrUpdateDoiModal,
-  UserAccessRequestHistoryPlace,
+  EntityForm: lazy(() => import('./components/EntityForm/EntityForm')),
```
nickgros, line 53: Rewrote all components as `lazy` - Vite will code-split, so the initial JS to download will be much smaller, depending on the page you're on (no more loading the Homepage assets when you never navigate to it!)

jay-hodgson (approved):

packages/synapse-react-client/src/components/AccountLevelBadges/AccountLevelBadges.tsx
```diff
@@ -86,3 +86,5 @@ export function AccountLevelBadges({ userId }: AccountLevelBadgesProps) {
     </>
   )
 }
+
+export default AccountLevelBadges
```
nickgros, line 90: Added `default` export to many component files to simplify dynamic import syntax

packages/synapse-react-client/src/components/SynapseHomepageV2/SynapseHomepageV2.tsx
```diff
@@ -274,7 +234,7 @@ export function SynapseHomepageV2({ gotoPlace }: SynapseHomepageV2Props) {
           </Box>
           {isDesktopView && (
             <Box sx={{ height: '100%', justifySelf: 'end' }}>
-              <Image1 />
+              <img src={image1} alt={''} />
```
nickgros, line 237: Decorative image is OK to have blank alt text, no a11y concern.

PR #1997, opened 2025-06-25
PORTALS-3646: For CCKP - create a Sustainability Scorecard component

kianamcc:
Jira: https://sagebionetworks.jira.com/jira/software/c/projects/PLFM/boards/238?assignee=712020%3Aa7fbbeb0-d7a6-443a-9dd4-246363ff09dd&selectedIssue=PORTALS-3646

Design: https://www.figma.com/design/C7wSn2ayPJBJgdMXdKbWQY/CCKP-Knowledge-Portal-v2?node-id=1753-4093&m=dev

Screenshot 2025-07-03 at 7 53 09 PM

Screenshot 2025-07-03 at 7 53 19 PM

I will create a follow-up ticket for the metric summaries and consult Adam about the design for when a tool has no sustainability data but this is what I have for now:

Screenshot 2025-07-07 at 11 01 00 AM


packages/synapse-react-client/src/components/SustainabilityScorecard/SustainabilityScorecard.tsx
```diff
@@ -0,0 +1,172 @@
...
+                borderRadius: '50%',
+                width: 18,
+                height: 18,
+                display: 'flex',
+                alignItems: 'center',
+                justifyContent: 'center',
+                marginLeft: 'auto',
+              }}
+            >
+              <CheckIcon
```
kianamcc, line 68: I was using check circle icon originally but realized I am unable to have the green background + white check combo like in the design because the check mark was cut out of the circle and so shows the background of whatever is behind it. Not sure what was the best way to approach this.

packages/synapse-react-client/src/components/SustainabilityScorecard/SustainabilityScorecardSummary.tsx
```diff
@@ -0,0 +1,191 @@
...
+  metricsConfig,
+  scoreDescriptorColumnName,
+  searchParamKey,
+  filterColumn,
+}: SustainabilityScorecardSummaryProps) => {
+  const theme = useTheme()
+  const [searchParams] = useSearchParams()
+  const searchValue = searchParams.get(searchParamKey)
+
+  console.log('entityId', entityId)
```
kianamcc, line 89: Whoops, let me clean console logs up.

packages/synapse-react-client/src/components/SustainabilityScorecard/SustainabilityScorecard.tsx
nickgros: I'll have to pull down the code to take a look so I'll get back to you. At first look, I don't see any issues with what you have here.

packages/synapse-react-client/src/components/CardContainerLogic/CardContainerLogic.tsx
```diff
@@ -252,6 +252,7 @@ export function CardContainerLogic(props: CardContainerLogicProps) {
             cardConfiguration={props.cardConfiguration}
             initialLimit={props.initialLimit}
             multiCardList={props.multiCardList}
+            searchParams={props.searchParams}
```
nickgros, line 255: Why do you need to pass this here? Think this could be removed

packages/synapse-react-client/src/components/SustainabilityScorecard/SustainabilityScorecardSummary.tsx
```diff
@@ -0,0 +1,185 @@
...
+const SustainabilityScorecardSummary = ({
+  entityId,
+  text,
+  metricsConfig,
+  scoreDescriptorColumnName,
+  searchParamKey,
+  filterColumn,
+}: SustainabilityScorecardSummaryProps) => {
+  const theme = useTheme()
+  const [searchParams] = useSearchParams()
```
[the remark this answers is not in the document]
kianamcc: By application/page, in this case, would it be the ToolsDetailsPage in CCKP? I would create the query here and pass it into a sustainabilityScorecard prop in CardContainerLogic then have it passed down to HeaderCard.

nickgros (approved):
Looks great!

PR #2000, opened 2025-06-26
SWC-7384

nickgros:
- Add UI to inform users MFA is required soon, with "skip" functionality
- When MFA_REQUIRED feature flag is enabled, "skip" functionality will be hidden
- Fix bad feature flag API path in synapse-react-client

Before MFA_REQUIRED is enabled: 

image

Once enabled:

image

### DIFF
--- a/apps/SageAccountWeb/src/hooks/useSkipMfaPrompt.test.ts
+++ b/apps/SageAccountWeb/src/hooks/useSkipMfaPrompt.test.ts
@@ -1,0 +1,252 @@
+import { renderHook, act } from '@testing-library/react'
+import { vi, describe, it, expect, beforeEach, afterEach } from 'vitest'
+import dayjs, { Dayjs } from 'dayjs'
+import { when } from 'vitest-when'
+import { useSkipMfaPrompt } from './useSkipMfaPrompt'
+import { useLocalStorageValue } from '@react-hookz/web'
+
+vi.mock('@react-hookz/web', () => ({
+  useLocalStorageValue: vi.fn(),
+}))
+
+vi.mock('dayjs', async importOriginal => {
+  return {
+    ...(await importOriginal()),
+    default: vi.fn(() => ({
+      subtract: vi.fn().mockReturnThis(),
+      isAfter: vi.fn(),
+    })),
+  }
+})
+
+const mockUseLocalStorageValue = vi.mocked(useLocalStorageValue)
+const mockDayjs = vi.mocked(dayjs)
+
+describe('useSkipMfaPrompt', () => {
+  const mockAccountId = '123'
+  const mockSetTimestamp = vi.fn()
+  const mockUseLocalStorageValueReturn = {
+    value: null,
+    set: mockSetTimestamp,
+    remove: vi.fn(),
+    fetch: vi.fn(),
+  }
+  // The dayjs mock for the current time should always return a consistent instance
+  const mockDayjsInstanceForCurrentTime = {
+    subtract: vi.fn().mockReturnThis(),
+  }
+
+  beforeEach(() => {
+    vi.clearAllMocks()
+    // Mock console.error to suppress error logs in tests
+    vi.spyOn(console, 'error').mockImplementation(() => {})
+
+    when(mockDayjs)
+      .calledWith() // No arguments for the current time
+      .thenReturn(mockDayjsInstanceForCurrentTime as unknown as Dayjs)
+  })
+
+  afterEach(() => {
+    vi.restoreAllMocks()
+  })
+
+  describe('hasSkippedRecently', () => {
+    it('should return false when no timestamp is stored', () => {
+      mockUseLocalStorageValue.mockReturnValue({
+        ...mockUseLocalStorageValueReturn,
+        value: null,
+      })
+
+      const { result } = renderHook(() => useSkipMfaPrompt(mockAccountId))
+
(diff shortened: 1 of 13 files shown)


apps/SageAccountWeb/src/hooks/useSkipMfaPrompt.test.ts
```diff
@@ -0,0 +1,251 @@
...
+      subtract: vi.fn().mockReturnThis(),
+      isAfter: vi.fn(),
+    })),
+  }
+})
+
+const mockUseLocalStorageValue = vi.mocked(useLocalStorageValue)
+const mockDayjs = vi.mocked(dayjs)
+
+describe('useSkipMfaPrompt', () => {
```
nickgros, line 25: Tests were originally written by Claude Sonnet 4 agent mode, reviewed & revised by me

apps/SageAccountWeb/src/hooks/useSkipMfaPrompt.test.ts
```diff
@@ -0,0 +1,251 @@
...
+        value: mockTimestamp,
+      })
+
+      const mockDayjsInstanceForStoredTimestamp = {
+        isAfter: vi.fn().mockReturnValue(true),
+      }
+
+      when(mockDayjs)
+        .calledWith(mockTimestamp)
+        .thenReturn(mockDayjsInstanceForStoredTimestamp as unknown as Dayjs)
```
nickgros, line 88: I thought using `vitest-when` to write a declarative/mockito-style mock was easier to write/read here. Just seems so much less error-prone & forces you to think about how the mock is being used compared to writing imperative conditional logic in a mock.

apps/SageAccountWeb/src/hooks/useSkipMfaPrompt.ts
```diff
@@ -0,0 +1,44 @@
...
+ * skips MFA.
+ *
+ * This is temporary code that can be deleted once 2FA is required for all users.
+ * @param accountId
+ */
+export function useSkipMfaPrompt() {
+  const { value: lastSkippedTimestamp, set: setTimestamp } =
+    useLocalStorageValue<string>(
+      SKIPPED_2FA_REQUIREMENT_TIMESTAMP_LOCALSTORAGE_KEY,
+    )
```
nickgros, line 27: I originally wanted to include the account ID in the localstorage key, so you would be prompted for each account that logs in to the same machine. I ended up running into issues, I think race conditions w/ how `useLocalStorageValue` worked, so I ended up not doing that.

packages/synapse-react-client/src/synapse-client/SynapseClient.ts
```diff
@@ -5313,7 +5313,7 @@ export function getPortalFileHandleServletUrl(
  */
 export const getFeatureFlags = () => {
   return doGet<FeatureFlags>(
-    `Portal/featureflags`,
+    `/Portal/featureflags`,
```
nickgros, line 5316: in the accounts app in development mode, pointing to production, we were making calls to `https://synapse.orgPortal/featureflags`, which threw an error. It seems like this actually works in real production/staging, so not sure why I was seeing it, but I would rather have 1-2 slashes than 0-1. It will be obvious if the MFA_REQUIRED feature flag breaks in staging, so this can be easily reverted if it causes problems.

apps/SageAccountWeb/src/pages/MfaRequiredPage.tsx
```diff
@@ -1,13 +1,43 @@
...
+const MFA_REQUIRED_DESCRIPTION =
+  'To enhance account security, two-factor authentication (2FA) is mandatory for all users. This extra layer ' +
+  'of protection helps keep your data safe by requiring both your password and a verification code sent to your device.'
 
-function MfaRequiredPage(props: mfaRequiredPageProps) {
+function MfaRequiredPage() {
+  const isMfaRequired = useGetFeatureFlag(FeatureFlagEnum.MFA_REQUIRED)
   const navigate = useNavigate()
 
+  // START TEMPORARY CODE: This code can be removed once MFA is required for all users.
```
jay-hodgson, line 25: You could point to reference SWC-7385 in the comment if you'd like

apps/SageAccountWeb/src/pages/TwoFactorAuthBackupCodesPage.tsx
```diff
@@ -10,10 +11,23 @@ function TwoFactorAuthBackupCodesPage() {
...
-      onClose={() => navigate('/authenticated/myaccount')}
+      onClose={() => {
+        // This is the last step in 2FA enrollment. Once the user is done, take them to the page they were on before being
+        // redirected to start 2FA enrollment.
+        // If the user was not originally redirected (e.g. the user voluntarily regenerated backup codes), redirect to the
+        // account settings page.
+        const redirectToAfterComplete =
+          pathBefore2faRedirect || '/authenticated/myaccount'
+        clearPathBefore2faRedirect()
+        navigate(redirectToAfterComplete)
```
jay-hodgson, line 29: Thank you for fixing this bug in the passthrough enhancement!

jay-hodgson:
Looks great.  By utilizing localstorage to capture the skip, and a timestamp, you're been able to maintain the UX workflow and introduce the requested (temporary) feature.  Thanks Nick!

PR #2011, opened 2025-07-01
SWC-7351 - Add a CRDT data grid

afwillia:
Add a CRDT data grid that establishes a websocket connection with a Synapse server via the Grid Services API.

### DIFF
--- a/packages/synapse-react-client/src/components/DataGrid/DataGrid.tsx
+++ b/packages/synapse-react-client/src/components/DataGrid/DataGrid.tsx
@@ -1,0 +1,705 @@
+import { useState, useRef, useEffect, useMemo } from 'react'
+import {
+  DataSheetGrid,
+  keyColumn,
+  Column,
+  createTextColumn,
+} from 'react-datasheet-grid'
+import 'react-datasheet-grid/dist/style.css'
+import '../../style/components/_data-grid-extra.scss'
+import { useSynapseContext } from '@/utils/context/SynapseContext'
+import { SynapseClient } from '@sage-bionetworks/synapse-client'
+import { useCreateGridSession } from './useCreateGridSession'
+import {
+  CreateGridRequest,
+  PostRepoV1GridSessionIdReplicaRequest,
+} from '@sage-bionetworks/synapse-client'
+import { Model, VecApi } from 'json-joy/lib/json-crdt'
+import { encode, decode } from 'json-joy/lib/json-crdt-patch/codec/compact'
+import { konst } from 'json-joy/lib/json-crdt-patch'
+import throttle from 'lodash.throttle'
+import { Encoder as VerboseEncoder } from 'json-joy/lib/json-crdt/codec/structural/verbose/Encoder'
+
+const verboseEncoder = new VerboseEncoder()
+
+const DataGrid = () => {
+  const synapseClient = useSynapseContext().synapseClient
+
+  // Grid session state
+  const [sessionId, setSessionId] = useState<string>('')
+  const [replicaId, setReplicaId] = useState<number | null>(null)
+  const [presignedUrl, setPresignedUrl] = useState<string>('')
+  const [gridSql, setGridSql] = useState<string>('')
+  const inputRef = useRef<HTMLInputElement>(null)
+  const webSocketRef = useRef<WebSocket | null>(null)
+
+  // WebSocket state
+  const [isConnected, setIsConnected] = useState<boolean>(false)
+  const [connectionStatus, setConnectionStatus] =
+    useState<string>('Disconnected')
+  const [messages, setMessages] = useState<string[]>([])
+
+  // CRDT state
+  const sequenceNumberRef = useRef<number>(1)
+  const modelRef = useRef<Model | null>(null)
+  type ModelColumnNames = string
+  type ModelColumnOrder = number
+  type ModelRow = string
+  type ModelSnapshot = {
+    columnNames: ModelColumnNames[]
+    columnOrder: ModelColumnOrder[]
+    rows: ModelRow[]
+  }
+  const modelSnapshotRef = useRef<ModelSnapshot>({
+    columnNames: [],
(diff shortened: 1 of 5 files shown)


packages/synapse-client/src/spec/openapispecification.json
```diff
@@ -19018,9 +19053,8 @@
             "type": "string",
             "enum": ["org.sagebionetworks.repo.model.grid.CreateGridRequest"]
           },
-          "csvFileHandleId": {
-            "type": "integer",
-            "format": "int32"
+          "initialQuery": {
```
afwillia, line 19056: Need this change to avoid a `CreateGridRequest` type-check error

packages/synapse-react-client/src/components/DataGrid/DataGrid.tsx
```diff
@@ -0,0 +1,695 @@
...
+  const modelSnapshotRef = useRef<ModelSnapshot>({
+    columnNames: [],
+    columnOrder: [],
+    rows: [],
+  })
+  const getModel = () => modelRef.current
+  const setModel = (newModel: Model) => {
+    if (newModel.view() !== modelRef.current?.view()) {
+      modelRef.current = newModel
+      modelSnapshotRef.current = modelRef.current.api.getSnapshot()
```
afwillia, line 64: The model itself will change with every patch, but the snapshot contains the essential information for rendering the grid - `columnNames`, `columnOrder`, and `rows`. So the snapshot should be computed each model update, but re-renders should only happen when the snapshot changes.

packages/synapse-react-client/src/components/DataGrid/DataGrid.tsx
```diff
@@ -0,0 +1,695 @@
...
+          createTextColumn({ continuousUpdates: false }),
+        ),
+        title: columnNames[index],
+      }
+    })
+    return gridCols
+  }
+
+  // Function to apply grid changes to a model
+  function gridToModel(gridRows: DataGridRow[], model: Model): Model {
```
afwillia, line 170: Model updates must use the API so that the clock is updated and patches can be computed.

packages/synapse-react-client/src/components/DataGrid/DataGrid.tsx
```diff
@@ -0,0 +1,695 @@
...
+      webSocketRef.current.send(messageToSend)
+      console.log('WebSocket message sent:', messageToSend)
+      setMessages(prev => [`Sent: ${messageToSend}`, ...prev])
+    } else {
+      console.error('WebSocket is not connected')
+      setMessages(prev => ['Failed to send: WebSocket not connected', ...prev])
+    }
+  }
+
+  const handleWebSocketMessage = (latestMsg: jsonJoyMessage) => {
```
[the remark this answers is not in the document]
afwillia: The server will also send `8, new-patch`, which client will need to respond by sending a `synchronize-clock` message following the pattern above.

packages/synapse-react-client/src/components/DataGrid/DataGrid.tsx
```diff
@@ -0,0 +1,695 @@
...
+    const trimmedInput = input.trim()
+    if (!trimmedInput) {
+      return { type: 'empty', input: '' }
+    } else if (trimmedInput.toUpperCase().startsWith('SELECT')) {
+      return { type: 'sql', input: trimmedInput }
+    } else if (/^N\w*=$/.test(trimmedInput)) {
+      return { type: 'sessionId', input: trimmedInput }
+    }
+    return { type: 'unknown', input: trimmedInput }
+  }
```
nickgros, line 113: Static utilities could be moved into a new file `DataGridUtils.ts` which would simplify this component, and those utilities could be tested in isolation.

packages/synapse-react-client/src/components/DataGrid/DataGrid.tsx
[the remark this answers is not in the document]
nickgros: Let's move this to the proposed `DataGridWebSocket` class. This refactored function could invoke callbacks that update the model/rows/columns as necessary.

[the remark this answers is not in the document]
nickgros: Does the `latestMsg[0]` correspond to [the list of 9 types of messages on this page](https://jsonjoy.com/specs/json-rx/introduction)? Can we create an `enum` to represent these for readability?

packages/synapse-react-client/src/components/DataGrid/DataGrid.tsx
```diff
@@ -0,0 +1,695 @@
...
+              if (updatedRowIds.has(rowData._rowId)) {
+                return 'row-updated'
+              } else return ''
+            }}
+            createRow={addRowToModel}
+            duplicateRow={({ rowData }: any) => ({
+              ...rowData,
+              _rowId: genId(),
+            })}
+            onChange={(newValue: any, operations: any) => {
```
nickgros, line 613: Do these have better types than `any` provided by react-datasheet-grid?

Can we pull this function out of the JSX and define it before the `return`?

nickgros (changes requested):

nickgros (approved):

packages/synapse-react-client/src/components/DataGrid/useCRDTState.ts
```diff
@@ -0,0 +1,22 @@
+import { Model } from 'json-joy/lib/json-crdt'
+import { useRef, useState } from 'react'
+import { ModelSnapshot } from './DataGridTypes'
+
+export function useCRDTState() {
+  // CRDT state
+  const modelRef = useRef<Model | null>(null)
```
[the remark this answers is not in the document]
afwillia: I was reluctant to make it a state variable at first because it is changes very often but the snapshot and/or visible data don't. We should benchmark this if it becomes an issue.

PR #2014, opened 2025-07-02
Rerun `msw init` to fix logged errors in Storybook

nickgros:


packages/synapse-react-client/public/mockServiceWorker.js
```diff
@@ -2,25 +2,26 @@
 /* tslint:disable */
 
 /**
- * Mock Service Worker (1.3.2).
+ * Mock Service Worker.
```
nickgros, line 5: Auto-generated file

packages/synapse-react-client/package.json
```diff
@@ -253,6 +253,8 @@
     ]
   },
   "msw": {
-    "workerDirectory": "public"
+    "workerDirectory": [
+      "public"
+    ]
```
nickgros, line 258: Script automatically changed this

PR #2024, opened 2025-07-08
PORTALS-3651: Portals FileEntity Page: Header card

kianamcc:
Jira: https://sagebionetworks.jira.com/browse/PORTALS-3651

Screenshot 2025-07-08 at 12 17 29 PM

### DIFF
--- a/apps/synapse-portal-framework/src/pages/FileEntityPage/FileEntityPage.tsx
+++ b/apps/synapse-portal-framework/src/pages/FileEntityPage/FileEntityPage.tsx
@@ -8,6 +8,7 @@ import { DetailsPageSectionLayoutType } from '@/components/DetailsPage/DetailsPa
 import { SynapseSpinner } from 'synapse-react-client/components/LoadingScreen/LoadingScreen'
 import DetailsPageMenu from '../../components/DetailsPageMenu'
 import DetailsPageLayout from '@/components/DetailsPageLayout'
+import FileEntityPageHeader from './FileEntityPageHeader'
 
 function FileEntityPage() {
   const searchParams = useGetPortalComponentSearchParams()
@@ -22,8 +23,11 @@ function FileEntityPage() {
     {
       includeEntity: true,
       includeRootWikiId: true,
+      includeDOIAssociation: true,
     },
   )
+
+  console.log('entityBundle', entityBundle)
 
   const fileEntityPageSections = [
     entityBundle?.rootWikiId && { id: 'wiki', title: 'Wiki' },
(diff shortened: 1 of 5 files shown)


apps/synapse-portal-framework/src/pages/FileEntityPage/FileEntityPageHeader.tsx
```diff
@@ -0,0 +1,75 @@
...
+import { EntityBundle } from '@sage-bionetworks/synapse-types'
+import { IconSvg } from 'synapse-react-client'
+import { FileEntity } from '@sage-bionetworks/synapse-types'
+import CitationPopover from 'synapse-react-client/components/CitationPopover'
+
+const FileEntityPageHeader = ({ data }: { data: EntityBundle }) => {
+  const entity = data.entity as FileEntity
+
+  return (
+    <Box
```
nickgros, line 11: Would it make sense to use the `HeaderCard` component here? Or could we extract some common styles/components? I think all of the table logic has been extracted out of HeaderCard, so I would think you could just use it here

packages/synapse-react-client/src/components/CitationPopover/CitationPopover.tsx
```diff
@@ -26,6 +26,7 @@ function CitationPopover(props: CitationProps) {
           onClick={handleClick}
           variant="outlined"
           startIcon={<DoubleQuotes />}
+          sx={{ ...props.buttonSx }}
```
nickgros, line 29:
```suggestion
          sx={props.buttonSx}
```

apps/synapse-portal-framework/src/pages/FileEntityPage/FileEntityPage.tsx
```diff
@@ -38,34 +40,37 @@ function FileEntityPage() {
...
-            flex: 1,
-            gap: '40px',
-          }}
-        >
-          {entityBundle?.rootWikiId && (
-            <CollapsibleSection title="Wiki" id="wiki">
-              <MarkdownSynapse
-                ownerId={entityId}
-                wikiId={entityBundle?.rootWikiId}
+    <>
```
nickgros, line 43: nit: could use this semantic HTML tag
```suggestion
    <article>
```

packages/synapse-react-client/src/components/HeaderCard.tsx
```diff
@@ -118,11 +121,20 @@ const HeaderCard = forwardRef(function HeaderCard(
...
+                {!hideIcon && icon}
+              </Box>
+              <Box
+                sx={{
                   width: '100%',
-                  ...(hideIcon && { display: 'flex' }),
+                  ...((doiUri || hideIcon) && {
+                    display: 'flex',
+                    justifyContent: 'space-between',
+                  }),
```
nickgros, line 137: If you can move the `CitationPopover` component to just slot into the `cardTopButtons`, I think you will probably be able to remove some of these conditional styles.

apps/synapse-portal-framework/src/pages/FileEntityPage/FileEntityPage.tsx
```diff
@@ -30,42 +32,76 @@ function FileEntityPage() {
...
+      <HeaderCard
+        type={type}
+        title={title}
+        description={''}
+        icon={icon}
+        doiUri={doiUri}
+      />
+      <Container>
+        <DetailsPageLayout>
+          <DetailsPageMenu menuSections={fileEntityPageSections} />
```
nickgros, line 78: Taking a look at these again since your last refactor, and I think it might be slightly cleaner to use the `DetailsPageContent` component and pass it a list of sections. It will render the menu and section headers for you, and ensure the style is more consistent with existing details pages.

apps/synapse-portal-framework/src/components/DetailsPageMenu.tsx
```diff
@@ -24,6 +24,10 @@ const DetailsPageMenu = ({
     }
   }
 
+  const filteredMenuSections = menuSections.filter(
```
kianamcc, line 27: I need a component in here alongside the other sections without it actually showing up as an option in the menu (SynapseFileEntityLinkCard)

nickgros: Do all of our existing details page sections have a truthy id and title? If so, then this works

apps/synapse-portal-framework/src/pages/FileEntityPage/FileEntityPage.tsx
```diff
@@ -1,13 +1,13 @@
...
 import { DetailsPageSectionLayoutType } from '@/components/DetailsPage/DetailsPageSectionLayout'
 import { SynapseSpinner } from 'synapse-react-client/components/LoadingScreen/LoadingScreen'
-import DetailsPageMenu from '../../components/DetailsPageMenu'
-import DetailsPageLayout from '@/components/DetailsPageLayout'
+import HeaderCard from 'synapse-react-client/components/HeaderCard'
+import CitationPopover from 'synapse-react-client/components/CitationPopover'
+import { DetailsPageContent } from '../../components/DetailsPage/DetailsPageContentLayout'
 import SynapseFileEntityLinkCard from './SynapseFileEntityLinkCard'
 
 function FileEntityPage() {
```
nickgros, line 13: If we need to change the styles (but not the layout/DOM) of the header card, e.g. to give the title a different font size, we could apply styles to this component that target those elements.

nickgros (approved):

PR #2028, opened 2025-07-10
SWC-7385: remove temp code and use of MFA_REQUIRED feature flag

jay-hodgson:


jay-hodgson:
After changes, pointing to staging logging in with account that does not have 2FA enabled:
Redirected, no option to skip
Screenshot 2025-07-10 at 3 08 39 PM

Account Settings page after enabling, no option to turn off
Screenshot 2025-07-10 at 3 09 56 PM

PR #2036, opened 2025-07-14
PORTALS-3412 - DOI tombstone for Portals

nickgros:
image

### DIFF
--- a/apps/portals/adknowledgeportal/src/pages/StudyDetailsPage/StudyDetailsPage.tsx
+++ b/apps/portals/adknowledgeportal/src/pages/StudyDetailsPage/StudyDetailsPage.tsx
@@ -8,6 +8,7 @@ import {
   STUDY_TABLE_COLUMN_NAMES,
   studyCardConfiguration,
   studyColumnAliases,
+  studyDoiConfiguration,
 } from '@/config/synapseConfigs/studies'
 import {
   DetailsPageTabConfig,
@@ -42,6 +43,7 @@ const studyDetailsPageTabs: DetailsPageTabConfig[] = [
 const studiesDetailsPageProps: DetailsPageProps = {
   sql: studiesSql,
   sqlOperator: ColumnSingleValueFilterOperator.LIKE,
+  portalDOIConfiguration: studyDoiConfiguration,
 }
 
 function StudyDetailsPage() {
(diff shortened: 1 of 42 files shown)


apps/portals/adknowledgeportal/src/pages/StudyDetailsPage/StudyDetailsPage.tsx
```diff
@@ -42,6 +43,7 @@ const studyDetailsPageTabs: DetailsPageTabConfig[] = [
 const studiesDetailsPageProps: DetailsPageProps = {
   sql: studiesSql,
   sqlOperator: ColumnSingleValueFilterOperator.LIKE,
+  portalDOIConfiguration: studyDoiConfiguration,
```
nickgros, line 46: Pass the DOI configuration to the DetailsPage so it can attempt to fetch the DOI

apps/synapse-portal-framework/src/components/DetailsPage/DetailsPage.tsx
```diff
@@ -83,40 +71,34 @@ export default function DetailsPage(props: DetailsPageProps) {
...
-          mt: 5,
-        }}
-      >
-        <Typography variant="headline1" gutterBottom>
-          Coming Soon!
-        </Typography>
-        <Typography variant={'smallText1'} gutterBottom>
-          This {pluralize.singular(name).toLowerCase()} is not yet available,
-          please check back soon.
-        </Typography>
```
nickgros, line 108: Removed this "Coming Soon" view that was shown when an error was encountered. It could be misleading if you followed a link to a resource that was moved or removed.

packages/synapse-react-client/src/components/GenericCard/PortalDOI/PortalDOIUtils.ts
```diff
@@ -33,12 +33,10 @@ export function useShowDoiCardLabel(opts: {
 export function getCandidateDoiId(args: {
   /** Configuration to display a DOI, as well as the ability to create one for users with such permission */
   portalDoiConfiguration?: TableToGenericCardMapping['portalDoiConfiguration']
-  /** Mapping of column name to row data index */
-  schema: Record<string, number>
-  /** The row data */
-  data: string[]
+  /** A record that represents a table row or DOI lookup, mapping from column name to value */
+  data: Record<string, string | null>
 }): string | undefined {
```
nickgros, line 38: In the 'tombstone' case, we don't actually have a row of data. We have search params, which is the direct mapping of column name to cell value.

So I changed this API to accommodate this case. In the existing place where we are using row data to generate this ID, we will transform the data to a record.

packages/synapse-react-client/src/components/error/ErrorPage.tsx
```diff
@@ -4,14 +4,27 @@ import { ReactComponent as UnavailableSvg } from '@/assets/icons/error_page/unav
...
 import SendMessageToEntityOwnerDialog from './SendMessageToEntityOwnerDialog'
+import { DoiObjectType } from '@sage-bionetworks/synapse-client/generated/models/DoiObjectType'
 
 export type ErrorPageProps = {
   type: SynapseErrorType
   message?: string // custom message to report
+  /** @deprecated use `id` and objectType: DoiObjectType.ENTITY */
   entityId?: string
+  /** @deprecated use `version` and objectType: DoiObjectType.ENTITY */
   entityVersion?: number
```
nickgros, line 17: Keep these props to avoid a breaking change that affects SynapseWebClient

jay-hodgson: Is there a ticket to update SWC to use id, objectType and version?

nickgros: I'll create one

apps/synapse-portal-framework/src/components/DetailsPage/DetailsPage.tsx
```diff
@@ -83,40 +71,34 @@ export default function DetailsPage(props: DetailsPageProps) {
...
-          Continue Exploring
-        </Button>
-      </Stack>
+      <ErrorPage
+        type={SynapseErrorType.NOT_FOUND}
+        gotoPlace={() => {}}
+        portalId={portalDOIConfiguration?.portalId}
+        id={candidateDoiId}
+        objectType={DoiObjectType.PORTAL_RESOURCE}
+      />
```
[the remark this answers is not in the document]
nickgros: Adam agreed that no changes to ErrorPage are needed at this time before using it on portals.

packages/synapse-react-client/src/components/error/ErrorPage.tsx
nickgros: Tracked in SWC-7400

PR #2039, opened 2025-07-15
PORTALS-3686, PORTALS-3687

nickgros:
Fix issues in SynapseForm where search params were no longer passed to the component.

We used to always inject the URLSearchParams as props to top-level portals components. This recently changed, breaking these components where these params must now be manually passed.


packages/synapse-react-client/src/components/SynapseForm/useGetFormPropsFromSearchParams.ts
```diff
@@ -0,0 +1,13 @@
+import { useSearchParams } from 'react-router'
+
+export function useGetFormPropsFromSearchParams() {
```
nickgros, line 3: Utility to get these props out of the URL search params

jay-hodgson (approved):

PR #2040, opened 2025-07-16
PORTALS-3693

jay-hodgson:
CCKP home page section updated to use the reusable Goals component
Screenshot 2025-07-16 at 1 18 56 PM

which also comes with a mobile viewport special rendering:
Screenshot 2025-07-16 at 1 19 44 PM

Component on NF (layout changed slightly):
Before:
Screenshot 2025-07-16 at 1 20 27 PM

After:
Screenshot 2025-07-16 at 1 20 51 PM

Component on BSMN:
Before:
Screenshot 2025-07-16 at 1 21 59 PM

After:
Screenshot 2025-07-16 at 1 22 07 PM

Component on dHealth:
Before:
Screenshot 2025-07-16 at 1 22 55 PM

After:
Screenshot 2025-07-16 at 1 23 03 PM


nickgros (approved):
The only potential 'regression' I see is the imbalanced appearance of 3+1 cards on dHealth -- if design is ok with that, then I am too.

PR #2047, opened 2025-07-18
SWC-7406 - Fix style precedence issue

nickgros:
Fix a few regressions introduced in #1921


apps/synapse-portal-framework/src/style/components/_DetailsPage.scss
```diff
@@ -151,12 +151,12 @@ $svg-icon-height: 30px;
   }
 
   .tab-groups + .tab-groups {
+    @include SrcMixins.underline-tabs($inactive-color: unset);
```
nickgros, line 154: fix other case

kianamcc (approved):

PR #2051, opened 2025-07-21
SWC-7410 - Handle error thrown by `registerSchema`

nickgros:
See SWC-7410

### DIFF
--- a/packages/synapse-react-client/src/utils/jsonschema/SchemaAnnotationUtils.test.ts
+++ b/packages/synapse-react-client/src/utils/jsonschema/SchemaAnnotationUtils.test.ts
@@ -1,0 +1,85 @@
+import { hasSchema, registerSchema } from '@hyperjump/json-schema'
+import { renderHook, waitFor } from '@testing-library/react'
+import { JSONSchema7 } from 'json-schema'
+import { useRegisterSchema } from './SchemaAnnotationUtils'
+
+vi.mock('@hyperjump/json-schema', async () => {
+  const actual = await vi.importActual('@hyperjump/json-schema')
+  return {
+    ...actual,
+    hasSchema: vi.fn(),
+    registerSchema: vi.fn(),
+  }
+})
+
+const mockHasSchema = vi.mocked(hasSchema)
+const mockRegisterSchema = vi.mocked(registerSchema)
+
+const MOCK_SCHEMA: JSONSchema7 = {
+  $id: 'https://example.com/test.schema.json',
+  type: 'object',
+  properties: {
+    foo: {
+      type: 'string',
+    },
+  },
+}
+
+describe('useRegisterSchema', () => {
+  beforeEach(() => {
+    vi.clearAllMocks()
+  })
+
+  it('should not register a schema if none is provided', () => {
+    renderHook(() => useRegisterSchema(undefined))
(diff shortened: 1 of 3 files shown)


packages/synapse-react-client/src/utils/jsonschema/SchemaAnnotationUtils.ts
```diff
@@ -86,6 +81,40 @@ export function getDisplayedAnnotationDescription(
...
+      if (!hasSchema(schema.$id)) {
+        try {
+          registerSchema(schema as SchemaObject)
+          setIsRegistered(true)
+        } catch (e) {
+          console.error(
+            'Unable to register JSON Schema in @hyperjump/json-schema:',
+            e,
+          )
+        }
```
nickgros, line 104: This `try`/`catch` handles the error causing the bug.

If this fails, we simply will not use the annotation enhancements (graceful degradation).

packages/synapse-react-client/src/utils/jsonschema/SchemaAnnotationUtils.ts
```diff
@@ -94,22 +123,17 @@ export function getDisplayedAnnotationDescription(
...
   return useQuery<JsonNode | null>({
     queryKey: ['getAnnotatedInstance', json, schema],
     queryFn: () => {
-      if (json && schema) {
-        return annotate(schema.$id!, json)
+      if (json && schema && schemaIsRegistered) {
+        try {
+          return annotate(schema.$id!, json)
+        } catch (e) {
+          console.error('Could not create annotated schema instance:', e)
```
nickgros, line 135: add error handling here for good measure.

packages/synapse-react-client/src/utils/jsonschema/SchemaAnnotationUtils.test.ts
```diff
@@ -0,0 +1,93 @@
...
+  $id: 'https://example.com/test.schema.json',
+  type: 'object',
+  properties: {
+    foo: {
+      type: 'string',
+    },
+  },
+}
+
+describe('useRegisterSchema', () => {
```
nickgros, line 36: Tests for an extracted hook to register a schema with the `@hyperjump/json-schema` store.

packages/synapse-react-client/src/utils/jsonschema/SchemaAnnotationUtils.ts
```diff
@@ -86,6 +81,40 @@ export function getDisplayedAnnotationDescription(
...
+          )
+        }
+      } else {
+        // Already registered, update local state.
+        setIsRegistered(true)
+      }
+    }
+    // TODO: Unregister the schema to avoid memory bloat
+    // This would require syncing with the current state of schemas being used to make sure that we don't unregister a
+    // schema being used by multiple hook instances
```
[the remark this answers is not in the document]
nickgros: Should we create a ticket to track this? I am partial to leaving it as-is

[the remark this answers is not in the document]
jay-hodgson: At the time of writing, this seems very unlikely.  You could open a ticket in Jira that we immediately Close as Won't Do if you think it would be useful for historical context

PR #2054, opened 2025-07-22
PORTALS-3698

jay-hodgson:
@jmvera255 would like to try using an EntityView for the Datasets source instead of a Dataset Collection.


apps/portals/arkportal/src/pages/DatasetDetailsPage.tsx
```diff
@@ -67,8 +67,20 @@ function DatasetDetailsPage() {
...
+                  let versionNumberToUse = 1
+                  if (
+                    hasVersionNumber &&
+                    context.rowData?.versionNumber !== undefined
+                  ) {
+                    if (context.rowData?.versionNumber > 1) {
+                      // If this is the first version, we want to show the files for the latest version
+                      versionNumberToUse = context.rowData?.versionNumber - 1
+                    }
+                  }
```
nickgros, line 81: If the latest snapshot version is deleted, this will probably result in an API error

jay-hodgson: Good call - what if I use the version history (use the first item)?

nickgros: We do a similar thing in the EntityFinder/Dataset Item editor for cases when the user has to select a version. This works!

PR #2055, opened 2025-07-22
PORTALS-3652: Portals FileEntity Page: File properties component

kianamcc:
Jira: https://sagebionetworks.jira.com/jira/software/c/projects/PLFM/boards/238?assignee=712020%3Aa7fbbeb0-d7a6-443a-9dd4-246363ff09dd&selectedIssue=PORTALS-3652

Screenshot 2025-07-21 at 5 09 46 PM


apps/synapse-portal-framework/src/pages/FileEntityPage/SynapseFileEntityPageProperties.tsx
```diff
@@ -0,0 +1,138 @@
...
+    fileEntity?.versionNumber,
+  )
+
+  const fileLocationName = fileHandle
+    ? getStorageLocationName(fileHandle, storageLocationUploadDestination)
+    : ''
+
+  if (!entityBundle || !fileEntity) {
+    return null
+  }
```
nickgros, line 91: One last optional improvement for your consideration:

1. If you passed the ID and version as props instead of the bundle, you could get the entityBundle out of `useGetEntityMetadata`
2. You could then update `useGetEntityMetadata` to return `isLoading` (derived from all of the queries in that hook)
3. You could render a SkeletonTable while this is loading.

nickgros (approved):

PR #2058, opened 2025-07-23
SWC-5439

jay-hodgson:


jay-hodgson:
Replace with https://github.com/Sage-Bionetworks/synapse-web-monorepo/pull/2061

PR #2059, opened 2025-07-24
Minor data grid demo story improvements

nickgros:
- Use replicaId from server
- Use ComplexJSONRenderer to display model snapshot
- Update ComplexJSONRenderer so 'expand'/'collapse' text is also interactive
- Add config options to Storybook to use localhost as a Synapse backend.

### DIFF
--- a/apps/SageAccountWeb/.env.localhost
+++ b/apps/SageAccountWeb/.env.localhost
@@ -1,0 +1,3 @@
+VITE_SYNAPSE_REPO_ENDPOINT=http://localhost:8080/services-repository-develop-SNAPSHOT
+VITE_SYNAPSE_PORTAL_ENDPOINT=http://localhost:8888
+
--- a/packages/synapse-react-client/.storybook/preview.tsx
+++ b/packages/synapse-react-client/.storybook/preview.tsx
@@ -172,6 +172,7 @@ const preview: Preview = {
           { value: 'staging', title: 'Staging' },
           { value: 'development', title: 'Development' },
           { value: 'mock', title: 'Mocked Data' },
+          { value: 'local', title: 'Local Instance of Synapse' },
         ],
       },
     },
--- a/packages/synapse-react-client/src/components/DataGrid/DataGrid.tsx
+++ b/packages/synapse-react-client/src/components/DataGrid/DataGrid.tsx
@@ -1,3 +1,4 @@
+import { ComplexJSONRenderer } from '@/components/SynapseTable/SynapseTableCell/JSON/ComplexJSONRenderer'
 import { useCallback, useRef, useEffect, useMemo, useState } from 'react'
 import {
   DataSheetGrid,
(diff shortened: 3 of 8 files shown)


apps/SageAccountWeb/.env.localhost
```diff
@@ -0,0 +1,3 @@
+VITE_SYNAPSE_REPO_ENDPOINT=http://localhost:8080/services-repository-develop-SNAPSHOT
+VITE_SYNAPSE_PORTAL_ENDPOINT=http://localhost:8888
+
```
nickgros, line 3: Allow starting SageAccountWeb with `--mode=localhost` to point to local instance of Synapse backend

packages/synapse-react-client/.storybook/preview.tsx
```diff
@@ -172,6 +172,7 @@ const preview: Preview = {
           { value: 'staging', title: 'Staging' },
           { value: 'development', title: 'Development' },
           { value: 'mock', title: 'Mocked Data' },
+          { value: 'local', title: 'Local Instance of Synapse' },
```
nickgros, line 175: Allow pointing Storybook at local instance of Synapse backend

packages/synapse-react-client/src/components/DataGrid/DataGrid.tsx
```diff
@@ -42,10 +43,10 @@ const DataGrid = () => {
     getModel,
   } = useDataGridWebSocket()
   useEffect(() => {
-    if (presignedUrl) {
-      createWebsocket(presignedUrl)
+    if (replicaId && presignedUrl) {
+      createWebsocket(replicaId, presignedUrl)
```
nickgros, line 47: Pass the `replicaId` from the server to the websocket class instance

packages/synapse-react-client/src/components/DataGrid/DataGrid.tsx
```diff
@@ -441,14 +442,17 @@ const DataGrid = () => {
...
           overflowY: 'auto',
         }}
       >
         <h3>Model snapshot</h3>
-        <p>
-          {modelSnapshot ? JSON.stringify(modelSnapshot) : 'No model available'}
-        </p>
+
+        {modelSnapshot ? (
+          <ComplexJSONRenderer value={modelSnapshot} />
```
nickgros, line 452: Interactive component for looking at JSON:

Image

packages/synapse-react-client/src/components/SynapseTable/SynapseTableCell/JSON/ComplexJSONRenderer.tsx
```diff
@@ -27,13 +27,18 @@ export function ComplexJSONRenderer(props: ComplexJSONRendererProps) {
...
       <Typography
         variant={'smallText1'}
-        sx={{ py: 0.5, color: 'grey.600', fontStyle: 'italic' }}
+        sx={{
+          py: 0.5,
+          color: 'grey.600',
+          fontStyle: 'italic',
+          cursor: 'pointer',
+        }}
+        onClick={() => setExpandAll(v => !v)}
```
nickgros, line 36: Minor improvement for this `ComplexJSONRenderer` where the "Expand all" text is now also interactive, not just the button next to the text.

packages/synapse-react-client/src/components/DataGrid/DataGrid.tsx
nickgros: We aren't using the `replicaId` that the server gives us, and passing it to the websocket class ensures that we do.

PR #2061, opened 2025-07-24
SWC-5439: change useGetFavorites() to get all pages of data (200 at a time)

jay-hodgson:


packages/synapse-react-client/src/synapse-queries/user/useFavorites.ts
```diff
@@ -134,7 +156,7 @@ export function useGetFavoritesInfinite<
         accessToken,
         // pass the context.pageParam for the new offset
         context.pageParam,
-        LIMIT,
+        undefined,
```
jay-hodgson, line 159: SynapseClient.getUserFavorites() default limit is already set to 200

packages/synapse-react-client/src/synapse-queries/user/useFavorites.ts
```diff
@@ -85,20 +84,45 @@ export function useGetFavorites(
     UseQueryOptions<PaginatedResults<EntityHeader>, SynapseClientError>
   >,
 ) {
-  const { accessToken, keyFactory } = useSynapseContext()
-
-  return useQuery({
-    ...options,
-    queryKey: keyFactory.getUserFavoritesQueryKey(sort, sortDirection),
-    queryFn: () =>
-      SynapseClient.getUserFavorites(
```
nickgros, line 94: In SynapseClient, we also have a `getAllOfPaginatedService` which may be preferable to implementing this in a hook

PR #2062, opened 2025-07-24
Update DataGrid model to reflect server side changes in PLFM-9088

nickgros:
- Data grid model rows are now an object with `data` and `metadata` properties
- Update grid types to use json-joy schema builder. Remove type-casts that this change made unnecessary.
- Update transformation functions to account for new schema

Depends on https://github.com/Sage-Bionetworks/Synapse-Repository-Services/pull/5309

### DIFF
--- a/packages/synapse-react-client/src/components/DataGrid/DataGridTypes.ts
+++ b/packages/synapse-react-client/src/components/DataGrid/DataGridTypes.ts
@@ -12,7 +12,7 @@ const gridRowSchema = s.obj({
   }),
 })
 
-const gridSchema = s.obj({
+const _gridSchema = s.obj({
   doc_version: s.con('0.1.0'),
   columnNames: s.vec(s.con('')),
   columnOrder: s.arr([s.con(0)]),
@@ -31,7 +31,7 @@ export interface QueryInput {
   input: string
 }
 
-export type GridModel = ReturnType<typeof Model.create<typeof gridSchema>>
+export type GridModel = ReturnType<typeof Model.create<typeof _gridSchema>>
 export type GridModelSnapshot = ReturnType<GridModel['api']['getSnapshot']>
 
 export interface Operation {


packages/synapse-react-client/src/components/DataGrid/DataGrid.tsx
```diff
@@ -69,17 +68,15 @@ const DataGrid = () => {
...
   const [prevRows, setPrevRows] = useState(rowValues)
 
   useEffect(() => {
-    setRowValues(modelRowsToGrid(modelSnapshot))
-    setColValues(modelColsToGrid(modelSnapshot))
+    if (modelSnapshot) {
+      setRowValues(modelRowsToGrid(modelSnapshot))
+      setColValues(modelColsToGrid(modelSnapshot))
+    }
   }, [modelSnapshot])
```
nickgros, line 80: Updated `modelSnapshot` so it may be null. So initialize the rows/cols to empty arrays and don't update them using a null snapshot.

packages/synapse-react-client/src/components/DataGrid/DataGrid.tsx
```diff
@@ -105,7 +102,7 @@ const DataGrid = () => {
   }
 
   // Convert model rows to a format suitable for DataSheetGrid
-  function modelRowsToGrid(modelSnapshot: ModelSnapshot): DataGridRow[] {
+  function modelRowsToGrid(modelSnapshot: GridModelSnapshot): DataGridRow[] {
```
nickgros, line 105: Use new types that include the model definition built-in

packages/synapse-react-client/src/components/DataGrid/DataGrid.tsx
```diff
@@ -114,7 +111,7 @@ const DataGrid = () => {
       columnOrder.forEach((index: number) => {
         const columnName = columnNames[index]
         if (columnName) {
-          rowObj[columnName] = row[index]
+          rowObj[columnName] = row.data[index]
```
nickgros, line 114: account for changes to the model in PLFM-9088

packages/synapse-react-client/src/components/DataGrid/DataGridTypes.ts
```diff
@@ -1,3 +1,24 @@
...
+    }),
+  }),
+})
+
+const gridSchema = s.obj({
+  doc_version: s.con('0.0.3'),
+  columnNames: s.vec(s.con('')),
+  columnOrder: s.arr([s.con(0)]),
+  rows: s.arr([gridRowSchema]),
+})
```
nickgros, line 20: Define the schema using json-joy. This will generate types that have our model built-in, giving us enhanced type safety.

packages/synapse-react-client/src/components/DataGrid/DataGridTypes.ts
```diff
@@ -1,3 +1,24 @@
...
+  metadata: s.obj({
+    synapseRow: s.obj({
+      rowId: s.con(0),
+      versionNumber: s.con(0),
+      etag: s.con(''),
+    }),
+  }),
+})
+
+const gridSchema = s.obj({
```
afwillia, line 15:
```suggestion
const _gridSchema = s.obj({
```

Nit: rename to _gridSchema to avoid [no-unused-vars rule](https://typescript-eslint.io/rules/no-unused-vars/): `'gridSchema' is assigned a value but only used as a type`.

packages/synapse-react-client/src/components/DataGrid/DataGridTypes.ts
```diff
@@ -10,11 +31,8 @@ export interface QueryInput {
   input: string
 }
 
-export interface ModelSnapshot {
-  columnNames: string[]
-  columnOrder: number[]
-  rows: string[]
-}
+export type GridModel = ReturnType<typeof Model.create<typeof gridSchema>>
```
afwillia, line 34:
```suggestion
export type GridModel = ReturnType<typeof Model.create<typeof _gridSchema>>
```
match the renamed var in the suggestion above

PR #2063, opened 2025-07-24
PORTALS-3702: Portals FileEntity Page: Cleanup Properties and Provenance Node Click

kianamcc:
- When a property has no value, do not show the property row.

Before:
Screenshot 2025-07-24 at 1 38 34 PM

After:
Screenshot 2025-07-24 at 1 39 05 PM

- Recognize when we are already on the FileEntity page. Since the provenance graph is displayed there, clicking a node should trigger the default redirect behavior to the appropriate Synapse page.

### DIFF
--- a/apps/synapse-portal-framework/src/components/RedirectDialog.tsx
+++ b/apps/synapse-portal-framework/src/components/RedirectDialog.tsx
@@ -173,13 +173,7 @@ const RedirectDialog = (props: RedirectDialogProps) => {
   }
 
   // Show loading state while fetching entity for FileEntity redirect
-  if (
-    redirectUrl &&
-    entityId &&
-    isLoading &&
-    isFeatureFlagEnabled &&
-    location.pathname !== '/FileEntity'
-  ) {
+  if (redirectUrl && entityId && isLoading && isFeatureFlagEnabled) {
     return (
       <Dialog
         open={true}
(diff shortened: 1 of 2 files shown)


apps/synapse-portal-framework/src/components/RedirectDialog.tsx
```diff
@@ -106,9 +107,15 @@ const RedirectDialog = (props: RedirectDialogProps) => {
       !isLoading &&
       isFeatureFlagEnabled
     ) {
+      const currentUrl = `${location.pathname}${location.search}`
       const internalUrl = `/FileEntity?entityId=${entityId}${
         versionNumber ? `&version=${versionNumber}` : ''
       }`
+
+      if (currentUrl === internalUrl) {
```
kianamcc, line 115: Do not navigate to internalUrl if we are already on the FileEntity page

apps/synapse-portal-framework/src/pages/FileEntityPage/SynapseFileEntityPageProperties.tsx
```diff
@@ -41,37 +46,60 @@ const FilePropertyValue: React.FC<FilePropertyValueProps> = ({
...
+      if (!fileHandle?.contentSize) return null
+      content = calculateFriendlyFileSize(fileHandle.contentSize)
+      break
     default:
-      return <>{fileEntity[keyName] as React.ReactNode}</>
+      if (!fileEntity[keyName]) return null
+      content = fileEntity[keyName] as React.ReactNode
   }
+
+  return (
```
kianamcc, line 93: Moved table row rendering in here to prevent row from being rendered if there is no value.

jay-hodgson (approved):

PR #2077, opened 2025-07-29
PORTALS-3669 - Custom access column for ALS portal

nickgros:
PORTALS-3669

- Add ability to supply custom columns to SynapseTable
- Create custom access column for ALS portal datasets based on dataset source

image


packages/synapse-react-client/src/components/HasAccess/AccessIcon.tsx
```diff
@@ -0,0 +1,70 @@
+import { Theme, useTheme } from '@mui/material'
```
nickgros, line 1: Contents of this file extracted from HasAccessV2.tsx for reuse with no changes.

PR #2079, opened 2025-07-29
SWC-7418

jay-hodgson:


packages/synapse-react-client/src/utils/context/SynapseContext.tsx
```diff
@@ -50,7 +50,8 @@ export type SynapseContextProviderProps = PropsWithChildren<{
  * @returns
  */
 export function SynapseContextProvider(props: SynapseContextProviderProps) {
-  const { children, synapseContext: providedContext } = props
+  // get children and context from props, falling back to defaultContext if not yet set
+  const { children, synapseContext: providedContext = defaultContext } = props
```
nickgros, line 54: I think with this implementation, the `defaultContext` would override some of the generated properties (such as the `keyFactory`). 

Recommend instead changing the prop type so it can be undefined using nullish coalescing where needed, and you could update the `synapseContext` `useMemo` to be something like


```ts
  const synapseContext: SynapseContextType = useMemo(
    () => ({
      ...defaultContext,
      ...providedContext,
      keyFactory: providedContext.keyFactory ?? queryKeyFactory,
      synapseClient: synapseApiClient,
    }),
    [
      providedContext,
      providedContext.keyFactory,
      queryKeyFactory,
      synapseApiClient,
    ],
  )
```

jay-hodgson: Updated the prop type to recognize that the provided SynapseContext might be null.  Rebuilt and tested in a local SWC and we no longer encounter the uncaught TypeError

PR #2083, opened 2025-08-01
SWC-7420 - Upgrade json-schema-ref-parser

nickgros:
Upgrade json-schema-ref-parser, allowing us to remove a patch for @jsdevtools/ono, which is no longer transitively imported.

With this change, we can remove all patches in synapse-web-monorepo and SynapseWebClient, so we are not affected by the change in pnpm 10 that causes transitively-overridden patches to be unused, which prevents running scripts.


apps/SageAccountWeb/index.html
```diff
@@ -31,16 +31,7 @@
         })(window,document,'script','dataLayer','GTM-KPW4KS62');
       </script>
       <!-- End Google Tag Manager -->
-    
-    <script type="module">
```
nickgros, line 35: Verified that SynapseForm still works in production/build mode after removing this from portals.

This polyfill has likely not been necessary for a while, as `vite-plugin-node-polyfills` has handled pretty much all cases`.

packages/synapse-react-client/package.json
```diff
@@ -32,7 +32,7 @@
     "url": "https://github.com/Sage-Bionetworks/synapse-web-monorepo/packages/synapse-react-client"
   },
   "dependencies": {
-    "@apidevtools/json-schema-ref-parser": "^9.1.2",
+    "@apidevtools/json-schema-ref-parser": "^14.1.1",
```
nickgros, line 35: Upgrade removes dependency on `@jsdevtools/ono`, so we can remove our patch for that package.

jay-hodgson (approved):

PR #2087, opened 2025-08-01
PORTALS-3709: Create a Storybook for a Synapse Grid

kianamcc:
Jira: https://sagebionetworks.jira.com/jira/software/c/projects/PLFM/boards/238?assignee=712020%3Aa7fbbeb0-d7a6-443a-9dd4-246363ff09dd&selectedIssue=PORTALS-3709

Logic should be the same.

Storybook:
Screenshot 2025-08-01 at 11 26 01 AM
Screenshot 2025-08-01 at 11 26 54 AM

SWC ex:
Screenshot 2025-08-01 at 11 22 18 AM


packages/synapse-react-client/src/components/DataGrid/SynapseGrid.tsx
```diff
@@ -239,95 +245,131 @@ const DataGrid = () => {
...
+                <SkeletonTable numRows={4} numCols={1} />
+              </div>
+            </>
+          )}
+
+          {/* Grid */}
+          {isGridReady && (
+            <div>
+              <button onClick={commit}>Commit</button>
+              <button onClick={cancel}>Cancel</button>
```
nickgros, line 324: As discussed I think these buttons will go away (doesn't have to be in this PR)

packages/synapse-react-client/src/components/DataGrid/StartGridSession.tsx
```diff
@@ -125,23 +129,31 @@ export const StartGridSession = ({
     }
   }
 
+  useEffect(() => {
+    if (!allowInput && query) {
+      handleStartSession(query)
+    }
+  }, [query])
```
nickgros, line 136: We should start the session imperatively (not in an effect). In StrictMode, running this in an effect will cause two sessions to be created.

Do we have a way to close (but not delete) a session today? If so, please update this to close the session in a cleanup function in this useEffect. If we don't have that today, then we can live with the current behavior.

I don't want this to block merging this PR. Can you add a `TODO` and create a Jira ticket for us to follow up on this issue?

packages/synapse-react-client/src/components/DataGrid/StartGridSession.tsx
```diff
@@ -125,23 +129,31 @@ export const StartGridSession = ({
...
+
   return (
```
nickgros, line 138: Could we simplify to something like this? Have a simple return here if `!allowInput` and remove the `allowInput` checks below
```suggestion
  if (!allowInput) { 
    return null
  }
  return (
```

nickgros (approved):

packages/synapse-react-client/src/components/DataGrid/StartGridSession.tsx
kianamcc: I believe we only have delete today.

PR #2101, opened 2025-08-14
SWC-7426 - Prevent empty patches created by grid scrolling

afwillia:
Horizontal and vertical scrolling in the `DataGrid` component triggers its `onChange` function. The current code marks newly appearing rows and columns as updated, which highlights the data as changed and creates a new empty patch. This PR adds an equality check between updated rows and existing data, ensuring that CSS changes and patches are only created when the grid data has truly changed.

The comparison treats nullish values like `null` and `""` as identical. It also converts values to `String()` before comparing to avoid false differences between numeric and string representations, like `1` and `"1"`. Currently all grid input is a string.

Now that the client can get the schema from the server, we should be able to create typed columns instead of coercing to string.


packages/synapse-react-client/src/components/DataGrid/SynapseGrid.tsx
```diff
@@ -209,9 +214,27 @@ const SynapseGrid = forwardRef<
       }
 
       if (operation.type === 'UPDATE') {
```
nickgros, line 216: Why is `handleChange` being called on scroll? I think adding this check is OK, but this check could slow down the browser if it's checking each row as you scroll past it

nickgros (approved):

PR #2103, opened 2025-08-15
SWC-6173

nickgros:
Update npm publish workflow to manually trigger releases.

This would no longer require code changes to publish new versions of packages, and would also simplify releasing 'patch' versions of packages that do not include all the latest changes.

image

### DIFF
--- a/.github/workflows/npm-publish.yml
+++ b/.github/workflows/npm-publish.yml
@@ -49,7 +49,7 @@ jobs:
         run: pnpm nx run ${{ github.event.inputs.packageName }}:build
       # If the package version changed, and the build succeeded, create the tag using the output of the previous step.
       - name: Create tag
-        if: github.event.inputs.dryRun == false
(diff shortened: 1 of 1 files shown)


packages/markdown-it-container/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@sage-bionetworks/markdown-it-container",
-  "version": "4.0.1",
+  "version": "0.0.0",
```
nickgros, line 3: Set all packages to v0.0.0 since the version is set in the workflow.  The engineer releasing the new version should reference NPM to identify the previous release version.

jay-hodgson:
Is there a relatively easy way to automatically generate the SRC release notes and set as the latest release?

nickgros:
Actually, I am not sure how this will work with the `workspace:*` identifiers in our package.json. This could break internal transitive dependencies

PR #2105, opened 2025-08-15
PORTALS-3726: add Explore Files back into the portal

jay-hodgson:


apps/portals/ampals/src/config/synapseConfigs/data.tsx
```diff
@@ -0,0 +1,27 @@
+import { QueryWrapperPlotNavProps } from 'synapse-react-client'
+import columnAliases from '../columnAliases'
+import { filesSql } from '../resources'
+
+const rgbIndex = 0
+// export const dataColumnLinks: LabelLinkConfig = [
```
kianamcc, line 6: Is this commented out code needed?

kianamcc (approved):

apps/portals/ampals/src/config/synapseConfigs/data.tsx
jay-hodgson: It seems possible that we will need to use a ColumnSpecifiedLink, so I'd like to keep this example in place for now (but this is likely going to change before release).  In general, we should remove commented out code though, so I agree.

PR #2109, opened 2025-08-18
PORTALS-3486

lakikowolfe:
Add a draggable dialog component

Jira: https://sagebionetworks.jira.com/browse/PORTALS-3486


packages/synapse-react-client/src/components/DraggableDialogue/DraggableDialogue.tsx
```diff
@@ -0,0 +1,38 @@
+import Draggable from 'react-draggable'
+import { Box, Paper } from '@mui/material'
+import { useRef } from 'react'
+import { SynapseChat } from '@/components/SynapseChat/index'
+
+export default function DraggableDialogue() {
```
nickgros, line 6: Rename for consistency with other components (e.g. MUI Dialog)
```suggestion
export default function DraggableDialog() {
```

packages/synapse-react-client/src/components/DraggableDialogue/DraggableDialogue.tsx
```diff
@@ -0,0 +1,38 @@
...
+        >
+          <Box
+            sx={{
+              maxHeight: '100%',
+              maxWidth: '100%',
+              overflow: 'auto',
+              padding: '16px',
+            }}
+          >
+            <SynapseChat />
```
nickgros, line 32: This would likely just be `{ children }` but ok for now as a proof of concept

nickgros (approved):

PR #2116, opened 2025-08-19
PORTALS-3715:  Add Our Partners to AMP-ALS home page

jay-hodgson:


jay-hodgson:
Screenshot 2025-08-19 at 1 19 07 PM

apps/portals/ampals/src/pages/HomePage.tsx
```diff
@@ -64,6 +68,17 @@ export default function HomePage() {
...
+      <SectionLayout
+        ContainerProps={{
+          className: 'home-spacer',
+          maxWidth: false,
+          sx: {
+            pb: '50px',
+          },
+        }}
+      >
+        <PortalFeaturedPartners sql={partnersSql} backgroundColor={'white'} />
```
nickgros, line 80: IMO move the background color to the `SectionLayout` sx and remove the prop. Would also have to update the EL portal

PR #2121, opened 2025-08-20
PORTALS-3723 - Grid autocomplete uses JSON schema enums

nickgros:
https://github.com/user-attachments/assets/d30f44a1-cf02-46b2-bec1-5b55c0ee1ed9

### DIFF
--- a/packages/synapse-react-client/src/components/DataGrid/SynapseGrid.tsx
+++ b/packages/synapse-react-client/src/components/DataGrid/SynapseGrid.tsx
@@ -417,6 +417,7 @@ const SynapseGrid = forwardRef<
                   <FullWidthAlert
                     variant="warning"
                     title="Validation Messages:"
+                    isGlobal={false}
                     description={
                       <ul>
                         {rowValues[


packages/synapse-react-client/src/components/DataGrid/StartGridSession.tsx
```diff
@@ -72,91 +73,89 @@ export const StartGridSession = forwardRef<
     const { mutateAsync: startGridSession } = useCreateGridSession()
     const { mutateAsync: createReplicaId } = useCreateGridReplica()
 
-    const handleStartSession = useCallback(
-      async (input: string) => {
```
nickgros, line 76: Removed input argument (just use `gridSql` state)

packages/synapse-react-client/src/components/DataGrid/StartGridSession.tsx
```diff
@@ -171,19 +170,20 @@ export const StartGridSession = forwardRef<
     }
 
     return (
-      <>
-        <div style={{ position: 'relative', width: '300px' }}>
-          <input
-            ref={inputRef}
+      <Grid container spacing={2} sx={{ mb: 2 }}>
+        <Grid size={{ xs: 12, lg: 6 }} sx={{ position: 'relative' }}>
+          <TextField
```
nickgros, line 175: Did a quick pass at adding/swapping to MUI components just to make it easier to see everything in the layout

packages/synapse-react-client/src/components/DataGrid/SynapseGrid.tsx
```diff
@@ -124,31 +129,19 @@ const SynapseGrid = forwardRef<
...
       const columnName = columnNames[index]
-      const enumeratedValues = getEnumeratedValuesForColumn(columnName)
+      const enumeratedValues = jsonSchema
+        ? getEnumeratedValues(getSchemaForProperty(jsonSchema, columnName)).map(
+            item => item.value,
+          )
+        : null
 
-      if (enumeratedValues) {
+      if (enumeratedValues && enumeratedValues.length > 0) {
```
nickgros, line 144: Use new utils to get enumerated values

packages/synapse-react-client/src/components/DataGrid/SynapseGrid.tsx
```diff
@@ -230,8 +223,12 @@ const SynapseGrid = forwardRef<
   }
 
   const handleChange = (newValue: DataGridRow[], operations: Operation[]) => {
+    if (!modelRef.current) {
+      console.error('Model is not initialized')
+      return
+    }
```
nickgros, line 229: Early return to fix occasional null object reference

packages/synapse-react-client/src/components/DataGrid/columns/AutocompleteColumn.tsx
```diff
@@ -1,28 +1,49 @@
...
       freeSolo
       disablePortal={false}
       options={choices}
+      getOptionLabel={option => castCellValueToString(option)}
       value={currentValue}
       onInputChange={(_, newInputValue) => {
-        setRowData(newInputValue || '')
+        setLocalInputState(newInputValue)
       }}
+      onBlur={() => setRowData(localInputState)}
```
nickgros, line 46: Prevent writing patches on each keystroke

afwillia:
@nickgros There's some weird behavior when selecting an existing session - the input for schema populates but not for session, then a blank grid loads.

kianamcc:
> @nickgros There's some weird behavior when selecting an existing session - the input for schema populates but not for session, then a blank grid loads.

I get the same issue as @afwillia. I think the issue might be in StartGridSession.tsx `onClick={() => {
                    setGridSql(session.sessionId || '')
                    setSchemaId(session.gridJsonSchema$Id || '')
                    void handleStartSession()
                  }}`

where we set the gridSql then call handleStartSession right after that uses gridSql that hasn't updated yet in React

packages/synapse-react-client/src/components/DataGrid/useCreateGridSession.ts
```diff
@@ -1,28 +0,0 @@
-import { SynapseClientError } from '@sage-bionetworks/synapse-client'
-import {
-  CreateGridResponse,
-  CreateGridRequest,
-} from '@sage-bionetworks/synapse-client'
-import { useMutation, UseMutationOptions } from '@tanstack/react-query'
-import { useSynapseContext } from '@/utils/context/SynapseContext'
-import startGridSession from '../../utils/functions/GridApiUtils'
-
-export const useCreateGridSession = (
```
nickgros, line 10: Moved to `useGridSession.ts` in `synapse-queries`

packages/synapse-react-client/src/components/DataGrid/useInitializeGridConnection.ts
```diff
@@ -0,0 +1,138 @@
...
+          )
+        } else {
+          console.log('Starting a new empty grid session.')
+        }
+
+        const gridSessionResponse = await startGridSession(gridRequest)
+        console.log('Grid session started:', gridSessionResponse)
+        session = gridSessionResponse.gridSession!
+      } else {
+        // parsedInput.type === 'sessionId'
```
kianamcc, line 61: If this comment is needed, could we make it a little more descriptive something like: `// Retrieve an existing grid session using the provided session ID`

nickgros:
```suggestion
        // The input is a sessionId, so retrieve the session object
```

PR #2124, opened 2025-08-20
SWC-7409 - Display validation messages in grid

afwillia:
This PR adds dynamic CSS to display the `isValid` state of each row in the index column. It also shows the validation messages for the selected column in a text box below the grid. If the row validation state is unknown, a spinner is shown.

image

### DIFF
--- a/packages/synapse-react-client/src/components/DataGrid/SynapseGrid.tsx
+++ b/packages/synapse-react-client/src/components/DataGrid/SynapseGrid.tsx
@@ -407,8 +407,10 @@ const SynapseGrid = forwardRef<
                     'row-created': createdRowIds.has(rowData._rowId),
                     'row-updated': updatedRowIds.has(rowData._rowId),
                     'row-valid': rowData.__validationResults?.isValid === true,
-                    'row-invalid': rowData.__validationResults?.isValid === false,
-                    'row-unknown': rowData.__validationResults?.isValid == undefined,
+                    'row-invalid':
+                      rowData.__validationResults?.isValid === false,
+                    'row-unknown':
+                      rowData.__validationResults?.isValid == undefined,
                     'row-selected': selectedRowIndex === rowIndex,
                   })
                 }
@@ -425,20 +427,26 @@ const SynapseGrid = forwardRef<
               {/* Show validation messages for selected row */}
               {selectedRowIndex !== null &&
                 rowValues[selectedRowIndex] &&
-                Array.isArray(rowValues[selectedRowIndex].__validationResults?.allValidationMessages) &&
-                rowValues[selectedRowIndex].__validationResults?.allValidationMessages.length > 0 && (
+                Array.isArray(
+                  rowValues[selectedRowIndex].__validationResults
+                    ?.allValidationMessages,
+                ) &&
+                rowValues[selectedRowIndex].__validationResults
+                  ?.allValidationMessages.length > 0 && (
                   <FullWidthAlert
-                    variant='warning'
-                    title='Validation Messages:'
-                    description={(
+                    variant="warning"
+                    title="Validation Messages:"
+                    description={
                       <ul>
-                        {rowValues[selectedRowIndex].validationMessages.map(
+                        {rowValues[
+                          selectedRowIndex
+                        ].__validationResults.allValidationMessages.map(
                           (msg: string) => (
                             <li key={msg}>{msg}</li>
                           ),
                         )}
                       </ul>
-                    )}
+                    }
                     sx={{
                       marginTop: '12px',
                     }}


packages/synapse-react-client/src/components/DataGrid/DataGridTypes.ts
```diff
@@ -9,6 +9,18 @@ const gridRowSchema = s.obj({
...
+      validationErrorMessage: s.con(''),
+      allValidationMessages: s.vec(s.con('')),
+      validationException: s.obj({
+        keyword: s.con(''),
+        pointerToViolation: s.con(''),
+        message: s.con(''),
+        schemaLocation: s.con(''),
+        causingExceptions: s.vec(s.con('')),
+      }),
+    }),
```
nickgros, line 23: This entire node is actually a constant. Can we do something like this?
```ts
import { ValidationResults } from '@sage-bionetworks/synapse-types'
```


```suggestion
    rowValidation: s.con<ValidationResults | undefined>(undefined),
```

packages/synapse-react-client/src/components/DataGrid/SynapseGrid.tsx
```diff
@@ -238,7 +244,7 @@ const SynapseGrid = forwardRef<
         // and adding rows to the model via the api
         for (let i = operation.fromRowIndex; i < operation.toRowIndex; i++) {
           const rowArr = model.api.arr(['rows'])
-          rowArr.ins(i, [{ data: s.vec(s.con('')), metadata: {} }])
+          rowArr.ins(i, [{ data: s.vec(s.con('')), metadata: s.obj({}) }])
```
nickgros, line 247: As John mentioned, we shouldn't need to supply this in the client.

```suggestion
          rowArr.ins(i, [{ data: s.vec(s.con('')) }])
```

packages/synapse-react-client/src/components/DataGrid/SynapseGrid.tsx
```diff
@@ -396,24 +405,64 @@ const SynapseGrid = forwardRef<
                 value={rowValues}
                 columns={colValues}
                 rowKey="_rowId"
-                rowClassName={({ rowData }: any) => {
+                rowClassName={({ rowData, rowIndex }: any) => {
```
nickgros, line 408: Ditto Copilot, but you can probably just remove the type altogether and it will infer it from `react-datasheet-grid`'s props
```suggestion
                rowClassName={({ rowData, rowIndex }) => {
```

packages/synapse-react-client/src/components/DataGrid/SynapseGrid.tsx
```diff
@@ -396,24 +405,64 @@ const SynapseGrid = forwardRef<
...
+                  let className = ''
```
nickgros, line 409: consider using the `classnames` util (used elsewhere in this project)
```ts
import classNames from 'classnames'
```

packages/synapse-react-client/src/components/DataGrid/SynapseGrid.tsx
```diff
@@ -396,24 +405,64 @@ const SynapseGrid = forwardRef<
...
+                    <ul style={{ margin: 0, paddingLeft: 18 }}>
+                      {rowValues[selectedRowIndex].validationMessages.map(
+                        (msg: string, i: number) => (
+                          <li key={i}>{msg}</li>
+                        ),
+                      )}
+                    </ul>
+                  </div>
+                )}
             </div>
```
nickgros, line 466: You could use our `FullWidthAlert` to get styling that will feel more 'native' to Synapse (and let you remove a lot of these inline styles)

packages/synapse-react-client/src/components/DataGrid/SynapseGrid.tsx
```diff
@@ -112,13 +112,19 @@ const SynapseGrid = forwardRef<
     const { columnNames, columnOrder, rows } = modelSnapshot
     const gridRows = rows.map(row => {
       const rowObj: { [key: string]: any } = {}
```
nickgros, line 114: I think we could improve the type/structure of `DataGridRow`; for now let's just use it here so it is clear that we need to fix it here if/when we update it
```suggestion
    const gridRows = rows.map((row): DataGridRow => {
      const rowObj: DataGridRow = {}
```


To clarify what I mean, a `DataGridRow` should probably explicitly declare that it has validation information. So it could look something like

```ts
type DataGridRow = {
  data: Record<string, string | number | boolean | null | undefined>
  validationInformation?: ValidationResults
}

```

packages/synapse-react-client/src/components/DataGrid/SynapseGrid.tsx
```diff
@@ -112,13 +112,19 @@ const SynapseGrid = forwardRef<
...
-      // Use columnOrder to determine which columnNames to use and in what order
       columnOrder.forEach((index: number) => {
         const columnName = columnNames[index]
         if (columnName) {
           rowObj[columnName] = row.data[index]
         }
       })
+      const validation = row.metadata?.rowValidation
+      rowObj.validationStatus = validation
```
nickgros, line 122: Suggest restructuring the `DataGridRow` type to nest the data so this validation info does not collide with it (see comment right above this). Or at the very least, prefix this with an underscore, e.g.
```suggestion
      rowObj._validationStatus = validation
```

packages/synapse-react-client/src/components/DataGrid/SynapseGrid.tsx
```diff
@@ -396,24 +407,54 @@ const SynapseGrid = forwardRef<
...
+                      (msg: string, i: number) => (
+                        <FullWidthAlert
+                          variant={'warning'}
+                          isGlobal={false}
+                          description={msg}
+                        />
+                      ),
+                    )}
+                  </div>
+                )}
```
nickgros, line 457: Use the __validationResults object. Also, I was thinking just one `FullWidthAlert`.
```suggestion
              {selectedRowIndex !== null &&
                rowValues[selectedRowIndex] &&
                Array.isArray(rowValues[selectedRowIndex].__validationResults?.allValidationMessages) &&
                rowValues[selectedRowIndex].__validationResults?.allValidationMessages.length > 0 && (
                  <FullWidthAlert
                    variant='warning'
                    title='Validation Messages:'
                    description={(
                      <ul>
                        {rowValues[selectedRowIndex].validationMessages.map(
                          (msg: string) => (
                            <li key={msg}>{msg}</li>
                          ),
                        )}
                      </ul>
                    )}
                    sx={{
                      marginTop: '12px',
                    }}
                  />
                )}
```

packages/synapse-react-client/src/components/DataGrid/DataGridTypes.ts
```diff
@@ -42,4 +44,7 @@ export interface Operation {
 
 export type DataGridRow = {
   [key: string]: string | number | boolean | null | undefined
+} & {
+  __validationResults?: ValidationResults
+  validationMessages?: string[]
```
nickgros, line 49: You don't need this since it's inside the `__validationResults` object
```suggestion
```

packages/synapse-react-client/src/components/DataGrid/DataGridTypes.ts
```diff
@@ -42,4 +44,7 @@ export interface Operation {
 
 export type DataGridRow = {
   [key: string]: string | number | boolean | null | undefined
+} & {
+  __validationResults?: ValidationResults
```
nickgros, line 48:
```suggestion
  /** The JSON Schema validation results for the row, generated by the hub */
  __validationResults?: ValidationResults
```

packages/synapse-react-client/src/components/DataGrid/DataGridTypes.ts
```diff
@@ -42,4 +44,7 @@ export interface Operation {
 
 export type DataGridRow = {
```
nickgros, line 45:
```suggestion
/**
 * Object that represents a single row of data in a Grid session that is compatible with react-datasheet-grid.
 * This objects acts as a "ViewModel" and can be translated to/from the Synapse data grid CRDT schema.
 */
export type DataGridRow = {
```

PR #2126, opened 2025-08-20
Release synapse-react-client v3.4.39

nickgros:
- Upgrade used version of @sage-bionetworks/synapse-client

Depends on #2125


jay-hodgson:
Released 3.4.39

PR #2131, opened 2025-08-21
SWC-7431

nickgros:
Do not query for entity bundle if `entityId` is falsy

### DIFF
--- a/packages/synapse-react-client/src/components/EntityFinder/EntityFinder.tsx
+++ b/packages/synapse-react-client/src/components/EntityFinder/EntityFinder.tsx
@@ -359,20 +359,22 @@ export function EntityFinder({
             <div style={searchActive ? { display: 'none' } : {}}>
               {treeOnly ? (
                 <div>
-                  <EntityTree
-                    toggleSelection={toggleSelection}
-                    showDropdown={true}
-                    visibleTypes={selectableAndVisibleTypesInTree}
-                    initialScope={initialScope}
-                    selectedEntities={selectedEntities}
-                    projectId={projectId}
-                    initialContainer={initialContainer}
-                    currentContainer={currentContainer}
-                    setCurrentContainer={setCurrentContainer}
-                    showScopeAsRootNode={false}
-                    treeNodeType={EntityTreeNodeType.SINGLE_PANE}
-                    selectableTypes={selectableTypes}
-                  />
+                  <SynapseErrorBoundary>
+                    <EntityTree
+                      toggleSelection={toggleSelection}
+                      showDropdown={true}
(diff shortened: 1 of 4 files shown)


packages/synapse-react-client/src/synapse-queries/entity/useEntityBundle.ts
```diff
@@ -6,26 +6,60 @@ import {
...
+            entityId,
+            bundleRequest,
+            version,
+            accessToken,
+          )
+      : skipToken,
+  }
+}
 
+export function useGetEntityBundleSuspenseQueryOptions<
```
nickgros, line 45: We can't use `skipToken` in suspense queries, so I had to create a separate function to ensure we return the correct type when we get the `UseSuspenseQueryOptions`

packages/synapse-react-client/src/synapse-queries/entity/useEntity.ts
```diff
@@ -513,18 +513,20 @@ function useGetEntityBenefactorACLQueryOptions(
...
+  return useGetEntityBundleSuspenseQueryOptions<
+    { includeBenefactorACL: true },
+    AccessControlList
+  >(
     entityId,
     undefined,
     {
       includeBenefactorACL: true,
     },
+    selectTransform,
```
nickgros, line 528: Pass `select` directly to `useGetEntityBundleSuspenseQueryOptions` to ensure the full options object has the correct generic type arguments.

packages/synapse-react-client/src/components/EntityFinder/EntityFinder.tsx
```diff
@@ -359,20 +359,22 @@ export function EntityFinder({
...
-                    selectedEntities={selectedEntities}
-                    projectId={projectId}
-                    initialContainer={initialContainer}
-                    currentContainer={currentContainer}
-                    setCurrentContainer={setCurrentContainer}
-                    showScopeAsRootNode={false}
-                    treeNodeType={EntityTreeNodeType.SINGLE_PANE}
-                    selectableTypes={selectableTypes}
-                  />
+                  <SynapseErrorBoundary>
```
nickgros, line 362: Added a bunch of error boundaries just to help isolate the problem to figure out where it was coming from

packages/synapse-react-client/src/synapse-queries/entity/useEntity.ts
kianamcc: Could we give a different name to 'selectTransform'? Would something like 'selectBenefactorAcl' be more descriptive? Ignore this if not appropriate here!

PR #2132, opened 2025-08-21
PORTALS-3749

jay-hodgson:


packages/synapse-react-client/src/synapse-queries/entity/useExportToCavatica.ts
```diff
@@ -14,10 +14,17 @@ export function addDrsUriToSql(
...
   fileNameColumnName?: string,
   fileVersionColumnName?: string,
+  removeSpacesFromSelectColumns: boolean = false,
 ): string {
   const selectColumnsList = selectColumns
     ?.filter(col => col.name != 'name')
-    .map(col => `"${col.name}"`)
+    .map(col => {
+      if (removeSpacesFromSelectColumns) {
+        return `"${col.name}" as "${col.name.replace(/\s/g, '')}"`
```
jay-hodgson, line 23: The bug fix - alias select columns and remove whitespace.
(removeSpacesFromSelectColumns set to true for pfb export)

nickgros: Couple of edge cases I verified:

- if there is a WHERE clause in the originalSql, it still works (e.g. `SELECT "my col" as "mycol" from syn123 WHERE "my col" = 'foo'` correctly applies a filter)
- there is no need to escape `"` (Synapse does not seem to allow it in a valid column name)

so this change LGTM

packages/synapse-react-client/src/synapse-queries/entity/useExportToTerra.test.tsx
```diff
@@ -0,0 +1,142 @@
...
+    sql: originalSql,
+    limit: DEFAULT_PAGE_SIZE,
+    offset: 0,
+  },
+}
+describe('useExportToTerra ()', () => {
+  beforeEach(() => {
+    vi.resetAllMocks()
+    ;(parseEntityIdFromSqlStatement as Mock).mockReturnValue(entityId)
+    ;(addDrsUriToSql as unknown as Mock).mockReturnValue(transformedSql)
```
nickgros, line 54: Removes unnecessary cast and provides a `mock` object that is aware of the original type

```suggestion
    mocked(parseEntityIdFromSqlStatement).mockReturnValue(entityId)
    mocked(addDrsUriToSql).mockReturnValue(transformedSql)
```

packages/synapse-react-client/src/synapse-queries/entity/useExportToTerra.test.tsx
```diff
@@ -0,0 +1,142 @@
...
+vi.mock('../../components/ToastMessage/ToastMessage', () => ({
+  displayToast: vi.fn(),
+}))
+
+import { parseEntityIdFromSqlStatement } from '../../utils/functions/SqlFunctions'
+vi.mock('../../utils/functions/SqlFunctions', () => ({
+  parseEntityIdFromSqlStatement: vi.fn(),
+}))
+
+import { addDrsUriToSql } from './useExportToCavatica'
```
nickgros, line 27: surprised these `import`s don't need to be done before the instantiation of `mockCreatePfb`

jay-hodgson: Moved imports

jay-hodgson:
Merging (despite EntityCitation test error)

PR #2142, opened 2025-08-22
AMPALS-143: display all acknowledgemenet statements in Requirements for Publicati…

jay-hodgson:
…on page


jay-hodgson:
This approach works for today, approved by Christine:  https://sagebionetworks.jira.com/browse/AMPALS-143?focusedCommentId=263953

PR #2147, opened 2025-08-27
SWC-7449 - Reduce size of grid patches

afwillia:
Refactor code for patching the model from grid updates. Only patch cells that have changed. Also, move all model CRUD operations into one function that `handleChange` calls.


nickgros (approved):
Looks great

PR #2153, opened 2025-08-28
SWC-7450

jay-hodgson:
Screenshot 2025-08-28 at 10 55 16 AM

No changes to the existing demo:
Screenshot 2025-08-28 at 10 54 36 AM

### DIFF
--- a/packages/synapse-react-client/src/components/dataaccess/SubmissionPage/SubmissionPage.tsx
+++ b/packages/synapse-react-client/src/components/dataaccess/SubmissionPage/SubmissionPage.tsx
@@ -25,7 +25,7 @@ import duration from 'dayjs/plugin/duration'
 import { isEmpty, toLower, upperFirst } from 'lodash-es'
 import { Fragment, ReactNode, useMemo, useState } from 'react'
 import { useNavigate } from 'react-router'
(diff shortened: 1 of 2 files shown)


packages/synapse-react-client/src/components/dataaccess/SubmissionPage/SubmissionPage.tsx
```diff
@@ -117,8 +117,11 @@ export default function SubmissionPage(props: SubmissionPageProps) {
 
   const { data: currentUserProfile } = useGetCurrentUserProfile()
 
-  const { data: submission, isLoading: isLoadingSubmission } =
-    useGetDataAccessSubmission(submissionId)
+  const {
+    data: submission,
+    isLoading: isLoadingSubmission,
+    error: submissionLoadingError,
+  } = useGetDataAccessSubmission(submissionId)
```
nickgros, line 124: An alternative would be to pass `throwOnError` as an option to the useQuery and let a parent error boundary handle it, but I think handling the error explicitly here is fine too

jay-hodgson: I like it!  Updated

PR #2164, opened 2025-09-02
Optimization/cleanup/tests for grid UI

nickgros:
- Update DataGridWebSocket to provide callback `onModelCreate` and remove `onModelChange`. To listen to the model changes, you can subscribe to the model object directly.
- Replace `useCRDTState` with `useCRDTModelView`, which subscribes to a nullable `json-joy` `Model` and returns the derived view result.
- Update `useDataGridWebsocket` to disconnect from the websocket on cleanup and reset state
- Update `SynapseGrid`
  - Use the derived model view returned by `useCRDTModelView` via `useDataGridWebsocket`
  - Remove unnecessary UI state tracking add/update/removes since those are applied directly to the model
  - Check for no-op updates returned in `handleChange` before committing changes to the model
  - Factor out code to check for no-op updates, moved to `DataGridUtils`
  - Removed unused argument to `autoCommit`


packages/synapse-react-client/src/components/DataGrid/DataGridUtils.ts
```diff
@@ -27,3 +27,30 @@ export function rowsAreIdentical(a: DataGridRow, b: DataGridRow): boolean {
   }
   return true
 }
+
+export function removeNoOpOperations(
```
nickgros, line 31: Factored out of SynapseGrid and updated to filter out no-ops and return the valid operations.

packages/synapse-react-client/src/components/DataGrid/SynapseGrid.tsx
```diff
@@ -245,31 +223,19 @@ const SynapseGrid = forwardRef<
     }
   }
 
-  // Grid editing functions
-  const createdRowIds = useMemo(() => new Set(), [])
-  const deletedRowIds = useMemo(() => new Set(), [])
-  const updatedRowIds = useMemo(() => new Set(), [])
-
-  const performCommit = (dataToCommit: DataGridRow[]) => {
+  const performCommit = () => {
```
nickgros, line 226: Remove unused argument

packages/synapse-react-client/src/components/DataGrid/SynapseGrid.tsx
```diff
@@ -279,76 +245,21 @@ const SynapseGrid = forwardRef<
...
-        const newVal = newValue.slice(
-          operation.fromRowIndex,
-          operation.toRowIndex,
-        )
-
-        const comparisonResults = oldVal.map((oldItem, idx) =>
-          rowsAreIdentical(oldItem, newVal[idx]),
-        )
-
-        if (comparisonResults.every(Boolean)) return
```
nickgros, line 312: Fix a potential bug here where this was an early return if we should ignore the update. If this no-op update was a batch of operations, we would skip all the other operations that we should apply!

PR #2168, opened 2025-09-03
Use row object logical timestamp as grid row React key

nickgros:


packages/synapse-react-client/src/components/DataGrid/SynapseGrid.tsx
```diff
@@ -237,7 +223,7 @@ const SynapseGrid = forwardRef<
   )
 
   function addRowToModel() {
```
afwillia, line 225: Do we need this function at all anymore?

afwillia (approved):

PR #2169, opened 2025-09-03
improve type handling, copy/paste for autocomplete fields

nickgros:


packages/synapse-react-client/src/components/DataGrid/columns/AutocompleteColumn.tsx
```diff
@@ -76,13 +91,20 @@ function AutocompleteCell({
...
   return {
-    component: (props: Omit<AutocompleteCellProps, 'choices'>) => (
-      <AutocompleteCell {...props} choices={choices} />
-    ),
+    component: ((props: Omit<AutocompleteCellProps, 'choices'>) => (
+      <AutocompleteCell {...props} choices={choices} colType={colType} />
+    )) as CellComponent,
+    // If we update our enums to support labels, then we can update copy to copy the label and paste to lookup the mapping from label -> value
+    copyValue: ({ rowData }) => rowData,
+    pasteValue: ({ value }) => value,
```
nickgros, line 107: add copy/paste support (required to make drag work)

if we support custom labels for enums, we would need to add additional logic here

packages/synapse-react-client/src/utils/jsonschema/getEnumeratedValues.ts
```diff
@@ -18,13 +18,6 @@ function rjsfOptionsListToValueList(option: EnumOptionsType): {
   return { value: option.value as EnumeratedValue }
 }
 
-function isSchemaDescribingNullType(s: any) {
-  return (
-    isObject(s) &&
-    'type' in s &&
-    (s.type == 'null' || isEqual(s.type, ['null']))
-  )
-}
```
nickgros, line 27: moved to new `getOnlyNonNullOneOfOption`

packages/synapse-react-client/src/components/DataGrid/columns/AutocompleteColumn.tsx
afwillia: we may need something similar for boolean columns as well.

PR #2183, opened 2025-09-08
B2ai issue 377 trying to remove empty sections in org detail

Sigfried:
In the middle of working on this, but started getting 
  image
Tracked it down to whether it's using the virtual table (_current) or the base table:
  image

I can't find any difference between the two in terms of schema or contents.

@nickgros: Any ideas?


apps/portals/b2ai.standards/src/hooks/useFetchTableData.ts
```diff
@@ -0,0 +1,217 @@
...
+) & // countsOnly is only an option with a single id
+  (
+    | { id: string; countsOnly?: boolean; ids?: never }
+    | { ids: string[]; id?: never; countsOnly?: never }
+  )
+
+/**
+ * General hook for fetching table data
+ */
+export function useFetchTableData(props: UseFetchAllRowCountsProps) {
```
nickgros, line 28: I am not opposed to adding utilities to simplify writing queries and transforming query results, but I don't think we should conflate the two and put it all in a hook that is somewhat redundant with `useGetQueryResultBundle`.

You could create a utility that handles creating the query bundle request:

```ts
export function getQueryBundleRequestWithIdFilter(
  tableName: string,
  selectExpressions?: string[],
  filterValues?: string[],
): QueryBundleRequest {
  const query: Query = {
    sql: getQueryString(tableName, selectExpressions),
  }

  if (filterValues && !isEmpty(filterValues)) {
    query.additionalFilters = [
      {
        concreteType:
          'org.sagebionetworks.repo.model.table.ColumnSingleValueQueryFilter',
        columnName: getIdCol(tableName),
        operator: ColumnSingleValueFilterOperator.IN,
        values: filterValues,
      },
    ]
  }

  return {
    entityId: getSynapseTableId(tableName),
    partMask:
      SynapseConstants.BUNDLE_MASK_QUERY_SELECT_COLUMNS |
      SynapseConstants.BUNDLE_MASK_QUERY_RESULTS,
    query,
  }
}
```

If you want to transform the QueryResultBundle to an array of objects, that could also be its own utility (which IMO should be separate from the `countsOnly` logic). Note that the API provides `selectColumns`, so you do not need to parse out the `expressions`.

```ts
export function getRowsAsObjects(data: QueryResultBundle) {
  const colIndexes =
    data.selectColumns?.map((column, index) => ({
      column: column.name,
      index: index,
    })) ?? []

  const rowsAsObjects: Record<string, string | null>[] =
    data?.queryResult?.queryResults.rows.map(el => {
      const values = el.values
      const result = {}
      colIndexes.forEach(({ column, index }) => {
        result[column] = values[index]
      })
      return result
    }) ?? []

  return rowsAsObjects
}
```

Transforming the result further should be done in a utility that is specific to each query, since semantic meaning of the result object will vary by query.

On the OrganizationDetailsPage, you could do:


```ts
  const queryBundleRequest = getQueryBundleRequestWithIdFilter('Organization_denormalized', colExpressions, id)
  const {
    data = [],
    error,
    isLoading,
  } = useGetQueryResultBundle(queryBundleRequest, {
    select: result => {
      const rows = getRowsAsObjects(result)
      // parse JSON as needed here and return
    }
  })

```

apps/portals/b2ai.standards/src/hooks/useHasQueryResults.ts
```diff
@@ -0,0 +1,53 @@
...
+    queryBundleRequest,
+    {
+      select: data => {
+        const hasResults = (data?.queryCount || 0) > 0
+        return hasResults
+      },
+    },
+  )
+
+  return { data, error, isLoading }
```
nickgros, line 52: Just return the hook result

```suggestion
  return useGetQueryResultBundle(
    queryBundleRequest,
    {
      select: data => {
        const hasResults = (data?.queryCount || 0) > 0
        return hasResults
      },
    },
  )
```

apps/portals/b2ai.standards/src/components/ChallengesCardDeck.tsx
```diff
@@ -1,114 +1,45 @@
...
-          return result
-        }) ?? []
-      return results
-    },
+  const { data, error, isLoading } = useFetchTableData({
+    tableName: 'Organization_denormalized',
+    colExpressions: COLS_NEEDED,
+    ids: GC_ORG_IDS,
   })
+  console.log({ data, error, isLoading })
```
nickgros, line 21: Remove console logs

Sigfried: did that

apps/portals/b2ai.standards/src/config/resources.ts
```diff
@@ -1,71 +1,93 @@
...
+    ${Object.values(ORG_TABLE_COLUMN_CONSTS).join(', ')},
     concat('/Explore/Organization/OrganizationDetailsPage?id=', ${
-      ORG_TABLE_COLUMN_NAMES.ID
+      ORG_TABLE_COLUMN_CONSTS.ID
     }) AS orgPageLink
-  FROM ${TABLE_IDS.Organization.id}`
+  FROM ${tableInfo.Organization_denormalized.id}`
+tableInfo.Organization_denormalized.queries ??= {}
+tableInfo.Organization_denormalized.queries.detailsSQL =
+  organizationDetailsPageSQL
```
nickgros, line 72: Kind of odd to dynamically set these, but I understand why you had to do it. Seems like this tableInfo needs to be 'normalized' itself

apps/portals/b2ai.standards/src/pages/OrganizationDetailsPage.tsx
```diff
@@ -1,48 +1,51 @@
...
-  ORG_TABLE_COLUMN_NAMES,
-  DATASET_DENORMALIZED_COLUMN_NAMES,
-  DST_TABLE_COLUMN_NAMES,
+  standardsSql,
 } from '@/config/resources'
+import {
+  getQueryBundleRequestWithIdFilter,
+  useFetchJsonArrayLengths,
+} from '@/hooks/fetchDataUtils'
+// import DetailsPage from '@sage-bionetworks/synapse-portal-framework/components/DetailsPage'
```
nickgros, line 37:
```suggestion
```

apps/portals/b2ai.standards/src/pages/OrganizationDetailsPage.tsx
```diff
@@ -77,184 +80,249 @@ function OrgHeaderCard({ id }) {
...
+
+  if (!id) {
+    return <ErrorPage type={SynapseErrorType.NOT_FOUND} gotoPlace={() => {}} />
+  }
+  if (isLoading) {
+    return <div>Loading...</div>
+  }
+  if (data.length !== 1) {
+    throw new Error(`Unexpected ${data.length} rows of Org data.`)
+  }
```
nickgros, line 117: Throwing an error here is not the best user experience. If there are 0, we should show a not found page. If there is more than 1, we can still show the first one.

```suggestion
  if (data.length == 0) {
    return <ErrorPage type={SynapseErrorType.NOT_FOUND} gotoPlace={() => {}} />
  } else if (data.length > 1) {
    console.error('Unexpected ${data.length} rows of Org data')
  }
```

apps/portals/b2ai.standards/src/pages/OrganizationDetailsPage.tsx
```diff
@@ -77,184 +80,249 @@ function OrgHeaderCard({ id }) {
...
+
+  if (error) {
+    return <ErrorBanner error={error} />
+  }
+
+  if (!id) {
+    return <ErrorPage type={SynapseErrorType.NOT_FOUND} gotoPlace={() => {}} />
+  }
+  if (isLoading) {
+    return <div>Loading...</div>
```
nickgros, line 113: Consider a more user-friendly loader, such as `LinearProgress` from `@mui/material`
```suggestion
    return <LinearProgress />
```

apps/portals/b2ai.standards/src/components/D4D.tsx
```diff
@@ -1,20 +1,55 @@
...
+    select: result => {
+      return getRowsAsObjects(result)
+    },
+  })
 
+  if (error) {
+    return <ErrorBanner error={error} />
+  }
+  if (isLoading) {
+    return <div>Loading...</div>
```
nickgros, line 39: Consider a more user-friendly loader, such as `LinearProgress` from `@mui/material`
```suggestion
    return <LinearProgress />
```

PR #2189, opened 2025-09-09
PORTALS-3799: Grid: Indicate required columns

afwillia:
This PR distinguishes required columns by giving the header cell a background color. It includes a utility to parse schemas and return required attributes. It sets `heaederClassName` of required columns.

### DIFF
--- a/packages/synapse-react-client/src/utils/jsonschema/getRequiredAttributes.test.ts
+++ b/packages/synapse-react-client/src/utils/jsonschema/getRequiredAttributes.test.ts
@@ -1,0 +1,102 @@
+import { describe, it, expect } from 'vitest'
+import getRequiredAttributes from './getRequiredAttributes'
+
+describe('getRequiredAttributes', () => {
+  it('returns empty array for undefined/null/non-object', () => {
+    expect(getRequiredAttributes(undefined)).toEqual([])
+    expect(getRequiredAttributes(null)).toEqual([])
+    // @ts-expect-error intentional wrong type
+    expect(getRequiredAttributes(5)).toEqual([])
+  })
+
+  it('returns required fields from simple schema', () => {
+    const schema = {
+      type: 'object',
+      required: ['a', 'b'],
+      properties: {
+        a: { type: 'string' },
+        b: { type: 'number' },
+        c: { type: 'boolean' },
+      },
+    }
+    expect(getRequiredAttributes(schema).sort()).toEqual(['a', 'b'])
+  })
+
+  it('deduplicates required fields', () => {
+    const schema = {
+      type: 'object',
+      required: ['a', 'b', 'a'],
+    }
+    expect(getRequiredAttributes(schema).sort()).toEqual(['a', 'b'])
+  })
+
+  it('handles allOf union of required', () => {
+    const schema = {
+      allOf: [
+        { required: ['a', 'b'] },
+        { required: ['b', 'c'] },
+        { properties: { d: { type: 'string' } } },
+      ],
+    }
+    expect(getRequiredAttributes(schema).sort()).toEqual(['a', 'b', 'c'])
+  })
+
+  it('handles anyOf union of required', () => {
+    const schema = {
+      anyOf: [{ required: ['x'] }, { required: ['y', 'z'] }, { required: [] }],
+    }
+    expect(getRequiredAttributes(schema).sort()).toEqual(['x', 'y', 'z'])
+  })
+
+  it('handles oneOf union of required', () => {
+    const schema = {
+      oneOf: [
+        { required: ['m'] },
+        { required: ['n'] },
+        { required: ['m', 'o'] },
+      ],
+    }
+    expect(getRequiredAttributes(schema).sort()).toEqual(['m', 'n', 'o'])
+  })
+
+  it('traverses nested combinators', () => {
+    const schema = {
+      allOf: [
+        {
+          anyOf: [{ required: ['a'] }, { required: ['b'] }],
+        },
+        {
+          oneOf: [
+            { required: ['c'] },
+            {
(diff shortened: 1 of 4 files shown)


packages/synapse-react-client/src/utils/jsonschema/getRequiredAttributes.test.ts
```diff
@@ -0,0 +1,102 @@
...
+    }
+    expect(getRequiredAttributes(schema).sort()).toEqual(['a', 'b', 'c'])
+  })
+
+  it('handles anyOf union of required', () => {
+    const schema = {
+      anyOf: [{ required: ['x'] }, { required: ['y', 'z'] }, { required: [] }],
+    }
+    expect(getRequiredAttributes(schema).sort()).toEqual(['x', 'y', 'z'])
+  })
```
nickgros, line 49: Not a valid test, with this schema, there is technically no requirement.
```suggestion
  it('handles anyOf union of required', () => {
    const schema = {
      anyOf: [{ required: ['x'] }, { required: ['y', 'z'] }, { required: [] }],
    }
    expect(getRequiredAttributes(schema).sort()).toEqual([])
  })
```

packages/synapse-react-client/src/utils/jsonschema/getRequiredAttributes.test.ts
```diff
@@ -0,0 +1,102 @@
...
+  it('handles oneOf union of required', () => {
+    const schema = {
+      oneOf: [
+        { required: ['m'] },
+        { required: ['n'] },
+        { required: ['m', 'o'] },
+      ],
+    }
+    expect(getRequiredAttributes(schema).sort()).toEqual(['m', 'n', 'o'])
+  })
```
nickgros, line 60: Not a valid test since it has conditional/alternative requirements
```suggestion
  it('handles oneOf union of required', () => {
    const schema = {
      oneOf: [
        { required: ['m'] },
        { required: ['n'] },
        { required: ['m', 'o'] },
      ],
    }
    expect(getRequiredAttributes(schema).sort()).toEqual([])
  })
```

packages/synapse-react-client/src/utils/jsonschema/getRequiredAttributes.test.ts
```diff
@@ -0,0 +1,102 @@
...
+      ],
+    }
+    expect(getRequiredAttributes(schema).sort()).toEqual([
+      'a',
+      'b',
+      'c',
+      'd',
+      'e',
+    ])
+  })
```
nickgros, line 85: This is way to complex for us to handle properly

packages/synapse-react-client/src/utils/jsonschema/getRequiredAttributes.test.ts
```diff
@@ -0,0 +1,102 @@
...
+    expect(getRequiredAttributes(schema).sort()).toEqual(['a', 'b', 'd'])
+  })
+
+  it('avoids infinite recursion on reused schema objects', () => {
+    const shared = { required: ['x'] }
+    const schema: any = {
+      allOf: [shared, { anyOf: [shared, { oneOf: [shared] }] }],
+    }
+    expect(getRequiredAttributes(schema)).toEqual(['x'])
+  })
```
nickgros, line 101: This schema doesn't have infinite recursion, but I think the test could be removed

packages/synapse-react-client/src/utils/jsonschema/getRequiredAttributes.test.ts
```diff
@@ -0,0 +1,102 @@
...
+  it('handles allOf union of required', () => {
+    const schema = {
+      allOf: [
+        { required: ['a', 'b'] },
+        { required: ['b', 'c'] },
+        { properties: { d: { type: 'string' } } },
+      ],
+    }
+    expect(getRequiredAttributes(schema).sort()).toEqual(['a', 'b', 'c'])
+  })
```
nickgros, line 42: This is correct, but in the requirements we don't have to handle this. So up to you if you want to preserve this.

packages/synapse-react-client/src/utils/jsonschema/getRequiredAttributes.ts
```diff
@@ -0,0 +1,42 @@
...
+    if (!isObject(schema) || visited.has(schema)) return
+    visited.add(schema)
+
+    if (isArray(schema.required)) {
+      for (const r of schema.required) {
+        if (typeof r === 'string') collected.add(r)
+      }
+    }
+
+    ;['allOf', 'anyOf', 'oneOf'].forEach(key => {
```
nickgros, line 32: I would just look at `allOf`, since the other cases are conditional/optional 
```suggestion
    ;['allOf'].forEach(key => {
      const group = (schema as any)[key]
      if (isArray(group)) {
        group.forEach(sub => visit(sub))
      }
    })
  }
```

packages/synapse-react-client/src/utils/jsonschema/getRequiredAttributes.ts
```diff
@@ -0,0 +1,42 @@
+import isArray from 'lodash-es/isArray'
+import isObject from 'lodash-es/isObject'
+import { JSONSchema7 } from 'json-schema'
+
+/**
+ * Returns a de-duplicated list of required attribute names from a JSON Schema.
+ * Traverses nested combinators (allOf, anyOf, oneOf).
+ * For:
+ *  - allOf: union of required fields
+ *  - anyOf / oneOf: union (pragmatic; intersection could be too restrictive for most UI uses)
```
nickgros, line 10: The union of required properties specified in `anyOf`/`oneOf` is misleading for the end user (they are not all actually required), so we shouldn't report it.

nickgros (changes requested):

PR #2200, opened 2025-09-12
Refactor to create JsonRx classes for websocket communication

nickgros:


afwillia (approved):
These types and tests are a big improvement!

PR #2205, opened 2025-09-16
PORTALS-3822

lakikowolfe:
Implement SynapseSearchResultsCard Download Button UI. Each entity type's download button actions outlined [here](https://sagebionetworks.jira.com/browse/PORTALS-3790). 

Work to implement button actions can be found here: https://sagebionetworks.jira.com/browse/PORTALS-3821

Jira ticket related to this work: https://sagebionetworks.jira.com/browse/PORTALS-3822


packages/synapse-react-client/src/components/SynapseSearchPageResults/SynapseSearchResultsCard.tsx
```diff
@@ -74,9 +153,15 @@ export function SynapseSearchResultsCard(props: SynapseSearchResultsCardProps) {
...
-          </Button>
+          <DropdownMenu
+            items={downloadMenuItems}
+            dropdownButtonText="Download"
+            buttonTooltip="Download options for this entity"
+            buttonProps={{
+              variant: 'outlined',
+              startIcon: <DownloadIcon />,
+            }}
+          />
```
nickgros, line 164: You may want to consider making this a reusable `EntityDownloadButton` (or some name to that effect) so the download logic is all captured there and decoupled from search. I could see us putting this in a lot of other places!

packages/synapse-react-client/src/components/SynapseSearchPageResults/SynapseSearchResultsCard.tsx
```diff
@@ -46,6 +46,85 @@ const SynapseSearchResultsCardContainer: StyledComponent<PaperProps> = styled(
 }))
 
 export function SynapseSearchResultsCard(props: SynapseSearchResultsCardProps) {
+  // Create menu items based on entity type
+  const createMenuItemsForEntityType = (
```
nickgros, line 50: This function could be created outside of the component and tested in isolation.

I think it may be hard to test this in its current state, especially when we add real logic to the `onClick`s. Just one idea to maybe improve this is to split it up into a couple of functions. You could do something like:

1. Add a new enum called something like `DownloadAction`. These enum values would be your keys in the `dynamicConfigs`, e.g. `'downloadFile'`, `'addToCart'`.
2. `getDownloadActionsForEntityType(type: EntityType): DownloadAction[][]` which would return what you are returning below. You can test this against each EntityType
3. `getMenuItemForAction(entityId: string, entityName: string, entityType: EntityType, downloadAction: DownloadAction): DropdownMenuItem` which will return the correct DownloadMenuItem. Here you could test that the text is correct and the callback works.

You'll still have to plug in a bunch of callbacks into `getMenuItemForAction` when you add that logic, but at least it would be decoupled from building the correct menu, letting you write tests that are more simple.

lakikowolfe:
@nickgros I noticed that the menu drop down has this outline to visually indicate selection on top of the standard synapse grey highlight. How do I remove this?

https://github.com/user-attachments/assets/d499c2a6-5b6b-4e77-8284-2d60762cdb62

nickgros:
@lakikowolfe Do you see the same issue in the base `DropdownMenu` component? It might be browser/OS specific with how focus is shown. I don't see anything in your code related to this, so I wouldn't worry about it

lakikowolfe:
> @lakikowolfe Do you see the same issue in the base `DropdownMenu` component? It might be browser/OS specific with how focus is shown. I don't see anything in your code related to this, so I wouldn't worry about it

Ah yup! It's there for the base component as well. Thanks!

packages/synapse-react-client/src/components/EntityDownloadButton/EntityDownloadButton.tsx
```diff
@@ -0,0 +1,121 @@
...
+    default:
+      return [[DownloadAction.exportTable, DownloadAction.programmaticAccess]]
+  }
+}
+
+export function EntityDownloadButton(props: {
+  entityId: string
+  name: string
+  entityType: string
+}) {
```
nickgros, line 95: entityType should be strongly typed to the enum when possible

```suggestion
export function EntityDownloadButton(props: {
  entityId: string
  name: string
  entityType: EntityType
}) {
```

packages/synapse-react-client/src/components/EntityDownloadButton/EntityDownloadButton.tsx
```diff
@@ -0,0 +1,122 @@
...
+    case 'table':
+      return [[DownloadAction.addToCart, DownloadAction.programmaticAccess]]
+    case 'dockerrepo':
+      return [[DownloadAction.programmaticAccessDocker]]
+    case 'datasetcollection':
+    case 'materializedview':
+    case 'submissionview':
+    case 'virtualtable':
+      return [[DownloadAction.exportTable, DownloadAction.programmaticAccess]]
+    case 'link':
```
lakikowolfe, line 90: @nickgros I can't find much information on the `recordset` entity type and Jay told me not to worry too much about handling the `link` entity type in stand up a couple weeks back. I want to check if you think this is the right menu to display for these entity types?

nickgros (approved):
LGTM pending tests pass!

PR #2207, opened 2025-09-16
SWC-7467

nickgros:
- Support creating a Grid from a RecordSet ID
- Modify SynapseGrid imperative handle
   - Change initializeGrid to accept the CreateGridRequest
   - Add `loadExistingSession` to load from an existing session ID.

In SynapseWebClient, to support instantiating a GridSession from a RecordSet, we will move the Grid UI to its own page. This page will be parameterized by the grid's sessionId. To enable this, SynapseGrid needs to be able to accept the session ID as an argument to the imperative handle function.


packages/synapse-react-client/src/components/DataGrid/SynapseGrid.tsx
```diff
@@ -43,380 +46,399 @@ import {
...
-  } | null>(null)
-
-  useImperativeHandle(
-    ref,
-    () => ({
-      initializeGrid: () => {
-        if (startGridSessionRef.current) {
-          startGridSessionRef.current.handleStartSession(query)
-        }
+export type SynapseGridHandle = {
```
nickgros, line 60: Suggest hiding whitespace changes -- the code formatter reduced indentation in this file, creating a large diff.

packages/synapse-react-client/src/components/DataGrid/StartGridSession.tsx
```diff
@@ -13,175 +19,213 @@ export interface StartGridSessionProps {
...
   onReplicaChange?: (replicaId: number | null) => void
   onPresignedUrlChange?: (url: string) => void
-  query?: string
+  show?: boolean
+}
+
+export type StartGridSessionHandle = {
+  handleStartSession: (request: CreateGridRequest) => void
+  handleLoadSession: (sessionId: string) => void
 }
```
nickgros, line 28: Update ImperativeHandle signature to support 2 cases:

- Start from a `CreateGridRequest`
- Load from a sessionId

packages/synapse-react-client/src/components/DataGrid/SynapseGrid.tsx
```diff
@@ -43,380 +46,399 @@ import {
...
-import { StartGridSession } from './StartGridSession'
+import { StartGridSession, StartGridSessionHandle } from './StartGridSession'
 import { useDataGridWebSocket } from './useDataGridWebsocket'
 import { useGridUndoRedo } from './hooks/useGridUndoRedo'
 import { applyModelChange, ModelChange } from './utils/applyModelChange'
 import { mapOperationsToModelChanges } from './utils/mapOperationsToModelChanges'
 import { getCellClassName } from './utils/getCellClassName'
 
 export type SynapseGridProps = {
-  query: string
```
nickgros, line 54: It was kind of disjointed to pass the query in as a prop, but start the session with an imperative handle call.  Updated this so you just pass the information needed to `initializeGrid`.

packages/synapse-react-client/src/components/DataGrid/useInitializeGridConnection.ts
```diff
@@ -100,15 +72,17 @@ export default function useInitializeGridConnection(
...
-      const session = await createOrGetSession({ gridSqlOrSessionId, schemaId })
-      const sessionId = session.sessionId!
+    mutationFn: async ({ createGridRequest, sessionId }) => {
+      if (!(createGridRequest || sessionId)) {
+        throw new Error(
+          'Must provide either a SQL query, recordSetId, or sessionId to initialize grid connection.',
+        )
+      }
+      const session = await createOrGetSession({ createGridRequest, sessionId })
+      sessionId = session.sessionId!
```
kianamcc, line 85: Reusing the same variable name here is a little confusing to me. Why do we use the same name as the function parameter?

packages/synapse-react-client/src/components/DataGrid/StartGridSession.tsx
```diff
@@ -13,175 +19,213 @@ export interface StartGridSessionProps {
...
+>((props, ref) => {
+  const {
+    show = false,
+    onSessionChange = noop,
+    onReplicaChange = noop,
+    onPresignedUrlChange = noop,
+  } = props
+  const [inputValue, setInputValue] = useState('')
+
+  const { mutate: startSession } = useInitializeGridConnection({
```
kianamcc, line 44: Would it make sense to change the naming of startSession as to not confuse it with handleStartSession? Something like mutateSession so we know it's a mutation function?

kianamcc (approved):

packages/synapse-react-client/src/components/DataGrid/useInitializeGridConnection.ts
nickgros: `sessionId` is the most appropriate name (for the caller to use) for the property in the function parameter. Internally, `sessionId` seems like the most appropriate name for the variable of the session ID that we plan to use to create a replica.

 In this mutation, I can resolve the confusion and locally alias the `args.sessionId` to a variable called `sessionIdArg`.

If the team feels strongly about this pattern then we could enable the [`no-param-reassign` eslint rule](https://eslint.org/docs/latest/rules/no-param-reassign).

PR #2209, opened 2025-09-17
SWC-7471 - Row selection in CRDT

nickgros:
Compute the user's selection and add it to the CRDT


packages/synapse-react-client/src/components/DataGrid/SynapseGrid.tsx
```diff
@@ -422,9 +446,9 @@ const SynapseGrid = forwardRef<SynapseGridHandle, SynapseGridProps>(
                     overflowY: 'auto',
                   }}
                 >
-                  <h3>Model snapshot</h3>
-                  {modelSnapshot ? (
-                    <ComplexJSONRenderer value={modelSnapshot} />
+                  <h3>Model</h3>
+                  {model ? (
+                    <ClickableJsonCrdt model={model} />
```
nickgros, line 451: The author of json-joy also has this `clickable-json` package which provides this utility, which also displays the CRDT ID of each node in the model:

Image

This was useful to validate that the selection object was correct.

afwillia (approved):
Looks good! I slacked you a question

PR #2210, opened 2025-09-17
SWC-7316

jay-hodgson:


jay-hodgson:
Output video attached as a comment in the Jira ticket:  https://sagebionetworks.jira.com/browse/SWC-7316?focusedCommentId=266989

PR #2224, opened 2025-09-22
PORTALS-3743: Prevent text content from showing behind title bar and input box when SynapseChat is inserted into the draggable dialog box component

kianamcc:
Jira: https://sagebionetworks.jira.com/browse/PORTALS-3743?atlOrigin=eyJpIjoiZTEwMTY3ZjI0NjY0NDI4OTlkY2M0YjRhOTU0YWEwYjYiLCJwIjoiaiJ9

Screenshot 2025-09-22 at 10 27 27 AM

No drag on mobile:
Screenshot 2025-09-22 at 10 27 53 AM

### DIFF
--- a/packages/synapse-react-client/src/components/DraggableDialog/DraggableDialog.tsx
+++ b/packages/synapse-react-client/src/components/DraggableDialog/DraggableDialog.tsx
@@ -30,20 +30,35 @@ export default function DraggableDialog({
   const isMobile = useMediaQuery(theme.breakpoints.down('sm'))
   const draggableRef = useRef<HTMLDivElement>(null)
   const [bounds, setBounds] = useState<DraggableBounds>()
-  const position = { x: 100, y: 100 }
+  const [position, setPosition] = useState({ x: 100, y: 100 })
 
   useEffect(() => {
+    // Calculate draggable bounds to keep dialog within viewport with margin
     function updateBounds() {
       if (draggableRef.current) {
         const { offsetWidth } = draggableRef.current
         const margin = 100
 
-        setBounds({
+        const newBounds = {
           left: -(offsetWidth - margin),
           top: 0,
           right: window.innerWidth - margin,
           bottom: window.innerHeight - margin,
-        })
+        }
+
+        setBounds(newBounds)
+
+        // Clamp dialog position to new bounds to keep it visible after window resize
+        setPosition(prevPosition => ({
+          x: Math.max(
+            newBounds.left,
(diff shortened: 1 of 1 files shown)


packages/synapse-react-client/src/components/DraggableDialog/DraggableDialog.tsx
```diff
@@ -23,52 +25,97 @@ export default function DraggableDialog({
...
+          top: -(offsetHeight - margin),
+          right: window.innerWidth - margin,
+          bottom: window.innerHeight - margin,
+        })
+      }
+    }
+    updateBounds()
+    window.addEventListener('resize', updateBounds)
+    return () => window.removeEventListener('resize', updateBounds)
+  }, [])
```
nickgros, line 51: What is the intent of this effect?

packages/synapse-react-client/src/components/DraggableDialog/DraggableDialog.tsx
```diff
@@ -23,52 +25,97 @@ export default function DraggableDialog({
...
+      <Box
+        sx={{
+          height: '100%',
+          maxWidth: '100%',
+          padding: '16px',
+          overflow: 'hidden',
+        }}
+      >
+        {children}
+      </Box>
```
nickgros, line 102: Seems like this could be a `DialogContent`, which already gracefully handles overflow

packages/synapse-react-client/src/components/DraggableDialog/DraggableDialog.tsx
```diff
@@ -23,52 +25,97 @@ export default function DraggableDialog({
...
+          left: '50%',
+          transform: 'translate(-50%, -50%)',
+          width: '90vw',
+          height: '80vh',
+          maxWidth: '600px',
+          maxHeight: '500px',
+        }),
+      }}
+    >
+      <Stack
```
nickgros, line 78: Could you make it so the dialog is only draggable when clicking this 'header' node? And maybe add an appropriate cursor style

I think you would do this using the `handle` prop in react-draggable

packages/synapse-react-client/src/components/DraggableDialog/DraggableDialog.tsx
kianamcc: The intent is to dynamically calculate and update bounds for dragging. I want to prevent users from dragging the dialog outside the viewport so it remains accessible even if they resize their window. Let me know if I should remove this entirely or if you have any suggestions to make it better!

PR #2243, opened 2025-09-30
PORTALS-3821

lakikowolfe:
Implement EntityDownloadButton Menu Item actions

https://sagebionetworks.jira.com/browse/PORTALS-3821

### DIFF
--- a/packages/synapse-react-client/src/components/EntityDownloadButton/EntityDownloadButton.tsx
+++ b/packages/synapse-react-client/src/components/EntityDownloadButton/EntityDownloadButton.tsx
@@ -13,6 +13,9 @@ syn.login(authToken="YOUR_TOKEN_HERE")`
 export const R_CLIENT_IMPORT_AND_LOGIN = `library(synapser)
 synLogin(authToken="YOUR_TOKEN_HERE")`
 
+export const DOCKER_LOGIN =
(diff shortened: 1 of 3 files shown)


packages/synapse-react-client/src/components/EntityDownloadButton/EntityDownloadButton.tsx
```diff
@@ -70,55 +142,186 @@ export function getDownloadActionsForEntityType(
...
+      sql: sql,
+    } as Query,
+    partMask: 0,
+  }
+}
+
 export function EntityDownloadButton(props: {
   entityId: string
   name: string
   entityType: EntityType
```
nickgros: For this to be reusable, I think we would also need to accept `versionNumber` as a prop. Where we would have the following behavior:

- If the version number is present, use it
- If the version number is `null`, do not use it,
- If the version number is `undefined`, then we would try to do the 'smart' thing -- get the latest version for datasets, otherwise don't use it.

However, I think for the search scenario, this is fine. You may want to add a comment to this effect, if you agree.

Some of the utilities you have written may need to change accordingly (e.g. to separate the logical steps of 'getting a version if this is a dataset' and 'including the version in the instructions/query request'), if/when we make that change.

nickgros (approved):

PR #2247, opened 2025-10-01
PORTALS-3848: Fix grid enum select

afwillia:
Fix issue where users typing a value into an enum column and then selecting a drop-down value did not update the cell's value.

### DIFF
--- a/packages/synapse-react-client/src/components/DataGrid/columns/AutocompleteColumn.tsx
+++ b/packages/synapse-react-client/src/components/DataGrid/columns/AutocompleteColumn.tsx
@@ -59,7 +59,7 @@ function AutocompleteCell({
         // Update local input state to match the selected/created value
         setLocalInputState(castCellValueToString(newVal))
       }}
-      blurOnSelect={true}
+      blurOnSelect={false}
       onBlur={_event => {
         // Only update on blur if the input value differs from the current rowData
         // and no option was selected (which would have already updated via onChange)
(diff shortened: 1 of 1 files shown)


packages/synapse-react-client/src/components/DataGrid/columns/AutocompleteColumn.tsx
```diff
@@ -56,9 +56,17 @@ function AutocompleteCell({
           // The value was selected, so explicitly set it
           setRowData(newVal)
         }
+        // Update local input state to match the selected/created value
+        setLocalInputState(castCellValueToString(newVal))
       }}
       blurOnSelect={true}
```
nickgros, line 62: When I try this locally, it still doesn't work the first time I try to pick an enum I have partially typed (but does work subsequent times).

It does seem totally fixed if you also remove `blurOnSelect`, but I'm not totally sure why

afwillia: thanks for trying that , `blurOnSelect={false}` works on my end, too. LLM suggested removing `localInputState` and `onBlur`, which seemed to work well, but it updated the cell value and sent patches on each keystroke.

nickgros (approved):

nickgros:
We should have tests for this, but since we don't already today, let's just go ahead and merge. This LGTM and I'll create a ticket to backfill the tests for this component

afwillia:
Agreed, I'm working on some tests but they are not ready yet. I'll just create another PR for them with that new ticket.

PR #2248, opened 2025-10-01
CLASSIC-6, CLASSIC-9, CLASSIC-12

jay-hodgson:
Ignoring CORS error locally:
Screenshot 2025-10-01 at 2 51 16 PM

Screenshot 2025-10-01 at 3 04 10 PM

Screenshot 2025-10-01 at 3 13 27 PM


apps/portals/classic/src/pages/HomePage.tsx
```diff
@@ -65,6 +73,34 @@ function HomePageInternal() {
         backgroundMp4Css="#024472"
         textAreaWidth="850px"
       />
+      <ImageCardGridWithLinks
+        sql={whatWeDoSql}
+        title="What We Do"
+        summaryText="Supported by the National Institute on Aging, we provide tools and training to make study metadata easier to create, share, and reuse.
```
kianamcc, line 79: Looks like the Figma uses bullet points here. Should we consider supporting markdown so we can render lists and style it closer to the Figma? Could also be a follow-up ticket!

PR #2254, opened 2025-10-02
SWC-7497: resizable draggable dialog

jay-hodgson:


packages/synapse-react-client/src/components/DraggableDialog/DraggableDialog.tsx
```diff
@@ -121,7 +115,7 @@ export default function DraggableDialog({
   )
 
   return (
-    <Box sx={{ position: 'fixed', zIndex: 1000, top: 0 }}>
+    <Box sx={{ position: 'fixed', zIndex: 1000, top: 0, left: 0 }}>
```
jay-hodgson, line 118: This was a tricky bug to fix, that only presented itself in the context of GridAgentChat (not standalone).  Resizing did not work properly without fixing it to left 0

nickgros (approved):

PR #2259, opened 2025-10-03
SWC 7500 - show cell selection when grid is not focused

afwillia:
Store the selected cells in state and add CSS to keep them highlighted when the grid loses focus. This allows the users to see which cells are highlighted when interacting with the chatbot.
image

### DIFF
--- a/packages/synapse-react-client/src/components/DataGrid/SynapseGrid.tsx
+++ b/packages/synapse-react-client/src/components/DataGrid/SynapseGrid.tsx
@@ -57,6 +57,9 @@ const SynapseGrid = forwardRef<SynapseGridHandle, SynapseGridProps>(
     const [session, setSession] = useState<GridSession | null>(null)
     const [replicaId, setReplicaId] = useState<number | null>(null)
     const [chatOpen, setChatOpen] = useState(false)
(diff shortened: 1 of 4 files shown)


packages/synapse-react-client/src/components/DataGrid/SynapseGrid.tsx
```diff
@@ -339,14 +347,28 @@ const SynapseGrid = forwardRef<SynapseGridHandle, SynapseGridProps>(
...
+
+                        // Add selection styling based on lastSelection
+                        const isInSelection =
+                          lastSelection &&
+                          rowIndex >= lastSelection.min.row &&
+                          rowIndex <= lastSelection.max.row &&
+                          colValues.findIndex(col => col.id === columnId) >=
+                            lastSelection.min.col &&
+                          colValues.findIndex(col => col.id === columnId) <=
+                            lastSelection.max.col
```
nickgros, line 366: Should probably just pass `isInSelection` to `getCellClassName`, and `getCellClassName` would apply the correct classname

afwillia: You read my mind! I just updated `getCellClassName` and its unit tests. Feel free to take a look, otherwise I'll merge when the actions finish.

PR #2262, opened 2025-10-03
SWC-7501

jay-hodgson:


jay-hodgson:
Debugging, it surprised me that isExternalFileHandle was set despite enabled=false:
```
const { data: isExternalFileHandle, isLoading: isLoadingExternalFile } =
    useGetIsExternalFileHandle(entityId, {
      enabled: useGetIsExternalFileHandleOptions.enabled,
    })
```
But it must have been queried for previously and cached.  Fixed the logic to respect useGetIsExternalFileHandleOptions.enabled

kianamcc (approved):
I like that you covered the React Query caching issue in your tests! Nice catch!

PR #2280, opened 2025-10-14
SWC-7509 - send client clock on grid reconnect

afwillia:
The grid client sends an empty model clock to the server upon reconnecting to the websocket. This makes the server send the entire patch history back to the client. The client **should** send its clock to the server to reduce the amount of data transferred when reconnecting. This PR implements the desired behavior by tracking the model state across the various websocket and grid components.

TODO - in the event of a long disconnection, users may receive patches from other users upon reconnection. This would manifest with the same undesired behavior. To fix, add a loading screen or notification while the client and server synchronize.

### DIFF
--- a/packages/synapse-react-client/src/components/DataGrid/DataGridWebSocket.ts
+++ b/packages/synapse-react-client/src/components/DataGrid/DataGridWebSocket.ts
@@ -236,10 +236,6 @@ export class DataGridWebSocket {
     this.throttledSendPatch()
   }
 
-  public flushPendingPatch() {
-    this.throttledSendPatch.flush()
-  }
-
   private sendPatchImmediate() {
     if (!this.model) {
       console.warn('Model is not initialized. Cannot send patch.')
--- a/packages/synapse-react-client/src/components/DataGrid/useDataGridWebsocket.ts
+++ b/packages/synapse-react-client/src/components/DataGrid/useDataGridWebsocket.ts
@@ -211,6 +211,7 @@ export function useDataGridWebSocket() {
       state.isConnecting ||
       isEstablishingWebsocketConnection ||
       errorEstablishingWebsocketConnection ||
+      state.connectionError ||
       !isDocumentVisible
     ) {
       return
--- a/packages/synapse-react-client/src/components/DataGrid/DataGridWebSocket.test.ts
+++ b/packages/synapse-react-client/src/components/DataGrid/DataGridWebSocket.test.ts
@@ -16,7 +16,7 @@ describe('DataGridWebSocket', () => {
     public onerror = () => {}
   }
 
-  it('sends multiple messages for a large patch', () => {
+  it('sends multiple messages for a large patch', async () => {
     const mockSocket = new MockWebSocket()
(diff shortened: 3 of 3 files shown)


packages/synapse-react-client/src/components/DataGrid/useDataGridWebsocket.ts
```diff
@@ -23,10 +23,24 @@ export function useDataGridWebSocket() {
     replicaId: number
     sessionId: string
   } | null>(null)
+  // Track the last connection parameters to detect changes - use ref to avoid callback recreation
```
afwillia: track the session info to determine when a model should be kept (when connecting to exisiting session) or cleared (connecting to a new session)

packages/synapse-react-client/src/components/DataGrid/useDataGridWebsocket.ts
```diff
@@ -70,9 +121,17 @@ export function useDataGridWebSocket() {
...
+    websocketOptionsWithoutModel,
   ])
 
+  /**
+   * Establish the WebSocket connection when conditions are met.
+   * Delegates to the mutation hook for presigned URL fetch, WebSocket creation, and retry logic.
+   */
+  useEffect(() => {
+    establishConnectionCallback()
+  }, [establishConnectionCallback])
```
nickgros: This is kind of an awkward pattern since it obscures which dependencies will trigger a connection. 

A nice medium might be to define the function in the effect:
```suggestion
  /**
   * Establish the WebSocket connection when conditions are met.
   * Delegates to the mutation hook for presigned URL fetch, WebSocket creation, and retry logic.
   */
  useEffect(() => {
    function establishConnection() {
      // ...
    }
    establishConnection()
  }, [
    shouldEstablishWebsocketConnection,
    connectionParams,
    establishWebsocketConnection,
    websocketOptionsWithoutModel
  ])
```

packages/synapse-react-client/src/components/DataGrid/SynapseGrid.tsx
```diff
@@ -125,14 +162,7 @@ const SynapseGrid = forwardRef<SynapseGridHandle, SynapseGridProps>(
...
 
-    const commit = useCallback(
-      throttle(() => {
-        console.log('Auto-committing changes')
-        // The model has already been updated in handleChange, just send the patch
-        websocketInstance?.sendPatch()
-      }, 500),
-      [websocketInstance],
-    )
+    const commit = useThrottledSendPatch(websocketInstance)
```
afwillia: split this into its own function with improved logic for handling debouncing. After reconnecting, kept getting console errors stating the websocket was closed when attempting to send messages.

packages/synapse-react-client/src/components/DataGrid/DataGridWebSocket.ts
```diff
@@ -101,6 +119,10 @@ export class DataGridWebSocket {
     }
   }
 
+  public getModel(): GridModel | null {
+    return this.model
+  }
```
nickgros: does not seem to be used

PR #2284, opened 2025-10-15
PORTALS-3869

jay-hodgson:


packages/synapse-react-client/src/components/Authentication/AuthenticationMethodSelection.tsx
```diff
@@ -2,20 +2,40 @@ import SynapseClient from '@/synapse-client'
...
+  if (crypto?.getRandomValues) {
+    const bytes = new Uint8Array(32)
+    crypto.getRandomValues(bytes)
+    return Array.from(bytes, byte => byte.toString(16).padStart(2, '0')).join(
+      '',
+    )
+  }
+  if (crypto?.randomUUID) {
+    return crypto.randomUUID().replace(/-/g, '')
+  }
```
nickgros: Which is preferred? `getRandomValues` is more widely supported than `randomUUID`, so I think `randomUUID` would almost never be used here.

packages/synapse-react-client/src/components/Authentication/AuthenticationMethodSelection.test.tsx
```diff
@@ -0,0 +1,142 @@
...
+}))
+
+const mockSynapseClient = vi.mocked(SynapseClient)
+
+afterEach(() => {
+  vi.clearAllMocks()
+  window.location.hash = ''
+})
+
+describe('AuthenticationMethodSelection', () => {
```
nickgros: Thanks for the test backfill

nickgros (approved):

PR #2296, opened 2025-10-21
SWC-7521: on tree state reset, delete cached queries.  listen for que…

jay-hodgson:
…ry invalidation on the tree (root), and reset the tree data state if encountered

### DIFF
--- a/packages/synapse-react-client/src/components/EntityTreeTable/hooks/useEntityTreeState.test.ts
+++ b/packages/synapse-react-client/src/components/EntityTreeTable/hooks/useEntityTreeState.test.ts
@@ -32,23 +32,18 @@ vi.mock('@/synapse-queries/entity/useGetEntityChildren', () => ({
 const mockUseGetEntityHeader = vi.mocked(useGetEntityHeader)
 const mockUseGetEntityChildren = vi.mocked(useGetEntityChildren)
 
-let wrapperSetup: ReturnType<typeof createWrapperAndQueryClient> | undefined
-
 const renderUseEntityTreeState = (
   rootId = 'syn123',
   expandRootByDefault = true,
   showRootNode = true,
 ) => {
-  const activeWrapperSetup = wrapperSetup ?? createWrapperAndQueryClient()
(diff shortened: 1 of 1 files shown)


packages/synapse-react-client/src/components/EntityTreeTable/hooks/useEntityTreeState.test.ts
```diff
@@ -3,16 +3,27 @@ import { useEntityTreeState } from './useEntityTreeState'
...
 // Mock the query hooks
-vi.mock('@/synapse-queries', () => ({
-  useGetEntityHeader: vi.fn(),
-}))
+vi.mock('@/synapse-queries', async () => {
+  const actual = await vi.importActual<typeof import('@/synapse-queries')>(
+    '@/synapse-queries',
+  )
+  return {
+    ...actual,
```
jay-hodgson: include actual KeyFactory implementation

packages/synapse-react-client/src/components/EntityTreeTable/hooks/useEntityTreeState.ts
```diff
@@ -73,36 +83,72 @@ export const useEntityTreeState = (
...
+    () => keyFactory.getEntityChildrenQueryKey(rootChildrenRequest, false),
+    [keyFactory, rootChildrenRequest],
+  )
+
   // Reset tree data when sorting changes
   const resetTreeData = useCallback(() => {
+    void queryClient.removeQueries({
+      queryKey: rootChildrenQueryKey,
+      exact: true,
+    })
```
nickgros: Why do we need to `removeQueries`? Never used that API

jay-hodgson: It's my attempt at removing cached results on `resetTreeData()` without triggering the new subscription to `invalidate` events
If the `useGetEntityChildrenInfinite` approach works, then I wouldn't need any of this complexity though

PR #2299, opened 2025-10-21
SWC-7518

jay-hodgson:


packages/synapse-react-client/src/components/DraggableDialog/DraggableDialog.tsx
```diff
@@ -115,7 +115,15 @@ export default function DraggableDialog({
...
 
   return (
-    <Box sx={{ position: 'fixed', zIndex: 1000, top: 0, left: 0 }}>
+    <Box
+      sx={{
+        position: 'fixed',
+        zIndex: 1000,
+        top: 0,
+        left: 0,
+        pointerEvents: isMobile ? 'auto' : 'none',
```
jay-hodgson: Allow pointer events through this fixed container, but continue to capture pointer events on the transformed/translated element

PR #2303, opened 2025-10-22
PORTALS-3872

lakikowolfe:
When "Add to Download Cart" Button is clicked files are added to the download cart.

https://sagebionetworks.jira.com/browse/PORTALS-3872

### DIFF
--- a/apps/portals/cancercomplexity/src/config/synapseConfigs/datasets.ts
+++ b/apps/portals/cancercomplexity/src/config/synapseConfigs/datasets.ts
@@ -41,7 +41,18 @@ export const datasetSchema: TableToGenericCardMapping = {
     'consortium',
   ],
   dataTypeIconNames: 'dataType',
+  // override Download Cart to use datasetAlias
   downloadCartSynId: 'datasetAlias',
+  synapseEntityConfig: {
+    id: {
+      source: 'column',
+      columnName: 'datasetId',
+    },
+    version: {
+      source: 'column',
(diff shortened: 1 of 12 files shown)


packages/synapse-react-client/src/components/EntityDownloadButton/EntityDownloadButton.tsx
```diff
@@ -265,6 +294,26 @@ export function EntityDownloadButton(props: {
     props.entityType,
   )
 
+  // Get context and download functionality
+  const { downloadCartPageUrl } = useSynapseContext()
+  const { mutate: addFileToDownloadList } = useAddFileToDownloadList({
+    onSuccess: () => {
+      displayFilesWereAddedToDownloadListSuccess(downloadCartPageUrl)
```
lakikowolfe: FIXME: This message displays to the user that files have been successfully added to their download cart even when there are zero files in a folder.

packages/synapse-react-client/src/components/EntityDownloadButton/EntityDownloadButton.tsx
```diff
@@ -265,6 +304,33 @@ export function EntityDownloadButton(props: {
...
+    },
```
nickgros: Probably also need to check the `numberOfFilesAdded` here since it could be an empty dataset

nickgros (approved):

PR #2312, opened 2025-10-24
PORTALS-3897

lakikowolfe:
Add chip that shows access level of entity

Used the HasAccessV2 component as the base of this chip. It is wrapped in a function so that restriction type is always displayed. Used css to modify HasAccessV2 to look like an mui chip.

https://sagebionetworks.jira.com/browse/PORTALS-3897

### DIFF
--- a/packages/synapse-react-client/src/components/HasAccess/HasAccessV2.tsx
+++ b/packages/synapse-react-client/src/components/HasAccess/HasAccessV2.tsx
@@ -37,9 +37,7 @@ export function HasAccessV2(props: HasAccessProps) {
   const {
     entityId,
     showButtonText = true,
-    wrap = true,
     showExternalAccessIcon = false,
-    sx,
   } = props
 
   const {
@@ -48,7 +46,7 @@ export function HasAccessV2(props: HasAccessProps) {
     icon,
     handleGetAccess,
     accessRequirementDialog,
-  } = useHasAccess(entityId, wrap, sx, { showExternalAccessIcon })
+  } = useHasAccess(entityId, true, undefined, { showExternalAccessIcon })
 
   const { data: restrictionInformation } = useGetRestrictionInformation({
     restrictableObjectType: RestrictableObjectType.ENTITY,
@@ -138,8 +136,6 @@ export function HasAccessV2(props: HasAccessProps) {
     props.className,
     showButtonText,
     accessRequirementDialog,
-    wrap,
-    sx,
     accessText,
     icon,
   ])
@@ -149,13 +145,11 @@ export function HasAccessV2(props: HasAccessProps) {
     return <></>
   }
 
-  if (wrap) {
-    return (
-      <span style={{ whiteSpace: 'nowrap' }}>
-        {accessRequirementsJsxOrIconContainer}
-      </span>
-    )
-  }
+  return (
+    <span style={{ whiteSpace: 'nowrap' }}>
+      {accessRequirementsJsxOrIconContainer}
+    </span>
+  )
   return accessRequirementsJsxOrIconContainer
 }
 


packages/synapse-react-client/src/components/SynapseSearchPageResults/HasAccessWithText.module.css
```diff
@@ -0,0 +1,71 @@
...
+  text-overflow: ellipsis;
+  padding-left: 0;
+  padding-right: 0;
+}
+
+/* Match EntityType chip styling exactly */
+/* No hover effects to match EntityType chip behavior */
+
+/* Disabled state */
+.hasAccessChip[data-disabled='true'] {
```
nickgros: When is this disabled?

packages/synapse-react-client/src/components/SynapseSearchPageResults/HasAccessWithText.module.css
```diff
@@ -0,0 +1,71 @@
...
+
+.hasAccessChip {
+  display: inline-flex;
+  align-items: center;
+  justify-content: center;
+  height: 32px;
+  border-radius: 16px;
+  padding: 0 12px;
+  margin-left: 8px;
+  font-family: 'DM Sans', 'Helvetica', 'Arial', sans-serif;
```
nickgros: Setting the font should not be necessary (it should be inherited)...or is it? 

If it is not necessary, let's remove it. If it is, let's introduce a new CSS Variable

packages/synapse-react-client/src/components/SynapseSearchPageResults/HasAccessWithText.tsx
```diff
@@ -0,0 +1,54 @@
...
+  size?: 'normal' | 'small'
+}
+
+export function HasAccessWithText({
+  entityId,
+  size = 'normal',
+}: HasAccessWithTextProps) {
+  const restrictionUiType = useGetRestrictionUiType(entityId, {
+    enabled: false,
+  })
```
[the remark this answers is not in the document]
lakikowolfe: Do you think it should be enabled?

packages/synapse-react-client/src/components/SynapseSearchPageResults/HasAccessWithText.module.css
```diff
@@ -0,0 +1,71 @@
...
+
+.hasAccessText {
+  overflow: hidden;
+  text-overflow: ellipsis;
+  padding-left: 0;
+  padding-right: 0;
+}
+
+/* Match EntityType chip styling exactly */
+/* No hover effects to match EntityType chip behavior */
```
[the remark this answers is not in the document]
lakikowolfe: I can't tell based on the figma file if clicking the chip does anything. Will follow up with Shysta about this.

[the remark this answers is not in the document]
lakikowolfe: ohh I can see that the button click is built in to the HasAccess icon

lakikowolfe:
removing the span from EntityTypeIcon messes with the layout in other places

With span
Screenshot 2025-10-29 at 1 46 34 PM

Without
Screenshot 2025-10-29 at 1 46 46 PM

Taking a closer look at the function the `wrap` param that you mentioned was not actually implemented in EntityTypeIcon. I'm going to go ahead and add a wrap prop that defaults to `wrap = true`.

packages/synapse-react-client/src/components/SynapseSearchPageResults/HasAccessWithText.module.css
```diff
@@ -0,0 +1,71 @@
...
+  justify-content: center;
+  height: 32px;
+  border-radius: 16px;
+  padding: 0 12px;
+  margin-left: 8px;
+  font-family: 'DM Sans', 'Helvetica', 'Arial', sans-serif;
+  font-size: 14px;
+  font-weight: 500;
+  color: #265149;
+  background-color: #dae9e7;
```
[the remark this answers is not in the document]
lakikowolfe: I took this route and refactored to use the MUI chip as a base and share sx props via chipStyles.ts

packages/synapse-react-client/src/components/SynapseSearchPageResults/HasAccessChip.tsx
```diff
@@ -0,0 +1,51 @@
...
+    return null
+  }
+  return (
+    <Chip
+      label={getAccessText()}
+      icon={
+        <HasAccessV2 entityId={entityId} showButtonText={false} wrap={false} />
+      }
+      sx={{ ...searchResultsCardChipStyles }}
+    />
```
nickgros: This ends up being a bit odd where it may say "View Terms" but you can only click the icon, and not the text/entire chip to actually view them, which is a weird user experience.

I think fixing this might require a bit of a refactor. Could we extract a hook from HasAccessV2 and reuse it in both places? The idea is that we could
* Have a `useHasAccess` hook that handles the logic for computing the correct access level, getting the text, and the icon
* Multiple components can use this to render it using whatever layout/design it needs to

So then this component would be something like:

```tsx
export function HasAccessChip(props: HasAccessChipProps) {
  const [showAccessRequirementModal, setShowAccessRequirementModal] = useState(false)
  const { restrictionUiType, icon, accessRequirementModal } = useHasAccess(entityId)


  return (<>
      <Chip 
        label={getAccessText(restrictionUiType)}
        icon={icon}
        sx={{ ...searchResultsCardChipStyles }}
        onClick={() => setShowAccessRequirementModal(true)}
      >
      {accessRequirementModal}
    </>)
}

```

packages/synapse-react-client/src/components/SynapseSearchPageResults/HasAccessChip.tsx
```diff
@@ -0,0 +1,51 @@
...
+  entityId: string
+  size?: 'normal' | 'small'
+}
+
+export function HasAccessChip({ entityId }: HasAccessChipProps) {
+  const restrictionUiType = useGetRestrictionUiType(entityId, {
+    enabled: false,
+  })
+
+  const getAccessText = () => {
```
nickgros: Move this out of the component and pass `restrictionUiType` as a parameter

lakikowolfe: done

packages/synapse-react-client/src/components/HasAccess/HasAccessV2.tsx
```diff
@@ -127,6 +129,116 @@ function useGetIsExternalFileHandle(
...
+      return 'External Access'
+    case undefined:
+      return 'Loading...'
+    default:
+      return ''
+  }
+}
+
+// A hook that handles the logic for computing the correct access level, getting the text, and the icon
+export function useHasAccess(entityId: string) {
```
nickgros: Move this to a new file `useHasAccess` (same dir as HasAccessV2 is fine)

nickgros (approved):
Great work!

PR #2321, opened 2025-10-28
SWC-7537: No need to set up portal callback cookie

jay-hodgson:


apps/synapse-portal-framework/src/components/AppInitializer.tsx
```diff
@@ -41,14 +32,9 @@ function AppInitializer(props: PropsWithChildren<Record<never, never>>) {
...
         }
       }
-      if (!cookies || !cookiePreferences.functionalAllowed) {
-        return
-      }
-      let isInvokingDownloadTable: boolean = false
+
       if (ev.target instanceof HTMLAnchorElement) {
         const anchorElement = ev.target
-        isInvokingDownloadTable =
```
jay-hodgson: At one point, I think there was this menu item to download files but it was not supported from the portal, so we redirected to Synapse.org and automatically opened the onDownload dialog in the table.  Menu item no longer exists, and I think this functionality has been implemented on the portal side after this code was written

PR #2548, opened 2026-03-03
SWC-7719 -  fix case where errorMessage is null

nickgros:


jay-hodgson (approved):
This seems like a safe change to merge

PR #2632, opened 2026-03-31
SWC-7761 - track edited cells in Curator grid

afwillia:
Use the CRDT model clock to determine edits that have been made since the user joined the grid session. Get replica IDs from the clock to attribute changes to a user or agent. Mark the upper right corner of edited cells; a different color for the user, their agent, remote users, and agents of remote users. Change the validation background to a gradient so it does not collide with change decorations.


packages/synapse-react-client/src/components/DataGrid/SynapseGrid.tsx
nickgros: Can we call the API to get all of the replicas before `modelRowsToGrid`, then pass that in here and avoid tracking replicas?

packages/synapse-react-client/src/components/DataGrid/hooks/useActiveReplicas.ts
```diff
@@ -0,0 +1,79 @@
...
+ *
+ * Strategy: when `hasCompletedInitialSync` first becomes true, snapshot every
+ * peer's clock time as a "baseline".  Any peer whose clock time increases past
+ * that baseline has written at least one CRDT operation (cell edit, row
+ * create/delete, selection update) and is considered active.
+ *
+ * This approach covers bots (isAgentReplica: true) which never call
+ * SET_SELECTION but do write row/cell data via UPDATE operations.
+ */
+export function useActiveReplicas(
```
nickgros: I am hopeful that we can remove this if we make the following changes

- Fetch replica information from API first ([using this API](https://rest-docs.synapse.org/rest/POST/grid/session/list.html))
- When transforming the model to rows, cross reference the replica IDs with the API information
- Invalidate the API data when a new replica connects (using the 'replica-connected' websocket event)

packages/synapse-react-client/src/components/DataGrid/DataGridWebSocket.ts
```diff
@@ -216,21 +237,75 @@ export class DataGridWebSocket {
           this.replicaId,
         ) as unknown as GridModel
         this.onModelCreate(this.model)
+        // Seed the baseline from the initial model
+        this.rowsBaseline = this.model.api.getSnapshot().rows
       } else {
+        // Extract author sid before applying (each patch has its own author;
+        // for the common single-patch case this is exact; for rare multi-author
+        // batches we attribute to the first patch's author).
+        const firstPatch = patches[0]
```
afwillia: @nickgros can you verify this assumption true, even with multi patches per message?

packages/synapse-react-client/src/components/DataGrid/hooks/useCellChangeTracker.ts
```diff
@@ -0,0 +1,58 @@
+import { useCallback, useRef, useState } from 'react'
+import type { CellChangeInfo } from '@/components/DataGrid/DataGridWebSocket'
+
+/** Map key encoding a specific cell position. */
+export function cellChangeKey(rowIndex: number, colName: string): string {
+  return `${rowIndex}:${colName}`
+}
```
nickgros: Not sure if `rowIndex` is the best way to identify a row. Maybe we should use the row's CRDT ID (rep,seq)?

If this works though, I think it's fine.

Test the case where you insert a row above another row that already has tracked changes.

packages/synapse-react-client/src/components/DataGrid/hooks/useCellChangeTracker.ts
```diff
@@ -0,0 +1,58 @@
...
+  // on each update. Returning a new reference is necessary so that useCallback
+  // dependency checks in consumers (e.g. DataGrid) detect the change and
+  // recreate the cell-class callback, triggering a re-render of the grid.
+  const cellChangesRef = useRef<Map<string, number>>(new Map())
+  const [cellChanges, setCellChanges] = useState<ReadonlyMap<string, number>>(
+    cellChangesRef.current,
+  )
+
+  const recordChanges = useCallback(
+    (authorSid: number, changes: CellChangeInfo[], isService: boolean) => {
```
nickgros:
```suggestion
    (gridReplicaInfo: GridReplicaInfo, changes: CellChangeInfo[]) => {
```

packages/synapse-react-client/src/components/DataGrid/hooks/useCellChangeTracker.test.ts
```diff
@@ -0,0 +1,146 @@
...
+
+    act(() => {
+      result.current.recordChanges(1, [], false)
+    })
+
+    // Same reference means no setState call was triggered
+    expect(result.current.cellChanges).toBe(mapBefore)
+  })
+
+  it('applies last-writer-wins: a second recordChanges call for the same cell overwrites the authorSid', () => {
```
nickgros: Is it possible for us to derive the cell change state from the current state of the model, instead of listening to changes (where our implementation could diverge from json-joy's)?

packages/synapse-react-client/src/components/DataGrid/hooks/useGridReplicaUsers.ts
```diff
@@ -0,0 +1,90 @@
...
+    ],
+    [attributableReplicas],
+  )
+
+  const profileQueries = useQueries({
+    queries: uniqueUserIds.map(userId => ({
+      queryKey: keyFactory.getUserProfileQueryKey(userId),
+      queryFn: () => SynapseClient.getUserProfileById(userId, accessToken),
+      staleTime: 5 * 60 * 1000, // profiles are stable within a session
+    })),
```
nickgros, line 51: Consider refactoring so `useGetUserProfile` uses `queryOptions` like here: https://github.com/Sage-Bionetworks/synapse-web-monorepo/blob/02cb0e78780798b2722dc2a14ceeaa4fefb7a257/packages/synapse-react-client/src/synapse-queries/grid/useGridSession.ts#L146

You could then map to create these same `queryOptions` objects here. Then the `queryKey` and `queryFn` are defined in only one place. You can keep the `staleTime` override here if it helps.

packages/synapse-react-client/src/components/DataGrid/hooks/useGridReplicaUsers.ts
```diff
@@ -0,0 +1,90 @@
...
+      result.set(replica.replicaId, {
+        replicaInfo: replica,
+        category,
+        profile: replica.createdBy
+          ? profileByUserId.get(replica.createdBy)
+          : undefined,
+      })
+    }
+    return result
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
nickgros: Add all deps to the array, or add a comment why one is missing.

packages/synapse-react-client/src/components/DataGrid/hooks/useGridReplicaUsers.ts
```diff
@@ -0,0 +1,90 @@
...
+  const profileQueries = useQueries({
+    queries: uniqueUserIds.map(userId => ({
+      queryKey: keyFactory.getUserProfileQueryKey(userId),
+      queryFn: () => SynapseClient.getUserProfileById(userId, accessToken),
+      staleTime: 5 * 60 * 1000, // profiles are stable within a session
+    })),
+  })
+
+  // Stable cache key: only changes when a query's data actually updates
+  const profileDataKey = profileQueries.map(q => `${q.dataUpdatedAt}`).join(',')
```
nickgros: Instead of this, try to use `useDeepCompareMemoize` (or something like this, we have a utility that does this) which will ensure `profileQueries` is stable assuming the actual query info isn't changing.

If that  doesn't work, then you can keep this, and make a note about `profileQueries` changing in the eslint ignore below

packages/synapse-react-client/src/components/DataGrid/DataGridWebSocket.ts
nickgros: Won't work; the batch may contain patches with different authors.

packages/synapse-react-client/src/components/DataGrid/hooks/useCellChangeTracker.ts
nickgros: See `modelRowsToGrid` - the logical timestamp of the row could be a better key than the row index

```
    rowObj[GRID_ROW_REACT_KEY_PROPERTY] = logicalTimestampToString(
      model.api.obj(['rows', String(rowIndex)]).node.id,
    )
```

packages/synapse-react-client/src/components/DataGrid/DataGridWebSocket.ts
nickgros: In `modelRowsToGrid`, when we process each cell, can we extract the replica ID with something like this (might not be exactly right)
```ts
model.api.obj(['rows', String(rowIndex), 'data', 'rows' i]).node.id,
```

Then derive/compute the map in that step rather than on each change

nickgros (changes requested):
General idea looks good, just need to nail down how we get the model replica information out and transform it

packages/synapse-react-client/src/style/components/_data-grid-extra.scss
```diff
@@ -228,6 +233,42 @@ body.col-resizing {
...
+    right: 0;
+    width: 0;
+    height: 0;
+    border-style: solid;
+    border-width: 0 7px 7px 0;
+    border-color: transparent $color transparent transparent;
+    z-index: 20;
+    pointer-events: none;
+  }
+}
```
nickgros, line 253: It will not be obvious to users what these indicators mean unless we can add alt/help/tooltip text. Has design considered this issue already and suggested a solution? 

Absent a recommendation from design, I think we would probably want tooltips. To do this, I think the indicator would need to be an actual DOM node to which we can attach a tooltip.

WDYT? Should such a change be in this PR, or in a future PR?

packages/synapse-react-client/src/components/DataGrid/DataGrid.tsx
```diff
@@ -40,6 +41,8 @@ type DataGridProps = {
     rowIndex: number | null,
     row: DataGridRow | null,
   ) => void
+  model?: GridModel | null
```
nickgros: One intent of the `DataGrid` component was to decouple it from the CRDT logic. Passing the model & replica IDs breaks this decoupling, so I would prefer if we could avoid it and pass in data that is already processed.

One option would be to extend `DataGridRow` to include cell-level attribution information, and process that information when those are created. But you could come up with something else

packages/synapse-react-client/src/components/DataGrid/utils/getCellClassName.ts
```diff
@@ -54,12 +68,26 @@ export function getCellClassName(params: {
...
+    if (colIndex !== -1) {
+      const authorSid = getCellAuthorSid(model, rowIndex, colIndex)
+      if (authorSid !== null) {
+        const info = replicaUserMap?.get(authorSid)
+        if (info) {
+          classList.push(`cell-changed--${info.category}`)
+        }
+      }
+    }
+  }
```
nickgros: As mentioned in a comment on `DataGrid.tsx`, this method should only translate precomputed attribution information into correct UI.

The actual attribution should be computed in SynapseGrid (or in a hook used by SynapseGrid)

packages/synapse-react-client/src/style/components/_data-grid-extra.scss
afwillia: I like doing this now. It also aligns with design's idea of showing who is editing a cell in a tooltip.

afwillia:
@nickgros I updated this PR to include a tooltip displaying who made the change. I also incorporated design's suggestion to indicate agent edits with an icon instead of a corner mark.

packages/synapse-react-client/src/components/DataGrid/utils/columnFactory.tsx
```diff
@@ -14,6 +16,66 @@ import {
...
+                transform: 'translateY(-50%)',
+                zIndex: 21,
+                pointerEvents: 'auto',
+                display: 'block',
+              }}
+            />
+          </Tooltip>
+        )}
+      </div>
+    )
```
nickgros: Some style suggestions:

- Replace the `div` with a fragment--these get wrapped in a flex container, which will maintain the alignment and give us a reasonable layout for the input and agent icon
- Update the agent icon to align within the flex layout so it doesn't doesn't overlap with the cell content
- Update the agent icon to hide when the input is focused so it doesn't overlap with the input content when editing
- Update the icon color to match design


```suggestion  
    <>
        <OriginalComponent {...props} />
        {changeInfo && !isAgent && (
          <Tooltip title={changeInfo.tooltipText} placement="top-end">
            <div
              style={{
                position: 'absolute',
                top: 0,
                right: 0,
                width: 7,
                height: 7,
                zIndex: 21,
                pointerEvents: 'auto',
              }}
            />
          </Tooltip>
        )}
        {isAgent && !props.focus && (
          <Tooltip title={changeInfo.tooltipText} placement="left">
            <SmartToyTwoTone
              sx={{
                color: 'grey.600',
                fontSize: '1em',
                mr: '5px',
                alignSelf: 'center',
                justifySelf: 'flex-end',
                zIndex: 21,
                pointerEvents: 'auto',
                display: 'block',
              }}
            />
          </Tooltip>
        )}
      </>
    )
```

Before:

Image

Image

After:

Image

Image

packages/synapse-react-client/src/components/DataGrid/utils/enrichRowsWithChangeInfo.ts
```diff
@@ -0,0 +1,54 @@
...
+ * Enriches each row in `rows` with a `__cellChangeInfo` map that records the
+ * change category and tooltip text for every cell whose last author is present
+ * in `replicaUserMap`.
+ *
+ * Short-circuits (returning the original array reference) when the model or
+ * snapshot is absent, or when no replica attribution data is available yet.
+ * Rows with no attributed cells are returned by reference to preserve React
+ * referential equality.
+ */
+export function enrichRowsWithChangeInfo(
```
nickgros, line 21: For performance, we may want to incorporate this in `modelRowsToGrid`, and just block loading until we have all the data we need. This is another case where we are iterating over the domain model (CRDT document) to transform it into a view model (DataGridRow[]). Ideally we would only do that once.

I think we could make that optimization in a future pass, and prioritize it based on how much this actually impacts performance. Not a blocker to merge this PR IMO.

packages/synapse-react-client/src/components/DataGrid/utils/applyModelChange.ts
```diff
@@ -73,10 +74,14 @@ export function applyModelChange(
...
 
     case 'UPDATE': {
+      const currentRowData = model.api.getSnapshot().rows[change.rowIndex]?.data
       Object.entries(change.updatedData).forEach(([key, value]) => {
         if (key.startsWith('_')) return // Skip internal properties like _rowId
         const colIndex = columnNames.indexOf(key)
         if (colIndex !== -1) {
+          // Only write cells whose value actually changed to avoid stamping
+          // the local replica's SID on unmodified cells.
+          if (isEqual(currentRowData?.[colIndex], value)) return
```
nickgros, line 84: Please ensure this is tested in `applyModelChange` and verifies that the timestamp of the cell does not change

nickgros (approved):

PR #2666, opened 2026-04-08
Vite 8/Vitest 4

nickgros:


apps/portals/adknowledgeportal/package.json
```diff
@@ -20,7 +20,8 @@
     "tslib": "^2.8.1",
     "typescript": "5.8.3",
     "vite-config": "workspace:*",
-    "vitest": "^3.2.2"
+    "vite-plugin-node-polyfills": "^0.26.0",
```
nickgros, line 24: Upgraded `vite-plugin-node-polyfills`. There was an internal change that requires us to include it as a direct dependency in each package that builds a browser application (all portals, SageAccountWeb, synapse-oauth-signin, and SRC for Storybook).

The breaking change that forces us to include it in each package is related to how the app sources the polyfill code and how pnpm hoists it. Basically, unless the app includes `vite-plugin-node-polyfills` as a direct dependency, the polyfill/shim cannot be resolved by pnpm.

apps/SageAccountWeb/src/components/TwoFactorAuth/ResetTwoFactorAuth.test.tsx
```diff
@@ -34,10 +35,9 @@ describe('ResetTwoFactorAuth', () => {
     expiresOn: new Date().toISOString(),
     createdOn: new Date().toISOString(),
   }
-  const hexEncodedResetToken = Buffer.from(
-    JSON.stringify(mockTwoFactorAuthResetToken),
-    'utf-8',
-  ).toString('hex')
```
nickgros, line 40: Re-implemented without using the Node.js Buffer API (added and ran tests before/after change so this should be safe)

apps/synapse-portal-framework/src/sitemap/queryTableForSitemap.test.ts
```diff
@@ -11,16 +11,18 @@ import {
 } from '@sage-bionetworks/synapse-client'
 
 // Mock the synapse-client module
-vi.mock('@sage-bionetworks/synapse-client', () => {
+vi.mock('@sage-bionetworks/synapse-client', function () {
```
nickgros, line 14: In order for Vitest to support mocking class constructors, Vitest 4 no longer lets us implement mocks using arrow functions

https://vitest.dev/guide/migration.html#spyon-and-fn-support-constructors

apps/SageAccountWeb/src/URLUtils.test.ts
```diff
@@ -0,0 +1,51 @@
+import { describe, it, expect } from 'vitest'
+import { hexDecodeAndDeserialize, serializeAndHexEncode } from './URLUtils'
+
+describe('hexDecodeAndDeserialize Compatibility', () => {
```
nickgros, line 4: New tests for existing method that was reimplemented because it unnecessarily used the Node.js Buffer API, which is not natively implemented in browsers (depended on a polyfill!). Ran tests against both implementations so this should be safe

apps/synapse-oauth-signin/src/mocks/handlers.ts
```diff
@@ -21,7 +21,7 @@ export function resetConsentedInMockService(newValue: boolean) {
 
 let mockPostConsentFn = vi.fn()
 
-export function setMockPostConsentFn(fn: Mock<any>) {
+export function setMockPostConsentFn(fn: Mock<() => void>) {
```
nickgros, line 24: Vitest upgrade requires more strict typing

apps/synapse-oauth-signin/src/index.tsx
```diff
@@ -17,8 +16,3 @@ root.render(
     <RouterProvider router={router} />
   </StrictMode>,
 )
-
-// If you want your app to work offline and load faster, you can change
-// unregister() to register() below. Note this comes with some pitfalls.
-// Learn more about service workers: https://bit.ly/CRA-PWA
-serviceWorker.unregister()
```
nickgros, line 24: Remove unused CRA service worker.

apps/synapse-portal-framework/package.json
```diff
@@ -14,7 +14,8 @@
       "types": "./dist/*.d.ts",
       "import": "./dist/*.js"
     },
-    "./style/*": "./src/style/*"
+    "./style/*": "./src/style/*",
+    "./*.svg": "./src/*.svg"
```
nickgros, line 18: allow NAMS portal to import SVG from this project

apps/synapse-portal-framework/vite.config.ts
```diff
@@ -36,10 +36,22 @@ const config = mergeConfig(
...
       plugins: [
-        tsconfigPathsPlugin(),
         ...reactPlugins(),
         ...libraryPlugins({ preserveModules: true }),
+        viteStaticCopy({
+          targets: [
+            {
+              src: 'src/components/assets/*',
+              dest: '',
+              rename: { stripBase: 1 },
```
nickgros, line 51: Copy SVGs to `dist` folder so NAMS portal can import SVG from this library

packages/synapse-react-client/src/components/OrientationBanner/OrientationBanner.tsx
```diff
@@ -8,35 +8,10 @@ import FullWidthAlert, {
   AlertButtonConfig,
   FullWidthAlertProps,
 } from '../FullWidthAlert/FullWidthAlert'
-
-const OrientationBannerNameStrings = [
```
nickgros, line 12: Moved to break dependency cycle

packages/synapse-react-client/vite.config.ts
```diff
@@ -41,25 +41,29 @@ const config = mergeConfig(
...
+              dest: '',
+              rename: { stripBase: 1 },
+            },
+            {
+              src: 'src/types/*.d.ts',
+              dest: '',
+              rename: { stripBase: 1 },
+            },
+          ],
+        }),
```
nickgros, line 59: Use a vite plugin to copy files instead of a separate build step

packages/synapse-react-client/vite.config.ts
```diff
@@ -41,25 +41,29 @@ const config = mergeConfig(
...
-        commonjsOptions: {
-          // react-datasheet-grid is common-js only and imports tanstack/react-virtual which is an ESM package
-          // for some reason this transitive import is treated as common-js but we can fix it with the config below:
-          esmExternals: (id: string) => {
-            if (id == '@tanstack/react-virtual') {
-              return true
-            }
-            return false
-          },
-        },
```
nickgros, line 61: strongly suspect we do not need this now that we have our own fork for react-datasheet-grid with a modern build system. Seems to work fine when linked to SWC

eslint.config.mjs
```diff
@@ -84,6 +84,8 @@ export default tseslint.config(
       'no-extra-semi': 'off',
       'prefer-const': 'warn',
       'vitest/expect-expect': 'off',
+      'vitest/no-conditional-expect': 'off',
+      'vitest/no-standalone-expect': 'off',
```
nickgros, line 88: Vitest eslint plugin upgrade enabled these rules by default. I don't think they are very helpful:

`vitest/no-conditional-expect` - we use this sparingly in `test.each` cases where we split conditional logic on the input data. The alternative would be much more test code.

`vitest/no-standalone-expect` - we use this in a few utility assertions that simplify test code

packages/synapse-react-client/src/components/SynapseForm/SynapseFormWrapper.stories.tsx
```diff
@@ -8,6 +8,7 @@ const meta = {
   component: SynapseFormWrapper,
   parameters: {
     withRouter: true,
+    requireLogin: true,
```
nickgros:
```suggestion
```

packages/vite-config/src/sitemap-vite-config.ts
```diff
@@ -1,8 +1,7 @@
 import { defineConfig, UserConfig, Plugin } from 'vite'
-import tsconfigPaths from 'vite-tsconfig-paths'
 
 /**
- * Creates a Vite/Rollup plugin that stubs out module imports that match the given patterns.
+ * Creates a Vite plugin that stubs out module imports that match the given patterns.
```
nickgros, line 4: We use this plugin to import and parse the portal routes file without having to load the entire module graph. It's kind of vibe-coded to begin with; I verified that the sitemap output is still what we expect.

apps/portals/namshub/package.json
```diff
@@ -12,7 +12,7 @@
     "react-dom": "^19.1.0",
     "react-router": "^7.13.1",
```
hallieswan: Should react-router be bumped too? 

```suggestion
    "react-router": "^7.14.0",
```

packages/synapse-react-client/package.json
hallieswan: Could `copyfiles` be removed now that `build-copy:assets` has been replaced with `viteStaticCopy`?

packages/vite-config/src/plugins.ts
```diff
@@ -1,15 +1,21 @@
...
-import react from '@vitejs/plugin-react'
+import react, { reactCompilerPreset } from '@vitejs/plugin-react'
+import babel from '@rolldown/plugin-babel'
 import svgr from 'vite-plugin-svgr'
 import { nodePolyfills } from 'vite-plugin-node-polyfills'
 import { externalizeDeps } from 'vite-plugin-externalize-deps'
 import dts from 'vite-plugin-dts'
-import tsconfigPaths from 'vite-tsconfig-paths'
+
+// const require = createRequire(import.meta.url)
```
hallieswan: Is this comment still useful?

hallieswan (approved):
Looks good to me! Just a couple small comments

packages/vite-config/src/plugins.ts
nickgros: Removed

hallieswan (approved):

PR #2671, opened 2026-04-09
Code review changes

nickgros:
- Fix inaccurate boolean comment
- Get user card out of the way of the edit button
- Add CLAUDE.md for metadata-task feature

See now that by default the edit button is not covered by the user card

image

### DIFF
--- a/packages/synapse-react-client/src/utils/hooks/useOverlay.tsx
+++ b/packages/synapse-react-client/src/utils/hooks/useOverlay.tsx
@@ -17,20 +17,22 @@ function resetTimer(timer: NodeJS.Timeout | null) {
 const DEFAULT_DELAY_SHOW_MS = 250
 const DEFAULT_DELAY_HIDE_MS = 500
 
+const defaultAnchorOrigin: PopoverProps['anchorOrigin'] = {
+  vertical: 'top',
+  horizontal: 'left',
+}
+
+const defaultTransformOrigin: PopoverProps['transformOrigin'] = {
+  vertical: 'bottom',
+  horizontal: 'left',
+}
+
 export function useOverlay(
   children: ReactNode,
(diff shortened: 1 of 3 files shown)


packages/synapse-react-client/src/utils/hooks/useOverlay.tsx
```diff
@@ -17,20 +17,22 @@ function resetTimer(timer: NodeJS.Timeout | null) {
...
   delayHide = DEFAULT_DELAY_HIDE_MS,
-  paperProps: PaperProps = {},
-  anchorOrigin: PopoverProps['anchorOrigin'] = {
-    vertical: 'top',
-    horizontal: 'right',
-  },
-  transformOrigin: PopoverProps['transformOrigin'] = {
-    vertical: 'center',
-    horizontal: 'left',
-  },
```
nickgros, line 33: These parameters were never used, so I pulled them out of the `useOverlay` signature (this could have been causing an infinite render bug because these new objects would be created each call). I changed the default values to reduce the likelihood the edit button is covered if it appears.

packages/synapse-react-client/src/utils/hooks/useOverlay.tsx
```diff
@@ -17,20 +17,22 @@ function resetTimer(timer: NodeJS.Timeout | null) {
...
+  vertical: 'top',
+  horizontal: 'left',
+}
+
+const defaultTransformOrigin: PopoverProps['transformOrigin'] = {
+  vertical: 'bottom',
+  horizontal: 'left',
+}
+
 export function useOverlay(
```
nickgros, line 30: This is a custom hook that controls the user card popover shown when hovering on the user badge. It is not used anywhere else.

afwillia (approved):

PR #2672, opened 2026-04-09
SWC-7756 part 2 - display concurrent user selections

afwillia:
Tints cells covered by another connected replica's active selection with a semi-transparent color (replicaId % 8 for a stable per-session hue), providing real-time awareness of where other users are working.

- replicaSelectionToGridSelection: converts stored CRDT-stable IDs back to a plain GridSelectionRange (row min/max + optional column set)
- useRemoteSelections: derives RemoteSelection entries from snapshot.selection[replicaId] for all connected non-SERVICE non-local replicas; no extra API calls needed
- getCellClassName / DataGrid / SynapseGrid: thread remoteSelections through so each cell can render cell-remote-selected--color-{i}
- _data-grid-extra.scss: 8-color palette and per-index tint rules

### DIFF
--- a/packages/synapse-react-client/src/components/DataGrid/hooks/useRemoteSelections.ts
+++ b/packages/synapse-react-client/src/components/DataGrid/hooks/useRemoteSelections.ts
@@ -1,5 +1,9 @@
 import { useMemo } from 'react'
-import type { GridModel, ReplicaSelectionModel } from '../DataGridTypes'
+import type {
+  GridModel,
+  GridModelSnapshot,
+  ReplicaSelectionModel,
+} from '../DataGridTypes'
 import type { GridReplicaInfo } from '@sage-bionetworks/synapse-client'
 import {
   replicaSelectionToGridSelection,
--- a/packages/synapse-react-client/src/components/DataGrid/hooks/useRemoteSelections.ts
+++ b/packages/synapse-react-client/src/components/DataGrid/hooks/useRemoteSelections.ts
@@ -19,7 +19,7 @@ export interface RemoteSelection {
  * `snapshot.selection[replicaId]`, so no extra API calls are needed.
  */
 export function useRemoteSelections(
-  modelSnapshot: ReturnType<GridModel['api']['getSnapshot']> | null | undefined,
+  modelSnapshot: GridModelSnapshot | null | undefined,
   model: GridModel | null | undefined,
   replicas: GridReplicaInfo[],
   localReplicaId: number | null,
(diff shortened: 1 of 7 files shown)


packages/synapse-react-client/src/components/DataGrid/utils/replicaSelectionToGridSelection.ts
```diff
@@ -0,0 +1,85 @@
...
+  let maxRow = numRows - 1
+
+  if (!selection.rowSelectAll) {
+    const rowIds = selection.rowSelection
+    if (!rowIds || rowIds.length === 0) return null
+
+    const rowIndexSet = crdtIdsToIndices(model.api.arr(['rows']), rowIds)
+    if (rowIndexSet.size === 0) return null
+    minRow = [...rowIndexSet].reduce((a, b) => Math.min(a, b))
+    maxRow = [...rowIndexSet].reduce((a, b) => Math.max(a, b))
```
nickgros, line 65: Kind of weird to optimize `crdtIdsToIndices` (with creating a set for O(1) lookup), but then you duplicate and iterate over the rowIndexSet twice. But if it works, that's ok. We can identify perf issues and optimize later, if there ever end up being any

packages/synapse-react-client/src/components/DataGrid/utils/replicaSelectionToGridSelection.ts
```diff
@@ -0,0 +1,85 @@
...
+  return result
+}
+
+/**
+ * Converts a stored ReplicaSelectionModel (CRDT-stable IDs) back into a
+ * plain grid selection range that can be used for cell class computation.
+ *
+ * This is the inverse of getCrdtIdsForArrayRange in computeReplicaSelectionModel.
+ */
+export function replicaSelectionToGridSelection(
```
nickgros, line 42: add tests

packages/synapse-react-client/src/components/DataGrid/hooks/useRemoteSelections.ts
```diff
@@ -0,0 +1,63 @@
...
+  colorIndex: number
+}
+
+/**
+ * Derives remote replica selection ranges from the CRDT model's selection map.
+ * The server mirrors each replica's selection into the model under
+ * `snapshot.selection[replicaId]`, so no extra API calls are needed.
+ */
+export function useRemoteSelections(
+  modelSnapshot: ReturnType<GridModel['api']['getSnapshot']> | null | undefined,
```
nickgros:
```suggestion
  modelSnapshot: GridModelSnapshot | null | undefined,
```

nickgros (approved):

PR #2673, opened 2026-04-09
PORTALS-4147 - derive more meta tags from loaders

nickgros:


apps/synapse-portal-framework/src/utils/detailPageRouteUtils.ts
```diff
@@ -115,38 +116,48 @@ export function createDetailPageRouteExports<
...
+        content: loaderData.description,
+      })
     }
-    if (extendMeta) {
-      descriptors.push(...extendMeta(loaderData))
+    if (extendMeta && loaderData) {
+      descriptors.push(...extendMeta(loaderData as TLoaderData))
     }
-    return descriptors
+    return mergeMeta(args, descriptors)
```
nickgros, line 165: Use `mergeMeta` which will replace ancestor route component meta tags with the most specific rendered route's tags.

apps/synapse-portal-framework/src/utils/detailPageRouteUtils.ts
```diff
@@ -177,6 +189,12 @@ export function createDetailPageRouteExports<
 export function createStaticMeta(
   pageTitle: string,
   portalName: string,
-): () => MetaDescriptor[] {
-  return () => [{ title: `${pageTitle} | ${portalName}` }]
+): (args: MetaArgs) => MetaDescriptor[] {
+  return args => {
+    const fullTitle = `${pageTitle} | ${portalName}`
+    return mergeMeta(args, [
```
nickgros: Use `mergeMeta`

apps/synapse-portal-framework/src/App.tsx
```diff
@@ -51,6 +53,7 @@ export default function App(props: AppProps) {
     <ChatDialogContext.Provider
       value={{ openChat, isChatAvailable: !!synapseChatProps }}
     >
+      <meta name="theme-color" content={palette.primary.main} />
```
nickgros, line 56: Moved this `meta` tag to somewhere where the palette was already available. React 19 puts it in the correct place for us: https://react.dev/reference/react-dom/components/meta

This particular change affects all portals, not just NF. Though according to MDN this tag is only used by Safari (desktop & iOS) and non-darkmode Android Chrome.

Even though React 19 has native `meta` support, we continue to use React Router for the other meta tags so we ensure they are rendered in SSG, and we can safely replace ancestor meta tags with child tags that should override them.

apps/portals/nf/src/root.tsx
```diff
@@ -9,15 +10,42 @@ import { Links, Meta, Outlet, Scripts, ScrollRestoration } from 'react-router'
...
  */
-export function meta(): MetaDescriptor[] {
+export function meta(args): MetaDescriptor[] {
   const portalName = import.meta.env.VITE_PORTAL_NAME
   const portalDescription = import.meta.env.VITE_PORTAL_DESCRIPTION
-  return [
+  const portalKey = import.meta.env.VITE_PORTAL_KEY ?? ''
+  const baseUrl = `https://${portalKey}.synapse.org`
+
+  const descriptors: MetaDescriptor[] = [
```
nickgros, line 19: Move meta tags out of layout and into `meta` so they can be deduplicated when a child route attempts to override

apps/portals/SSG_MIGRATION.md
```diff
@@ -194,7 +194,14 @@ export default {
...
 
-      return [...getStaticPaths(), ...dynamicRoutes]
+      // Exclude legacy /DetailsPage redirect routes from prerendering.
+      // Prerendering them creates S3 directory structures that cause S3 to
+      // drop the query string on redirect, breaking the client-side redirect logic.
+      const staticPaths = getStaticPaths().filter(
+        p => !p.endsWith('/DetailsPage'),
+      )
+
+      return [...staticPaths, ...dynamicRoutes]
```
nickgros, line 209: Update migration guide to account for change in an earlier PR.

(A lot of the code in this migration guide could be abstracted into reusable utilities, where the guide just references the utility. That refactor is out of scope for this ticket, and makes more sense to do when we migrate a 2nd portal to use SSG)

apps/synapse-portal-framework/src/App.tsx
```diff
@@ -51,6 +53,7 @@ export default function App(props: AppProps) {
...
       <SynapseToastContainer />
```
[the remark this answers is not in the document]
nickgros: https://react.dev/reference/react-dom/components/meta

apps/synapse-portal-framework/src/utils/detailPageRouteUtils.ts
jay-hodgson: Reminds me of React Helmet

jay-hodgson (approved):

PR #2677, opened 2026-04-10
SWC-7778

jay-hodgson:
Screenshot 2026-04-10 at 10 51 12 AM

### DIFF
--- a/packages/synapse-react-client/src/components/TextField/TextField.module.scss
+++ b/packages/synapse-react-client/src/components/TextField/TextField.module.scss
@@ -1,0 +1,18 @@
+.labelRow {
+  display: flex;
+  align-items: baseline;
+  gap: 6px;
+  margin-bottom: 4px;
+}
+
+.inputLabel {
+  position: relative !important;
+  font-weight: 700;
+  margin-bottom: 0;
+  pointer-events: unset;
+}
+
+.helperText {
+  font-weight: 500;
+  color: var(--synapse-gray-600);
+}
--- a/packages/synapse-react-client/src/components/TextField/TextField.tsx
+++ b/packages/synapse-react-client/src/components/TextField/TextField.tsx
@@ -1,9 +1,12 @@
 import StyledFormControl from '@/components/styled/StyledFormControl'
 import {
+  Box,
   InputBase,
   InputLabel,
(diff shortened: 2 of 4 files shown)


packages/synapse-react-client/src/components/TextField/TextField.module.scss
```diff
@@ -0,0 +1,18 @@
+.labelRow {
```
jay-hodgson, line 1: Refactor many styles to a scss module

packages/synapse-react-client/src/components/TextField/TextField.tsx
```diff
@@ -13,6 +16,7 @@ export type TextFieldProps = Pick<
   | 'disabled'
   | 'error'
   | 'fullWidth'
+  | 'helperText'
```
jay-hodgson, line 19: Added ability to specify helper text for a text field

nickgros: Could you just use the MUI `TextField` instead of this weird wrapper we have? OK to keep if this is actually using some wrapper feature.

Not opposed to having wrapper components if we want to make our code independent of MUI but I don't think that was the original intent of this component

nickgros (approved):

packages/synapse-react-client/src/components/TextField/TextField.tsx
jay-hodgson: That's a great question.  I was happy that I had a place to add this helper text enhancement that may have general reuse

jay-hodgson: So it may not have been justified before, but I see reason to keep it now (and for additional potential enhancements)

nickgros: MUI TextField already has a builtin `helperText` prop: 

https://mui.com/material-ui/api/text-field/#props

jay-hodgson: > MUI TextField already has a builtin `helperText` prop:
> 
> https://mui.com/material-ui/api/text-field/#props

Then perhaps I don't need to declare?  MUIs implementation of the helper text differs from our custom design, so I think this wrapper is still needed

PR #2679, opened 2026-04-13
SWC-7776: ability to set the Project description on creation

jay-hodgson:
Screenshot 2026-04-13 at 9 51 20 AM

Screenshot 2026-04-13 at 9 53 49 AM


packages/synapse-react-client/src/components/CreateProjectModal/CreateProjectModal.tsx
```diff
@@ -72,6 +75,22 @@ export function CreateProjectModal({
           },
         }}
       />
+      <Box mt={2}>
```
nickgros: Could use `Stack` with `gap={2}` (which could also ensure appropriate spacing for the `Alert`)

nickgros (approved):

PR #2685, opened 2026-04-14
SWC-7777

jay-hodgson:
Screenshot 2026-04-14 at 3 56 10 PM


packages/synapse-react-client/src/components/CreateProjectModal/CreateProjectModal.tsx
```diff
@@ -20,22 +27,88 @@ export function CreateProjectModal({
...
+            principalId: Number(publicGroupId),
+            accessType: [ACCESS_TYPE.READ],
+          })
+        }
+        if (authenticatedUsersId) {
+          newResourceAccess.push({
+            principalId: Number(authenticatedUsersId),
+            accessType: [ACCESS_TYPE.READ, ACCESS_TYPE.DOWNLOAD],
+          })
+        }
```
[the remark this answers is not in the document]
jay-hodgson: This is a brand new entity, so this should not matter.  However, I see no harm in implementing an upsert by principalId

packages/synapse-react-client/src/components/CreateProjectModal/CreateProjectModal.tsx
```diff
@@ -89,7 +187,16 @@ export function CreateProjectModal({
...
       />
+      <ProjectVisibilityRadioGroup
+        value={visibility}
+        onChange={setVisibility}
+      />
       {errorMessage && <Alert severity="error">{errorMessage}</Alert>}
+      {aclErrorMessage && <Alert severity="error">{aclErrorMessage}</Alert>}
+      <Alert severity="warning">
+        You can change visibility later in Project Tools and also at a
+        granular/file level
```
nickgros: Copy feels kind of clunky imo. Some options generated with LLM help: 

> Visibility can be changed anytime in Project Tools, and can also be set individually on files and folders.

or

> You can update project visibility at any time in Project Tools, or manage access at a more granular level for individual files and folders.

packages/synapse-react-client/src/components/CreateProjectModal/CreateProjectModal.tsx
```diff
@@ -20,22 +27,116 @@ export function CreateProjectModal({
...
+      setAclErrorMessage(
+        `Project was created, but visibility could not be set: ${
+          err.reason ?? err.message
+        }`,
+      )
+      return false
+    }
+  }
+
   const onCreateProject = async () => {
```
nickgros: Can you wrap `onCreateProject` in `useMutation`? That would let you easily derive loading state, handle failure cases, etc.

packages/synapse-react-client/src/components/CreateProjectModal/CreateProjectModal.tsx
```diff
@@ -20,22 +27,116 @@ export function CreateProjectModal({
...
   const hide = () => {
     setProjectName('')
     setDescription('')
+    setVisibility('DISCOVERABLE')
     setErrorMessage(undefined)
+    setAclErrorMessage(undefined)
     onClose()
   }
+
+  const applyVisibilityAcl = async (projectId: string): Promise<boolean> => {
```
nickgros: You might also make this its own mutation (called by the create project mutation), but that's up to you. We get most of the benefits from ensuring the user's 'entrypoint' (`onCreateProject`) is wrapped.

packages/synapse-react-client/src/components/CreateProjectModal/CreateProjectModal.tsx
jay-hodgson: updated!

nickgros (approved):

PR #2686, opened 2026-04-15
SWC-7790 - Fix Autocomplete cell hover selection

afwillia:
Fix a bug in Autocomplete columns where hovering over a dropdown item and closing the dropdown mistakenly selects the value. Now if the user clicks outside the dropdown, it will close without selecting a value.

Create a new utility, useGridAutocompleteState to manage shared state between the two Autocomplete columns.


packages/synapse-react-client/src/components/DataGrid/columns/AutocompleteColumn.tsx
```diff
@@ -62,32 +63,37 @@ export function AutocompleteCell({
   active,
   clearValue = undefined,
 }: AutocompleteCellProps) {
-  const [menuIsOpen, setMenuIsOpen] = useState(false)
-  const textInputRef = useRef<HTMLInputElement>(null)
+  // Stable refs so memo(AutocompleteCell) can use default shallow comparison —
+  // react-datasheet-grid recreates these functions on every render, but their
+  // behavior is stable for a given cell position.
+  const setRowDataRef = useRef(setRowData)
+  setRowDataRef.current = setRowData
```
nickgros: Latest ref pattern should set `.current` in a `useLayoutEffect`: https://www.epicreact.dev/the-latest-ref-pattern-in-react

packages/synapse-react-client/src/components/DataGrid/columns/AutocompleteColumn.tsx
```diff
@@ -123,84 +129,94 @@ export function AutocompleteCell({
...
+    }
+  }, [])
+
+  const handleClose = useCallback(() => {
+    // clear the guard in case a mousedown on an option
+    // was abandoned before onChange could fire (e.g. drag-off).
+    optionMouseDownRef.current = false
+    setMenuIsOpen(false)
+    stopEditingRef.current({ nextRow: false })
+  }, [setMenuIsOpen])
```
nickgros: `handleClose` is the same implementation in both components. Propose `useGridAutocompleteState` returns this as `handleMenuClose`.

To match this, you can also expose a simple `handleMenuOpen` which just calls `() => setMenuIsOpen(true)`

packages/synapse-react-client/src/components/DataGrid/columns/AutocompleteColumn.tsx
```diff
@@ -123,84 +129,94 @@ export function AutocompleteCell({
...
+
+  const handleChange = useCallback(
+    (_e: React.SyntheticEvent, newVal: AutocompleteOption, reason: string) => {
+      // Reset the portal-click guard now that the selection has been committed
+      optionMouseDownRef.current = false
```
nickgros: Both components toggle this ref value before executing their own `handleChange` logic. Could we instead have the hook return a function that cleans up the abstraction, (called something like `notifyOptionCommited`)?

E.g. 

```ts
const handleChange = useCallback((_e: React.SyntheticEvent, newVal: AutocompleteOption, reason: string) => {
  notifyOptionCommitted()
  // ...component specific logic here...
})

```

then in the hook:

```ts
const notifyOptionCommitted = useCallback(() => { optionMouseDownRef.current = false }, []);
```

With this change and the suggestion above, the hook no longer needs to expose `optionMouseDownRef` so the abstraction is a bit cleaner

packages/synapse-react-client/src/components/DataGrid/columns/AutocompleteColumn.test.tsx
```diff
@@ -531,3 +685,90 @@ describe('AutocompleteColumn', () => {
     })
   })
 })
+
+describe('MemoizedAutocompleteCell', () => {
```
nickgros, line 510: Are these tests capturing any known bugs/regressions? If not, I am ok with just testing the newly-factored-out `areAutocompleteCellPropsEqual`. Especially if the copilot comment below is true that the 'no-rerender' check in this test is not valid

packages/synapse-react-client/src/components/DataGrid/columns/AutocompleteMultipleEnumColumn.tsx
```diff
@@ -202,32 +245,13 @@ function AutocompleteMultipleEnumCell({
...
-              return typeof item === 'string'
-                ? parseFreeTextGivenJsonSchemaType(item, colType)
-                : item.value
-            })
-            // Use clearValue when all items are removed
-            setRowData(values.length === 0 ? clearValue : values)
-            setLocalInputState('')
-          }}
+          onClose={handleClose}
+          onChange={handleChange as any}
```
nickgros: can we remove the `as any` here? probably just need to fix the types in `handleChange`

packages/synapse-react-client/src/components/DataGrid/columns/AutocompleteColumn.tsx
```diff
@@ -62,32 +71,41 @@ export function AutocompleteCell({
...
-  const textInputRef = useRef<HTMLInputElement>(null)
+  // Stable refs so memo(AutocompleteCell) can use default shallow comparison —
+  // react-datasheet-grid recreates these functions on every render, but their
+  // behavior is stable for a given cell position.
+  const setRowDataRef = useRef(setRowData)
+  useLayoutEffect(() => {
+    setRowDataRef.current = setRowData
+  })
+  const stopEditingRef = useRef(stopEditing)
+  stopEditingRef.current = stopEditing
```
nickgros: also `useLayoutEffect` for `stopEditing`

packages/synapse-react-client/src/components/DataGrid/columns/AutocompleteMultipleEnumColumn.tsx
```diff
@@ -88,21 +94,23 @@ function AutocompleteMultipleEnumCell({
   stopEditing,
   clearValue = undefined,
 }: AutocompleteMultipleEnumCellProps) {
-  const ref = useRef<HTMLInputElement>(null)
-  const [menuIsOpen, setMenuIsOpen] = useState(false)
+  // Stable refs so memo(AutocompleteMultipleEnumCell) can use default shallow
+  // comparison — react-datasheet-grid recreates these functions on every render,
+  // but their behavior is stable for a given cell position.
+  const setRowDataRef = useRef(setRowData)
+  setRowDataRef.current = setRowData
```
nickgros: also wrap setter in `useLayoutEffect`

packages/synapse-react-client/src/components/DataGrid/columns/useGridAutocompleteState.ts
```diff
@@ -0,0 +1,132 @@
...
+  stopEditing,
+  onDeactivate,
+}: UseGridAutocompleteStateProps): UseGridAutocompleteStateReturn {
+  const inputRef = useRef<HTMLInputElement>(null)
+  const [menuIsOpen, setMenuIsOpen] = useState(false)
+  const optionMouseDownRef = useRef(false)
+
+  // Always reflects the latest `active` value — safe to read in closures.
+  const activeRef = useRef(active)
+  activeRef.current = active
```
nickgros: also wrap setter in `useLayoutEffect`

packages/synapse-react-client/src/components/DataGrid/columns/useGridAutocompleteState.ts
```diff
@@ -0,0 +1,132 @@
...
+
+  // Ref-stabilize callbacks so effects/handlers never need them in deps arrays,
+  // yet always call the latest versions (avoiding stale closures).
+  const onDeactivateRef = useRef(onDeactivate)
+  onDeactivateRef.current = onDeactivate
+  const stopEditingRef = useRef(stopEditing)
+  stopEditingRef.current = stopEditing
```
nickgros: also wrap setters in `useLayoutEffect`

nickgros: you could consider adding a utility hook and using it in all of these places

```js
function useLatestRef(value) {
	const ref = useRef(value)
	useLayoutEffect(() => {
		ref.current = value;
	})

	return ref;
}

```

packages/synapse-react-client/src/components/DataGrid/columns/AutocompleteColumn.tsx
```diff
@@ -123,84 +141,85 @@ export function AutocompleteCell({
...
-          )
-        } else {
-          // The value was selected, so explicitly set it
-          setRowData(newVal)
-        }
-        setTimeout(() => stopEditing({ nextRow: false }), 0)
-      }}
+      onKeyDown={handleKeyDown}
+      onClose={handleClose}
+      onChange={handleChange as any}
```
nickgros: Would also be nice to get rid of this `as any`

packages/synapse-react-client/src/components/DataGrid/columns/AutocompleteMultipleEnumColumn.tsx
```diff
@@ -158,6 +166,42 @@ function AutocompleteMultipleEnumCell({
...
+          : item.value
+      })
+      // Use clearValue when all items are removed
+      setRowDataRef.current(values.length === 0 ? clearValue : values)
+      setLocalInputState('')
+      if (!activeRef.current) {
+        // The cell was deactivated while the option click was in progress (portal
+        // click was seen as an outside click by the grid). Now that the value is
+        // committed, close the menu.
+        setMenuIsOpen(false)
```
nickgros, line 197: Was hoping we could remove `setMenuIsOpen` from the hook return but if we really need it here for this case, then it's fine

nickgros (approved):

PR #2687, opened 2026-04-15
PORTALS-4158: ADKP Homepage Redesign:  Add new Search place

kianamcc:
Jira: https://sagebionetworks.jira.com/browse/PORTALS-4158

/Search route already exists, just needed some cleanup.

Only studies tables seemed to have FTS enabled:
Screenshot 2026-04-15 at 2 22 10 PM


jay-hodgson (approved):

jay-hodgson:
May want to just comment these out, instead of deleting, so we can more easily migrate to the new Search API components when they are available (soon!).

PR #2689, opened 2026-04-15
PORTALS-4176: from realms, do not assume user is logged in if an acce…

jay-hodgson:
…ssToken is set


nickgros (approved):
Nice catch! I missed this one in #2501

PR #2691, opened 2026-04-16
Audit GitHub Actions

nickgros:
Update GitHub Actions to use specific version tags for dependencies and improve permissions

Audited with Claude Opus 4.7

### DIFF
--- a/.github/actions/pnpm-setup-action/action.yml
+++ b/.github/actions/pnpm-setup-action/action.yml
@@ -9,11 +9,11 @@ runs:
   using: 'composite'
   steps:
     - name: Use Node.js 22
-      uses: actions/setup-node@v4
+      uses: actions/setup-node@49933ea5288caeca8642d1e84afbd3f7d6820020 # v4
       with:
         node-version: 22
         registry-url: https://registry.npmjs.org/
-    - uses: pnpm/action-setup@v4
+    - uses: pnpm/action-setup@b906affcce14559ad1aafd4ab0e942779e9f58b1 # v4
       name: Install pnpm
       id: pnpm-install
       with:
@@ -24,7 +24,7 @@ runs:
       run: |
         echo "STORE_PATH=$(pnpm store path)" >> $GITHUB_OUTPUT
     - name: Setup pnpm cache
-      uses: actions/cache@v3
+      uses: actions/cache@0057852bfaa89a56745cba8c7296529d2fc39830 # v4
       with:
         path: ${{ steps.pnpm-cache.outputs.STORE_PATH }}
         key: ${{ runner.os }}-pnpm-store-${{ hashFiles('**/pnpm-lock.yaml') }}
@@ -35,7 +35,7 @@ runs:
       # It should be safe to use when running tests, but do not use it in deploy workflows.
       # See: https://nx.dev/recipes/troubleshooting/unknown-local-cache#you-share-cache-with-another-machine-using-a-network-drive
       if: ${{ inputs.STORE_NX_CACHE == 'true' }}
-      uses: actions/cache@v3
+      uses: actions/cache@0057852bfaa89a56745cba8c7296529d2fc39830 # v4
       with:
         path: .nx/cache
         key: ${{ runner.os }}-nx-${{ github.ref }}-${{ github.sha }}
--- a/.github/actions/pnpm-setup-action/action.yml
+++ b/.github/actions/pnpm-setup-action/action.yml
@@ -9,11 +9,11 @@ runs:
   using: 'composite'
   steps:
     - name: Use Node.js 22
-      uses: actions/setup-node@49933ea5288caeca8642d1e84afbd3f7d6820020 # v4
+      uses: actions/setup-node@49933ea5288caeca8642d1e84afbd3f7d6820020 # v4.4.0
       with:
         node-version: 22
         registry-url: https://registry.npmjs.org/
-    - uses: pnpm/action-setup@b906affcce14559ad1aafd4ab0e942779e9f58b1 # v4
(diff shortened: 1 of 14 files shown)


.github/actions/pnpm-setup-action/action.yml
```diff
@@ -9,11 +9,11 @@ runs:
   using: 'composite'
   steps:
     - name: Use Node.js 22
-      uses: actions/setup-node@v4
+      uses: actions/setup-node@49933ea5288caeca8642d1e84afbd3f7d6820020 # v4
```
nickgros: Pin SHAs for all 3rd party actions to reduce risk of using a compromised action.

.github/workflows/build-storybook-for-chromatic.yml
```diff
@@ -0,0 +1,38 @@
+name: 'Build Storybook for Chromatic'
```
nickgros: To remove an unsafe `pull_request_target` workflow trigger, split the deployment to chromatic into two steps:

1. Unconditionally build Storybook and create an artifact
2. Change `deploy-to-chromatic.yml` to be triggered by a `workflow_run`

.github/workflows/build-test.yml
```diff
@@ -59,38 +63,26 @@ jobs:
...
     steps:
-      - uses: actions/checkout@v4
+      - uses: actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5 # v4
         with:
           fetch-depth: 0
       - uses: ./.github/actions/pnpm-setup-action
         with:
           STORE_NX_CACHE: 'true'
       - run: pnpm nx affected --target=type-check --base=remotes/origin/main
-  storybook:
```
nickgros, line 69: Now redundant since `build-storybook-for-chromatic.yml` was created.

.github/workflows/deploy-to-chromatic.yml
```diff
@@ -1,47 +1,40 @@
 name: 'Deploy Storybook to Chromatic'
 
-env:
-  NODE_OPTIONS: '--max-old-space-size=4096'
-
 on:
-  push:
-    branches:
-      - main
-  pull_request_target:
```
nickgros, line 10: Risk here was that:

1. A malicious actor could craft a PR that exfiltrates secrets in a step in this workflow (e.g. by adding code that runs in `pnpm nx run synapse-react-client:build`)
2. A web engineer allows the actor's actions to run (reasonable if the attacker obfuscates the malicious code in the PR)
3. A web engineer applies this `chromatic` label to the PR (reasonable if the attacker suggested reviewing in Storybook before merging!)

The attacker then gets the chromatic deploy secret and possibly GH secrets. This change eliminates that attack vector.

.github/actions/pnpm-setup-action/action.yml
```diff
@@ -9,11 +9,11 @@ runs:
...
       with:
         node-version: 22
         registry-url: https://registry.npmjs.org/
-    - uses: pnpm/action-setup@v4
+    - uses: pnpm/action-setup@b906affcce14559ad1aafd4ab0e942779e9f58b1 # v4
```
hallieswan: Should we make [this version](https://github.com/pnpm/action-setup/releases) more specific? I think this is v4.3.0

.github/workflows/build-storybook.yml
```diff
@@ -0,0 +1,38 @@
...
+  NODE_OPTIONS: '--max-old-space-size=4096'
+
+on:
+  push:
+    branches:
+      - main
+  pull_request:
+    types: [labeled]
+    branches:
+      - main
```
hallieswan, line 17: I think the current configuration will only trigger on push to main or on labeling a PR to main. If this job replaces the `build-test.storybook` job, should we update the trigger to include the [default `pull_request` triggers](https://docs.github.com/en/actions/reference/workflows-and-actions/events-that-trigger-workflows#pull_request) too?


```suggestion
  pull_request:
    types: 
      - labeled
      - opened
      - synchronize
      - reopened
    branches:
      - main
```

.github/workflows/build-storybook-for-chromatic.yml
```diff
@@ -0,0 +1,38 @@
...
+
+permissions:
+  contents: read
+
+jobs:
+  build-storybook:
+    name: Build Storybook
+    runs-on: ubuntu-latest
+    # On push, always build. On pull_request, only build when the 'chromatic' label is applied.
+    if: ${{ github.event_name == 'push' || github.event.label.name == 'chromatic' }}
```
hallieswan: Similarly, if we're replacing the `build-test.storybook` job, should this run for other pull_request events?

.github/workflows/build-storybook-for-chromatic.yml
hallieswan: Should the workflow title be generic (e.g. "Build Storybook"), since we're not only using this workflow for Chromatic, we're also using it to test the Storybook build in general?

.github/workflows/deploy-to-chromatic.yml
```diff
@@ -1,47 +1,40 @@
...
+permissions:
+  actions: read # needed to download artifacts from the triggering workflow run
+  contents: read
 
 jobs:
   deploy-to-chromatic:
     name: Deploy to Chromatic
     runs-on: ubuntu-latest
+    # Only proceed if the upstream build workflow succeeded
+    if: ${{ github.event.workflow_run.conclusion == 'success' }}
```
hallieswan: If we change the "Build Storybook" workflow. then will probably need to gate on the PR label as well job success

hallieswan (commented):
Looks great! Just a couple of suggestions

hallieswan (approved):
Looks good to me!

PR #2709, opened 2026-04-20
PORTALS-4144

nickgros:
- Retry transient network errors
- Add jitter to retry delay

Not sure if this is fixes the cause of the problem, but if the issue is that we are saturating the available connections in the container, this will hopefully alleviate the issue.


jay-hodgson:
~~We used to take an exponential backoff approach.  Maybe we don't want to reimplement this approach (with or without jitter), but would it be helpful to increase the delay on retry?  Maybe a linear backoff (1s, 2s, 3s)?  If so, I think you could just pass in the current delay plus 1s (perhaps with jitter).~~
This was misguided (just looked at the diff rather than the file).  Please ignore

jay-hodgson (approved):

PR #2715, opened 2026-04-20
PORTALS-4164

jay-hodgson:
Synapse home page Hot Drops
Screenshot 2026-04-20 at 3 04 45 PM

WordPress News component
Screenshot 2026-04-20 at 3 04 57 PM


apps/portals/adknowledgeportal/src/pages/HomePageV2.tsx
```diff
@@ -14,6 +16,18 @@ function HomePageV2() {
       >
         <AdknowledgeContributeCard />
       </SectionLayout>
+      <SectionLayout
+        title={'News Releases'}
```
jay-hodgson, line 20: titles should probably all get styled differently before release

packages/synapse-react-client/src/components/SynapseHomepageV2/FloatingBlobsBackground.tsx
```diff
@@ -0,0 +1,125 @@
+import React from 'react'
```
jay-hodgson, line 1: Renamed existing SynapseHotdropsBackground, and made stop colors configurable.  Now used in 2 places

packages/synapse-react-client/src/components/SynapseHomepageV2/HotdropsAndNews.module.scss
```diff
@@ -0,0 +1,200 @@
+@use 'sass:map';
```
jay-hodgson: Pulled out inline styles from SynapseHotDrops (and Item) to a scss module shared by both Hotdrops and WordPressNews.

packages/synapse-react-client/src/components/SynapseHomepageV2/FloatingBlobsBackground.module.scss
```diff
@@ -0,0 +1,92 @@
+// ---------------------------------------------------------------------------
```
jay-hodgson, line 1: styles refactored out of the existing Background component

packages/synapse-react-client/src/components/SynapseHomepageV2/HotdropsAndNews.module.scss
```diff
@@ -0,0 +1,104 @@
+// ---------------------------------------------------------------------------
+// Grid container (SynapseHotDrops / WordPressNews outer wrapper)
```
jay-hodgson, line 2: Shared styles for Hotdrops and WordPress News

packages/synapse-react-client/src/components/SynapseHomepageV2/SynapseHomepageV2.tsx
```diff
@@ -400,7 +400,10 @@ export function SynapseHomepageV2({ gotoPlace }: SynapseHomepageV2Props) {
               </Typography>
             </Box>
             <Box sx={{ position: 'relative', minHeight: '500px' }}>
-              <SynapseHotdropsBackground />
+              <FloatingBlobsBackground
```
jay-hodgson, line 403: Use reusable FloatingBlobsBackground with (previously hard-coded) colors

packages/synapse-react-client/src/components/SynapseHomepageV2/WordPressNewsItem.tsx
```diff
@@ -0,0 +1,56 @@
...
+        window.open(post.link, '_blank', 'noopener,noreferrer')
+      }}
+      className={styles.item}
+    >
+      <Box className={styles.itemContent}>
+        <Typography
+          variant="body1"
+          className={styles.itemTitle}
+          noWrap
+          dangerouslySetInnerHTML={{ __html: post.title.rendered }}
```
jay-hodgson, line 44: This is dangerous, but we only configure our portals to show content from WordPress sites that Sage (or trusted Sage partners) control

apps/portals/adknowledgeportal/src/pages/HomePageV2.tsx
kianamcc: We can probably use the titleNode prop in SectionLayout that I made for the namshub about section here. But yes, that can be done later!

packages/synapse-react-client/src/components/SynapseHomepageV2/WordPressNewsItem.tsx
```diff
@@ -0,0 +1,56 @@
...
+  post: WordPressPost
+  categoryName: string | undefined
+}
+
+export function WordPressNewsItem({
+  post,
+  categoryName,
+}: WordPressNewsItemProps) {
+  return (
+    <Box
```
kianamcc, line 32: Can we make this more accessible by adding role button and tabIndex 0?

kianamcc (approved):

packages/synapse-react-client/src/components/SynapseHomepageV2/HotdropsAndNews.module.scss
```diff
@@ -0,0 +1,104 @@
...
+      border-top: 1px solid var(--synapse-gray-400);
+    }
+  }
+
+  width: 100%;
+  background-color: rgba(255, 255, 255, 0.9);
+  padding: 15px 0;
+
+  &:hover {
+    background-color: rgba(233, 243, 254, 0.95);
```
kianamcc, line 65: Is this the light blue color we see on the synapse homepage? ADKP seems to be using purple gradient which would match with the purple blobs.

jay-hodgson: Easiest cross-theme solution is to probably remove the alpha on hover (resulting in just white)

packages/synapse-react-client/src/components/SynapseHomepageV2/WordPressNewsItem.tsx
jay-hodgson: Claude is telling the the following.  So plan will be to change it to an anchor.
```
Using role="button" isn't ideal here since the interaction navigates to a URL — that's semantically a link, not a button. Screen readers announce the role to users, so getting it wrong is misleading.

The most accessible fix is to render the Box as an <a> element. This gives you:

Correct semantics (role="link" implicitly)
Native keyboard focus and tabIndex for free
Enter key activates it natively (no onKeyDown needed)
Right-click → "Open in new tab" works
```

jay-hodgson: Also
```
No tabIndex needed. <a> elements with an href are natively focusable and are automatically included in the tab order by browsers.
```

packages/synapse-react-client/src/components/SynapseHomepageV2/HotdropsAndNews.module.scss
jay-hodgson: I really didn't like the white (since it made no difference on the parts with white background, so it did a poor job at highlighting).  Reverted to the light blue.  We'll see what ADKP stakeholders and Adam say

packages/synapse-react-client/src/components/SynapseHomepageV2/HotdropsAndNews.module.scss
```diff
@@ -59,9 +59,12 @@
   background-color: rgba(255, 255, 255, 0.9);
   padding: 15px 0;
 
+  text-decoration: none;
+
   &:hover {
     background-color: rgba(233, 243, 254, 0.95);
     cursor: pointer;
+    text-decoration: none !important;
```
jay-hodgson, line 67: There's a high specificity selector for anchors, and I think it's ok to override for this item hover

PR #2719, opened 2026-04-21
PORTALS-4162: ADKP Homepage Redesign: New Programs component/visualization

kianamcc:
Jira: https://sagebionetworks.jira.com/jira/software/c/projects/PLFM/boards/238?assignee=712020%3Aa7fbbeb0-d7a6-443a-9dd4-246363ff09dd&selectedIssue=PORTALS-4162

Screenshot 2026-04-21 at 11 00 46 AM

Screenshot 2026-04-21 at 10 58 57 AM

Screenshot 2026-04-21 at 10 58 44 AM

Clicking a hex on mobile or clicking "view program" button on desktop:
Screenshot 2026-04-21 at 10 52 57 AM


packages/synapse-react-client/src/components/HexGrid/HexGrid.module.scss
```diff
@@ -0,0 +1,243 @@
+.hexSlat {
```
kianamcc, line 1: Majority of styles taken from here: https://github.com/adamhindman/adkp-hex-test/blob/main/src/style.css

apps/portals/adknowledgeportal/src/pages/HomePageV2.tsx
```diff
@@ -17,27 +19,28 @@ function HomePageV2() {
...
-        centerTitle={true}
         ContainerProps={{
           className: 'home-spacer',
         }}
       >
-        <div style={{ position: 'relative', minHeight: '500px' }}>
-          <FloatingBlobsBackground stopColor1="#3A5AA8" stopColor2="#6B5FD9" />
-          <WordPressNews wordpressSiteUrl="https://news.adknowledgeportal.org" />
-        </div>
+        <MailchimpSubscribeSection
```
kianamcc, line 27: Should just be order change, for new releases section to appear below programs as we see in the Figma.

kianamcc: Image

kianamcc:
Forgot to add button: 
Screenshot 2026-04-21 at 1 09 36 PM

apps/synapse-portal-framework/src/components/adknowledge/AdknowledgePrograms/AdknowledgePrograms.module.scss
```diff
@@ -0,0 +1,30 @@
+.adknowledgeProgramsTitle:global(.MuiTypography-root) {
+  color: var(--adkp-accent-color);
```
jay-hodgson: I think all ADKP section titles need to be styled in this way, is that correct?  
If so, is there a way to set these styles so we can just use the existing `SectionLayout` to set the title and description?

apps/synapse-portal-framework/src/components/adknowledge/AdknowledgePrograms/AdknowledgePrograms.module.scss
```diff
@@ -0,0 +1,30 @@
...
+.adknowledgeProgramsContainer {
+  width: 100%;
+}
+
+.adknowledgeProgramsButton {
+  align-self: center;
+  border-radius: 8px;
+  background: var(--adkp-button-gradient);
+  border: 1px solid;
+  border-color: var(--adkp-button-border-color);
```
jay-hodgson: From the design, it looks like all buttons in the portal have been restyled (background for primary, border color for primary, border-radius for all MUI buttons).  Maybe apply to the MUI theme?

apps/synapse-portal-framework/src/components/adknowledge/AdknowledgePrograms/HexGrid.tsx
```diff
@@ -0,0 +1,148 @@
...
+import { useState } from 'react'
+import { Button, ButtonBase, Typography, useMediaQuery } from '@mui/material'
+import { useNavigate } from 'react-router'
+import { Link as RouterLink } from 'react-router'
+
+type HexGridProps = {
+  sql: string
+}
+
+function HexGrid({ sql }: HexGridProps) {
```
jay-hodgson: I love this component!  Do we prepare it for reuse now or later?  To reuse, I think it would need the column names for each part (titleColName, imageColName, descriptionColName), and perhaps an onClick function (which is given the selected data row).  Oh, and the component moved to the SRC library project, or somewhere else in the synapse-portal-framework.

apps/synapse-portal-framework/src/components/adknowledge/AdknowledgePrograms/AdknowledgePrograms.module.scss
kianamcc: We can create a new theme provider and wrap the homepagev2 component in it!

apps/synapse-portal-framework/src/components/adknowledge/AdknowledgePrograms/HexGrid.tsx
kianamcc: Added props titleColName, imageColName, and descriptionColName and moved the component to SRC. I left out the onClick function for now, I think we can revisit and refactor as needed when we see the case for reuse!

jay-hodgson: >I left out the onClick function for now, I think we can revisit and refactor as needed when we see the case for reuse!

I think that's fine for now, but could you pass in the target base URL (that you currently append the title column name value)?  
- Add prop for `detailPageParamColValue` (I'm not set on the name!). Find the column index for this column, and use it in `getDetailsUrl`.  In the current usage it will be the same as `titleColName` ("Program")
- Add prop `baseDetailPageURL`, in the current usage it will be set to '/Explore/Programs/DetailsPage?Program='
- Rename `programColumnIndex` to `titleColumnIndex`

kianamcc: > > I left out the onClick function for now, I think we can revisit and refactor as needed when we see the case for reuse!
> 
> I think that's fine for now, but could you pass in the target base URL (that you currently append the title column name value)?
> 
> * Add prop for `detailPageParamColValue` (I'm not set on the name!). Find the column index for this column, and use it in `getDetailsUrl`.  In the current usage it will be the same as `titleColName` ("Program")
> * Add prop `baseDetailPageURL`, in the current usage it will be set to '/Explore/Programs/DetailsPage?Program='
> * Rename `programColumnIndex` to `titleColumnIndex`

I was thinking about this more last night and was already reconsidering! I think this further confirms that we should leave the 'click' logic to the parent component.

kianamcc:
Screenshot 2026-04-22 at 10 45 40 AM

Screenshot 2026-04-22 at 10 46 02 AM

jay-hodgson (approved):

PR #2733, opened 2026-04-24
Update D4D content on B2AI Standards Explorer 

Bankso:
- added new CSS for styling D4D HTML
- updated table version numbers

FYI @milen-sage


Bankso:
@jay-hodgson / @nickgros I think this is ready, if you could take a quick look and approve.

apps/portals/b2ai.standards/src/assets/d4d/d4d.css
```diff
@@ -1,531 +1,538 @@
 /* D4D Datasheet Human-Readable HTML Common Stylesheet */
-.d4d-styles {
```
nickgros, line 2: @Bankso Removing this class caused the styles to be applied to the entire page, I don't think we want this:

Image

Bankso: Oh, nope we do not want that

Bankso: @nickgros Okay, I added back the .d4d-styles class and things look right on my local build

nickgros (approved):

PR #2737, opened 2026-04-24
SWC-7601: ACL editor should handle multiple resource access objects with the same principals

kianamcc:
Jira: https://sagebionetworks.jira.com/issues?jql=project%20IN%20(SWC%2C%20PORTALS)%20AND%20status%20%3D%20%22Ready%20for%20Team%22%20AND%20assignee%20%3D%20empty%20ORDER%20BY%20created%20DESC&selectedIssue=SWC-7601

Issue: Backend can return multiple resourceAccess entries with the same principalId but different access types but only the last entry would be used, meaning users could be shown the wrong permission level.

Create a utility function that unions accessType arrays for entries sharing the same principalId.

Before:
image

After:
Screenshot 2026-04-24 at 2 31 35 PM

### DIFF
--- a/packages/synapse-react-client/src/components/AccessRequirementAclEditor/AccessRequirementAclEditor.tsx
+++ b/packages/synapse-react-client/src/components/AccessRequirementAclEditor/AccessRequirementAclEditor.tsx
@@ -9,12 +9,14 @@ import { PermissionLevel } from '@/utils/PermissionLevelToAccessType'
 import { Alert, Box, Stack, Typography } from '@mui/material'
 import { SynapseClientError } from '@sage-bionetworks/synapse-client/util/SynapseClientError'
 import { ACCESS_TYPE, AccessControlList } from '@sage-bionetworks/synapse-types'
+import { consolidateResourceAccessList } from '@/utils/functions/AccessControlListUtils'
 import { isEqual } from 'lodash-es'
 import {
   ForwardedRef,
   forwardRef,
   useEffect,
   useImperativeHandle,
+  useMemo,
   useState,
 } from 'react'
 import { AclEditor } from '../AclEditor/AclEditor'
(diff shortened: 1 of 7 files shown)


packages/synapse-react-client/src/components/AccessRequirementAclEditor/AccessRequirementAclEditor.tsx
```diff
@@ -70,6 +72,11 @@ export const AccessRequirementAclEditor = forwardRef(
         staleTime: Infinity,
       })
 
+    const originalResourceAccess = useMemo(
```
kianamcc: For each ACL editor, we want to pass originalAcl.resourceAccess through consolidateResourceAccessList before setting the state so we show the correct access types in the UI.

jay-hodgson: Suggest renaming `originalResourceAccess` to `consolidatedOriginalResourceAccess`

packages/synapse-react-client/src/components/EntityAclEditor/EntityAclEditor.tsx
```diff
@@ -177,7 +178,12 @@ const EntityAclEditor = forwardRef(function EntityAclEditor(
     },
   )
 
-  const originalResourceAccess = entityBundle.benefactorAcl.resourceAccess
+  const originalResourceAccess = useMemo(
```
jay-hodgson: Same comment for rename

packages/synapse-react-client/src/utils/functions/AccessControlListUtils.test.ts
```diff
@@ -261,4 +262,47 @@ describe('AccessControlListUtils', () => {
...
+          principalId: 1,
+          accessType: [ACCESS_TYPE.UPDATE, ACCESS_TYPE.DELETE],
+        },
+        { principalId: 2, accessType: [ACCESS_TYPE.READ] },
+      ]
+
+      const result = consolidateResourceAccessList(list)
+      expect(result).toHaveLength(2)
+
+      const entry = result.find(item => item.principalId === 1)
```
jay-hodgson, line 292: Do you want to also verify the entry for principalId 2 was unchanged?

jay-hodgson (approved):

jay-hodgson (approved):

PR #2742, opened 2026-04-28
PORTALS-4195: Upgrade TypeScript to version 6.0.3

nickgros:
Upgrade TypeScript to version 6.0.3

Additionally:

Modify markdown-it-synapse build to use Vite, emit ESM-only artifacts 
- Bump package version to 2.0.0.
- Changed build script to use Vite for building and TypeScript for declaration generation.
- Removed `tsup` as a dependency and configuration.
- Added a new `tsconfig.build.json` for build-specific TypeScript settings.
- Updated Vite configuration to include all source files while preserving module structure (similar to synapse-react-client, synapse-portal-framework)

The motivation for this change is that `tsup` v8.4.0 is not compatible with TypeScript 6. While we could try upgrading `tsup` or migrating to `tsdown`, it makes more sense to drop it and adopt the same build conventions we are using elsewhere in the monorepo.


packages/markdown-it-synapse/tsconfig.build.json
```diff
@@ -0,0 +1,18 @@
+{
```
nickgros, line 1: tsconfig used to create type declarations at build-time

packages/markdown-it-synapse/vite.config.ts
```diff
@@ -1,11 +1,37 @@
...
+  exclude: (f: string) =>
+    f.includes('.test.') ||
+    f.includes('.stories.') ||
+    f.endsWith('.d.ts') ||
+    f.includes('/tests/'),
+})
+  .filter(file => !file.endsWith('sitemap/cli.ts'))
+  .map(file => resolve(__dirname, file))
 
 export default mergeConfig(
```
nickgros, line 22: Similar build strategy to synapse-react-client, synapse-portal-framework

packages/markdown-it-synapse/package.json
```diff
@@ -1,14 +1,13 @@
 {
   "name": "markdown-it-synapse",
-  "version": "1.1.17",
+  "version": "2.0.0",
```
nickgros, line 3: Build is now ESM-only (breaking change), so we bump the major version.

If/when SWC needs to update this, SWC's Vite build can consume the ESM and re-export it as a global (it already does this for MUI, React Query, and SRC)

packages/markdown-it-synapse/package.json
```diff
@@ -25,7 +24,7 @@
   ],
   "scripts": {
     "type-check": "tsc --build",
-    "build": "tsup",
```
nickgros, line 28: Remove tsup. The version we had installed is not compatible with TypeScript 6. It made more sense to use Vite to build just like our other libs instead of upgrading or picking a different build tool

jay-hodgson (approved):

PR #2743, opened 2026-04-28
SWC-7770: Curator grid validation message enhancements

afwillia:
- Display the Validation Message banner at all times on top of the grid.
- Display total number of validation errors.
- Scroll through all messages by row, column, or message.
- Link to the invalid cell.
- Hide validation messages during initial grid sync

Collapsed
collapsed

</br>
Expanded
expanded

</br>
No errors
clean

</br>
Sync
sync

### DIFF
--- a/packages/synapse-react-client/src/components/DataGrid/DataGrid.tsx
+++ b/packages/synapse-react-client/src/components/DataGrid/DataGrid.tsx
@@ -36,10 +36,6 @@ type DataGridProps = {
   lastSelection: SelectionWithId | null
   handleChange: (newValue: DataGridRow[], operations: Operation[]) => void
   handleSelectionChange: (opts: { selection: SelectionWithId | null }) => void
-  onSelectedRowChange?: (
-    rowIndex: number | null,
-    row: DataGridRow | null,
-  ) => void
   remoteSelections?: readonly RemoteSelection[]
 }
 
@@ -60,7 +56,6 @@ export default function DataGrid(props: DataGridProps) {
     lastSelection,
     handleChange,
     handleSelectionChange,
-    onSelectedRowChange,
     remoteSelections,
   } = props
 
@@ -212,15 +207,9 @@ export default function DataGrid(props: DataGridProps) {
   // Wrap onActiveCellChange in useCallback
   const handleActiveCellChange = useCallback(
     ({ cell }: { cell: { row: number; col: number } | null }) => {
-      if (cell) {
-        setSelectedRowIndex(cell.row)
-        onSelectedRowChange?.(cell.row, rowValues[cell.row])
-      } else {
-        setSelectedRowIndex(null)
-        onSelectedRowChange?.(null, null)
-      }
+      setSelectedRowIndex(cell ? cell.row : null)
     },
-    [onSelectedRowChange, rowValues],
+    [],
   )
 
   // Wrapper ref for the grid container
--- a/packages/synapse-react-client/src/components/DataGrid/SynapseGrid.tsx
+++ b/packages/synapse-react-client/src/components/DataGrid/SynapseGrid.tsx
@@ -324,18 +324,9 @@ const SynapseGrid = forwardRef<SynapseGridHandle, SynapseGridProps>(
       [applyAndCommitChanges, model, replicaId],
     )
 
-    // Track selected row index for validation display
-    const selectedRowIndexRef = useRef<number | null>(null)
-    const [, forceUpdate] = useState({})
-
-    const handleSelectedRowChange = useCallback(
-      (rowIndex: number | null, _row: DataGridRow | null) => {
-        // Only update when a real row is selected — don't clear on blur/click-away
-        // so the ValidationAlert stays open while the user interacts with it.
-        if (rowIndex !== null) {
-          selectedRowIndexRef.current = rowIndex
-          forceUpdate({})
-        }
+    const handleNavigateToCell = useCallback(
+      (rowIndex: number, colIndex: number) => {
+        gridRef.current?.setActiveCell({ col: colIndex, row: rowIndex })
       },
       [],
     )
@@ -453,6 +444,14 @@ const SynapseGrid = forwardRef<SynapseGridHandle, SynapseGridProps>(
               {hasSufficientData && (
                 <>
                   <Grid size={12}>
+                    <ValidationAlert
+                      rowValues={rowValues}
+                      columnNames={modelSnapshot?.columnNames ?? []}
+                      columnOrder={modelSnapshot?.columnOrder ?? []}
+                      onNavigateToCell={handleNavigateToCell}
+                    />
+                  </Grid>
+                  <Grid size={12}>
                     <Stack
                       direction={'row'}
                       spacing={1}
@@ -509,14 +508,7 @@ const SynapseGrid = forwardRef<SynapseGridHandle, SynapseGridProps>(
                       lastSelection={lastSelection}
                       handleChange={handleChange}
                       handleSelectionChange={handleSelectionChange}
-                      onSelectedRowChange={handleSelectedRowChange}
                       remoteSelections={remoteSelections}
-                    />
-                  </Grid>
-                  <Grid size={12}>
-                    <ValidationAlert
-                      selectedRowIndex={selectedRowIndexRef.current}
-                      rowValues={rowValues}
                     />
                   </Grid>
                 </>
--- a/packages/synapse-react-client/src/components/DataGrid/SynapseGrid.tsx
+++ b/packages/synapse-react-client/src/components/DataGrid/SynapseGrid.tsx
@@ -455,7 +455,7 @@ const SynapseGrid = forwardRef<SynapseGridHandle, SynapseGridProps>(
                       columnNames={modelSnapshot?.columnNames ?? []}
                       columnOrder={modelSnapshot?.columnOrder ?? []}
                       onNavigateToCell={handleNavigateToCell}
-                      isSyncing={!hasCompletedInitialSync}
+                      isLoading={!hasCompletedInitialSync}
                     />
                   </Grid>
                   <Grid size={12}>
--- a/packages/synapse-react-client/src/components/DataGrid/SynapseGrid.tsx
+++ b/packages/synapse-react-client/src/components/DataGrid/SynapseGrid.tsx
@@ -9,7 +9,13 @@ import { SkeletonTable } from '@/components/index'
 import { useGetEntity } from '@/synapse-queries/index'
 import { getSchemaPropertiesInfo } from '@/utils/jsonschema/getSchemaPropertyInfo'
 import { SmartToyTwoTone } from '@mui/icons-material'
-import { Stack, Tooltip } from '@mui/material'
+import {
+  Box,
+  CircularProgress,
+  Stack,
+  Tooltip,
+  Typography,
+} from '@mui/material'
 import Grid from '@mui/material/Grid'
 import {
   CreateGridRequest,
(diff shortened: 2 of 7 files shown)


packages/synapse-react-client/src/components/DataGrid/DataGrid.tsx
```diff
@@ -36,10 +36,6 @@ type DataGridProps = {
   lastSelection: SelectionWithId | null
   handleChange: (newValue: DataGridRow[], operations: Operation[]) => void
   handleSelectionChange: (opts: { selection: SelectionWithId | null }) => void
-  onSelectedRowChange?: (
```
afwillia, line 39: `onSelectedRowChange` was only used to display the validation message banner for the correct row. Can remove entirely.

packages/synapse-react-client/src/components/DataGrid/SynapseGrid.tsx
```diff
@@ -324,18 +324,9 @@ const SynapseGrid = forwardRef<SynapseGridHandle, SynapseGridProps>(
       [applyAndCommitChanges, model, replicaId],
     )
 
-    // Track selected row index for validation display
```
afwillia, line 327: `handleSelectedRowChange` was only used to display the validation message banner for the correct row. Can remove entirely.

packages/synapse-react-client/src/components/DataGrid/components/ValidationAlert.tsx
```diff
@@ -1,41 +1,395 @@
...
+              message={firstError.message}
+              rowIndex={firstError.rowIndex}
+            />
+          </Typography>
+        )}
+      </Box>
+
+      {/* Expanded panel */}
+      <Collapse in={isExpanded}>
+        <Tabs
```
afwillia, line 418: Add tabs to arrange messages by row, column, or message

packages/synapse-react-client/src/style/components/_data-grid-extra.scss
```diff
@@ -127,6 +127,8 @@ $remote-selection-colors: #1677ff, #52c41a, #fa8c16, #eb2f96, #13c2c2, #9254de,
   }
 
   &.cell-unknown {
+    background: $color-unknown;
```
afwillia, line 130: Change the background of recently edited cells awaiting validation status updates.

packages/synapse-react-client/src/components/DataGrid/utils/getCellClassName.ts
```diff
@@ -49,8 +52,15 @@ export function getCellClassName(params: {
     }
   }
 
-  if (isInvalid) {
-    classList.push('cell-invalid')
+  if (hasCellError) {
```
afwillia, line 55: Set edited cells to pending, like the row gutter.

packages/synapse-react-client/src/components/DataGrid/SynapseGrid.tsx
```diff
@@ -452,12 +449,36 @@ const SynapseGrid = forwardRef<SynapseGridHandle, SynapseGridProps>(
...
+                      isSyncing={!hasCompletedInitialSync}
+                    />
+                  </Grid>
                   <Grid size={12}>
                     <Stack
                       direction={'row'}
                       spacing={1}
                       sx={{ justifyContent: 'flex-end' }}
                     >
+                      {!hasCompletedInitialSync && (
```
afwillia, line 467: Additional loading spinner may be overkill. Currently only displays when connecting to session, not for sending/receiving patches mid-session. What do you think?

packages/synapse-react-client/src/components/DataGrid/components/ValidationAlert.tsx
```diff
@@ -1,41 +1,483 @@
...
+    if (allErrors.length === 0) setIsExpanded(false)
+  }, [allErrors.length])
 
-  const validationMessages =
-    selectedRow.__validationResults?.allValidationMessages
-  if (!Array.isArray(validationMessages) || validationMessages.length === 0) {
-    return null
+  const firstError = allErrors[0]
+
+  if (isSyncing) {
```
afwillia: Don't show messages until the client's clock is synced. Avoids displaying stale messages and sudden updates (especially with slow connections).

packages/synapse-react-client/src/components/DataGrid/SynapseGrid.tsx
nickgros: I think this is good, and could be tweaked in the future to show anytime we're downloading new patches if we decide we want it.

At the very least, it communicates that the initial data is still loading, even if the grid is visible, so it's an improvement.

packages/synapse-react-client/src/components/DataGrid/components/ValidationAlert.tsx
```diff
@@ -1,41 +1,483 @@
...
+      msgGroup.get(columnName)!.push({ rowIndex, pending })
+    })
+    return groups
+  }, [allErrors])
 
-  const selectedRow = rowValues[selectedRowIndex]
-  if (!selectedRow) return null
+  useEffect(() => {
+    if (allErrors.length === 0) setIsExpanded(false)
+  }, [allErrors.length])
```
nickgros: Instead of `useEffect` you could derive isExpanded:

```ts
const shouldExpand = isExpanded && allErrors.length === 0
```

packages/synapse-react-client/src/components/DataGrid/components/ValidationAlert.tsx
```diff
@@ -1,41 +1,483 @@
...
+  }
+  return (
+    <>
+      <Box component="span" sx={{ fontWeight: 'bold' }}>
+        {columnName}
+      </Box>
+      <Box component="span" sx={{ mx: 0.5 }}>
+        —
+      </Box>
+      {message}
```
nickgros: Design issue: while this is a single link, the span margin breaks up the underline decoration, which makes it look like multiple links:

Image

You could probably fix this by using space characters `{' '}` instead of margin

packages/synapse-react-client/src/components/DataGrid/components/ValidationAlert.tsx
```diff
@@ -1,41 +1,483 @@
...
+          borderLeftColor: 'divider',
+          borderRadius: 1,
+          mb: 1,
+          px: 2,
+          py: 1,
+          display: 'flex',
+          alignItems: 'center',
+          gap: 1,
+        }}
+      >
```
nickgros, line 324: Lots of duplicate/repeated styles for the container. To prevent drift/layout changes, these styles should be shared as much as possible (or just render a single container with dynamic styles/content).

packages/synapse-react-client/src/components/DataGrid/components/ValidationAlert.tsx
```diff
@@ -1,41 +1,483 @@
...
+        <Link
+          component="button"
+          variant="body2"
+          onClick={() => setIsExpanded(p => !p)}
+          sx={{ cursor: 'pointer', flexShrink: 0 }}
+        >
+          {isExpanded ? 'Collapse' : 'Expand'}
+        </Link>
+        {!isExpanded && (
+          <Link
```
nickgros, line 382: IMO there should be more space between the clickable message and the left side title/chip/expand button. You could justify this towards the end of the flex container.

packages/synapse-react-client/src/components/DataGrid/components/ValidationAlert.tsx
```diff
@@ -1,41 +1,483 @@
...
+      <Box
+        sx={{
+          display: 'flex',
+          alignItems: 'center',
+          px: 2,
+          py: 1,
+          gap: 1,
+        }}
+      >
+        <Typography variant="body2" fontWeight="bold">
```
nickgros: fonts in general should be bigger imo, consider using `body1`

packages/synapse-react-client/src/components/DataGrid/components/ValidationAlert.tsx
nickgros: Also I think the container needs more vertical padding

packages/synapse-react-client/src/components/DataGrid/components/ValidationAlert.tsx
```diff
@@ -1,41 +1,483 @@
...
+            label="By message"
+            value="message"
+            sx={{ minHeight: 36, py: 0.5, fontSize: '0.75rem' }}
+          />
+        </Tabs>
+
+        <Box sx={{ px: 2, py: 1, maxHeight: 200, overflowY: 'auto' }}>
+          {/* ── By row ─────────────────────────────────────────── */}
+          {groupBy === 'row' && (
+            <Box component="ul" sx={{ m: 0, p: 0, listStyle: 'none' }}>
```
nickgros: To get closer to Adam's mockup (and visually separate each item), consider using a MUI `Stack` with the [`divider` prop](https://mui.com/material-ui/react-stack/#dividers)

I did a similar thing for the [OpenInvitationsToUser](https://sage-bionetworks.github.io/synapse-web-monorepo/?path=/story/features-team-openinvitationstouser--with-pending-invitations) component:

Image

This suggestion applies to the content of all tabs

packages/synapse-react-client/src/components/DataGrid/components/ValidationAlert.tsx
```diff
@@ -1,41 +1,483 @@
...
+          {groupBy === 'column' &&
+            Array.from(byColumn.entries()).map(([colName, messages]) => {
+              const totalCount = Array.from(messages.values()).reduce(
+                (sum, rows) => sum + rows.length,
+                0,
+              )
+              const navCol = Math.max(
+                0,
+                getColDisplayIndex(colName, columnNames, columnOrder),
+              )
```
nickgros: This is a lot of inline logic for the component, consider precomputing this groupBy or extracting to a helper component

packages/synapse-react-client/src/components/DataGrid/components/ValidationAlert.tsx
```diff
@@ -1,41 +1,483 @@
...
+                    return (
+                      <Box
+                        key={colName}
+                        sx={{
+                          pl: 2,
+                          display: 'flex',
+                          alignItems: 'baseline',
+                          flexWrap: 'wrap',
+                          gap: 0.5,
+                          py: 0.25,
```
nickgros: Lots of duplicate style declarations throughout this module. Recommend using an SCSS module file with a shared classname or extracting the styles as constant `SxProps`

packages/synapse-react-client/src/components/DataGrid/components/ValidationAlert.tsx
nickgros: I prefer a different name over `isSyncing` because I don't think it's quite accurate. We could go one of two directions:

1. Just call it `isLoading` and the caller passes in `hasCompletedInitialSync`. The internals of this component do not care how `isLoading` was determined.
2. Call it `hasCompletedInitialSync` because it is more correct, and this component needs to consider the sync process.

I am partial to (1) but I think either is an improvement

nickgros (changes requested):

packages/synapse-react-client/src/components/DataGrid/SynapseGrid.tsx
afwillia: I was thinking the same regarding incoming patches, but decided the implementation was out of scope for this ticket.

packages/synapse-react-client/src/components/DataGrid/components/ValidationAlert.tsx
afwillia: funny, I considered renaming to `isLoading` - will do now! `isSyncing` in the future can refer to messages after the initial load. But this component won't need to handle that.

nickgros (approved):

PR #2745, opened 2026-04-29
PORTALS-4198

jay-hodgson:


apps/portals/cancercomplexity/src/config/synapseConfigs/datasets.tsx
```diff
@@ -101,18 +107,5 @@ export const datasetsQueryWrapperPlotNavProps: QueryWrapperPlotNavProps = {
       textMatchesMode: 'BOOLEAN',
       distance: 3,
     },
-    searchable: [
```
jay-hodgson, line 104: FTS is on, specifying searchable columns is unnecessary (and not referenced)

packages/synapse-react-client/src/components/EntityDownloadConfirmation/EntityDownloadConfirmation.tsx
```diff
@@ -50,7 +52,10 @@ export function EntityDownloadConfirmation({
   handleClose,
   onIsLoadingChange,
 }: EntityDownloadConfirmationProps) {
-  const { data: entity, isLoading: isLoadingEntity } = useGetEntity(entityId)
+  const { isAuthenticated } = useSynapseContext()
```
jay-hodgson, line 55: Bug fix.  When anonymous, and EntityDownloadConfirmation was shown, the useGetEntity may not load and the user would never be prompted to log in.  Now, explicitly look for isAuthenticated, and show the DownloadConfirmationUI (which already handles the anonymous case)

Image

jay-hodgson:
With card adornment for the downloadType:
Screenshot 2026-04-29 at 10 21 27 AM

After clicking download when logged in:
Screenshot 2026-04-29 at 10 22 39 AM

packages/synapse-react-client/src/components/EntityDownloadConfirmation/EntityDownloadConfirmation.tsx
```diff
@@ -92,6 +97,16 @@ export function EntityDownloadConfirmation({
     }, [entityType, entityId, versionNumber])
 
   if (!addToDownloadListRequest) {
+    if (!isAuthenticated) {
+      return (
+        <div>
+          <DownloadConfirmationUI
+            onAddToDownloadCart={() => {}}
```
nickgros:
```suggestion
            onAddToDownloadCart={() => {
            	// Only displaying UI to prompt sign-in, so no-op is OK
            }}
```

nickgros (approved):
LGTM, suggest adding a comment because I was looking at this funny for a couple of minutes until I understood

PR #2746, opened 2026-04-29
SWC-7794: Replace GWT CSV upload wizard with React CsvPreviewDialog

kianamcc:
Jira: https://sagebionetworks.jira.com/browse/SWC-7794?atlOrigin=eyJpIjoiNDI1NzlmYjg4ZjYxNDc5NDkxZGQ0MzEwZmQyODZmYWIiLCJwIjoiaiJ9

Previously, CsvPreviewDialog only handled file upload and csv preview steps. Extend CsvPreviewDialog to be able to create new tables and append to existing tables.

Create: If parentId is provided, creates column models, a new table entity, then uploads the csv rows. Has an extra TABLE_NAME step (table name and schema editor) to cover what we see in the GWT implementation.

Append: If tableId is provided, upload csv rows directly to existing table.

Grid: onConfirm is provided, existing behavior should be unchanged

CreateTableViewWizard.tsx often referenced.

New table steps:
Screenshot 2026-04-29 at 1 09 11 PM
Screenshot 2026-04-29 at 1 09 32 PM
Screenshot 2026-04-29 at 1 09 53 PM

Existing table steps:
Screenshot 2026-04-29 at 1 10 23 PM
Screenshot 2026-04-29 at 1 10 29 PM

Grid upload:
Screenshot 2026-04-29 at 2 45 31 PM
Screenshot 2026-04-29 at 2 46 36 PM
Screenshot 2026-04-29 at 2 49 49 PM


packages/synapse-react-client/src/components/DataGrid/components/UploadCsvToGridDialog.test.tsx
```diff
@@ -51,7 +51,7 @@ describe('UploadCsvToGridDialog', () => {
       })
 
       act(() => {
-        mockCsvPreviewDialog.mock.lastCall![0].onConfirm(
+        mockCsvPreviewDialog.mock.lastCall![0].onConfirm!(
```
kianamcc: onConfirm is now optional (only grid uses it via UploadCsvToGridDialog)

packages/synapse-react-client/src/components/table/CsvPreview/CsvPreviewDialog.stories.tsx
```diff
@@ -19,6 +19,16 @@ const meta: Meta<CsvPreviewDialogProps> = {
 export default meta
 type Story = StoryObj<typeof meta>
 
-export const Demo: Story = {
-  name: 'CSV Preview Dialog',
+export const GridMode: Story = {
```
kianamcc: I didn't make a story for append because the UI was the same as Grid.

packages/synapse-react-client/src/components/table/CsvPreview/CsvPreviewDialog.tsx
```diff
@@ -125,6 +320,35 @@ export default function CsvPreviewDialog(props: CsvPreviewDialogProps) {
...
+        label={'Table Name'}
+        value={tableName}
+        onChange={e => setTableName(e.target.value)}
+        fullWidth
+      />
+      <Typography variant={'body1'}>
+        Use the schema options button to make changes to the columns of the
+        table. Use the create button to finish building the table.
+      </Typography>
+      <Accordion>
```
kianamcc: Similar UI as "Show Options" in csv preview page for consistency

packages/synapse-react-client/src/components/table/CsvPreview/CsvPreviewDialog.tsx
```diff
@@ -75,6 +151,125 @@ export default function CsvPreviewDialog(props: CsvPreviewDialogProps) {
...
+      let resolvedTableId = tableId
+
+      if (!tableId) {
+        let createdColumnModelResults: SynapseTypesColumnModel[]
+        try {
+          createdColumnModelResults = await createColumnModels(
+            columnModelsParam,
+          )
+        } catch {
+          return
```
kianamcc: handled by each hook's error and displayed via ErrorBanner

packages/synapse-react-client/src/components/table/CsvPreview/CsvPreviewDialog.tsx
```diff
@@ -75,6 +151,125 @@ export default function CsvPreviewDialog(props: CsvPreviewDialogProps) {
   const uploadRef = useRef<FileUploadHandle | null>(null)
   const csvDescriptorFormRef = useRef<CsvTableDescriptorFormHandle | null>(null)
 
+  /**
+   * If no tableId is provided, creates column models and a new table entity, then uploads the CSV to the new table.
+   * If a tableId is provided, uploads the csv to the existing table.
+   * @param columnModelsParam - The schema defined by the user. Defaults to empty array for append flow.
+   */
+  const handleFinish = useCallback(
```
kianamcc: Currently, this handles two different tasks, creating and uploading; Would it be better to split these into two distinct functions or can we leave it as is?

packages/synapse-react-client/src/components/table/CsvPreview/CsvPreviewDialog.tsx
```diff
@@ -75,6 +151,125 @@ export default function CsvPreviewDialog(props: CsvPreviewDialogProps) {
...
+    async (
+      columnModelsParam: SetOptional<SynapseTypesColumnModel, 'id'>[] = [],
```
kianamcc: We have the schema editor (TableColumnSchemaForm) in the TABLE_NAME step (like in original gwt implementation) so we pass the final edited values directly from the onSubmit

packages/synapse-react-client/src/components/table/CsvPreview/CsvPreviewDialog.tsx
nickgros: If you create a new mutation that itself calls the `mutateAsync` functions, then you could just let these throw and the wrapper mutation will catch and give you a single error message. Some example mutations that call other mutations in our codebase (unfortunately most are in the grid logic):  `useInitializeGridConnection`, `useMergeGridWithTable`, `useMergeGridWithSource`

packages/synapse-react-client/src/components/table/CsvPreview/CsvPreviewDialog.tsx
```diff
@@ -8,24 +11,38 @@ import CsvPreview from '@/components/table/CsvPreview/CsvPreview'
...
+import {
+  Entity,
+  ColumnModel as SynapseTypesColumnModel,
+  Table,
+} from '@sage-bionetworks/synapse-types'
 
 enum CsvPreviewDialogStep {
   UPLOAD_CSV = 0,
   COLUMN_PREVIEW = 1,
+  TABLE_NAME = 2,
```
nickgros: In SWC, the CSV preview step is decoupled from the actual table update logic. Can we follow that convention? Instead of making table creation a part of CSV Preview, let's add new components that each have a single responsibility. One component for creating a TableEntity using a CSV, and one component for updating a Table (interface) using a CSV.

These new components will likely manage their own dialog state, and can use the `CSVPreview` component.

nickgros (changes requested):

packages/synapse-react-client/src/components/table/CsvPreview/CsvPreviewDialog.tsx
kianamcc: Created new useUploadCsvToTable wrapper to handle the mutateAsync functions. Was able to get rid of the nested try/catch blocks.

packages/synapse-react-client/src/components/table/CsvPreview/CsvPreviewDialog.tsx
kianamcc: CsvPreviewDialog only handles preview now. Split the create and append flow into separate  components.

kianamcc:
Screenshot 2026-05-22 at 11 17 29 AM
Screenshot 2026-05-22 at 11 14 59 AM

packages/synapse-react-client/src/components/table/CsvPreview/CsvPreviewDialog.tsx
```diff
@@ -67,6 +67,22 @@ export default function CsvPreviewDialog(props: CsvPreviewDialogProps) {
...
+      setCsvPreviewData(null)
+      setCsvTableDescriptor({
+        separator: ',',
+        quoteCharacter: '"',
+        escapeCharacter: '\\',
+        lineEnd: '\n',
+        isFirstLineHeader: true,
+      })
+    }
+  }, [open])
```
nickgros: IMO, it's easier if the caller just passes `open` as the `key` prop instead of having this effect. It would be less code to maintain since this effect needs to be updated if the set of state tracked by this component changes.

packages/synapse-react-client/src/components/table/CsvPreview/CreateTableFromCsvDialog.tsx
```diff
@@ -0,0 +1,297 @@
...
+      setColumnModels([])
+      setCsvTableDescriptor({
+        separator: ',',
+        quoteCharacter: '"',
+        escapeCharacter: '\\',
+        lineEnd: '\n',
+        isFirstLineHeader: true,
+      })
+    }
+  }, [open])
```
nickgros: As mentioned in another file, I prefer we drop the `useEffect` and just use `open` as the `key` prop where this component is used

packages/synapse-react-client/src/components/table/CsvPreview/CreateTableFromCsvDialog.tsx
```diff
@@ -0,0 +1,297 @@
...
+      ref={uploadRef}
+      allowMultipleUpload={false}
+      onFileUploadComplete={fileHandleId => {
+        onFileUploaded(fileHandleId)
+      }}
+      disableDragAndDrop={true}
+    />
+  )
+
+  const previewStepContent = (
```
nickgros: This is highly duplicated with CsvPreviewDialog.tsx and UpdateTableWithCsvDialog.tsx

Could we extract a common shared component, `CsvPreviewWithOptions`?

We might also benefit from creating a custom hook to share stateful logic if necessary, but that depends on if the parent components need to 'own' a lot of state that just gets passed to `CsvPreviewWithOptions`. If most state ends up being owned by `CsvPreviewWithOptions`, then that becomes less useful.

packages/synapse-react-client/src/components/table/CsvPreview/useUploadCsvToTable.ts
```diff
@@ -0,0 +1,103 @@
...
+  tableName: string
+  columnModels: SetOptional<SynapseTypesColumnModel, 'id'>[]
+}
+
+type UploadCsvToTableProps = {
+  tableId?: string
+  csvTableDescriptor: CsvTableDescriptor
+  fileHandleId: string
+  createTableProps?: CreateTableProps
+}
```
nickgros: Overloading create + update is kind of awkward, it makes the arguments confusing (`tableId` and `createTableProps` become mutually exclusive, but it's not super obvious).


 Best thing might be to split this into two hooks, `useUploadCsvToExistingTable` and 
`useCreateTableFromCsv`?

packages/synapse-react-client/src/components/table/CsvPreview/UpdateTableWithCsvDialog.tsx
```diff
@@ -0,0 +1,188 @@
...
+      onSuccess() {
+        onSuccess?.()
+      },
+      onError(error) {
+        displayToast(error.message, 'danger')
+      },
+    })
+
+  // Reset local state when dialog is closed
+  useEffect(() => {
```
nickgros: Ditto other files, prefer using `key` over resetting state with `useEffect` if we can avoid it

packages/synapse-react-client/src/components/table/CsvPreview/UpdateTableWithCsvDialog.tsx
```diff
@@ -0,0 +1,188 @@
...
+      ref={uploadRef}
+      allowMultipleUpload={false}
+      onFileUploadComplete={fileHandleId => {
+        onFileUploaded(fileHandleId)
+      }}
+      disableDragAndDrop={true}
+    />
+  )
+
+  const previewStepContent = (
```
nickgros: Ditto other files, add something like a `CsvPreviewWithOptions` for consistent UI and less duplicate code

packages/synapse-react-client/src/components/table/CsvPreview/CreateTableFromCsvDialog.tsx
```diff
@@ -0,0 +1,297 @@
...
+  const { open, onClose, parentId, onSuccess } = props
+  const [step, setStep] = useState(CreateTableFromCsvDialogStep.UPLOAD_CSV)
+  const [csvTableDescriptor, setCsvTableDescriptor] =
+    useState<CsvTableDescriptor>({
+      separator: ',',
+      quoteCharacter: '"',
+      escapeCharacter: '\\',
+      lineEnd: '\n',
+      isFirstLineHeader: true,
+    })
```
nickgros: I think this duplicated initial state might get collapsed into a single `CsvPreviewWithOptions` component, but if not, consider adding a `getDefaultCsvTableDescriptor()` function used by multiple components

nickgros (changes requested):

packages/synapse-react-client/src/components/table/CsvPreview/useCreateTableFromCsv.ts
```diff
@@ -0,0 +1,103 @@
...
+        ;(entityToCreate as Table)['columnIds'] = createdColumnModelResults.map(
+          cm => cm.id,
+        )
+
+        const newEntity = await createEntity(entityToCreate)
+
+        resolvedTableId = newEntity.id!
+      }
+
+      const tableUpdateRequest: TableUpdateTransactionRequest = {
```
nickgros, line 70: In SWC (`UploadCSVFinishPageImpl`), if the descriptor indicated that the first line was the header, then `linesToSkip` was set to 1 (or incremented if it was some other value).

Can you investigate to see we need to maintain that logic? It's possible the backend CSV upload logic is smart enough to not require this, but it would be nice to know for sure.

packages/synapse-react-client/src/components/table/CsvPreview/CreateTableFromCsvDialog.tsx
kianamcc: Removed the useEffect, looks like SWC will need need an update.

packages/synapse-react-client/src/components/table/CsvPreview/useCreateTableFromCsv.ts
kianamcc: I seem to be getting the same behavior in prod and local (pnpm linked src), Line 1 of CSV becomes data when isFirstLineHeader=false:
Screenshot 2026-05-29 at 11 12 13 AM

Confirming that the new SRC version does not set linesToSkip, only passes csvTableDescriptor (which contains isFirstLineHeader).

Copilot pointed me to CSVToRowIterator.java in Synapse Repository Services (https://github.com/Sage-Bionetworks/Synapse-Repository-Services/blob/develop/lib/jdomodels/src/main/java/org/sagebionetworks/repo/model/dbo/dao/table/CSVToRowIterator.java), I think the backend already automatically handles the logic for headers, for both isFirstLineHeader=true and isFirstLineHeader=false without needing to set linesToSkip. 

isFirstLineHeader=true the header line is automatically consumed (starting around line 86) and when false the headers are built from the schema instead (around line 125).

packages/synapse-react-client/src/components/table/CsvPreview/UpdateTableWithCsvDialog.tsx
kianamcc: Same here, removed, will update SWC.

packages/synapse-react-client/src/components/table/CsvPreview/CreateTableFromCsvDialog.tsx
```diff
@@ -0,0 +1,223 @@
...
+  open: boolean
+  /** Callback when the dialog is closed */
+  onClose: () => void
+  /** Destination project/folder to upload to */
+  parentId: string
+  /** Invoked after all backend operations successfully complete */
+  onSuccess?: () => void
+}
+
+export default function CreateTableFromCsvDialog(
```
nickgros, line 47: Add a component-level comment describing the purpose of this component

packages/synapse-react-client/src/components/table/CsvPreview/UpdateTableWithCsvDialog.tsx
```diff
@@ -0,0 +1,124 @@
...
+    <CsvPreviewWithOptions
```
nickgros: is it intentional for this one to omit `onCsvPreviewDataChange={setCsvPreviewData}`?

packages/synapse-react-client/src/components/table/CsvPreview/useCsvUploadPreview.ts
```diff
@@ -0,0 +1,53 @@
...
+  quoteCharacter: '"',
+  escapeCharacter: '\\',
+  lineEnd: '\n',
+  isFirstLineHeader: true,
+}
+
+export enum CsvUploadPreviewStep {
+  UPLOAD_CSV = 0,
+  COLUMN_PREVIEW = 1,
+  TABLE_NAME = 2,
```
nickgros: TABLE_NAME is not always used, so I think you should remove it from this step enum. The caller that uses this should manage that state by itself.

I am wondering if we could capture the entire upload -> preview flow (state + component) in just this hook? Details below...

packages/synapse-react-client/src/components/table/CsvPreview/useCsvUploadPreview.ts
```diff
@@ -0,0 +1,53 @@
...
+    csvTableDescriptor,
+    setCsvTableDescriptor,
+    uploadedFileHandleId,
+    csvPreviewData,
+    setCsvPreviewData,
+    isLoadingPreview,
+    setIsLoadingPreview,
+    onFileUploaded,
+  }
+}
```
nickgros, line 52: It looks like some of these state variables are only used by `BasicFileHandleUpload` or `CsvPreviewWithOptions` and you never seem to need to pass other component info in, so we could instead create the components in the hook (you will have to rename this file to `.tsx`).

Then, this hook manages everything related to CSV Preview and Upload, and exposes only the bare minimum state needed for the caller to use it!


```suggestion
  const csvUpload = (
    <BasicFileHandleUpload
      ref={uploadRef}
      allowMultipleUpload={false}
      onFileUploadComplete={fileHandleId => {
        onFileUploaded(fileHandleId)
      }}
      disableDragAndDrop={true}
    />
  )

  const csvPreview = (<CsvPreviewWithOptions
      fileHandleId={uploadedFileHandleId}
      csvTableDescriptor={csvTableDescriptor}
      onCsvTableDescriptorChange={setCsvTableDescriptor}
      onCsvPreviewDataChange={setCsvPreviewData}
      onIsLoadingChange={setIsLoadingPreview}
    />)
    
 const csvUploadPreview = step === CsvUploadPreviewStep.UPLOAD_CSV ? csvUpload : csvPreview
 
 return {
    step,
    csvTableDescriptor,
    uploadedFileHandleId,
    csvPreviewData,
    isLoadingPreview,
    onFileUploaded,
    csvUploadPreview
 } 
}
```

So we can remove `setStep`, `setCsvTableDescriptor`, `setCsvPreviewData`, `setIsLoadingPreview`, and `onFileUploaded` since they are all only used internally.

Then a caller itself could just do

```tsx
  const {    
    step: csvUploadPreviewStep,
    csvTableDescriptor,
    uploadedFileHandleId,
    isLoadingPreview,
    onFileUploaded,
  } = useCsvUploadPreview()

// ...render the component. Note that that the step enum is a different one managed by the caller, e.g. CreateTableFromCsvDialog
<>
  {step === CreateTableFromCsvStep.CSV_UPLOAD_PREVIEW && csvUploadPreview}
</>

// ... to render the "Next" button (maybe you can find a simpler way to express this)
<>
  {step === CreateTableFromCsvStep.CSV_UPLOAD_PREVIEW &&
      csvUploadPreviewStep === CsvUploadPreviewStep.COLUMN_PREVIEW &&
      !isLoadingPreview && (
        <Button {/* ...you get the idea... */}>Next</Button>
      )}
</>

```

packages/synapse-react-client/src/components/table/CsvPreview/CsvPreviewWithOptions.tsx
```diff
@@ -0,0 +1,80 @@
...
+  csvTableDescriptor: CsvTableDescriptor
+  /** Called when the user clicks "Refresh Preview" with the form's current values */
+  onCsvTableDescriptorChange: (descriptor: CsvTableDescriptor) => void
+  /** Called when the CSV preview data is fetched */
+  onCsvPreviewDataChange?: (data: UploadToTablePreviewResult) => void
+  /** Called when the preview load state changes */
+  onIsLoadingChange?: (isLoading: boolean) => void
+}
+
+export default function CsvPreviewWithOptions({
```
nickgros, line 38: Add a component level comment. If you decide to follow my recommendation and render this from the hook (and it isn't used in any other production code), then I think the comment should also instruct the person reading it to use `useCsvUploadPreview`, because it also handles upload and state tracking/sharing.

packages/synapse-react-client/src/components/table/CsvPreview/UpdateTableWithCsvDialog.tsx
```diff
@@ -0,0 +1,124 @@
...
+    useUploadCsvToExistingTable({
+      onSuccess() {
+        onSuccess?.()
+      },
+      onError(error) {
+        displayToast(error.message, 'danger')
+      },
+    })
+
+  const uploadRef = useRef<FileUploadHandle | null>(null)
```
nickgros: I think `uploadRef` can also go into the useCsvUploadPreview hook

packages/synapse-react-client/src/components/table/CsvPreview/useCreateTableFromCsv.ts
```diff
@@ -0,0 +1,88 @@
...
+      TableUpdateTransactionResponse,
+      SynapseClientError,
+      CreateTableFromCsvProps
+    >,
+    'mutationFn'
+  >,
+) {
+  const { mutateAsync: createColumnModels } = useCreateColumnModels()
+  const { mutateAsync: createEntity } = useCreateEntity()
+  const { mutateAsync: tableTransaction } = useTableUpdateTransaction()
```
nickgros: Nit: rename to be more clear that this is an update operation

```suggestion
  const { mutateAsync: updateTableInTransaction } = useTableUpdateTransaction()
```

nickgros (changes requested):

packages/synapse-react-client/src/components/table/CsvPreview/useCsvUploadPreview.ts
nickgros: I had a conversation with an agent and it had an enhancement to this approach (combine the upload + preview states), but keep the render logic in its own component to keep state/UI separate and testable.

So you would keep the hook mostly the same as you have it (still making the recommended `step` changes):

```ts
import { useCallback, useState } from 'react'
import {
  CsvTableDescriptor,
  UploadToTablePreviewResult,
} from '@sage-bionetworks/synapse-client'

export enum CsvUploadPreviewStep {
  UPLOAD_CSV = 0,
  COLUMN_PREVIEW = 1,
}

const DEFAULT_CSV_TABLE_DESCRIPTOR: CsvTableDescriptor = {
  separator: ',',
  quoteCharacter: '"',
  escapeCharacter: '\\',
  lineEnd: '\n',
  isFirstLineHeader: true,
}

export default function useCsvUploadPreview() {
  const [step, setStep] = useState(CsvUploadPreviewStep.UPLOAD_CSV)
  const [csvTableDescriptor, setCsvTableDescriptor] =
    useState<CsvTableDescriptor>(DEFAULT_CSV_TABLE_DESCRIPTOR)
  const [uploadedFileHandleId, setUploadedFileHandleId] = useState<
    string | null
  >(null)
  const [csvPreviewData, setCsvPreviewData] =
    useState<UploadToTablePreviewResult | null>(null)
  const [isLoadingPreview, setIsLoadingPreview] = useState(false)

  const onFileUploaded = useCallback((fileHandleId: string) => {
    setUploadedFileHandleId(fileHandleId)
    setStep(CsvUploadPreviewStep.COLUMN_PREVIEW)
  }, [])

  return {
    step,
    setStep,
    csvTableDescriptor,
    setCsvTableDescriptor,
    uploadedFileHandleId,
    csvPreviewData,
    setCsvPreviewData,
    isLoadingPreview,
    setIsLoadingPreview,
    onFileUploaded,
  }
}
```

You would add a component that handles rendering the correct upload/preview UI

```tsx
import {
  BasicFileHandleUpload,
  FileUploadHandle,
} from '@/components/file/upload/BasicFileHandleUpload'
import CsvPreviewWithOptions from '@/components/table/CsvPreview/CsvPreviewWithOptions'
import useCsvUploadPreview, {
  CsvUploadPreviewStep,
} from '@/components/table/CsvPreview/useCsvUploadPreview'
import { useRef } from 'react'

export type CsvUploadPreviewContentProps = {
  workflow: ReturnType<typeof useCsvUploadPreview>
}

export default function CsvUploadPreviewContent({
  workflow,
}: CsvUploadPreviewContentProps) {
  const {
    step,
    uploadedFileHandleId,
    csvTableDescriptor,
    setCsvTableDescriptor,
    setCsvPreviewData,
    setIsLoadingPreview,
    onFileUploaded,
  } = workflow

  const uploadRef = useRef<FileUploadHandle | null>(null)

  if (step === CsvUploadPreviewStep.UPLOAD_CSV) {
    return (
      <BasicFileHandleUpload
        ref={uploadRef}
        allowMultipleUpload={false}
        onFileUploadComplete={onFileUploaded}
        disableDragAndDrop={true}
      />
    )
  }

  return (
    <CsvPreviewWithOptions
      fileHandleId={uploadedFileHandleId}
      csvTableDescriptor={csvTableDescriptor}
      onCsvTableDescriptorChange={setCsvTableDescriptor}
      onCsvPreviewDataChange={setCsvPreviewData}
      onIsLoadingChange={setIsLoadingPreview}
    />
  )
}
```

And a caller uses both the hook and the component to render the result.

```tsx
import { DialogBase } from '@/components/DialogBase'
import CsvUploadPreviewContent from './CsvUploadPreviewContent'
import useCsvUploadPreview, {
  CsvUploadPreviewStep,
} from './useCsvUploadPreview'
import Button from '@mui/material/Button'
import { useCallback, useState } from 'react'

enum CreateTableFromCsvStep {
  CSV_UPLOAD_PREVIEW = 0,
  TABLE_NAME = 1,
}

export default function CreateTableFromCsvDialog(props: {
  open: boolean
  onClose: () => void
  parentId: string
  onSuccess?: () => void
}) {
  const { open, onClose, parentId, onSuccess } = props
  const csvUploadWorkflow = useCsvUploadPreview()
  const [step, setStep] = useState(CreateTableFromCsvStep.CSV_UPLOAD_PREVIEW)
  const [tableName, setTableName] = useState('')

  const handlePreviewConfirm = useCallback(() => {
    const suggestedColumns = csvUploadWorkflow.csvPreviewData?.suggestedColumns
    if (!suggestedColumns) return

    // initialize schema editor, etc.
    setStep(CreateTableFromCsvStep.TABLE_NAME)
  }, [csvUploadWorkflow.csvPreviewData])

  return (
    <DialogBase
      open={open}
      onCancel={onClose}
      title="Upload CSV"
      content={
        <>
          {step === CreateTableFromCsvStep.CSV_UPLOAD_PREVIEW && (
            <CsvUploadPreviewContent workflow={csvUploadWorkflow} />
          )}

          {step === CreateTableFromCsvStep.TABLE_NAME && (
            <>
              {/* create-specific UI: table name + schema form */}
            </>
          )}
        </>
      }
      actions={
        <>
          <Button onClick={onClose}>Cancel</Button>

          {step === CreateTableFromCsvStep.CSV_UPLOAD_PREVIEW &&
            csvUploadWorkflow.step === CsvUploadPreviewStep.COLUMN_PREVIEW && (
              <Button
                variant="contained"
                disabled={csvUploadWorkflow.isLoadingPreview}
                onClick={handlePreviewConfirm}
              >
                Next
              </Button>
            )}

          {step === CreateTableFromCsvStep.TABLE_NAME && (
            <Button variant="contained" disabled={!tableName}>
              Create
            </Button>
          )}
        </>
      }
    />
  )
}
```

packages/synapse-react-client/src/components/table/CsvPreview/UpdateTableWithCsvDialog.tsx
kianamcc: Yes, omitting this part for update since the schema already exists, onCsvPreviewDataChange needed for create flow for suggestedColumns:
Screenshot 2026-06-01 at 10 17 10 AM

packages/synapse-react-client/src/components/table/CsvPreview/UpdateTableWithCsvDialog.tsx
kianamcc: Will now live in UploadPreviewContent where BasicFileHandleUpload exists.

packages/synapse-react-client/src/components/table/CsvPreview/CsvPreviewWithOptions.tsx
kianamcc: I think this was from earlier refactor direction? We now use CsvPreviewWithOptions from CsvUploadPreviewContent.

nickgros: Yep, I wrote this before arriving on the `CsvUploadPreviewContent` component

packages/synapse-react-client/src/components/table/CsvPreview/useUploadCsvToExistingTable.ts
```diff
@@ -0,0 +1,56 @@
...
+  options?: Omit<
+    UseMutationOptions<
+      TableUpdateTransactionResponse,
+      SynapseClientError,
+      UploadCsvToExistingTableProps
+    >,
+    'mutationFn'
+  >,
+) {
+  const { mutateAsync: tableTransaction } = useTableUpdateTransaction()
```
nickgros: Also update variable name here

```suggestion
  const { mutateAsync: updateTableInTransaction } = useTableUpdateTransaction()
```

packages/synapse-react-client/src/components/table/CsvPreview/UpdateTableWithCsvDialog.tsx
```diff
@@ -0,0 +1,84 @@
...
+  open: boolean
+  /** Callback when the dialog is closed */
+  onClose: () => void
+  /** If provided, upload csv rows to existing table */
+  tableId: string
+  /** Invoked after all backend operations successfully complete */
+  onSuccess?: () => void
+}
+
+export default function UpdateTableWithCsvDialog(
```
nickgros, line 28: Add a comment describing this component

packages/synapse-react-client/src/components/table/CsvPreview/CsvPreviewDialog.tsx
```diff
@@ -87,42 +75,13 @@ export default function CsvPreviewDialog(props: CsvPreviewDialogProps) {
...
-                )
-              }
-            }}
-          >
-            Refresh Preview
-          </Button>
-        </AccordionDetails>
-      </Accordion>
-    </Stack>
+    <CsvPreviewWithOptions
```
nickgros: Can we replace this & the uploader with `CsvUploadPreviewContent` just like in the new components?

packages/synapse-react-client/src/components/table/CsvPreview/CreateTableFromCsvDialog.tsx
nickgros: Still want one here!

kianamcc: I thought I put one here! Sorry about that.

nickgros (approved):
Nice work, this looks great!

PR #2748, opened 2026-04-30
PORTALS-4202

jay-hodgson:


packages/synapse-react-client/src/components/HeaderCard.tsx
```diff
@@ -103,48 +114,94 @@ const HeaderCardClassic = forwardRef(function HeaderCardClassic(
                   }),
                 }}
               >
-                <div className="SRC-cardContent" style={{ marginLeft: '15px' }}>
```
jay-hodgson, line 106: Wrapped in flex row

packages/synapse-react-client/src/components/HeaderCard.tsx
```diff
@@ -103,48 +114,94 @@ const HeaderCardClassic = forwardRef(function HeaderCardClassic(
...
+                  >
+                    <Box
                       sx={{
-                        background: 'rgba(0, 0, 0, 0.10)',
-                        marginTop: '30px',
+                        display: 'flex',
+                        flexDirection: 'row',
+                        alignItems: 'center',
+                        gap: '10px',
+                        color: 'var(--synapse-gray-300)',
```
jay-hodgson: Special styles to make text visible in the header (also targeted MuiChip text, since this tends to be what is used in this area)

packages/synapse-react-client/src/components/HeaderCard.tsx
```diff
@@ -103,48 +114,94 @@ const HeaderCardClassic = forwardRef(function HeaderCardClassic(
...
+                          color: 'var(--synapse-gray-300)',
+                        },
                       }}
+                    >
+                      <div className="SRC-type">{type}</div>
+                      {cardTypeAdornment}
+                    </Box>
+                    <Box
+                      sx={{
+                        '& .icon-list.themed path, & .icon-list.themed circle':
```
jay-hodgson, line 137: Special color overrides for icons rendered in the header

jay-hodgson:
Started up CCKP, NF, and ADKP and verified explore card rendering and detail page rendering looked ok with these changes.  Here are a few examples:

Screenshot 2026-04-30 at 10 49 06 AM

Screenshot 2026-04-30 at 1 45 13 PM



Screenshot 2026-04-30 at 10 57 54 AM

packages/synapse-react-client/src/style/components/_cards.scss
```diff
@@ -153,6 +153,10 @@
     text-transform: uppercase;
   }
 
+  .SRC-type {
+    margin-right: calc(var(--baseline-grid) * 0.5);
```
jay-hodgson, line 157: Add a bit more space to the right of the card type (left of the adornment)

packages/synapse-react-client/src/components/HeaderCard.tsx
jay-hodgson: Will remove

packages/synapse-react-client/src/components/HeaderCard.tsx
nickgros: Is it a lot of work to fix these icons so they use `currentColor`? Or is this a special 'header card' theme issue?

nickgros (approved):

packages/synapse-react-client/src/components/HeaderCard.tsx
jay-hodgson: It's a good idea, but it is 40+ icon svgs to update, and we may want to spin up CCKP, Challenges, and NF to verify each icon output in the context of each detail page

PR #2750, opened 2026-04-30
SWC-7621

jay-hodgson:


packages/synapse-react-client/src/utils/hooks/useLogin.ts
```diff
@@ -200,7 +200,12 @@ export default function useLogin(opts: UseLoginOptions): UseLoginReturn {
       mutationFn: SynapseClient.loginWith2fa,
       onError: e => {
         setErrorMessage(e.reason)
-        if (
+        if (e.reason.includes('The provided code is invalid')) {
+          // The most common cause of an invalid TOTP code is clock skew between the user's device and the server.
+          setErrorMessage(
+            `${e.reason} Please ensure automatic date and time is enabled in your device settings and try again.`,
+          )
```
nickgros: The same issue could arise when setting up 2fa. Should we add this logic there as well?

nickgros (approved):

packages/synapse-react-client/src/components/Authentication/Constants.ts
```diff
@@ -3,3 +3,6 @@ export const TOTP_GUIDANCE_TEXT =
 
 export const RECOVERY_CODE_GUIDANCE_TEXT_SHORT =
   'Enter a one-time backup code. Your backup code is a 16 digit code, with groups of 4 letters or numbers separated by hyphens.'
+
+export const TOTP_CLOCK_SKEW_ERROR_APPENDAGE =
```
jay-hodgson, line 7: "Appendage" is weird, but I just accepted it

nickgros (approved):

PR #2753, opened 2026-05-01
PORTALS-4203 : follow up and set up this conditional nav bar item for many other portals

jay-hodgson:


kianamcc (approved):
Glad the component could be of use for other portals!

PR #2755, opened 2026-05-01
SWC-7791 fix null paste in array columns

afwillia:


packages/synapse-react-client/src/components/DataGrid/columns/AutocompleteColumn.tsx
```diff
@@ -67,6 +67,25 @@ export function castCellValueToString(toCast: any): string {
...
+// Free-text typed/pasted into the cell parses to "" for empty input. Committing
+// "" leaves the cell as an empty string, which fails enum validation with a
+// terse "is not a valid enum value" message. Substitute the column's clearValue
+// (null for required, undefined for optional) so the commit path matches the
+// X-button clear path and the validator surfaces the substantive message.
+function commitParsedValue(
+  parsed: unknown,
+  clearValue: undefined | null,
+): unknown {
+  if (
```
afwillia, line 79: The same nullish check logic as in `coerceModelCellValue`, but duplicating it here avoids needing to pass the schema to this component.

packages/synapse-react-client/src/components/DataGrid/utils/columnFactory.tsx
nickgros: Let's introduce a new util like

```ts
function getEmptyValue(isRequired: boolean) {
	return isRequired ? null : undefined
}
```

And use it in every place where we need the empty value (here and in `schemaAwarePasteValue`, maybe other places too).

packages/synapse-react-client/src/components/DataGrid/utils/schemaAwarePasteValue.ts
```diff
@@ -0,0 +1,70 @@
...
+): unknown {
+  if (!schemaPropertyInfo) {
+    return value
+  }
+  if (
+    value === null ||
+    value === undefined ||
+    (typeof value === 'string' && value.trim() === '')
+  ) {
+    return schemaPropertyInfo.isRequired ? null : undefined
```
nickgros:
```suggestion
    return getEmptyValue(schemaPropertyInfo.isRequired)
```

packages/synapse-react-client/src/components/DataGrid/columns/AutocompleteColumn.tsx
nickgros: We already have `clearValue` available so this makes sense

packages/synapse-react-client/src/components/DataGrid/columns/AutocompleteColumn.test.tsx
```diff
@@ -132,6 +132,107 @@ describe('AutocompleteColumn', () => {
...
+        />,
+      )
+    })
+
+    await waitFor(() => {
+      expect(mockSetRowData).toHaveBeenCalledWith(undefined)
+    })
+  })
+
+  it('commits clearValue via the createOption path when the parsed result coerces to clearValue', async () => {
```
nickgros, line 208: this is kind of a weird case, but i guess the weird behavior is really owned by `parseFreeTextGivenJsonSchemaType`. So I think this is fine

packages/synapse-react-client/src/components/DataGrid/utils/applyModelChange.ts
```diff
@@ -59,7 +60,14 @@ export function applyModelChange(
...
         s.con(
-          getDefaultValueForProperty(change.rowData, name, schemaPropertyInfo),
+          coerceModelCellValue(
+            getDefaultValueForProperty(
+              change.rowData,
+              name,
+              schemaPropertyInfo,
+            ),
+            schemaPropertyInfo[name],
+          ),
```
nickgros: When does `getDefaultValueForProperty` return something that doesn't match the model? It already has the schema, so it shouldn't ever return something invalid.

At the very least, probably call `coerceModelCellValue` inside `getDefaultValueForProperty` (if you even need to call it).

If there are cases where `getDefaultValueForProperty` is returning something incorrect, add test(s) to cover those cases.

packages/synapse-react-client/src/components/DataGrid/columns/AutocompleteColumn.test.tsx
```diff
@@ -132,6 +132,107 @@ describe('AutocompleteColumn', () => {
     })
   })
 
+  it('commits null (clearValue) when the user clears a required cell and deactivates', async () => {
```
nickgros: These two tests don't consider the schema, but I get that you want to cover both cases just in case. Maybe convert to a parameterized test?


```suggestion
  it.each([null, undefined])('commits clearValue (%s) when the user clears a required cell and deactivates', async (clearValue) => {
```

nickgros (approved):

PR #2760, opened 2026-05-04
[Challenge Portal] Add "Back to Top" button on details page

vpchung:
On a Challenge Details page, navigation between the tabs currently require manual scrolling, which can get pretty tedious for the user when the tab content is lengthy, e.g. https://challenges.synapse.org/Challenges/DetailsPage/Task1?id=syn74274097

This PR introduces a "Back to Top" button to allow users to quickly return to the tabs.

## Changelog

* Add a "Back to Top" button to the side navbar
* Only show button when user has scrolled 500px
   * Otherwise, button will be hidden

## Preview

https://github.com/user-attachments/assets/862fa224-23d8-4132-98e4-86274b4ed87f

### DIFF
--- a/apps/synapse-portal-framework/src/components/DetailsPageMenu.tsx
+++ b/apps/synapse-portal-framework/src/components/DetailsPageMenu.tsx
@@ -1,4 +1,4 @@
-import React from 'react'
+import React, { useEffect, useState } from 'react'
 import { DetailsPageSectionLayoutType } from '@/components/DetailsPage/DetailsPageSectionLayout'
 import { scrollToWithOffset } from '@/utils'
 import { Button, Stack, Typography } from '@mui/material'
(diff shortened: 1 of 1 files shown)


apps/synapse-portal-framework/src/components/DetailsPageMenu.tsx
```diff
@@ -11,6 +11,15 @@ const DetailsPageMenu = ({
 }): React.ReactNode => {
   const location = useLocation()
   const navigate = useNavigate()
+  const [showBackToTop, setShowBackToTop] = useState(false)
+
+  useEffect(() => {
+    const onScroll = () => {
+      setShowBackToTop(window.scrollY > 500)
```
jay-hodgson: Performance issue.
The scroll handler fires on every single scroll event (potentially hundreds per second). While React's useState setter does bail out if the boolean value doesn't change (no re-render when already past 500px), the handler function itself still executes on every event. Wrapping the handler in requestAnimationFrame or a simple throttle is worth doing:
```
const onScroll = () => {
  requestAnimationFrame(() => {
    setShowBackToTop(window.scrollY > 500)
  })
}
```

jay-hodgson:
I asked Adam H to also review the UX, and it has his stamp of approval.  So I have no issues with introducing this UX to all portal detail pages (thank you for proposing this, @vpchung !)

apps/synapse-portal-framework/src/components/DetailsPageMenu.tsx
vpchung: Thank you, @jay-hodgson ! I have added `requestAnimationFrame` per your recommendation.

PR #2769, opened 2026-05-08
PORTALS-4205 : Mint/resolve DOIs for NF Study resources

jay-hodgson:
Mirrors setting up ELITE portal for DOIs (PR here: https://github.com/Sage-Bionetworks/synapse-web-monorepo/pull/2408 ,
ticket https://sagebionetworks.jira.com/browse/PORTALS-3690 ).  

Tested minting a Study resource DOI on NF, and resolution.

### DIFF
--- a/apps/portals/nf/src/pages/DoiRedirect.tsx
+++ b/apps/portals/nf/src/pages/DoiRedirect.tsx
@@ -1,29 +1,11 @@
-import { Navigate, useSearchParams } from 'react-router'
+import { DoiRedirectComponent } from '@sage-bionetworks/synapse-portal-framework/shared-config/DoiRedirectRoute'
 import { doiRedirector, doiSerializer } from '@/config/doiRedirector'
 
-const OBJECT_ID_QUERY_PARAM = 'objectId'
-
 export default function DoiRedirect() {
-  const [searchParams] = useSearchParams()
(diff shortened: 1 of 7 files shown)


apps/portals/nf/src/pages/DoiRedirect.tsx
```diff
@@ -0,0 +1,29 @@
+import { Navigate, useSearchParams } from 'react-router'
+import { doiRedirector, doiSerializer } from '@/config/doiRedirector'
+
+const OBJECT_ID_QUERY_PARAM = 'objectId'
+
+export default function DoiRedirect() {
```
kianamcc, line 4: There is similar code to this in DoiRedirectRoute.tsx in the synapse-portal-framework. Is it possible to export DoiRedirectComponent and use that here?

jay-hodgson: Nice catch!  We should be reusing DoiRedirectRoute in nf

jay-hodgson: Because of SSG, updated DoiRedirect to be a thin no-parameter wrapper around DoiRedirectComponent

kianamcc (approved):

PR #2772, opened 2026-05-11
SWC-7815 - curator agent prompts

afwillia:
Add a few pill prompts to the grid chat window.
</br>
On open/chat start
ss1

</br>
after starting a chat, prompts disappear
ss

### DIFF
--- a/packages/synapse-react-client/src/components/SynapseChat/GridAgentChat.tsx
+++ b/packages/synapse-react-client/src/components/SynapseChat/GridAgentChat.tsx
@@ -5,6 +5,12 @@ import { useState } from 'react'
 import DraggableDialog from '../DraggableDialog/DraggableDialog'
 import { SynapseChat } from './index'
 
+const suggestedPrompts = [
+  'Help me fill this out',
+  'Help me understand this',
+  'Find missing fields',
+]
+
 export type GridAgentChatProps = {
   gridSessionId: string
   usersReplicaId: number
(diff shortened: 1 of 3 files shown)


packages/synapse-react-client/src/components/SynapseChat/SynapseChat.tsx
```diff
@@ -277,6 +282,24 @@ export function SynapseChat({
           backgroundColor: 'white',
         }}
       >
+        {suggestedPrompts && suggestedPrompts.length > 0 && (
```
jay-hodgson: It was my understanding that these initial hardcoded suggestion prompts would sit above all of the chat interactions (would be the first thing that the user sees when the interactions are empty, but would not always be the suggested next prompts in the conversation).

afwillia: Right - I'll modify the prompts so they disappear after a chat begins. But the current start state matches expectations?

packages/synapse-react-client/src/components/SynapseChat/GridAgentChat.tsx
```diff
@@ -37,6 +37,12 @@ export function GridAgentChat({
     usersReplicaId,
   }
 
+  const suggestedPrompts = [
```
jay-hodgson: nit : these are recreated on every render, could be refactored

jay-hodgson (approved):

PR #2777, opened 2026-05-12
PORTALS-4119 - add github workflow precommit linter

afwillia:
use actionlint to lint github workflow/action yaml files. Add script to install actionlint when running pnpm prepare. Add lint:actions script.
ss

### DIFF
--- a/scripts/install-actionlint.sh
+++ b/scripts/install-actionlint.sh
@@ -1,0 +1,27 @@
+#!/usr/bin/env bash
+# Downloads the actionlint binary to ./bin/actionlint for the pre-commit hook.
+# Pinned to a specific version for reproducibility.
+set -euo pipefail
+
+ACTIONLINT_VERSION="1.7.12"
+REPO_ROOT="$(cd "$(dirname "$0")/.." && pwd)"
+BIN_DIR="$REPO_ROOT/bin"
+BIN="$BIN_DIR/actionlint"
+
+if [ -x "$BIN" ] && [ "$("$BIN" -version 2>/dev/null | head -n1)" = "$ACTIONLINT_VERSION" ]; then
+  exit 0
+fi
+
+mkdir -p "$BIN_DIR"
+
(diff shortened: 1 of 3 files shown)


scripts/install-actionlint.sh
```diff
@@ -0,0 +1,27 @@
+#!/usr/bin/env bash
```
afwillia, line 1: actionlint needs to be installed. Alternatives are homebrew or a Node.js script instead of bash for Windows support.

zaro0508:
Instead of setting up individual linters I would recommend using the [pre-commit framework](https://pre-commit.com/) and adding the [check-jsonschema's check-github-actions](https://github.com/python-jsonschema/check-jsonschema) linter similar to how it's done in the [agora-infra-v3 repo](https://github.com/Sage-Bionetworks-IT/agora-infra-v3/blob/dev/.pre-commit-config.yaml#L45-L49).  This will allow you to run all your linter jobs locally with a git commit hook.  You can run all linters in GH actions by just running the pre-commit command.

zaro0508:
hmm, it looks like actionslint is also supported as a pre-commit linter, https://github.com/rhysd/actionlint/blob/v1.7.12/docs/usage.md#pre-commit

afwillia:
Thanks @zaro0508. Looks like a bit of work to replace our husky + lint-staged pipeline and add a python dependency. But pre-commit includes some additional linters and we could easily add a secret scanner.

scripts/install-actionlint.sh
nickgros: Looks like they publish a docker image, could we use that instead of installing locally?

https://hub.docker.com/r/rhysd/actionlint

sagely1:
As @zaro0508 mentioned, you could use our Agora as an example.  It does take a bit more effort to add initially.  Also ensure that if you decide on this route to consider security as well.  You can use: https://github.com/Sage-Bionetworks-IT/agora-infra-v3/blob/dev/.github/dependabot.yml as a guide.

scripts/install-actionlint.sh
nickgros: https://github.com/rhysd/actionlint/blob/main/docs/usage.md#docker

PR #2778, opened 2026-05-12
PORTALS-4224

jay-hodgson:


packages/synapse-react-client/src/utils/functions/calculateFriendlyFileSize.ts
```diff
@@ -12,16 +12,18 @@ const suffixes: string[] = [
...
-  fractionDigits: number = 2,
+  fractionDigits?: number,
 ) {
   if (bytes == null) {
     return ''
   }
   // https://stackoverflow.com/questions/10420352/converting-file-size-in-bytes-to-human-readable-string
   const i = Math.floor(Math.log(bytes) / Math.log(1024))
+  // 0 for Bytes/KB, 1 for MB, 2 for GB, 3 for TB etc
+  const digits = fractionDigits ?? Math.max(0, i - 1)
```
kianamcc, line 23: I wonder if this should be capped at 3 decimal places 🤔 Not sure how often we even do surpass a TB.

kianamcc (approved):

packages/synapse-react-client/src/utils/functions/calculateFriendlyFileSize.ts
jay-hodgson: Yes, you're probably right.  Hopefully we'll get a bug about too many decimal places when downloading PB+ size files from Synapse!

PR #2779, opened 2026-05-13
SWC-7718 - fix SynapsePlot test

afwillia:
import 'plotly.js-basic-dist' to avoid lazy load and set timoeout to 10s instead of default 1s to ensure SynapsePlots test has time to complete


jay-hodgson:
Thanks @afwillia !

jay-hodgson (approved):

PR #2780, opened 2026-05-13
PORTALS-4204

jay-hodgson:
Summary:

### New: `AdaptiveTwoColumnLayout` component (SRC)

A new component (`AdaptiveTwoColumnLayout.tsx`) has been added. It renders two
columns side-by-side, measures their heights after the initial render, and
switches to a stacked single-column layout if the left content is too short
relative to the right (controlled by a `stackRatioThreshold` prop, default
`0.45`). Storybook stories (`AdaptiveTwoColumnLayout.stories.tsx`) are also
included.

---

### Refactored: `HeaderCard` (`HeaderCardClassic`) (SRC)

`HeaderCard.tsx` has been substantially restructured:

- Layout migrated from Bootstrap grid (`container-fluid`, `col-md-offset-1`,
  etc.) to MUI `Container maxWidth="lg"` with explicit `Box` rows.
- The card now breaks out of constrained parents using
  `width: 100vw` / `marginLeft: calc(50% - 50vw)`.
- Content is split into logical rows: type/adornment/buttons → title →
  icon list → adaptive two-column section (description left, metadata/values
  right) using the new `AdaptiveTwoColumnLayout`.
- `showFullDescriptionByDefault` now defaults to `false` (was `true`).
- The icon is no longer hidden when a `sustainabilityScorecard` is present.

---

### Updated: `TableRowGenericCard` (SRC)

For header cards with no image or explicit icon value, the icon is now rendered
as `undefined` (suppressed) rather than rendering a `GenericCardIcon` with no
content.

---

### Refactored: `ChallengeDetailPageWrapper` (portal framework)

Same Bootstrap-to-MUI migration — replaced `div.container-fluid` / `div.row` /
`div.col-md-offset-1` with `Container maxWidth="lg"` and `Box`. The negative
`marginTop` offset was adjusted from `-150px` to `-130px`.

---

### Style tweaks

- `_DetailsPage.scss`: `.SRC-portalCardHeader` `padding-bottom` increased from
  `45px` → `65px`.
- `_cards.scss`:
  - Added `font-size: 24px` for `h3` inside header cards.
  - Added width constraints for thumbnail images (`max-width: 110px`).
  - Fixed a margin issue for `> table` inside card metadata.

### Example output from various portals
Screenshot 2026-05-13 at 9 13 29 AM
Screenshot 2026-05-13 at 9 01 31 AM
Screenshot 2026-05-12 at 5 31 02 PM
Screenshot 2026-05-12 at 5 08 14 PM
Screenshot 2026-05-12 at 5 00 43 PM
Screenshot 2026-05-12 at 4 30 50 PM
Screenshot 2026-05-12 at 4 30 12 PM
Screenshot 2026-05-12 at 4 30 27 PM


packages/synapse-react-client/src/components/AdaptiveTwoColumnLayout/AdaptiveTwoColumnLayout.tsx
```diff
@@ -0,0 +1,134 @@
...
+export function AdaptiveTwoColumnLayout({
+  leftContent,
+  rightContent,
+  stackRatioThreshold = 0.45,
+  gap = '16px',
+  sx,
+  onMeasure,
+}: AdaptiveTwoColumnLayoutProps) {
+  // null = not yet measured (container is hidden while side-by-side heights are read)
+  const [isStacked, setIsStacked] = useState<boolean | null>(null)
```
jay-hodgson, line 54: Notes:  Intentional no re-measurement on resize.  Calculate the ratio once only (so doing things like clicking Show More does not cause the layout to jump).

packages/synapse-react-client/src/components/HeaderCard.tsx
```diff
@@ -89,141 +89,156 @@ const HeaderCardClassic = forwardRef(function HeaderCardClassic(
...
+          >
+            {renderedIconList}
+          </Box>
+        )}
+
+        {/* Row 4: left content, optionally alongside right content in an adaptive two-column layout */}
+        {(() => {
+          const leftContent = (
+            <>
+              {/* Column 1: icon (floated left) with subTitle and description flowing around */}
```
jay-hodgson, line 162: Note - Intentional float to allow text to flow around the icon

packages/synapse-react-client/src/components/HeaderCard.tsx
```diff
@@ -69,11 +70,10 @@ const HeaderCardClassic = forwardRef(function HeaderCardClassic(
     sx,
   } = props
 
-  const hideIcon = Boolean(sustainabilityScorecard)
   const descriptionConfiguration: DescriptionConfig = {
     ...descriptionConfig,
     showFullDescriptionByDefault:
-      descriptionConfig?.showFullDescriptionByDefault ?? true,
+      descriptionConfig?.showFullDescriptionByDefault ?? false,
```
jay-hodgson, line 76: Design requested this showFullDescriptionByDefault default value be changed for this updated implementation

packages/synapse-react-client/src/components/HeaderCard.tsx
```diff
@@ -69,11 +70,10 @@ const HeaderCardClassic = forwardRef(function HeaderCardClassic(
     sx,
   } = props
 
-  const hideIcon = Boolean(sustainabilityScorecard)
```
jay-hodgson, line 72: New design supports showing the icon and sustainability scorecard.  No need to dynamically hide.

packages/synapse-react-client/src/style/components/_cards.scss
```diff
@@ -52,6 +52,9 @@
   &.SRC-portalCardHeader {
     color: white;
     margin-top: 0px;
+    h3 {
+      font-size: 24px !important;
```
jay-hodgson, line 56: Needed the `!important`, unfortunately, to compete with another old `!important` selector

packages/synapse-react-client/src/components/AdaptiveTwoColumnLayout/AdaptiveTwoColumnLayout.tsx
```diff
@@ -0,0 +1,134 @@
...
+   * Measured once after initial render with both columns side-by-side.
+   * Default: 0.45
+   */
+  stackRatioThreshold?: number
+  gap?: string | number
+  sx?: SxProps
+  /**
+   * Called once after the initial measurement completes. Useful for debugging.
+   */
+  onMeasure?: (info: MeasurementInfo) => void
```
kianamcc, line 25: Is the storybook debug sufficient or should we additionally have some tests for this component? Test to ensure onMeasure is fired once, clicking a "Show more" button does not trigger re-measurement

packages/synapse-react-client/src/components/AdaptiveTwoColumnLayout/AdaptiveTwoColumnLayout.tsx
```diff
@@ -0,0 +1,134 @@
...
+  const leftRef = useRef<HTMLDivElement>(null)
+  const rightRef = useRef<HTMLDivElement>(null)
+  // Stable ref so onMeasure is never a useEffect dependency.
+  const onMeasureRef = useRef(onMeasure)
+  onMeasureRef.current = onMeasure
+
+  useEffect(() => {
+    let cancelled = false
+
+    // Two rAFs: the first waits for the browser to commit the layout, the
```
kianamcc: Had to look this one up! Is it better to use rAFs rather than useLayoutEffect in this case? Or can we get the same results using useLayoutEffect and get rid of the visibility: hidden?

packages/synapse-react-client/src/components/AdaptiveTwoColumnLayout/AdaptiveTwoColumnLayout.tsx
```diff
@@ -0,0 +1,134 @@
...
+  gap = '16px',
+  sx,
+  onMeasure,
+}: AdaptiveTwoColumnLayoutProps) {
+  // null = not yet measured (container is hidden while side-by-side heights are read)
+  const [isStacked, setIsStacked] = useState<boolean | null>(null)
+  const leftRef = useRef<HTMLDivElement>(null)
+  const rightRef = useRef<HTMLDivElement>(null)
+  // Stable ref so onMeasure is never a useEffect dependency.
+  const onMeasureRef = useRef(onMeasure)
```
kianamcc, line 58: Should this be in a useEffect to avoid writing to the ref during render? useEffect(() => { onMeasureRef.current = onMeasure })

packages/synapse-react-client/src/components/AdaptiveTwoColumnLayout/AdaptiveTwoColumnLayout.tsx
```diff
@@ -0,0 +1,134 @@
...
+ * Flexbox's default align-items:stretch causes both children to report the same
+ * offsetHeight (the taller column's height), making the ratio always 1.0 and
+ * stacking never triggering. Grid cells with align-items:start only take the
+ * height of their own content, giving accurate independent measurements.
+ */
+export function AdaptiveTwoColumnLayout({
+  leftContent,
+  rightContent,
+  stackRatioThreshold = 0.45,
+  gap = '16px',
```
kianamcc: I noticed the gap between the columns is quite small. I think Adam wants 40px here: https://www.figma.com/design/C7wSn2ayPJBJgdMXdKbWQY/CCKP-Knowledge-Portal-v2?node-id=2331-4596&m=dev

kianamcc:
Since it's in the same area, do you mind changing the endIcon for the share button to startIcon?: https://www.figma.com/design/C7wSn2ayPJBJgdMXdKbWQY/CCKP-Knowledge-Portal-v2?node-id=2331-4581&m=dev

jay-hodgson:
> Since it's in the same area, do you mind changing the endIcon for the share button to startIcon?: https://www.figma.com/design/C7wSn2ayPJBJgdMXdKbWQY/CCKP-Knowledge-Portal-v2?node-id=2331-4581&m=dev

Screenshot 2026-05-13 at 2 03 19 PM

packages/synapse-react-client/src/components/AdaptiveTwoColumnLayout/AdaptiveTwoColumnLayout.tsx
jay-hodgson: Good question.  Sonnet is suggesting that this is the correct approach, but let me know if you disagree!

> The concern about writing to refs during render applies when the ref is read during render (which would make it a side effect that could interfere with concurrent rendering). Here the ref is only ever read inside [useLayoutEffect](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html) (i.e., after commit), so the inline update is safe. React's own docs endorse this exact pattern as the standard "stable callback ref" approach before useEffectEvent becomes stable.

packages/synapse-react-client/src/components/AdaptiveTwoColumnLayout/AdaptiveTwoColumnLayout.tsx
jay-hodgson: Nice catch.  Updated so it still has no flash.  Removed `visibility: hidden` and double rAF

jay-hodgson:
Gap updated:
Screenshot 2026-05-13 at 3 38 59 PM

And fixed a bug, around the max width of the sustainability scorecard when stacked:
Screenshot 2026-05-13 at 3 38 36 PM

packages/synapse-react-client/src/components/AdaptiveTwoColumnLayout/AdaptiveTwoColumnLayout.tsx
jay-hodgson: Added unit tests

kianamcc (approved):

PR #2782, opened 2026-05-13
SWC-7057 Move local sharing settings button

afwillia:
move local sharing settings button immediately below inheritance message. Modify InheritanceMessage text.
ss


jay-hodgson (approved):
Requested from Stockard, nice!

PR #2787, opened 2026-05-14
PORTALS-4232: hook up the NF Tools search and explore pages to the new SearchQueryWrapper

jay-hodgson:
### NF Portal — Explore pages migrated to `SearchQueryWrapperPlotNav`

The six main "Explore" pages in the NF portal have been switched from
`QueryWrapperPlotNav` to `SearchQueryWrapperPlotNav`, wiring each page to a
dedicated Synapse full-text search index:

| Page | Search Index ID |
|---|---|
| Datasets | `syn75081630` |
| Publications | `syn75081631` |
| Studies | `syn75081633` |
| Initiatives | `syn75081635` |
| Tools | `syn75081636` |
| Hackathons | `syn75081639` |

For each entity type, a new `*Search` config export (e.g. `datasetsSearch`,
`toolsSearch`) was added alongside the existing `QueryWrapperPlotNavProps`
config. The search config includes `searchIndexId`, `autocompleteFieldName`,
and card/column-alias settings. The portal search page (`searchConfig.tsx`)
was updated to use the new search variants.

A new `organizationsSearch` config was also added for funders
(`syn75081638`), and search index IDs for all NF entity types were
centralized in `resources.ts` as named constants.

---

### SRC — `FTSConfig.minSearchQueryLength` (new option)

A new optional `minSearchQueryLength` field was added to `FTSConfig`
(`SearchV2.tsx`). When set, it overrides the default minimum of 3 characters
required before a search executes. `SearchQueryWrapperPlotNav` now sets this
to `1` so single-character autocomplete queries work. A new test was added
to `FullTextSearch.test.tsx` verifying the override is respected.

---

### SRC — Facet ordering fix in `searchQueryResultsToQueryResultBundle`

Facets returned from the search API are now reordered to match the
`columnModels` order before being forwarded downstream. This ensures
`FacetFilterControls` renders facets in the expected column order. Facets
with no matching `columnModel` are appended at the end. Three new unit tests
cover reordering and the no-match fallback.

---

### SRC — `entityId` fix in `SearchQueryWrapper`

`entityId` in the `QueryContextType` built by `SearchQueryWrapper` was
previously hardcoded to `undefined`; it now correctly passes `synapseId`.

---

### b2ai Standards portal — minor cleanup

The hardcoded `'syn74909093'` search index ID in `standards.ts` was
extracted to a named constant `standardsSearchIndexId` in `resources.ts`.

---

### Miscellaneous

A commented-out endpoint override block was added to `nf/src/entry.client.tsx`
to make it easier to test against the staging environment.


apps/portals/nf/src/config/synapseConfigs/tools.tsx
```diff
@@ -75,4 +76,16 @@ const tools: QueryWrapperPlotNavProps = {
   },
 }
 
+export const toolsSearch: SearchQueryWrapperPlotNavProps = {
+  rgbIndex,
+  name: 'Tools',
+  shouldDeepLink: false,
+  cardConfiguration: toolsCardConfiguration,
+  columnAliases,
+  searchIndexId: 'syn74937101', //TODO: update to the production Synapse ID once created
```
jay-hodgson: Note - NF e2e tests expected to fail until production Synapse ID is used

nickgros (approved):

PR #2788, opened 2026-05-14
PORTALS-4232: Support deep linking in SearchQueryWrapper

jay-hodgson:


jay-hodgson:
Verified in the Standards Registry that the Query is properly pushed to the URL after this change, and if that URL is loaded in a new page then the expected Search is executed (with the selected filters)

hallieswan (approved):

PR #2802, opened 2026-05-17
PORTALS-4189 - Misc SSG fixes/improvements

nickgros:
- Refactor ExploreWrapperTabs
  - Use React Router Link component for tabs
  - Refactor styles to use SCSS
- Provide opt-out for detail page prerender. CCKP has >4000 publications, which would create extremely long build times.
- Update useSetCanonicalUrl to prevent issue where a DOM element managed by another component was deleted, resulting in obscure React errors.

### DIFF
--- a/apps/synapse-portal-framework/src/components/Explore/ExploreWrapperTabs.module.scss
+++ b/apps/synapse-portal-framework/src/components/Explore/ExploreWrapperTabs.module.scss
@@ -1,0 +1,25 @@
+:global(.MuiTab-root.MuiButtonBase-root).tab {
+  transition: all 400ms;
+  font-size: 16px;
+  font-weight: 700;
+  color: var(--synapse-gray-700);
+  min-width: 100%;
+  padding: 8px 0;
+  @media (max-width: 600px) {
+    min-width: 100%;
+  }
+  @media (min-width: 601px) {
+    min-width: unset;
+  }
+  border-bottom: 4px solid;
+  border-bottom-color: transparent;
+  text-decoration: none !important;
+  &:global(.Mui-selected) {
+    color: var(--synapse-secondary-action-color);
+    border-bottom-color: var(--synapse-secondary-action-color);
+  }
+
+  &:hover {
+    border-bottom-color: var(--synapse-secondary-action-color);
+  }
+}
--- a/apps/synapse-portal-framework/src/components/Explore/ExploreWrapperTabs.module.scss
+++ b/apps/synapse-portal-framework/src/components/Explore/ExploreWrapperTabs.module.scss
@@ -5,9 +5,7 @@
   color: var(--synapse-gray-700);
   min-width: 100%;
   padding: 8px 0;
-  @media (max-width: 600px) {
-    min-width: 100%;
-  }
+
   @media (min-width: 601px) {
     min-width: unset;
   }
--- a/apps/synapse-portal-framework/src/components/Explore/ExploreWrapperTabs.tsx
+++ b/apps/synapse-portal-framework/src/components/Explore/ExploreWrapperTabs.tsx
@@ -7,9 +7,10 @@ import {
   useTheme,
 } from '@mui/material'
 import { useMemo } from 'react'
-import { useLocation, useNavigate } from 'react-router'
+import { Link, LinkProps, useLocation, useNavigation } from 'react-router'
 import { ExploreWrapperProps } from './ExploreWrapperProps'
 import { matchPath } from 'react-router'
+import styles from './ExploreWrapperTabs.module.scss'
 
 export function CustomScrollButton(props: TabScrollButtonProps) {
   if (props.disabled) {
@@ -34,9 +35,10 @@ export function CustomScrollButton(props: TabScrollButtonProps) {
 export function ExploreWrapperTabs(props: ExploreWrapperProps) {
   const { explorePaths } = props
   const theme = useTheme()
-  const { pathname } = useLocation()
-  const navigate = useNavigate()
+  const navigation = useNavigation()
+  const location = useLocation()
   const isMobileView = useMediaQuery(theme.breakpoints.down('sm'))
+  const pathname = navigation?.location?.pathname ?? location.pathname
 
   const pathnameWithoutTrailingSlash = useMemo(
     () => pathname.replace(/\/$/, ''),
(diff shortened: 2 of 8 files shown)


apps/synapse-portal-framework/src/components/Explore/ExploreWrapperTabs.tsx
```diff
@@ -81,32 +83,15 @@ export function ExploreWrapperTabs(props: ExploreWrapperProps) {
...
-              borderBottom: '4px solid',
-              borderBottomColor: 'transparent',
-              '&.Mui-selected': {
-                color: 'secondary.main',
-                borderBottomColor: 'secondary.main',
-              },
-              '&:hover:not(.Mui-selected)': {
-                color: 'grey.800',
-              },
-            }}
```
nickgros, line 109: Move styles to scss module

apps/synapse-portal-framework/src/components/Explore/ExploreWrapperTabs.tsx
```diff
@@ -81,32 +83,15 @@ export function ExploreWrapperTabs(props: ExploreWrapperProps) {
...
+            LinkComponent={Link}
```
nickgros, line 93: Use React Router `Link` component instead of navigation `onClick`. 

I made this change to try to track down a bug. This didn't fix the bug, but is the right way to do this. Using the native components lets us take advantage of RR framework features, such as prefetching the route component and its data before the user clicks this.

apps/synapse-portal-framework/src/components/Explore/ExploreWrapperTabs.tsx
```diff
@@ -81,32 +83,15 @@ export function ExploreWrapperTabs(props: ExploreWrapperProps) {
     >
       {explorePaths.map(({ path, displayName = path }) => {
         path = `/Explore/${path}`
+        const linkProps: any = { to: path } satisfies LinkProps
```
nickgros, line 86: Passing RR `Link` props to the MUI `Tab` component yields unexpected TypeScript errors.

`any` allows us to pass the RR props to the MUI component, while `satisfies LinkProps` gives us type safety.

apps/synapse-portal-framework/src/sitemap/types.ts
```diff
@@ -30,6 +30,24 @@ export interface DetailPageConfig {
...
+   *   by `sql`) and leave the rest to the SPA fallback at runtime. Use for
+   *   high-cardinality detail pages (e.g. >1000 records) where the build-time
+   *   cost outweighs the SEO benefit. The single prerendered page is required
+   *   to satisfy React Router's `ssr:false` validator, which rejects `loader`
+   *   exports on routes that have zero prerendered paths.
+   *
+   * Either way, every record is still listed in `sitemap.xml` — this flag
+   * only affects build-time HTML generation, not sitemap inclusion.
+   */
+  prerender?: boolean
```
nickgros, line 50: Add an opt-out for detail page prerendering. CCKP has >4000 publications, each with a details page, which takes way too long to prerender.

I don't think SEO is important for these pages. If it is, we'll need to consider using SSR.

apps/synapse-portal-framework/src/ssg/createReactRouterConfig.ts
```diff
@@ -54,14 +54,18 @@ export function createReactRouterConfig(
...
+        // Either way React Router's ssr:false validator is satisfied because
+        // at least one path matches each loader-exporting route — the
+        // remaining paths fall through to the SPA fallback at runtime.
         for (const detailPage of sitemapConfig.detailPages) {
-          const result = await fetchResourceIds(
-            detailPage,
-            isProduction ? undefined : 1,
-          )
+          const limit =
+            !isProduction || detailPage.prerender === false ? 1 : undefined
```
nickgros, line 67: Respect the `prerender: false` opt-out

apps/synapse-portal-framework/src/utils/useSetCanonicalUrl.ts
```diff
@@ -1,31 +1,31 @@
...
 
 /**
  * Hook to set the canonical URL of the page.
  *
  * The canonical URL helps search crawlers understand which URL(s) to index when there are multiple URLs that point to
  * the same content.
  * @param canonicalUrl
+ * @deprecated Prefer using exporting `meta` from the page's route module to set the canonical URL. This approach is brittle and can conflict with other components that manipulate the same tag, leading to obscure React errors.
  */
 export function useSetCanonicalUrl(canonicalUrl?: string) {
```
nickgros, line 11: This hook was causing an error that was hard to track down. To set the canonical URL, it would delete an existing canonical URL tag before replacing it with a new tag.

If another React component created this tag (such as react-router), then when that component tried to clean itself up, React would fail to find the tag and throw a null reference error. The app's next route would fail to load, but no UI error would be thrown.

Fix the error by only deleting the tag if this component created it; otherwise, just update the tag's `href` attribute. 

Also, mark this as `@deprecated` so we can eventually migrate to remove this and only use React Router for this functionality.

packages/synapse-react-client/src/utils/hooks/useShowDesktop.ts
```diff
@@ -6,21 +6,18 @@ const MOBILE_VIEWPORT_MAX_WIDTH_PX = 768
 
 export default function useShowDesktop(breakpoint?: UseShowDesktopProps) {
```
nickgros, line 7: This hook used by the `Programs` home page component caused a hydration error because the first client render did not match the prerendered content if the viewport was narrow.

apps/synapse-portal-framework/src/components/Explore/ExploreWrapperTabs.tsx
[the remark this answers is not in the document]
nickgros: This is specifically a workaround to allow passing these props to `Tab`

apps/synapse-portal-framework/src/components/Explore/ExploreWrapperTabs.module.scss
```diff
@@ -0,0 +1,25 @@
...
+  border-bottom: 4px solid;
+  border-bottom-color: transparent;
+  text-decoration: none !important;
+  &:global(.Mui-selected) {
+    color: var(--synapse-secondary-action-color);
+    border-bottom-color: var(--synapse-secondary-action-color);
+  }
+
+  &:hover {
+    border-bottom-color: var(--synapse-secondary-action-color);
```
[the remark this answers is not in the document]
nickgros: Minor change where an unselected tab's bottom border changes instead of its text on hover. This is minor and I think makes it more clear that the other tabs are clickable (the prior text color change was very subtle).

jay-hodgson (approved):

PR #2803, opened 2026-05-17
PORTALS-4189 - SSG for CCKP

nickgros:


apps/portals/cancercomplexity/src/config/sitemapConfig.ts
```diff
@@ -2,23 +2,55 @@ import {
...
+  },
+  {
+    path: 'Explore/Publications',
+    sql: publicationSql,
+    primaryKeyColumn: 'pubMedId',
+    urlParamStyle: 'path-segment',
+    // >4000 publications — too many to justify statically generating each
+    // page. Only one sample page is prerendered; the rest fall through to
+    // the SPA fallback. All entries still appear in sitemap.xml.
+    prerender: false,
```
nickgros, line 35: As mentioned in #2802 , this would cause very long build times. Let's not do this unless we need it.

If we do need it, then we should consider changing our CI pipeline or deploying a server and doing live SSR

apps/portals/cancercomplexity/src/root.tsx
```diff
@@ -0,0 +1,25 @@
...
+
+export const Layout = createLayout({
+  headChildren: (
+    <script
+      data-jsd-embedded
+      data-key="885c2536-2aaa-4f84-a9af-f39f15d44e4a"
+      data-base-url="https://jsd-widget.atlassian.com"
+      src="https://jsd-widget.atlassian.com/assets/embed.js"
+    />
+  ),
```
nickgros, line 21: Include Jira Service Desk script (required adding this `createLayout` function)

apps/synapse-portal-framework/src/ssg/root.tsx
```diff
@@ -101,33 +105,54 @@ export function links() {
...
-      </body>
-    </html>
-  )
+/**
+ * Returns a `Layout` component for the root route. Pass `headChildren` to
+ * inject supplemental tags (e.g. a third-party widget `<script>`) into
+ * `<head>`. Portals that don't need extras can re-export the default `Layout`
+ * named export below.
+ */
+export function createLayout(options: LayoutOptions = {}) {
```
nickgros, line 123: allows portals to include their own additional head content such as `script`s. I think CCKP is the only one that does this right now.

apps/portals/cancercomplexity/src/pages/DatasetsDetailsPage.tsx
```diff
@@ -23,9 +28,34 @@ import { grantsCardConfiguration } from '../config/synapseConfigs/grants'
...
+  {
+    extendLoader: async (_base, params) => ({
+      croissantJsonLd: params.datasetId
+        ? await fetchCroissantMetadata(params.datasetId)
+        : null,
+    }),
+    extendMeta: data =>
+      data.croissantJsonLd ? [{ 'script:ld+json': data.croissantJsonLd }] : [],
+  },
+)
```
nickgros, line 52: A couple of datasets in the CCKP are referenced in our croissant catalog. Get the JSON LD the same way we do it in the NF portal

jay-hodgson (approved):

PR #2804, opened 2026-05-18
SWC-7050 Table upload wizard with new schema editor

afwillia:


packages/synapse-react-client/src/components/file/upload/BasicFileHandleUpload.tsx
```diff
@@ -23,7 +23,7 @@ export type BasicFileHandleUploadProps = {
   /** Callback that is invoked when component is ready to upload */
   onUploadReady?: () => void
   /** Callback that is invoked when an individual upload is complete */
-  onFileUploadComplete?: (fileHandleIds: string) => void
+  onFileUploadComplete?: (fileHandleId: string, file: File) => void
```
afwillia, line 26: add file to preserve behavior of setting the table name as the filename

PR #2805, opened 2026-05-18
PORTALS-4242: Navbar with Search causes mobile menu to show up in an odd location

kianamcc:
Jira: https://sagebionetworks.jira.com/browse/PORTALS-4242?search_id=f254d349-2a29-42fd-923e-51d926b7ca68

Could not reproduce locally in 'pnpm dev' but could reproduce in 'pnpm preview' (pre-rendered). I think module CSS loads later in 'pnpm dev' so modules win the cascade. But in 'pnpm preview', the prerendred HTML has the order as the CSS modules first then MUI/Emotion classes later, so MUI style wins. By using injectFirst, like the other SPA portals already do, we can let module CSS take precedence.

Sticky search + nav mobile:
Screenshot 2026-05-18 at 2 15 13 PM
Screenshot 2026-05-18 at 2 15 19 PM

Sticky search + nav desktop and header search box styling:
Before:
A home for
After:
A home for

### DIFF
--- a/apps/synapse-portal-framework/src/components/HeaderSearchBoxV3.module.scss
+++ b/apps/synapse-portal-framework/src/components/HeaderSearchBoxV3.module.scss
@@ -14,7 +14,9 @@
 }
 
 .exampleSearchesSection {
-  display: none;
+  &.exampleSearchesSection {
+    display: none;
+  }
 }
 
 .exampleTermsContainer {
--- a/apps/synapse-portal-framework/src/ssg/PortalRoot.tsx
+++ b/apps/synapse-portal-framework/src/ssg/PortalRoot.tsx
@@ -55,7 +55,7 @@ export default function PortalRoot(props: PortalRootProps) {
   return (
     <PortalContextProvider value={portalContext}>
       <CookiesProvider>
-        <ThemeProvider theme={{ palette }} injectFirst>
+        <ThemeProvider theme={{ palette }}>
           <CssBaseline />
           <QueryClientProvider client={queryClient}>
             <DocumentMetadataProvider defaultTitle={portalContext.portalName}>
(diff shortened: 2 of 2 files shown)


kianamcc:
Checking some pages for breaking changes:
 
 Explore
￼
Screenshot 2026-05-18 at 2 08 36 PM
Screenshot 2026-05-18 at 2 08 45 PM

 Details
￼
Screenshot 2026-05-18 at 2 09 55 PM
Screenshot 2026-05-18 at 2 09 43 PM


 Search 
￼
Screenshot 2026-05-18 at 2 10 47 PM

Screenshot 2026-05-18 at 2 10 40 PM

jay-hodgson (approved):
Vite vs MUI (style injection).  
Thank you for fixing so the output specificity is consistent with other (non SSG) portals.  If there are additional style bugs (relating to this order), more time may be needed.

nickgros (approved):

kianamcc:
The culprit!

example searches section's display none getting overridden:
Screenshot 2026-05-19 at 4 52 00 PM

increase specificity (temporary fix for now):
Screenshot 2026-05-19 at 4 49 50 PM

jay-hodgson (approved):

apps/synapse-portal-framework/src/components/HeaderSearchBoxV3.module.scss
```diff
@@ -14,7 +14,9 @@
 }
 
 .exampleSearchesSection {
-  display: none;
+  &.exampleSearchesSection {
```
jay-hodgson, line 17: Prefer to define a specific selector naturally (e.g., scope to a parent, to increase specificity), but the doubled-class should do the trick

PR #2818, opened 2026-05-20
PORTALS-4239 - Add nonprofit info to footer

nickgros:
Screenshot of the NF portal with JavaScript disabled:

image


jay-hodgson:
Could you update the layout based on this quick design pass from Adam?

image

jay-hodgson (changes requested):
See other comment (with mockup)

nickgros:
@jay-hodgson the Version Info information ends up populating the left part, it just doesn't show up when JavaScript is disabled:

image

Would you like me to fold that information into the new content? Or put it in the middle and have it wrap at narrow screen widths?

jay-hodgson (approved):

jay-hodgson:
> Would you like me to fold that information into the new content? Or put it in the middle and have it wrap at narrow screen widths?

In that case, let's merge what you have!

adamhindman:
Makes sense to me. Thanks Nick + Jay!

PR #2820, opened 2026-05-20
PORTALS-4248, PORTALS-4249

jay-hodgson:
## PR #2820 — PORTALS-4248, PORTALS-4249
---

### Summary

Two bug fixes related to portal search UX and the NF (NeuroFibromatosis) portal's Browse Tools page.

---

#### 1. `HeaderSearchBox.tsx` — Fix navigation when already on the search path (PORTALS-4248)

Previously, navigating from the search box always set `pathname: path`, causing a full route re-render even when the user was already on the search results page. This broke the UX because the current object context was lost (I was searching Tools, and then another tab was auto-selected).

**Fix:** Uses `useLocation` to check if the browser is already on the target path. If so, omits `pathname` from the `navigate()` call so only the search params are updated.

```tsx
const alreadyOnPath = location.pathname.startsWith(path)
navigate({
  ...(alreadyOnPath ? {} : { pathname: path }),
  search: `?${params.toString()}`,
})
```

---

#### 2. `SearchQueryWrapperPlotNav.tsx` (SRC) — Expose `hideSearchBarControl` and `defaultShowSearchBar` as props

These two props were previously hardcoded (`false` and `true` respectively). They are now passed through as optional props with the same defaults, making the component more configurable.

---

#### 3. `SearchParamAwareQueryWrapperPlotNav.tsx` — Hide the search bar in param-aware variants

Both `SearchParamAwareSearchQueryWrapperPlotNav` and `SearchParamAwareStandaloneQueryWrapperPlotNav` now explicitly pass `hideSearchBarControl={true}` and `defaultShowSearchBar={false}`. This hides the redundant in-table search bar when the portal header search box is the intended entry point.

---

#### 4. `NFBrowseToolsPage.tsx` — Remove outdated search guidance text (PORTALS-4249)

Removes two `<Typography>` blocks containing search tips referencing MySQL full-text search quirks (spelling, pluralization, suffixes) and a link to an NF docs page. This copy is no longer accurate given updates to the search implementation.

### DIFF
--- a/apps/synapse-portal-framework/src/components/HeaderSearchBox.tsx
+++ b/apps/synapse-portal-framework/src/components/HeaderSearchBox.tsx
@@ -14,7 +14,7 @@ import {
 } from '@mui/material'
 import React from 'react'
 import PortalFullTextSearchField from './PortalSearch/PortalFullTextSearchField'
-import { useNavigate } from 'react-router'
+import { useLocation, useNavigate } from 'react-router'
 import defaultStyles from './HeaderSearchBox.module.scss'
 import v2Styles from './HeaderSearchBoxV2.module.scss'
 import v3Styles from './HeaderSearchBoxV3.module.scss'
@@ -67,6 +67,7 @@ const HeaderSearchBox = ({
   const [mode, setMode] = useState<'Chat' | 'Search'>('Search')
   const theme = useTheme()
   const navigate = useNavigate()
+  const location = useLocation()
   const { isAuthenticated } = useSynapseContext()
   const chatDialogContext = useChatDialogContext()
   const isChatAvailable = chatDialogContext?.isChatAvailable
@@ -93,8 +94,9 @@ const HeaderSearchBox = ({
         if (role) {
           params.set(SEARCH_ROLE, role)
         }
+        const alreadyOnPath = location.pathname.startsWith(path)
         navigate({
-          pathname: path,
+          ...(alreadyOnPath ? {} : { pathname: path }),
           search: `?${params.toString()}`,
         })
       }
--- a/apps/synapse-portal-framework/src/components/nf/NFBrowseToolsPage.tsx
+++ b/apps/synapse-portal-framework/src/components/nf/NFBrowseToolsPage.tsx
@@ -7,7 +7,7 @@ import advancedCellularModelsUrl from '@/components/assets/advanced-cellular-mod
 import clinicalAssessmentToolsUrl from '@/components/assets/clinical-assessment-tools.svg?url'
 import computationalToolsUrl from '@/components/assets/computational-tools.svg?url'
 import patientDerivedModelsUrl from '@/components/assets/patient-derived-models.svg?url'
-import { Box, Link, Typography } from '@mui/material'
+import { Box, Typography } from '@mui/material'
 import { Query } from '@sage-bionetworks/synapse-types'
 import pluralize from 'pluralize'
 import { ReactElement, ReactNode } from 'react'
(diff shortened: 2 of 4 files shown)


apps/synapse-portal-framework/src/components/nf/NFBrowseToolsPage.tsx
```diff
@@ -306,24 +306,6 @@ const NFBrowseToolsPage = (props: NFBrowseToolsPageProps): ReactNode => {
         >
           What Tools Can We Help You Find?
         </Typography>
-        <Typography
```
jay-hodgson, line 309: Removed, as it is no longer relevant

apps/synapse-portal-framework/src/components/HeaderSearchBox.tsx
```diff
@@ -93,8 +94,9 @@ const HeaderSearchBox = ({
         if (role) {
           params.set(SEARCH_ROLE, role)
         }
+        const alreadyOnPath = location.pathname.startsWith(path)
         navigate({
-          pathname: path,
+          ...(alreadyOnPath ? {} : { pathname: path }),
```
kianamcc, line 99: To be more explicit, instead of {}, should we use location.pathname when alreadyOnPath is true instead of relying of router fallback behavior?

kianamcc (approved):

apps/synapse-portal-framework/src/components/HeaderSearchBox.tsx
jay-hodgson: I have no preference, so I asked my friend Claude.  Here's what it told me:

---

It's a reasonable question, but using {} (omitting [pathname](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html)) is actually the more correct approach here.

In React Router v6, [navigate({ search: '...' })](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html) without a [pathname](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html) performs a relative navigation — it preserves the current [location.pathname](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html) exactly, including any dynamic segments (e.g. /explore/tools/ToolDetail/syn123).

If you passed [pathname: location.pathname](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html) explicitly, it would be functionally equivalent in most cases, but there's a subtle difference: [location.pathname](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html) is the fully resolved current path, while omitting [pathname](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html) tells the router "stay wherever I am." The end result is the same, but the omission is semantically more precise — it says "I don't want to change the path, only the search params."

The only scenario where [location.pathname](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html) would be meaningfully different is if the router resolved [pathname](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html) differently than [location.pathname](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html) (e.g. trailing slash normalization), which would be a router quirk rather than a feature.

So the current {} approach is fine and idiomatic. If you wanted to be maximally explicit for readability, [pathname: location.pathname](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html) works, but it's not more correct — just more verbose.

PR #2824, opened 2026-05-21
SWC-7836

jay-hodgson:
Add error boundary wrapper:
Screenshot 2026-05-21 at 9 46 30 AM
Screenshot 2026-05-21 at 9 46 46 AM
Fix actual error, supporting the new searchindex entity type:
Screenshot 2026-05-21 at 9 47 08 AM

### DIFF
--- a/packages/synapse-react-client/src/components/EntityDownloadButton/EntityDownloadButton.tsx
+++ b/packages/synapse-react-client/src/components/EntityDownloadButton/EntityDownloadButton.tsx
@@ -98,6 +98,12 @@ read.table(query$filepath, sep=",")`,
 query = syn.tableQuery("SELECT * FROM ${id}")
 query.asDataFrame()`,
       }
+    case EntityType.searchindex:
+      return {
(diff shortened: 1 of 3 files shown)


nickgros (approved):

packages/synapse-react-client/src/components/EntityDownloadButton/EntityDownloadButton.tsx
```diff
@@ -98,6 +98,12 @@ read.table(query$filepath, sep=",")`,
...
 query.asDataFrame()`,
       }
+    case EntityType.searchindex:
+      return {
+        cliCode: undefined,
+        rCode: undefined,
+        pythonCode: undefined,
+      }
     default:
       throw new Error(`Unhandled EntityType: ${type}`)
```
nickgros, line 108: maybe add a test that iterates through every `EntityType` value and asserts that no error is thrown, so when a new EntityType is added, the test fails

nickgros (approved):

PR #2829, opened 2026-05-22
PORTALS-4252: Use new CCKP SearchIndex entities

jay-hodgson:


jay-hodgson:
At the time of writing, tools cannot be found - syn74937079.
So this PR must not be merged until this issue has been sorted out

nickgros (approved):

PR #2835, opened 2026-05-24
PORTALS-4256

jay-hodgson:
### Problem
`SearchQueryWrapper` uses the Synapse Search API (OpenSearch) as its data source. Unlike the standard Table Query API, the Search API cannot return `FacetColumnResultRange` results (min/max stats) for range-type columns, only terms aggregations. This caused range facet filters to be invisible, and DATE range values were sent to the API in the wrong format (`YYYY-MM-DD` instead of unix milliseconds).

---

#### `toSearchIndexQuery`
- **Exclude range columns from `facetRequests`**: changed filter from `cm.facetType != null` to `cm.facetType === 'enumeration'`. The Search API's `FacetRequest` only supports terms aggregations; including range columns produced garbage bucket results.
- **DATE range filter conversion**: when building `rangeFilters` from `selectedFacets`, DATE column values formatted as `YYYY-MM-DD` are now converted to unix millisecond strings (as OpenSearch expects). Non-DATE and already-numeric values are passed through unchanged. One-sided ranges (`undefined` min or max) are preserved as `undefined`.

#### `searchQueryResultsToQueryResultBundle`
- **Synthetic `FacetColumnResultRange` facets**: since the Search API never returns range facets, they are now synthesised client-side for all `DATE`, `DOUBLE`, and `INTEGER` range columns that are absent from the server response. Synthesised facets use `columnMin: ''` / `columnMax: ''` (bounds unavailable) and populate `selectedMin` / `selectedMax` from the current query's `rangeFilters`. For DATE columns, unix ms values in `rangeFilters` are converted back to `YYYY-MM-DD` for display.
- **Facet ordering**: updated to merge server facets and synthetic facets before sorting by `columnModels` order.

---

### `RangeFacetFilterUI.tsx`

- **`facetResult` prop type**: `columnMin` and `columnMax` are now optional (`string | undefined`) instead of required. This reflects that the Search API cannot supply column statistics.
- **INTEGER fallback**: changed `columnMin != null && columnMax != null` guard to a truthiness check (`columnMin && columnMax`). When bounds are absent (empty string or undefined — as in synthesised facets), the component now renders a plain `Range type="number"` input instead of a `RangeSlider` that requires a valid numeric domain.
- **DATE `initialValues`**: changed from always passing `{ min: '...', max: '...' }` to passing `undefined` when no bounds or selected values are present. Avoids passing empty strings to `dayjs`, which would produce `"Invalid Date"`.
- **DOUBLE `initialValues`**: same guard — only passes `initialValues` when at least one bound is present, preventing `parseFloat('')` → `NaN`.
- **Single-column guard**: updated `columnMin === columnMax` label shortcut to also require both values to be truthy.
- **`getRadioValue` call**: `currentMin ?? ''` passed instead of bare `currentMin` (which could be `undefined` when bounds are absent).

---

### `Range.tsx`

- **`useState` initialiser**: guards `dayjs(...)` calls with a truthiness check on `initialValues.min` / `initialValues.max`. Previously, `dayjs(undefined).format('YYYY-MM-DD')` produced the string `"Invalid Date"`, which would then be sent to the backend causing a `number_format_exception`.
- **`isValid`**: treats empty string (`''`) the same as `null`/`undefined` — one-sided ranges (only min or only max filled) are always valid. Previously an empty max triggered `Number('') === 0`, causing a false `min > max` error.
- **`handleAppyChanges`**: normalises `''` to `undefined` in the callback so downstream code never receives empty strings.
- **DatePicker `error={false}`**: added `error: false` to `slotProps.textField` on both From and To `DatePicker` instances. Suppresses MUI's built-in red error outline when a date field is empty (valid for one-sided ranges).

---

### Tests

New tests added across three files (17 new test cases total):

| File | New tests |
|---|---|
| `SearchQueryUseQueryOptions.test.ts` | Range column exclusion from `facetRequests`; DATE YYYY-MM-DD → unix ms conversion; non-DATE passthrough; `undefined` bound preservation; synthesis for DATE, DOUBLE, INTEGER; no duplicate synthesis; unix ms → YYYY-MM-DD in `selectedMin`/`selectedMax`; absent `selectedMin`/`selectedMax`; one-sided synthesised range |
| `Range.test.tsx` | One-sided number apply (min only, max only); date init with `undefined` min shows empty input (not `"Invalid Date"`); one-sided date apply |
| `RangeFacetFilter.test.tsx` | INTEGER without bounds shows `Range` number input; INTEGER with empty-string bounds shows `Range` (not `RangeSlider`); INTEGER pre-filled from `selectedMin`/`selectedMax`; INTEGER with real bounds still shows `RangeSlider` |

### DIFF
--- a/packages/synapse-react-client/src/components/SearchQueryWrapper/SearchQueryUseQueryOptions.ts
+++ b/packages/synapse-react-client/src/components/SearchQueryWrapper/SearchQueryUseQueryOptions.ts
@@ -250,16 +250,20 @@ export function searchQueryResultsToQueryResultBundle(
 
   const rawFacets = results.facets as QueryResultBundle['facets']
 
-  // Synthesize FacetColumnResultRange for range-type DATE/DOUBLE columns. The Search API's
+  // Synthesize FacetColumnResultRange for range-type DATE/DOUBLE/INTEGER columns. The Search API's
   // FacetRequest only produces terms aggregations and cannot return min/max stats, so these
   // entries will never appear in results.facets. columnMin/columnMax are empty strings
   // (unavailable from the server); selectedMin/selectedMax are derived from the query's
   // rangeFilters so any applied range filter is still reflected in the UI.
+  // Note: for INTEGER columns the RangeSlider requires real bounds, so the UI falls back to a
+  // plain Range number input when columnMin/columnMax are absent.
   const syntheticRangeFacets: FacetColumnResultRange[] = (columnModels ?? [])
     .filter(
       cm =>
         cm.facetType === 'range' &&
-        (cm.columnType === 'DATE' || cm.columnType === 'DOUBLE') &&
+        (cm.columnType === 'DATE' ||
+          cm.columnType === 'DOUBLE' ||
+          cm.columnType === 'INTEGER') &&
         !(rawFacets ?? []).some(
           f => f.columnName === cm.name && f.facetType === 'range',
         ),
--- a/packages/synapse-react-client/src/components/SearchQueryWrapper/SearchQueryUseQueryOptions.ts
+++ b/packages/synapse-react-client/src/components/SearchQueryWrapper/SearchQueryUseQueryOptions.ts
@@ -263,7 +263,8 @@ export function searchQueryResultsToQueryResultBundle(
         cm.facetType === 'range' &&
         (cm.columnType === 'DATE' ||
           cm.columnType === 'DOUBLE' ||
-          cm.columnType === 'INTEGER'),
+          cm.columnType === 'INTEGER') &&
+        !(rawFacets ?? []).some(f => f.columnName === cm.name),
     )
     .map(cm => {
       const rangeFilter = query.searchQuery?.rangeFilters?.find(
--- a/packages/synapse-react-client/src/components/SearchQueryWrapper/SearchQueryUseQueryOptions.ts
+++ b/packages/synapse-react-client/src/components/SearchQueryWrapper/SearchQueryUseQueryOptions.ts
@@ -38,6 +38,7 @@ import {
   UseSuspenseQueryOptions,
 } from '@tanstack/react-query'
 import { omit } from 'lodash-es'
+import dayjs from 'dayjs'
 import { useCallback, useMemo } from 'react'
 import { KeyFactory } from '@/synapse-queries/KeyFactory'
 
@@ -119,18 +120,33 @@ export function toSearchIndexQuery(
       )
       .map(f => ({ key: f.columnName, values: f.facetValues }))
 
-  // Build rangeFilters from range selectedFacets
+  // Build rangeFilters from range selectedFacets.
+  // DATE columns: the Range UI component emits formatted dates ("YYYY-MM-DD") but the
+  // Search API (OpenSearch) expects unix millisecond timestamps. Convert here.
   const rangeFilters: SearchKeyRange[] | undefined =
     queryBundleRequest.query.selectedFacets
       ?.filter(
         (f): f is FacetColumnRangeRequest =>
           f.concreteType === FACET_COLUMN_RANGE_REQUEST_CONCRETE_TYPE_VALUE,
       )
-      .map(f => ({
-        key: f.columnName,
-        min: f.min,
-        max: f.max,
-      }))
+      .map(f => {
+        const columnType = columnModels?.find(
+          cm => cm.name === f.columnName,
+        )?.columnType
+        const toApiValue = (v: string | undefined): string | undefined => {
+          if (!v || columnType !== 'DATE') return v
+          // Convert "YYYY-MM-DD" → unix ms; leave already-numeric values untouched
+          if (/^\d{4}-\d{2}-\d{2}$/.test(v)) {
+            return String(dayjs(v).valueOf())
+          }
+          return v
+        }
+        return {
+          key: f.columnName,
+          min: toApiValue(f.min),
+          max: toApiValue(f.max),
+        }
+      })
 
   // Extract queryText by concatenating all TextMatchesQueryFilter searchExpression values
   const textMatchesExpressions = queryBundleRequest.query.additionalFilters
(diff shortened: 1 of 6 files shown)


packages/synapse-react-client/src/components/widgets/Range.tsx
```diff
@@ -40,12 +44,8 @@ export function Range(props: RangeProps): React.ReactNode {
...
     type: ControlType = 'number',
   ) => {
-    if (
-      min === null ||
-      min === undefined ||
-      max === null ||
-      max === undefined
-    ) {
+    // Treat null, undefined, and empty string as "no value" — one-sided ranges are always valid
+    if (min == null || min === '' || max == null || max === '') {
```
nickgros: Could use lodash `isEmpty`
```suggestion
    if (isEmpty(min) || isEmpty(max)) {
```

packages/synapse-react-client/src/components/SearchQueryWrapper/SearchQueryUseQueryOptions.ts
```diff
@@ -116,18 +120,33 @@ export function toSearchIndexQuery(
...
-      }))
+      .map(f => {
+        const columnType = columnModels?.find(
+          cm => cm.name === f.columnName,
+        )?.columnType
+        const toApiValue = (v: string | undefined): string | undefined => {
+          if (!v || columnType !== 'DATE') return v
+          // Convert "YYYY-MM-DD" → unix ms; leave already-numeric values untouched
+          if (/^\d{4}-\d{2}-\d{2}$/.test(v)) {
+            return String(dayjs(v).valueOf())
```
nickgros, line 140: The REST docs say:

> [...] [date columns accept ISO-8601 strings](https://rest-docs.synapse.org/rest/org/sagebionetworks/repo/model/search/KeyRange.html) [...]

Does @BryanFauble plan to make the API and REST docs consistent?

packages/synapse-react-client/src/components/SearchQueryWrapper/SearchQueryUseQueryOptions.ts
```diff
@@ -229,24 +248,75 @@ export function searchQueryResultsToQueryResultBundle(
...
+  const syntheticRangeFacets: FacetColumnResultRange[] = (columnModels ?? [])
+    .filter(
+      cm =>
+        cm.facetType === 'range' &&
+        (cm.columnType === 'DATE' ||
+          cm.columnType === 'DOUBLE' ||
+          cm.columnType === 'INTEGER') &&
+        !(rawFacets ?? []).some(
+          f => f.columnName === cm.name && f.facetType === 'range',
+        ),
```
nickgros: I am very confused by this check. Will the backend ever give us range facets in `results.facets`? Under what conditions?

packages/synapse-react-client/src/components/SearchQueryWrapper/SearchQueryUseQueryOptions.ts
```diff
@@ -229,24 +248,75 @@ export function searchQueryResultsToQueryResultBundle(
...
+      )
+      // rangeFilters carry unix ms for DATE columns (converted in toSearchIndexQuery).
+      // Convert back to "YYYY-MM-DD" so the Range UI component can display them.
+      const toDisplayValue = (v: string | undefined): string | undefined => {
+        if (!v || cm.columnType !== 'DATE') return v
+        if (/^\d+$/.test(v)) {
+          return dayjs(parseInt(v)).format('YYYY-MM-DD')
+        }
+        return v
+      }
```
nickgros, line 281: Is the range component smart enough to drop the min/max value if the user selected the largest rendered value? If not, there is a chance that this transformation causes us to lose resolution that unexpectedly filters data.

For example, consider a date column whose max value is `1779888149396` (Wed May 27 2026 13:22:29). The facet maximum, after converted to `YYYY-MM-DD` and back just becomes `1779854400000` (Wed May 27 2026 00:00:00). If a user applies this facet and manually sets the maximum to be `2026-05-27`, and we don't clear the max, rows where the value is greater than `1779854400000` could be unexpectedly excluded from the results.

packages/synapse-react-client/src/components/SearchQueryWrapper/SearchQueryUseQueryOptions.ts
```diff
@@ -229,24 +248,75 @@ export function searchQueryResultsToQueryResultBundle(
...
+    ...(rawFacets ?? []),
+    ...syntheticRangeFacets,
+  ]
+
+  // Reorder all facets to match the columnModels order so downstream components
+  // (e.g. FacetFilterControls) render facets in the expected column order.
+  const orderedFacets: QueryResultBundle['facets'] =
+    rawFacets == null && syntheticRangeFacets.length === 0
+      ? undefined
+      : columnModels
```
nickgros: Can you remove the first ternary? Seems to be unnecessary unless there is a real case where `orderedFacets` must be `null` instead of an empty array

packages/synapse-react-client/src/components/SearchQueryWrapper/SearchQueryUseQueryOptions.ts
jay-hodgson: The Search API does not return the column maximum value (or the max of the column value for the result set).  So in your example, how would the value `1779888149396` ever be set in the component?

jay-hodgson: I was thinking that it could be possible for a user to save the URL from a page that used the Query API, but as soon as the user applied a new date range filter from the UI the value would have been rounded to the day

nickgros: Makes sense. It could be an issue with a slider (and might even be a real issue with the Table Query API and UI today), but seems like a non-issue for these search cases where there is no slider.

packages/synapse-react-client/src/components/SearchQueryWrapper/SearchQueryUseQueryOptions.ts
jay-hodgson: I don't think this will ever be the case, I'll remove.

packages/synapse-react-client/src/components/SearchQueryWrapper/SearchQueryUseQueryOptions.ts
jay-hodgson: Yes, I think we can remove.

packages/synapse-react-client/src/components/SearchQueryWrapper/SearchQueryUseQueryOptions.ts
jay-hodgson: I reached out to Bryan to highlight this documentation issue.  I verified that passing an ISO-8601 string causes an error, so it's not expecting that format.

packages/synapse-react-client/src/components/SearchQueryWrapper/SearchQueryUseQueryOptions.ts
jay-hodgson: The Range Slider component is only used if the column type is Integer (and with this PR if the columnMin and columnMax are known).  Will not be used if the column type is Date

nickgros (approved):

PR #2836, opened 2026-05-26
SWC-7833 Fix Autocomplete cell selection

afwillia:
Fixes a bug where the dropdown cell value and label diverge because the content appears selected but does not respond to delete.

### DIFF
--- a/packages/synapse-react-client/src/components/DataGrid/columns/useGridAutocompleteState.ts
+++ b/packages/synapse-react-client/src/components/DataGrid/columns/useGridAutocompleteState.ts
@@ -105,13 +105,11 @@ export function useGridAutocompleteState({
   // grid focus (our effect would open the menu) and (b) toggles via MUI's
   // own onClick (which would then close because we just opened it) — flicker.
   // Letting MUI handle the open via its onOpen avoids the dueling state.
+  // Cleared by handleMenuOpen (successful click) or by the focus/active effects
+  // when focus drops (abandoned press).
   const popupClickInProgressRef = useRef(false)
   const handlePopupIndicatorMouseDown = useCallback(() => {
     popupClickInProgressRef.current = true
-    // Clear on the next macro task — after MUI's click handler has run.
-    window.setTimeout(() => {
-      popupClickInProgressRef.current = false
-    }, 0)
   }, [])
 
   useEffect(() => {
@@ -123,6 +121,7 @@ export function useGridAutocompleteState({
     } else if (!optionMouseDownRef.current) {
       // Only close when there is no pending option click; if there is one,
       // onChange will reset optionMouseDownRef and handle cleanup itself.
+      popupClickInProgressRef.current = false
       setMenuIsOpen(false)
       inputRef.current?.blur()
     }
@@ -153,6 +152,7 @@ export function useGridAutocompleteState({
       // not enough — close on deactivation too. Otherwise a deactivated cell
       // can leave a stale menu open that re-anchors and "moves" as the user
       // scrolls the grid, or stack up alongside the next cell's menu.
+      popupClickInProgressRef.current = false
       setMenuIsOpen(false)
       inputRef.current?.blur()
       onDeactivateRef.current?.()
@@ -178,6 +178,7 @@ export function useGridAutocompleteState({
   }, [])
 
   const handleMenuOpen = useCallback(() => {
+    popupClickInProgressRef.current = false
     setMenuIsOpen(true)
   }, [])
 
--- a/packages/synapse-react-client/src/components/DataGrid/columns/useGridAutocompleteState.ts
+++ b/packages/synapse-react-client/src/components/DataGrid/columns/useGridAutocompleteState.ts
@@ -8,6 +8,7 @@ import {
 
 export type UseGridAutocompleteStateProps = {
   active: boolean
+  focus: boolean
   /**
    * `stopEditing` from react-datasheet-grid's CellProps. The hook ref-stabilizes
    * this internally, so callers can pass it directly without memoizing.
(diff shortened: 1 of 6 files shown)


packages/synapse-react-client/src/components/DataGrid/columns/useGridAutocompleteState.ts
```diff
@@ -88,17 +100,63 @@ export function useGridAutocompleteState({
...
+    const input = inputRef.current
+    if (!input) return
+    const handleFocus = () => {
+      input.blur()
+    }
+    input.addEventListener('focus', handleFocus)
+    return () => {
+      input.removeEventListener('focus', handleFocus)
+    }
+  }, [focus])
```
nickgros, line 146: This defensive focus/blur/event listener seems pretty complicated. Do we need to do all this? Would it fix the bug without meaningful loss in functionality/complexity if the AutocompleteCell rendered a different, readonly component when the cell is not focused? i.e.

```tsx
  if (!focus)
    return <SomeReadOnlyStyledCellComponent value={value} />
  }

  return <Autocomplete 
      {...}
  />
```

packages/synapse-react-client/src/components/DataGrid/columns/useGridAutocompleteState.ts
```diff
@@ -88,17 +100,63 @@ export function useGridAutocompleteState({
...
+  const popupClickInProgressRef = useRef(false)
+  const handlePopupIndicatorMouseDown = useCallback(() => {
+    popupClickInProgressRef.current = true
+  }, [])
+
   useEffect(() => {
-    if (active) {
+    if (focus) {
       inputRef.current?.focus()
+      if (!popupClickInProgressRef.current) {
```
nickgros, line 118: Instead of a ref tracking 'popupClickInProgressRef', can we capture the click event and check to see if the 'target' of the click was the popup icon button?

This wouldn't be possible in the 'effect', because it would have to happen in the click handler

nickgros:
From copilot discussion:

---------------

> Is this solution excessively complicated? Is there a more simple way to approach this?

No — I don’t think it’s *excessively* complicated, but it **is more complex than the ideal fix**, and most of that complexity comes from working around interaction quirks between **react-datasheet-grid**, **MUI Autocomplete**, and the portal-rendered menu.

My take:

- **The core idea is sound:** distinguish **`active`** from **`focus`**, and only show caret/open menu in real edit mode.
- **The complexity spike is in the defensive glue:**  
  - `popupClickInProgressRef`
  - `optionMouseDownRef`
  - the extra DOM `focus` listener that blurs when `focus === false`
  - multiple effects coordinating menu open/close and commit timing

That’s a lot of moving pieces for a UX bug in a cell editor.

### What feels justified
The PR description says the bug is that the cell can *look editable* while not actually being in edit mode, and the code now ties UI behavior to `focus` instead of `active`. That’s a good correction.

In particular, these changes seem reasonable:
- making menu open/close depend on `focus`
- syncing local input state when `focus` drops
- preserving special handling for portal clicks on options

### What feels like the “overcomplicated” part
The part I’d scrutinize is the **caret-defense listener** in `useGridAutocompleteState.ts`:

- when `focus` is false, it adds a native `focus` listener to the input
- if the input gets focused anyway, it immediately blurs it

That is clever, but it’s also the most fragile part of the patch. It suggests the component tree is still allowing focus to land in places it ideally wouldn’t. It also introduces behavior that’s harder to reason about than normal React/MUI state flow.

The popup-indicator guard also adds complexity, though it’s more understandable: it prevents “our effect opens it, then MUI toggles and closes it” flicker.

## Simpler approaches I’d consider

### Option 1: Split “selected view” and “editing view” more explicitly
The simplest conceptual model is:

- if `!focus`: render a **non-editing presentation**
- if `focus`: render the real `Autocomplete`

That avoids fighting browser focus in a half-editing state.

Example shape:

```tsx name=AutocompleteCell.tsx
return focus ? (
  <Autocomplete ... />
) : (
  <ReadOnlyCellDisplay
    value={rowDataAsString}
    onMouseDown={/* maybe let grid handle activation */}
  />
)
```

**Why this is simpler**
- no need to blur stray focus
- no need to suppress caret manually
- no auto-open/focus race while inactive
- much less dependency on MUI internals

**Tradeoff**
- you may lose the ability to click the popup indicator while only `active === true`
- if product wants “first click selects cell, arrow can still open menu before full edit mode,” then this won’t preserve current behavior

So this is the best simplification **if UX can be slightly tightened**.

---

### Option 2: Keep one component, but remove the popup-indicator special case
If clicking the popup indicator on an active-only cell is not essential, then require the cell to enter edit mode first, then let the menu open.

That would likely let you remove:
- `popupClickInProgressRef`
- `handlePopupIndicatorMouseDown`
- some related test bulk

Then the rule becomes:

- `focus=true` → focus input + maybe open menu
- `focus=false` → blur/close
- indicator clicks only matter once already editing

This is probably the cleanest simplification **without fully splitting render modes**.

---

### Option 3: Control `open` purely from state transitions, not DOM focus defense
If you keep the current model, I’d still ask whether the native focus listener can be eliminated by making the rendered input non-focusable when not editing.

For example, when `!focus`, see whether you can set input props to prevent real focus landing:
- `readOnly`
- `tabIndex={-1}`
- possibly disable pointer events on the inner input and let the wrapper/grid own activation

If that works, it would be cleaner than “let focus happen, then blur it immediately.”

## My recommendation
If you want the **smallest code with the least long-term maintenance cost**, the best approach is:

1. **Only render the actual Autocomplete in edit mode (`focus === true`)**
2. Render a lightweight read-only cell otherwise

If you must preserve the current first-click / second-click / popup-arrow behavior exactly, then the PR’s complexity is more understandable. In that case I’d still try to simplify by:

- **dropping the popup-indicator active-only path**, if acceptable
- or at minimum **removing the native focus-listener blur hack** if there’s any way to prevent the input from becoming focusable in the first place

## Bottom line
**Answer:** not wildly overengineered, but probably **more complicated than ideal** because it preserves nuanced existing interaction behavior.  
**Simpler alternative:** make the cell truly non-editable unless `focus === true`, ideally by rendering a read-only view outside edit mode. That’s the cleanest design.

If you want, I can also give you:
1. a **review comment draft** to leave on the PR, or  
2. a **concrete refactor sketch** showing what the simpler split-render approach would look like in this codebase.

nickgros (approved):

PR #2837, opened 2026-05-26
PORTALS-4257: rename NAMs Hub to NAM Hub

jay-hodgson:


apps/portals/namhub/src/config/scripts/exportS3ProductionBucketName.sh
```diff
@@ -0,0 +1,2 @@
+export S3_PRODUCTION_BUCK_LOCATION=s3://prod.namhub.synapse.org
```
jay-hodgson: @xschildw : Are these bucket names also being updated?  If not, then I'll revert the content change to  exportS3ProductionBucketName.sh and exportS3StagingBucketName.sh

xschildw: I was not planning on getting the buckets/distributions change. 
Hang on before you roll back, I'm not sure it's going to work.

xschildw (approved):
Never would have thought it'd be so many changes...

PR #2838, opened 2026-05-26
PORTALS-4222: Keyword search bar and URL aren't updated when search filter is removed

kianamcc:
Jira: https://sagebionetworks.jira.com/jira/software/c/projects/PLFM/boards/238?assignee=712020%3Aa7fbbeb0-d7a6-443a-9dd4-246363ff09dd&selectedIssue=PORTALS-4222

Screenshot 2026-05-26 at 2 21 37 PM
Screenshot 2026-05-26 at 2 27 48 PM
Screenshot 2026-05-26 at 2 38 55 PM


apps/synapse-portal-framework/src/components/PortalSearch/PortalSearchPage.tsx
```diff
@@ -15,6 +15,14 @@ export type PortalSearchPageProps = {
...
   searchPageTabs: PortalSearchTabConfig[]
   roleMapping?: Record<string, string>
+  /**
+   * Whether to hide the keyword-search (TextMatchesQueryFilter) chiclet in the
+   * "results filtered by" row. Defaults to true because removing the chiclet
+   * does not currently clear `SEARCH_TERM` from the URL nor the keyword input,
+   * which leaves the page in an inconsistent state. A portal can opt back in
+   * by passing `hideKeywordSearchPill={false}`.
+   */
+  hideKeywordSearchPill?: boolean
```
jay-hodgson: I think we can hardcoded to true because the portal search page should always hide.
Consider renaming the param throughout,.since we're really hiding all "TextMatchesQueryFilters".
The QueryWrapperPlotNav has real cases of adding multiple search terms.

kianamcc:
Are we ok with the "results filtered by" text still there when we have no pills showing?

Screenshot 2026-05-28 at 10 45 48 AM

jay-hodgson:
> Are we ok with the "results filtered by" text still there when we have no pills showing?
> 
Maybe we show the pill but not the ability to remove it?  Also, ideally the Clear All Filters would not remove this "locked" filter

kianamcc:
Agreed. Added lock icon if that's ok? 

Screenshot 2026-05-28 at 2 35 17 PM
Screenshot 2026-05-28 at 2 31 54 PM
Screenshot 2026-05-28 at 2 31 26 PM

packages/synapse-react-client/src/components/widgets/facet-nav/SelectionCriteriaPill.tsx
```diff
@@ -14,19 +16,30 @@ export type SelectionCriteriaPillProps = {
...
+        {!isLocked && onRemoveFilter && (
+          <button
+            onClick={onRemoveFilter}
+            className="SelectionCriteriaPill__btnRemove"
+            title="deselect"
+          >
+            <Close />
+          </button>
+        )}
+        {isLocked && (
```
jay-hodgson: Looks good to me, but we'll be responsive to user feedback as always.  
It's possible that this will be confused with our governance-related iconography

jay-hodgson (approved):

packages/synapse-react-client/src/components/widgets/facet-nav/SelectionCriteriaPill.tsx
kianamcc: Good point. I did not think of this. I think we should remove the lock icon for now then, keep it simple and less confusing, but will be ready to implement any feedback as needed!

kianamcc: Screenshot 2026-05-28 at 6 32 28 PM
Screenshot 2026-05-28 at 6 32 15 PM

jay-hodgson (approved):

PR #2839, opened 2026-05-26
PORTALS-4260: migrate AMP ALS portal to use search indices

jay-hodgson:


apps/portals/ampals/src/config/synapseConfigs/datasets.tsx
```diff
@@ -67,3 +68,16 @@ export const datasetCardConfiguration: CardConfiguration = {
 
   labelLinkConfig: datasetColumnLinks,
 }
+
+export const datasetsSearch: SearchQueryWrapperPlotNavProps = {
```
nickgros, line 72: I see 

> No enum constant org.sagebionetworks.repo.model.search.SearchQueryPart.FACETS

Image

apps/portals/ampals/src/config/searchConfig.ts
```diff
@@ -21,8 +21,8 @@ export const searchPageTabs = [
 ] as const satisfies PortalSearchTabConfig[]
 
 export const portalSearchPageConfigs = [
```
nickgros, line 23: On this page, the result never loads, and I see the same error in the network log:

```
{
    "concreteType": "org.sagebionetworks.repo.model.ErrorResponse",
    "reason": "No enum constant org.sagebionetworks.repo.model.search.SearchQueryPart.FACETS"
}
```

This error is never shown to the user, so it seems like it's just loading forever

jay-hodgson: I think this should be fixed by merging upstream `main` (after the Portal Search API update), but let me give it a go and push the merged commit if successful

jay-hodgson: Merged upstream `main`, and rebuilt (with no merge conflicts).  I do not see any errors after this update

nickgros (approved):

PR #2840, opened 2026-05-27
SWC-7608 memoize cell components

afwillia:
Manually memoize the Autocomplete and DateTime column components that are missed by react router.

### DIFF
--- a/packages/synapse-react-client/src/components/DataGrid/columns/DateTimeColumn.tsx
+++ b/packages/synapse-react-client/src/components/DataGrid/columns/DateTimeColumn.tsx
@@ -1,6 +1,8 @@
 import DateTimePicker from '@/components/DateTimePicker/DateTimePicker'
+import { SxProps, Theme } from '@mui/material'
 import dayjs, { Dayjs } from 'dayjs'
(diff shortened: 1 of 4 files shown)


packages/synapse-react-client/src/components/DataGrid/columns/DateTimeColumn.tsx
```diff
@@ -11,41 +13,87 @@ export type DateTimeCellProps = CellProps & {
...
   disabled,
   colType,
 }: DateTimeCellProps) {
+  // Stable ref so memo(DateTimeCell) can use default shallow comparison —
+  // react-datasheet-grid recreates setRowData on every render, but its behavior
+  // is stable for a given cell position.
+  const setRowDataRef = useRef(setRowData)
+  useLayoutEffect(() => {
+    setRowDataRef.current = setRowData
+  })
```
nickgros, line 37: Did these changes solve measurable performance issues? I am hesitant to start tracking callbacks in refs if not...I think this increases complexity and maintenance burden and we need to make sure it's worth it before making that choice

PR #2848, opened 2026-05-29
TextField improvements

nickgros:
- Refactor custom TextField to be based on MUI text field
- Remove usages of StyledFormControl wrapping TextField (which contains its own FormControl)
- [a11y] Link labels to inputs in valid uses of StyledFormControl

### DIFF
--- a/apps/SageAccountWeb/src/components/ProfileValidation/ProfileFieldsEditor.tsx
+++ b/apps/SageAccountWeb/src/components/ProfileValidation/ProfileFieldsEditor.tsx
@@ -1,10 +1,10 @@
 import React from 'react'
-import { StyledFormControl } from '@/components/StyledComponents'
-import { Box, TextField, useTheme } from '@mui/material'
+import { Stack, useTheme } from '@mui/material'
 import { VerificationSubmission } from '@sage-bionetworks/synapse-types'
 import { ChangeEvent, useState } from 'react'
 import RORInstitutionField from 'synapse-react-client/components/RORInstitutionField/RORInstitutionField'
 import { ContinueButton } from './ContinueButton'
+import { TextField } from 'synapse-react-client/components/TextField/index'
 
 const keysToValidate = ['firstName', 'lastName', 'location', 'company']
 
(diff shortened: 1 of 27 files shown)


apps/SageAccountWeb/src/components/ProfileValidation/ProfileFieldsEditor.tsx
```diff
@@ -55,75 +55,53 @@ export const ProfileFieldsEditor = (
...
   return (
     <>
-      <Box
+      <Stack
+        gap={2}
         sx={{
           marginBottom: theme.spacing(2),
         }}
       >
-        <StyledFormControl
```
nickgros, line 63: Removed usages of `StyledFormControl` where it was inappropriate; `TextField` (both ours and MUI's) includes its own internal FormControl

apps/SageAccountWeb/src/components/ProfileValidation/ProfileFieldsEditor.tsx
```diff
@@ -55,75 +55,53 @@ export const ProfileFieldsEditor = (
 
   return (
     <>
-      <Box
+      <Stack
+        gap={2}
```
nickgros, line 59: Changes to TextField and removal of StyledFormControl required adding spacing between form elements in many cases. I think this is fine; spacing/layout should be governed by the parent component.

packages/synapse-react-client/src/components/Authentication/TwoFactorEnrollmentForm.tsx
```diff
@@ -260,9 +260,8 @@ export default function TwoFactorEnrollmentForm(
                 }}
               >
                 <TextField
-                  noWrapInFormControl
```
nickgros, line 263: This was the only usage of `noWrapInFormControl`. The purpose of it was to control margins; instead we changed the component to ensure there is no margin if there's not a built-in label.

packages/synapse-react-client/src/components/Authentication/TwoFactorEnrollmentForm.tsx
```diff
@@ -260,9 +260,8 @@ export default function TwoFactorEnrollmentForm(
...
                   autoFocus
-                  inputProps={{ maxLength: totpSecret?.digits }}
+                  slotProps={{ htmlInput: { maxLength: totpSecret?.digits } }}
```
nickgros, line 264: MUI deprecated `inputProps` in favor of `slotProps`, so do so here

packages/synapse-react-client/src/components/TextField/TextField.tsx
```diff
@@ -1,98 +1,96 @@
...
-      noWrapInFormControl
-        ? (props: PropsWithChildren<object>) => <>{props.children}</>
-        : (props: PropsWithChildren<object>) => (
-            <StyledFormControl fullWidth sx={{ my: 1 }}>
-              {props.children}
-            </StyledFormControl>
-          ),
-    [noWrapInFormControl],
-  )
+export default function TextField({
```
nickgros, line 29: Refactored to

- Use MUI TextField instead of trying to roll-our-own using other MUI components. Rolling our own is a valid approach, but our current implementation is so similar that we benefit more from modifying theirs.
- Support any MUI TextFieldProp
- Remove some unused props such as `noWrapInFormControl`, deprecated `inputProps`

packages/synapse-react-client/src/components/TextField/TextField.tsx
```diff
@@ -1,98 +1,96 @@
...
-  | 'value'
-> & { noWrapInFormControl?: boolean; maxCharacterCount?: number }
+export type TextFieldProps = MuiTextFieldProps & {
+  /**
+   * Neutral supplementary text shown below the label and above the input (similar to an inline tooltip).
+   * Always rendered in a neutral color regardless of form state.
+   * For error or status messages, use `helperText` instead.
+   */
+  description?: ReactNode
+  maxCharacterCount?: number
```
nickgros, line 18: Renamed our custom `helperText` (which is overloaded with a MUI prop for text that appears below the input, and is often used for things like error messages) to `description`. We can use this and `helperText` at the same time.

packages/synapse-react-client/src/components/doi/CreateOrUpdateDoiModal.tsx
```diff
@@ -358,7 +358,7 @@ export function CreateOrUpdateDoiModal(props: CreateOrUpdateDoiModalProps) {
             </MarkdownPopover>
           </InputLabel>
           <Select
-            id="doi-version-select"
+            labelId="doi-version-select"
```
nickgros, line 361: For some reason, MUI Select a11y only works if you reference the labelId from the component (usually the other way around). Fixed here, and in all places where we were wrapping `Select` in a `StyledFormControl`

packages/synapse-react-client/src/components/styled/StyledFormControl.tsx
```diff
@@ -7,22 +7,29 @@ import {
...
+ * ```
+ */
 export const StyledFormControl: StyledComponent<FormControlProps> = styled(
   FormControl,
   {
     label: 'StyledFormControl',
   },
 )(({ theme }) => ({
   '& label': {
+    position: 'static',
```
nickgros, line 32: Fixes the labels being positioned inside the input container:

Image

packages/synapse-react-client/src/components/TextField/TextField.module.scss
```diff
@@ -1,20 +1,27 @@
...
+    order: 1;
+  }
 }
 
-.helperText {
+.description {
+  order: 2;
+  flex-basis: 100%;
   font-weight: 500;
+  // Explicit color prevents inheriting error/focused color from the parent InputLabel.
```
nickgros, line 23:
```suggestion
  text-wrap: wrap;
  // Explicit color prevents inheriting error/focused color from the parent InputLabel.
```

apps/SageAccountWeb/src/components/ProfileValidation/ProfileFieldsEditor.tsx
```diff
@@ -1,10 +1,10 @@
 import React from 'react'
-import { StyledFormControl } from '@/components/StyledComponents'
-import { Box, TextField, useTheme } from '@mui/material'
+import { Stack, useTheme } from '@mui/material'
 import { VerificationSubmission } from '@sage-bionetworks/synapse-types'
 import { ChangeEvent, useState } from 'react'
 import RORInstitutionField from 'synapse-react-client/components/RORInstitutionField/RORInstitutionField'
 import { ContinueButton } from './ContinueButton'
+import { TextField } from 'synapse-react-client/components/TextField/index'
```
jay-hodgson, line 7: Change to direct import?

jay-hodgson (approved):

apps/SageAccountWeb/src/components/ProfileValidation/ProfileFieldsEditor.tsx
nickgros: Oh, yes, I think I started doing that but changed my mind since it could have been a much larger change! Should be fine here, I inspected each affected component.

PR #2852, opened 2026-05-29
SWC-7838 display websocket errors in grid

afwillia:
Display websocket error messages as an alert in the grid. Hide the grid on error and prompt to refresh.


packages/synapse-react-client/src/components/DataGrid/SynapseGrid.tsx
```diff
@@ -420,8 +421,37 @@ const SynapseGrid = forwardRef<SynapseGridHandle, SynapseGridProps>(
...
+                      : typeof websocketError === 'object' &&
+                        websocketError !== null &&
+                        'message' in websocketError &&
+                        typeof (websocketError as { message: unknown })
+                          .message === 'string'
+                      ? (websocketError as { message: string }).message
+                      : 'An error occurred while communicating with the server. Please refresh the page to continue.'}
+                  </Alert>
+                </Grid>
+              )}
```
nickgros: This doesn't differ too much from our `SynapseErrorBoundary` component. You could trigger it with 
```ts
import { useErrorHandler } from 'react-error-boundary'

// ...
  const handleError = useErrorHandler()

  if (websocketError) {
    handleError(websocketError)
  }
```

You can add a more narrow error boundary parent wrapper if appropriate. The 'reload' button in the error boundary will reload the tree that errored, so the user doesn't have to refresh the entire page.

packages/synapse-react-client/src/components/DataGrid/SynapseGrid.tsx
```diff
@@ -559,4 +564,12 @@ const SynapseGrid = forwardRef<SynapseGridHandle, SynapseGridProps>(
   },
 )
 
+const SynapseGrid = forwardRef<SynapseGridHandle, SynapseGridProps>(
```
nickgros: we are using react 19 so [forwardRef is not necessary](https://react.dev/blog/2024/12/05/react-19#ref-as-a-prop) 

You could change `SynapseGridProps` to this and remove `forwardRef` from both the outer and inner components:

```ts
export type SynapseGridProps = {
  agentRegistrationId?: string
  showDebugInfo?: boolean
  ref?: Ref<SynapseGridHandle>
}
```

nickgros (approved):

nickgros (approved):

PR #2868, opened 2026-06-02
SWC-7843: Do not show portal banner if appId is unset

jay-hodgson:


jay-hodgson:
After the change, the banner does not show up.
Screenshot 2026-06-02 at 10 57 10 AM
But rows that have appId set still cause the banner to be shown:
Screenshot 2026-06-02 at 10 56 54 AM

nickgros (approved):

PR #2878, opened 2026-06-04
PORTALS-4166: ADKP Homepage Redesign: Related consortia and repositories component

kianamcc:
Jira: https://sagebionetworks.jira.com/browse/PORTALS-4166

I have not uploaded all the SVGs but it would look like this:
Related Consortia and Repositories

Confirming other similar cards are unaffected:
OUR PARTNERS

Screenshot 2026-06-04 at 11 16 34 AM


packages/synapse-react-client/src/components/GenericCard/GenericCard.tsx
```diff
@@ -258,6 +260,9 @@ export const GenericCard = forwardRef(function GenericCard(
                 {titleAreaRightContent}
               </div>
             )}
+            {ctaLinkPosition === 'right' && ctaLinkConfig && (
+              <div>{ctaLinks}</div>
```
jay-hodgson: It looks like the positioning in the figma for this is always "pinned" to the right .  Is additional layout needed for this case?

Image

kianamcc: Screenshot 2026-06-04 at 4 13 23 PM

Screenshot 2026-06-04 at 4 15 33 PM

apps/portals/adknowledgeportal/src/pages/HomePageV2.module.scss
```diff
@@ -34,8 +34,10 @@
     border: 1px solid var(--UI-Gray-400, #dfe2e6);
     background: var(--Colors-white, #fff);
     box-shadow: 0 4px 8px 0 rgba(0, 0, 0, 0.1);
+    padding: 15px;
     h3 {
       color: var(--adkp-accent-color);
+      margin-top: 0 !important;
```
kianamcc, line 40: The base .SRC-portalCard h3 rule uses !important on margin so using !important here to override

jay-hodgson (approved):

PR #2815, opened 2026-05-20
PORTALS-4239 - NF Home page hydrated data components

nickgros:
- Refactor components featured on NF Portal home page to support prerendering with data

UserCardList:
  - Updated to use hooks for fetching user profiles.
  - Replaced class component with functional component and React hooks.
  - Added prefetch function for UserCardListRotate to support prerendering with data.
  - Removed sessionStorage cache (had little benefit with react-query cache in use)

Goals
  - Added associated prefetch function to support prerendering with data

CardContainerLogic
  - Added associated prefetch function to support prerendering with data
  - Remove suspense and blocking loaders from various child components that prevented prerendering
  - Enhanced session initialization handling in SessionInitializedGuard for better hydration stability.
 
RssFeedCards
  - Added associated prefetch function to support prerendering with data
  - Refactor to function component
  - Refactor data fetching to use react-query

For all affected components
- Added tests for changed components to verify prefetching functionality.

These changes support displaying portal data and relevant information on the pre-rendered home page, which will improve SEO and allow crawlers without JavaScript support to collect more information about the NF portal.

Full NF portal page with JavaScript disabled after these changes:

image

PR #2880, opened 2026-06-04
PORTALS-4239 - SSR-compatible Markdown component

nickgros:
Make MarkdownSynapse SSR-compatible by removing usage of DOMParser. DOMParser was originally only used to fix invalid HTML, but by introducing html-react-parser we could simplify that logic and also replace our custom parsing logic.

-----

When trying to get the NF portal home page to render under SSG (#2815), I ran into an issue where an error would be thrown by the MarkdownSynapse component because it used a browser API that was not available in node.

To fix it, I brought in a new dependency, html-react-parser. In addition to fixing the immediate problem of relying on browser-only APIs, this library also replaces our recursive HTML-to-React logic with its own.

### DIFF
--- a/packages/synapse-react-client/src/components/Markdown/MarkdownSynapse.tsx
+++ b/packages/synapse-react-client/src/components/Markdown/MarkdownSynapse.tsx
@@ -26,7 +26,15 @@ import markdownitSynapsePlugin from 'markdown-it-synapse'
 import markdownitSynapseHeading from 'markdown-it-synapse-heading'
 import markdownitMath from 'markdown-it-synapse-math'
 import markdownitSynapseTable from 'markdown-it-synapse-table'
-import { JSX, useEffect, useMemo, useRef } from 'react'
+import { ComponentProps, ElementType, useEffect, useMemo, useRef } from 'react'
+import {
+  htmlToDOM,
+  domToReact,
+  attributesToProps,
+  Element as DomElement,
+  type DOMNode,
+  type HTMLReactParserOptions,
+} from 'html-react-parser'
 import { ErrorBanner } from '../error/ErrorBanner'
 import { SkeletonTable } from '../Skeleton'
 import MarkdownWidget from './MarkdownWidget'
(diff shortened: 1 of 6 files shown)


packages/synapse-react-client/src/components/Markdown/MarkdownSynapse.tsx
```diff
@@ -160,6 +168,105 @@ function MapParamsToMarkdownWidget(props: {
...
+ * through to html-react-parser's default conversion, which maps attributes
+ * (e.g. `class` -> `className`, inline `style` string -> object), handles void
+ * elements, and assigns React keys to sibling elements.
+ *
+ * @param markdown The original (post-processing) markup string, forwarded to
+ * widgets that need the full markup (e.g. table of contents).
+ */
+function getParserOptions(markdown: string): HTMLReactParserOptions {
+  const options: HTMLReactParserOptions = {
+    replace: domNode => {
```
nickgros, line 186: Before, we had a `RecursiveRender` component that walked the DOM and created React components matching the DOM, replacing certain DOM nodes with alternative React components when appropriate.

Now, `html-react-parser` creates the React tree for us. We can hook into it and do our component replacement (including injecting widgets) with this `replace` function.

packages/synapse-react-client/src/components/Markdown/MarkdownSynapse.tsx
```diff
@@ -185,20 +292,23 @@ function RenderMarkdown(props: {
...
 
-  const processedDocuments = useMemo(() => {
+  // Parse the markup into React elements. This uses html-react-parser, which
+  // is isomorphic (native DOMParser in the browser, htmlparser2 during
+  // SSR/SSG prerender) and produces `domhandler` nodes in both environments,
+  // so there is no dependency on `DOMParser`/`Node`/`HTMLElement` globals.
+  const content = useMemo(() => {
     if (markup.length > 0) {
-      const domParser = new DOMParser()
-      const document = domParser.parseFromString(markup, 'text/html')
```
nickgros, line 191: This was the main issue. `DOMParser` is not available in Node (e.g. during SSR/SSG), so when we tried to SSG a component that used `MarkdownSynapse`, an error was thrown.

packages/synapse-react-client/src/components/Markdown/MarkdownSynapse.tsx
```diff
@@ -212,145 +321,6 @@ function RenderMarkdown(props: {
...
 
-/**
- * recursiveRender will render react tree from HTML tree
- *
- * @param {Node} element This will be either a text Node or an HTMLElement
- * @param {string} markdown The original markdown, its kept as a special case for the table of contents widget
- * @returns {*}
- */
-
-function RecursiveRender(props: { element: Node; markdown: string }) {
```
nickgros, line 223: html-react-parser now handles the recursion--we just provide a transform so we render our React component variants and widgets.

packages/synapse-react-client/src/components/Markdown/MarkdownUtils.ts
```diff
@@ -300,33 +333,19 @@ export function fixInvalidNesting(
...
-      // replace the ancestor with the new div
-      if (ancestor.parentNode) {
-        ancestor.parentNode.replaceChild(div, ancestor)
-      }
-
-      ancestors[i] = div // Update our local stack to reference to the new div
+      // Demote the ancestor to a <div> in place. Renaming preserves the
+      // existing attributes (`attribs`) and children, so no node surgery is
+      // required.
+      ancestor.name = 'div'
```
nickgros, line 339: The html-react-parser API emits a mutable tree that has not been converted to real DOM objects yet, so this transform step simplifies greatly

jay-hodgson (approved):

PR #2887, opened 2026-06-08
SWC-7781 Delete Curation Task UI

afwillia:
Add a button to delete a curation task from the Edit dialogue on the Tasks and Action page.
ss
ss1

### DIFF
--- a/packages/synapse-react-client/src/features/entity/metadata-task/components/CreateOrUpdateCurationTaskDialog.test.tsx
+++ b/packages/synapse-react-client/src/features/entity/metadata-task/components/CreateOrUpdateCurationTaskDialog.test.tsx
@@ -2,6 +2,7 @@ import { render, screen, waitFor } from '@testing-library/react'
 import userEvent from '@testing-library/user-event'
 import {
   useCreateCurationTask,
+  useDeleteCurationTask,
   useUpdateCurationTask,
 } from '@/synapse-queries/curation/task/useCurationTask'
 import {
@@ -14,6 +15,7 @@ import { EntityIdTextFieldProps } from '@/components/EntityFinder/EntityIdTextFi
 import CreateOrUpdateCurationTaskDialog from './CreateOrUpdateCurationTaskDialog'
 import {
   AUTH_MODE_SOURCE_BENEFACTOR_TITLE,
+  DELETE_CURATION_TASK_DIALOG_TITLE,
   FILE_BASED_TASK_TITLE,
   RECORD_BASED_TASK_TITLE,
 } from '../utils/constants'
@@ -21,6 +23,7 @@ import {
 vi.mock('@/synapse-queries/curation/task/useCurationTask', () => ({
   useCreateCurationTask: vi.fn(),
   useUpdateCurationTask: vi.fn(),
+  useDeleteCurationTask: vi.fn(),
 }))
 
 vi.mock('@/components/UserSearchBox/UserSearchBox', () => ({
@@ -54,9 +57,13 @@ vi.mock('@/components/EntityFinder/EntityIdTextField', () => ({
 
 const mockUseCreateCurationTask = vi.mocked(useCreateCurationTask)
 const mockUseUpdateCurationTask = vi.mocked(useUpdateCurationTask)
+const mockUseDeleteCurationTask = vi.mocked(useDeleteCurationTask)
 
 const mockCreateMutate = vi.fn()
 const mockUpdateMutate = vi.fn()
+const mockDeleteMutate = vi.fn()
+// Capture the options passed to useDeleteCurationTask so tests can invoke onSuccess
+let lastDeleteOptions: Parameters<typeof useDeleteCurationTask>[0] | undefined
 
 const defaultProps = {
   open: true,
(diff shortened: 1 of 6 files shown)


packages/synapse-react-client/src/features/entity/metadata-task/components/CreateOrUpdateCurationTaskDialog.tsx
```diff
@@ -609,18 +640,65 @@ export default function CreateOrUpdateCurationTaskDialog(
...
+                </Button>
+              )}
+            </Box>
+            <Stack direction="row" gap={1}>
+              {dialogActions}
+            </Stack>
+          </Stack>
+        }
+      />
+      {canDelete && (
```
nickgros: nit: This `canDelete` is not really necessary, you can't get into this state (where `open` is true) if you don't have that permission

packages/synapse-react-client/src/features/entity/metadata-task/components/CreateOrUpdateCurationTaskDialog.tsx
```diff
@@ -609,18 +640,65 @@ export default function CreateOrUpdateCurationTaskDialog(
...
+        actions={
+          <Stack
+            direction="row"
+            gap={1}
+            justifyContent="space-between"
+            alignItems="center"
+            width="100%"
+          >
+            <Box>
+              {canDelete && step === 'COMMON_MUTABLE_FIELDS' && (
```
nickgros: nit: I don't think checking the `step` is necessary; it's the only step we allow right now if you have delete permission. If we let you go to the other steps, I don't see why we would want to block 'delete' on those either.

```suggestion
              {canDelete && (
```

packages/synapse-react-client/src/features/entity/metadata-task/components/CreateOrUpdateCurationTaskDialog.tsx
```diff
@@ -105,8 +118,11 @@ type CreateOrUpdateCurationTaskDialogStep =
 export default function CreateOrUpdateCurationTaskDialog(
   props: CreateOrUpdateCurationTaskDialogProps,
 ) {
-  const { open, onCancel, onSuccess, projectId, task } = props
+  const { open, onCancel, onSuccess, onDeleteSuccess, projectId, task } = props
```
nickgros: nit: If this is not required (which seems reasonable, the create flow will never support delete), then I prefer to give it a default constant value so we don't need the `?.()` syntax later:

```ts
import noop from 'lodash-es/noop'
```

```suggestion
  const { open, onCancel, onSuccess, onDeleteSuccess = noop, projectId, task } = props
```

packages/synapse-react-client/src/features/entity/metadata-task/components/CreateOrUpdateCurationTaskDialog.tsx
```diff
@@ -195,7 +211,22 @@ export default function CreateOrUpdateCurationTaskDialog(
     },
   })
 
-  const isPending = isCreatePending || isUpdatePending
+  const { mutate: deleteTask, isPending: isDeletePending } =
+    useDeleteCurationTask({
+      onSuccess: () => {
+        displayToast(DELETE_CURATION_TASK_SUCCESS_TOAST, 'success')
+        setShowDeleteConfirmation(false)
+        onDeleteSuccess?.()
```
nickgros: See above comment

```suggestion
        onDeleteSuccess()
```

packages/synapse-react-client/src/features/entity/metadata-task/components/CreateOrUpdateCurationTaskDialog.test.tsx
```diff
@@ -340,5 +364,64 @@ describe('CreateOrUpdateCurationTaskDialog', () => {
...
+
+    describe('Delete', () => {
+      it('does not render the Delete button when onDeleteSuccess is not provided', () => {
+        renderEditDialog(fileBasedTask)
+        expect(
+          screen.queryByRole('button', { name: /^delete$/i }),
+        ).not.toBeInTheDocument()
+      })
+
+      it('renders the Delete button when onDeleteSuccess is provided', () => {
```
nickgros, line 385: Can you add a test to verify the button is not shown when the prop is provided but `canDelete` is false?

packages/synapse-react-client/src/features/entity/metadata-task/components/CreateOrUpdateCurationTaskDialog.tsx
```diff
@@ -105,8 +118,11 @@ type CreateOrUpdateCurationTaskDialogStep =
 export default function CreateOrUpdateCurationTaskDialog(
   props: CreateOrUpdateCurationTaskDialogProps,
 ) {
-  const { open, onCancel, onSuccess, projectId, task } = props
+  const { open, onCancel, onSuccess, onDeleteSuccess, projectId, task } = props
   const isEditMode = task !== undefined
+  const canDelete = isEditMode && task?.taskId != null && !!onDeleteSuccess
```
nickgros: Should probably check `canDelete` in the project permissions (see `MetadataTasksPage` for example usage of `useGetEntityPermissions`)

nickgros (approved):

PR #2889, opened 2026-06-08
PORTALS-4279: Update CLASSIC to use the study column. id has been rem…

jay-hodgson:
…oved, and studyName is now linked to the study details page

### DIFF
--- a/apps/portals/classic/src/config/synapseConfigs/instruments.ts
+++ b/apps/portals/classic/src/config/synapseConfigs/instruments.ts
@@ -1,14 +1,26 @@
 import type { QueryWrapperPlotNavProps } from 'synapse-react-client/components/QueryWrapperPlotNav/QueryWrapperPlotNav'
+import { TargetEnum } from 'synapse-react-client/utils/html/TargetEnum'
 import { instrumentsSql } from '../resources'
 
 const instrumentsPlotNavProps: QueryWrapperPlotNavProps = {
   sql: instrumentsSql,
   //   name: 'Instruments',
   fileIdColumnName: 'id',
+  columnAliases: { study: 'On Synapse' },
   tableConfiguration: {
     showAccessColumn: true,
     showDownloadColumn: false,
     hideAddToDownloadListColumn: true,
+    columnLinks: [
+      {
+        matchColumnName: 'studyName',
+        isMarkdown: false,
+        overrideLinkURLColumnName: 'study',
+        overrideLinkURLColumnTransform: study =>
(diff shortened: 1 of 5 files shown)


apps/portals/classic/src/config/synapseConfigs/instruments.ts
```diff
@@ -1,14 +1,26 @@
...
+    columnLinks: [
+      {
+        matchColumnName: 'studyName',
+        isMarkdown: false,
+        overrideLinkURLColumnName: 'study',
+        overrideLinkURLColumnTransform: study =>
+          `/Explore/Studies/DetailsPage?study=${encodeURIComponent(study)}`,
+        target: TargetEnum.CURRENT_WINDOW,
+      },
+    ],
```
nickgros, line 23: Same comment as on #2890 
```suggestion
      {
        matchColumnName: 'studies',
        isMarkdown: false,
        overrideLinkURLColumnName: 'study',
        baseURL: `/Explore/Studies/DetailsPage`
      },
```

apps/portals/classic/src/config/synapseConfigs/variables.ts
```diff
@@ -1,14 +1,26 @@
...
     hideAddToDownloadListColumn: true,
+    columnLinks: [
+      {
+        matchColumnName: 'studyName',
+        isMarkdown: false,
+        overrideLinkURLColumnName: 'study',
+        overrideLinkURLColumnTransform: (study: string) =>
+          `/Explore/Studies/DetailsPage?study=${encodeURIComponent(study)}`,
+        target: TargetEnum.CURRENT_WINDOW,
+      },
```
nickgros, line 22: ditto

nickgros (approved):

apps/portals/classic/src/config/synapseConfigs/instruments.ts
jay-hodgson: Your example matches on the `studies` column.
`overrideLinkURLColumnName` tells us to use that column's raw value as the entire href, but `study` does not contain the actual link in this case.
In theory we could use the `baseURL`+ `URLColumnName` combo, but this would result in `/<baseURL>?<URLColumnName>=<matchColumnName value>`, which is also not what we want.  
I chose the configuration that is most flexible, because we don't have another way to accomplish it today

nickgros: Got it, study name and study ID are different columns in the source table, so we need to consider both columns to create a link to the target table. Not possible with the `baseURL` link style

PR #2890, opened 2026-06-08
PORTALS-4280

jay-hodgson:
Add Metadata explore section to the Classic portal.  Also fix a small bug

### DIFF
--- a/apps/portals/classic/src/config/synapseConfigs/metadata.ts
+++ b/apps/portals/classic/src/config/synapseConfigs/metadata.ts
@@ -1,0 +1,27 @@
+import type { QueryWrapperPlotNavProps } from 'synapse-react-client/components/QueryWrapperPlotNav/QueryWrapperPlotNav'
+import { TargetEnum } from 'synapse-react-client/utils/html/TargetEnum'
+import { metadataSql } from '../resources'
+
+const metadataPlotNavProps: QueryWrapperPlotNavProps = {
+  sql: metadataSql,
+  name: 'Metadata',
+  shouldDeepLink: true,
+  columnAliases: { study: 'On Synapse' },
+  tableConfiguration: {
+    showAccessColumn: true,
+    showDownloadColumn: false,
+    hideAddToDownloadListColumn: true,
+    columnLinks: [
+      {
+        matchColumnName: 'studyName',
+        isMarkdown: false,
+        overrideLinkURLColumnName: 'study',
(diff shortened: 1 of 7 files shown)


packages/synapse-react-client/src/components/GenericCard/SynapseCardLabel.tsx
```diff
@@ -275,7 +275,7 @@ export function SynapseCardLabel(props: SynapseCardLabelProps) {
                 selectColumns,
                 columnModels,
               )
-              if (overrideHrefIndex) {
+              if (overrideHrefIndex != null) {
```
jay-hodgson, line 278: `study` is the first column returned by the query, so its index is 0. JavaScript treats 0 as falsy

nickgros: Created https://sagebionetworks.jira.com/browse/PORTALS-4281 to enable a lint rule to prevent this class of bug

apps/portals/classic/src/config/synapseConfigs/metadata.ts
```diff
@@ -0,0 +1,27 @@
...
+    hideAddToDownloadListColumn: true,
+    columnLinks: [
+      {
+        matchColumnName: 'studyName',
+        isMarkdown: false,
+        overrideLinkURLColumnName: 'study',
+        overrideLinkURLColumnTransform: (study: string) =>
+          `/Explore/Studies/DetailsPage?study=${encodeURIComponent(study)}`,
+        target: TargetEnum.CURRENT_WINDOW,
+      },
```
nickgros, line 22: Isn't this the more conventional style? And would ensure the SPA doesn't reload the page (maybe the overrideLinkURLColumnTransform is sophisticated already, but I am not sure)

```suggestion
      {
        matchColumnName: 'studies',
        isMarkdown: false,
        URLColumnName: 'study',
        baseURL: `/Explore/Studies/DetailsPage`
      },
```

nickgros (approved):

apps/portals/classic/src/config/synapseConfigs/metadata.ts
jay-hodgson: See [this comment](https://github.com/Sage-Bionetworks/synapse-web-monorepo/pull/2889#discussion_r3377340390).  
There might be a SPA reload issue, that's another good point

PR #2891, opened 2026-06-09
Add Query Builder prototype

jay-hodgson:
Screenshot 2026-06-08 at 6 27 23 PM

This functionality will likely be incompatible with the SearchIndex implementation


jay-hodgson:
UX is ok, but manipulating the sql will be fragile.  Opened https://sagebionetworks.jira.com/browse/PLFM-9746

jay-hodgson:
This PR should never be merged, but I'd like to keep it as a draft until we implement something similar (for cohort builder v2)

PR #2894, opened 2026-06-11
Upgrade dependencies, replace universal-cookie with js-cookie

nickgros:
We are widely using `@react-hookz/web`, which has an integration with js-cookie. Opt to use it and replace all usages of universal-cookie with `js-cookie`. Drop `react-cookie`, which was not used.


apps/portals/cancercomplexity/package.json
```diff
@@ -4,33 +4,32 @@
...
     "@sage-bionetworks/synapse-portal-framework": "workspace:*",
     "@sage-bionetworks/synapse-types": "workspace:*",
-    "@tanstack/react-query": "^5.80.7",
-    "isbot": "^5",
+    "@tanstack/react-query": "^5.101.0",
+    "isbot": "^5.1.42",
     "jsdom": "^28.1.0",
-    "katex": "^0.16.22",
-    "react": "^19.2.5",
-    "react-cookie": "^7.2.2",
```
nickgros, line 16: Drop react-cookie. It was not used

apps/SageAccountWeb/src/pages/AccountSettings.tsx
```diff
@@ -19,11 +19,10 @@ import {
   VerificationState,
   VerificationStateEnum,
 } from '@sage-bionetworks/synapse-types'
-import React from 'react'
+import React, { useCallback } from 'react'
 import { RefObject, useEffect, useRef, useState } from 'react'
 import { Link as RouterLink, useLocation, useNavigate } from 'react-router'
 import RORInstitutionField from 'synapse-react-client/components/RORInstitutionField/RORInstitutionField'
-import UniversalCookies from 'universal-cookie'
```
nickgros, line 26: `@react-hookz/web` (already installed) gives us a hook interface for cookies if we install `js-cookie`. Let's use it and drop `universal-cookie`.

apps/synapse-portal-framework/src/ssg/createReactRouterConfig.ts
```diff
@@ -100,7 +100,7 @@ export function createReactRouterConfig(
 
         return [...staticPaths, ...dynamicRoutes]
       },
-      unstable_concurrency: concurrency,
+      concurrency,
```
nickgros, line 103: Some react-router APIs are now stable

apps/synapse-portal-framework/src/ssg/PortalRoot.tsx
```diff
@@ -54,16 +53,14 @@ export default function PortalRoot(props: PortalRootProps) {
 
   return (
     <PortalContextProvider value={portalContext}>
-      <CookiesProvider>
```
nickgros, line 57: CookiesProvider was added here but was not actually used

packages/synapse-react-client/src/components/dataaccess/SubmissionPage/SubmissionPage.test.tsx
```diff
@@ -55,7 +55,6 @@ import {
   within,
 } from '@testing-library/react'
 import userEvent from '@testing-library/user-event'
-import failOnConsoleError from 'jest-fail-on-console'
```
nickgros, line 58: jest-fail-on-console started causing errors after something was upgraded. There is a vitest alternative, but it does not seem to be well-maintained. I think we can just remove it.

apps/SageAccountWeb/src/pages/AccountSettings.tsx
```diff
@@ -104,25 +104,32 @@ const AccountSettings = (): React.ReactNode => {
...
-      expires: nextYear,
-      domain: hostname.endsWith('.synapse.org') ? 'synapse.org' : undefined,
-    })
-  }, [isUTCTime])
+      domain:
+        typeof window !== 'undefined' &&
+        window.location.hostname.toLowerCase().endsWith('.synapse.org')
+          ? 'synapse.org'
+          : undefined,
+      expires: 365,
```
jay-hodgson, line 121: useCookieValue expires accepts a number, and interprets as days?  I expected it would take a js Date

jay-hodgson (approved):

apps/SageAccountWeb/src/pages/AccountSettings.tsx
nickgros: It uses `js-cookie`: https://github.com/js-cookie/js-cookie/blob/13e4ac8d7451588a9056868e9e4cee65d8737f22/README.md#cookie-attributes

`expires` can be a number (days) or a `Date`. It more simple to express these with a number than to do the date math

PR #2895, opened 2026-06-12
PORTALS-4285: Add Sankey plot with very a specific configuration (roo…

jay-hodgson:
…t, broken into x rows, each with an associated count value), driven by a Synapse query

Storybook: 
Screenshot 2026-06-11 at 4 52 57 PM

Screenshot 2026-06-11 at 4 53 06 PM

ALS KP:
Screenshot 2026-06-11 at 5 02 59 PM

### DIFF
--- a/packages/synapse-react-client/src/components/Plot/Plot.tsx
+++ b/packages/synapse-react-client/src/components/Plot/Plot.tsx
@@ -2,12 +2,16 @@ import { PlotParams } from 'react-plotly.js'
 import createPlotlyComponent from './safe-react-plotly-factory'
 import { lazy, Suspense } from 'react'
 
-// Lazy-load plotly.js since it is very large!
-const safeESModule = <T,>(a: T | { default: T }): T => {
+// We want to bundle browser-compatible code, but use ESM (`"type": "module"` in package.json) when we use Node (e.g. tests).
+// esbuild (used by Vite in development mode) has a quirk where this will cause packages react-plotly.js/factory to be
+// imported in 'node mode', which breaks imports in the browser.
+// For further explanation and source of workaround, see: https://github.com/evanw/esbuild/issues/2480#issuecomment-1970337003
+export const safeESModule = <T,>(a: T | { default: T }): T => {
   const b = a as any
   return b.__esModule || b[Symbol.toStringTag] === 'Module' ? b.default : b
 }
 
+// Lazy-load plotly.js-basic-dist (scatter, bar, pie, scattergl traces).
 const LazyLoadedPlot = lazy(async () => {
   const Plotly = await import('plotly.js-basic-dist')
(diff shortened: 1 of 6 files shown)


packages/synapse-react-client/src/components/Plot/SynapseSankeyPlot.tsx
```diff
@@ -0,0 +1,121 @@
...
+import type { Data, Layout, SankeyData } from 'plotly.js'
+import { PlotParams } from 'react-plotly.js'
+import createPlotlyComponent from './safe-react-plotly-factory'
+
+// Lazy-load the full plotly.js bundle (not the basic dist) since Sankey is not
+// included in plotly.js-basic-dist.
+const safeESModule = <T,>(a: T | { default: T }): T => {
+  const b = a as any
+  return b.__esModule || b[Symbol.toStringTag] === 'Module' ? b.default : b
+}
```
nickgros: `safeESModule` is duplicated from Plot.tsx.

Maybe Plot.tsx should export both `Plot` and `PlotFullBundle` (maybe you can think of a better name than I can)?

packages/synapse-react-client/src/components/Plot/SynapseSankeyPlot.tsx
```diff
@@ -0,0 +1,121 @@
...
+      },
+    },
+  ]
+
+  const layout: Partial<Layout> = {
+    title,
+  }
+
+  return (
+    <SankeyPlot
```
nickgros:
```suggestion
    <PlotFullBundle
```

or a better name you think of

nickgros (approved):
Just some minor cleanup before merging, LGTM othewise

packages/synapse-react-client/src/components/Plot/Plot.tsx
```diff
@@ -2,20 +2,25 @@ import { PlotParams } from 'react-plotly.js'
 import createPlotlyComponent from './safe-react-plotly-factory'
 import { lazy, Suspense } from 'react'
 
-// Lazy-load plotly.js since it is very large!
-const safeESModule = <T,>(a: T | { default: T }): T => {
+// We want to bundle browser-compatible code, but use ESM (`"type": "module"` in package.json) when we use Node (e.g. tests).
+// esbuild (used by Vite in development mode) has a quirk where this will cause packages react-plotly.js/factory to be
+// imported in 'node mode', which breaks imports in the browser.
+// For further explanation and source of workaround, see: https://github.com/evanw/esbuild/issues/2480#issuecomment-1970337003
```
nickgros: I'm not sure about this comment, since Vite does not use esbuild as of v8. Still might need the workaround though for rolldown (the bundler that replaced esbuild) if it has the same quirk

nickgros (approved):

PR #2896, opened 2026-06-12
SWC-7835 - Fix Task modal requiring two clicks to to open

nickgros:
Fix bug where global editing state 'edit' button remounted before dialog could open

### DIFF
--- a/packages/synapse-react-client/src/features/entity/metadata-task/hooks/useMetadataTaskTable.tsx
+++ b/packages/synapse-react-client/src/features/entity/metadata-task/hooks/useMetadataTaskTable.tsx
@@ -109,9 +109,11 @@ export function useMetadataTaskTable(opts: UseMetadataTaskTableOptions) {
     [data],
   )
 
+  const columns = useMemo(() => getColumns(canEditTasks), [canEditTasks])
+
   const table: Table<TaskBundle> = useReactTable<TaskBundle>({
     data: tasks,
-    columns: getColumns(canEditTasks),
+    columns,
     getRowId: row => String(row.task!.taskId!),
     getCoreRowModel: getCoreRowModel(),
     columnResizeMode: 'onChange',
--- a/packages/synapse-react-client/src/testutils/IsEditingStoreTestUtils.ts
+++ b/packages/synapse-react-client/src/testutils/IsEditingStoreTestUtils.ts
@@ -1,0 +1,32 @@
+import { IsEditingStore } from '@/utils/context/GlobalIsEditingContext'
+
+/**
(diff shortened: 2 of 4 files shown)


packages/synapse-react-client/src/testutils/IsEditingStoreTestUtils.ts
```diff
@@ -0,0 +1,32 @@
...
+/**
+ * Creates a functional, in-memory {@link IsEditingStore} for use in tests.
+ *
+ * This store holds real state and **notifies its subscribers when the value
+ * changes** — mirroring the behavior of the GlobalApplicationState-backed
+ * store used in the Synapse Web Client (SWC).
+ *
+ * This is essential for reproducing editing-state-driven re-renders in tests.
+ */
+export function createTestIsEditingStore(initialValue = false): IsEditingStore {
```
nickgros, line 12: One of the conditions that is needed to reproduce the bug is that toggling the editing state needs to actually trigger a re-render. The store used by our test wrapper did not do this, so add a simple state toggle utility that does.

packages/synapse-react-client/src/testutils/TestingLibraryUtils.tsx
```diff
@@ -32,6 +33,7 @@ export const createWrapperAndQueryClient = (
           synapseContext={wrapperProps}
           queryClient={queryClient}
           applicationSessionContext={MOCK_APPLICATION_SESSION_CONTEXT}
+          isEditingStore={createTestIsEditingStore()}
```
nickgros, line 36: Use the new test store in all of our SRC component tests that use context.

packages/synapse-react-client/src/features/entity/metadata-task/hooks/useMetadataTaskTable.tsx
```diff
@@ -109,9 +109,11 @@ export function useMetadataTaskTable(opts: UseMetadataTaskTableOptions) {
     [data],
   )
 
+  const columns = useMemo(() => getColumns(canEditTasks), [canEditTasks])
```
nickgros, line 112: Memoizing the columns fixes the bug. The issue can be summarized as 
 
1. The first Edit click sets isDialogOpen=true locally, then flips the global isEditing flag.
2. That flag change notifies the useSyncExternalStore subscriber, re-rendering the editing-context provider and thus MetadataTasksPage. 
3. Because useMetadataTaskTable rebuilds the columns array, every row's cell gets a new component identity and remounts — throwing away the instance that just set isDialogOpen=true.
 
Memoizing the columns keeps the cell's identity stable, so the instance survives the re-render and the dialog opens on the first click.

afwillia (approved):

PR #2897, opened 2026-06-12
SWC-7780 - disable adding new row with enter key

afwillia:
set DataGrid autoAddRow prop to false to disable adding new rows with the enter key


jay-hodgson (approved):

jay-hodgson:
Ok!

PR #2905, opened 2026-06-15
SWC-7854 grid column updates

afwillia:
- remove column help icon and replace the column title tooltip with the schema description
- change background of header cells to gray
- required columns have a red asterisk
columns

### DIFF
--- a/packages/synapse-react-client/src/components/DataGrid/components/ColumnHeaderWithTooltip.test.tsx
+++ b/packages/synapse-react-client/src/components/DataGrid/components/ColumnHeaderWithTooltip.test.tsx
@@ -1,44 +1,42 @@
 import { render, screen } from '@testing-library/react'
-import userEvent from '@testing-library/user-event'
 import { describe, expect, it } from 'vitest'
 import { ColumnHeaderWithTooltip } from './ColumnHeaderWithTooltip'
 
 describe('ColumnHeaderWithTooltip', () => {
-  it('renders column name without tooltip when no description', () => {
+  it('renders column name', () => {
     render(<ColumnHeaderWithTooltip name="myColumn" />)
-
     expect(screen.getByText('myColumn')).toBeInTheDocument()
   })
 
-  it('renders column name with help icon when description is provided', async () => {
-    const user = userEvent.setup()
+  it('uses column name as native title when no description', () => {
+    render(<ColumnHeaderWithTooltip name="myColumn" />)
+    expect(screen.getByTitle('myColumn')).toBeInTheDocument()
+  })
 
+  it('uses description as native title when description is provided', () => {
     render(
       <ColumnHeaderWithTooltip
         name="myColumn"
         description="This is a helpful description"
       />,
     )
-
-    const columnHeader = screen.getByText('myColumn')
-    expect(columnHeader).toBeInTheDocument()
-
-    // Help icon should be present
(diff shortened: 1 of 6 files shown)


packages/synapse-react-client/src/components/DataGrid/components/ColumnHeaderWithTooltip.test.tsx
```diff
@@ -1,44 +1,42 @@
...
   })
 
-  it('renders only the name when description is empty string', () => {
+  it('falls back to column name as title when description is empty string', () => {
     render(<ColumnHeaderWithTooltip name="myColumn" description="" />)
+    expect(screen.getByTitle('myColumn')).toBeInTheDocument()
+  })
 
-    expect(screen.getByText('myColumn')).toBeInTheDocument()
+  it('does not render a help icon', () => {
```
nickgros: This would be useful if there was conditional logic used to hide the icon, but personally I don't think we need tests that verify UI was totally removed

packages/synapse-react-client/src/components/DataGrid/hooks/useColumnResizeHandles.ts
```diff
@@ -231,14 +231,18 @@ class ColumnResizeManager {
 
       const cellElement = cell as HTMLElement
 
-      // Get the column name from the cell's text content
-      // This is more reliable than using DOM index with virtualization
-      const cellText = cellElement.textContent?.trim() || ''
+      // Prefer the data-column-id attribute set by ColumnHeaderWithTooltip over
+      // textContent, which can include extra characters (e.g. the required "*").
+      const columnId =
```
nickgros, line 236: Nice robust improvement. It might be worth considering if we should generate IDs for each column instead of using the name, to ensure we are always operating with a safe, reliable identifier.

packages/synapse-react-client/src/components/DataGrid/hooks/useColumnResizeHandles.test.ts
```diff
@@ -31,6 +31,41 @@ describe('useColumnResizeHandles', () => {
...
+
+    const textSpan = document.createElement('span')
+    textSpan.textContent = columnName
+    headerRoot.appendChild(textSpan)
+
+    if (isRequired) {
+      const asterisk = document.createElement('span')
+      asterisk.textContent = '*'
+      headerRoot.appendChild(asterisk)
+    }
```
nickgros: This test & its hook don't need to care about requiredness at all. Instead of baking this 'requiredness' logic into the test setup, I recommend changing the test and setup to simply verify that the hook works for a column that has a `data-column-id` that does not match its name.

nickgros (changes requested):
Core changes look good, just requesting a few test changes

packages/synapse-react-client/src/components/DataGrid/hooks/useColumnResizeHandles.ts
afwillia: Not sure if generating a column ID will provide much of an improvement:
> The one scenario where generated IDs would genuinely help is if column names could contain characters that break DOM APIs — but data-column-id is only ever compared as a plain string (=== columnId), not used in a selector, so that's not a concern here.

nickgros (approved):

PR #2918, opened 2026-06-17
Redesign AMP-ALS homepage Sankey (themed, interactive, butterfly d3-sankey)

allaway:
Replaces the generic Plotly Sankey on the AMP-ALS homepage with a custom, theme-driven, interactive `d3-sankey` visualization, and reworks it into a **butterfly** layout: **All Datasets → each Source → All Files**.

### `synapse-react-client` — `SynapseSankeyPlot`
- Drops the full `plotly.js` bundle for this chart; lays out with `d3-sankey` and renders custom SVG skinned to the portal's MUI theme (no hardcoded colors — adopts each portal's palette automatically).
- **Configurable / portal-agnostic:** `labelColumn` / `valueColumn` / `rightValueColumn` (resolved by header name or index), `rootLabel` / `rightLabel` / `unitLabel` / `rightUnitLabel` / `breakdownLabel`, and click callbacks (`onCategoryClick`, `onRightCategoryClick`, `onRootClick`, `onRightEndClick`). No table names, routes, or facet columns live in the component.
- **Butterfly mode** (opt-in via `rightValueColumn`): a second flow mirrors from each center node to a right end node. The two flows are normalized to a common total so a small-magnitude side (e.g. ~25 datasets vs thousands of files) doesn't collapse to hairlines — ribbon widths read as each side's share while labels keep the real counts.
- Focus/dim on hover, a brief "swell" before click navigation (with proper timer cleanup on unmount), and label legibility tuning (darker subtitles/eyebrows + a white text halo over ribbons).
- Hub-and-spoke (single-flow) mode is preserved as the default.

### `ampals` — `HomePage`
- Removes the GoalsV3 stat buttons; encodes that info in the Sankey instead.
- `sankeyPlotSql` selects `source, count(source), sum(datasetItemCount)`.
- Wires clicks: root → `/Explore/Datasets`, All Files → `/Explore/Files`, source → datasets filtered by source, right flow → files filtered by source (shared helper). Confirmed the Files view `syn66271104` has a `source` column.

### Testing
- New `SynapseSankeyPlot.test.tsx` (9 tests): render, column resolution by name/index, swell-then-fire click behavior, butterfly right-end totals + per-source counts, and empty results.
- Type-checks clean, lint clean, and verified live in the `ampals` dev server with real Synapse data.


jay-hodgson:
Pulling in a new dependency that was last updated 7 years ago :/
Are there any vulnerabilities that this introduces?
It is covered by the BSD-3-Clause, which is probably ok

jay-hodgson (approved):

jay-hodgson:
E2E test failures are due to SearchIndex build failures on production, which Bryan is investigating.  I love this sankey plot update, and don't see any security concerns (or licensing concerns).
The only minor issue that I'd note is around keyboard accessibility.  Plotly allows for keyboard navigation (although perhaps poor), where this new plot seems to have no support. 
In theory the plot could be reused, but it does seem to have a unique shape.
Approved!

allaway:
Thanks Jay! 
The plot component also supports the one-sided type of sankey plot that is currently on the portal today, too, so hopefully this will be useful elsewhere! 

Claude also warned me about keyboard accessibility too, I'll take a look into this

allaway:
I was working on the keyboard accessibility, but then decided to put that work in a different branch, and this now should be back to the state where it was when Jay approved.

jay-hodgson:
This fixes https://sagebionetworks.jira.com/browse/PORTALS-4303

PR #2922, opened 2026-06-18
PORTALS-4295

jay-hodgson:
### Problem

Previously, when a user selected facet values, all selections were placed directly into `query.bool.filter`. This caused every aggregation to be computed against the already-filtered dataset, so facet option counts reflected only the current selection rather than the full range of options. This is the "frozen facets" problem — selecting `Program = AMP-AD` would collapse all other facet counts to only documents matching that filter.

---

### Solution: OpenSearch Faceted Search Pattern

Implements the standard OpenSearch [faceted search pattern](https://docs.opensearch.org/latest/tutorials/faceted-search/#maintaining-facet-options-during-filtering) using `post_filter` and per-facet filter aggregations.

#### Request changes (`toSearchIndexQuery`)

**Before:**
- Facet selections → `query.bool.filter` (affected both hits and aggregations)
- Text search + facet selections → `query.bool.must` + `query.bool.filter`
- Aggregations → plain `terms` regardless of active selections

**After:**
- Facet selections → `post_filter` (narrows hits/total only, does **not** affect aggregation counts)
- Text search → `query` field only (`multi_match` or `match_all`); facets never touch the query
- Aggregations → **filter-wrapped** when selections are active:
  - Each enumeration column's `terms` agg is wrapped in a `filter` aggregation
  - The filter = AND of all **other** columns' active selections (own selection is dropped)
  - This keeps the selected column "wide" (all options remain visible with full counts) while unselected columns narrow to the current filter
  - No wrapping when there are no active selections (baseline remains plain `terms`)

**Example with two selections (`category=X`, `mature=Y`):**
```json
{
  "query": { "match_all": {} },
  "post_filter": { "bool": { "must": [
    { "terms": { "category": ["X"] } },
    { "terms": { "mature": ["Y"] } }
  ]}},
  "aggregations": {
    "category": {
      "filter": { "terms": { "mature": ["Y"] } },
      "aggregations": { "category": { "terms": { "field": "category" } } }
    },
    "mature": {
      "filter": { "terms": { "category": ["X"] } },
      "aggregations": { "mature": { "terms": { "field": "mature" } } }
    }
  }
}
```

#### Response parsing changes (`searchQueryResultsToQueryResultBundle`)

When filter-wrapping is active, OpenSearch nests the buckets under a `sterms#<columnName>` key (its internal aggregation type prefix) rather than directly in `buckets`. The parser now handles both shapes:

- **Plain terms** (no selections): `agg.buckets`
- **Filter-wrapped** (selections active): `agg["sterms#columnName"].buckets`

A defensive fallback to `agg[columnName].buckets` is also included.

-----

Showing the selection of a facet value, where the current facet is not filtered by that value (but all other facets, and the search hits, are filtered):
Screenshot 2026-06-18 at 9 00 08 AM


packages/synapse-react-client/src/components/SearchQueryWrapper/SearchQueryUseQueryOptions.ts
```diff
@@ -214,23 +211,72 @@ export function toSearchIndexQuery(
...
+  // (see aggregationEntries below) so each column stays wide for its own selection while
+  // all other columns narrow.
+  let postFilter: SelectionFilter | undefined
+  if (filterClauses.length === 1) {
+    postFilter = filterClauses[0]
+  } else if (filterClauses.length > 1) {
+    postFilter = { bool: { must: filterClauses } }
+  }
+
+  // Build aggregations. When selections are active, wrap each terms aggregation in an
```
nickgros:
```suggestion
  // Build aggregations. When selections are active, wrap each term's aggregation in an
```

packages/synapse-react-client/src/components/SearchQueryWrapper/SearchQueryUseQueryOptions.ts
```diff
@@ -315,18 +362,33 @@ export function searchQueryResultsToQueryResultBundle(
...
   > | null
   const rawFacets: FacetColumnResultValues[] = Object.entries(
     aggregationResults ?? {},
   ).map(([columnName, agg]) => {
+    // Resolve the actual terms result: OpenSearch prefixes the agg type when it is
+    // nested inside a filter aggregation, producing a "sterms#columnName" key.
+    // Fall back to a direct [columnName] match (defensive), then to the agg itself
+    // for the plain (no-selection) terms shape.
+    const aggRecord = agg as Record<string, OpenSearchTermsAgg>
+    const nested = aggRecord[`sterms#${columnName}`] ?? aggRecord[columnName]
```
nickgros, line 389: Is this always `sterms`? https://docs.opensearch.org/latest/api-reference/grpc-apis/search/#aggregate-fields

nickgros (approved):

packages/synapse-react-client/src/components/SearchQueryWrapper/SearchQueryUseQueryOptions.ts
jay-hodgson: The bucket keys are based on the column names, which are always Strings.  So I think it will always be `sterms`, right?

nickgros: My assumption was that it's based on values not the key (columnName). Best way to check would be with an enumerated integer column, like publication 'Year'.

jay-hodgson:
Tested locally before merging.  Noticed a SearchIndex build issue (and opened up https://sagebionetworks.jira.com/browse/PLFM-9764), but for SearchIndexes that built properly this seems to work as expected.

PR #2923, opened 2026-06-18
Fix AMP-ALS Sankey files deep-link to filter by contributor

allaway:
The AMP-ALS homepage Sankey's right-flow (files) click deep-linked to `/Explore/Files` filtered on the `source` facet, but the files table records the originating group as `contributor`. This parameterizes the deep-link helper's column so:

- center node / left flow → `/Explore/Datasets` filtered by `source`
- right flow → `/Explore/Files` filtered by `contributor`

Verified the files view (`syn66271104`) has a `contributor` column, and `ampals` type-checks.


allaway:
Note, the file counts are still not accurate in the files-explore queries as compared to what's available in the datasets, but they should be after we fix the `Contributor` annotation.

jay-hodgson (approved):

PR #2930, opened 2026-06-19
DUO data-use tags on NF dataset cards and facet (DESIGN-1740 / PORTALS-4282)

allaway:
> [!IMPORTANT]
> **Draft — contains mockup artifacts that MUST be removed before review/merge.**
> The DUO **mockup/demo page is exploration scaffolding only** and should be deleted before this is reviewed:
> - `apps/portals/nf/src/pages/DuoMockup/DuoMockup.tsx`
> - the `DUO Mockup` route in `apps/portals/nf/src/routes.ts`
>
> Everything else (the reusable component + the card/facet integration) is the intended change. Also pending: sign-off from the DUO owners and confirmation of whether this should ship to the live NF cards now or stay behind further review.

## What this does
Surfaces Data Use Ontology (DUO) terms in the NF Data Portal so users can see data-use conditions (especially commercial-use restrictions) at a glance.

**Component (`synapse-react-client`)**
- `DuoTermTags` + a DUO vocabulary sourced from [`Sage-Bionetworks/governanceDUO`](https://github.com/Sage-Bionetworks/governanceDUO): all 24 GA4GH DUO terms (incl. the three-way commercial family NPUNCU/NCU/NPU) plus the Sage DUO-Plus **Attribution** term. Keyed by ontology code; resolves from a code or a term name.
- EGA-style labelled tags: full term name, colored by category (green = permitted use, red = commercial restriction, blue = obligation, amber = use limit), each with a per-code icon and a tooltip (abbreviation, ontology code linked to Bioregistry, definition). Long names truncate with an ellipsis.

**Card integration**
- Optional `dataUseModifiersColumnName` on the generic card schema renders the tags as a **"Data Usage Restrictions"** row in the card metadata section (consistent with how DOI / croissant / scorecard are configured). Wired on the NF dataset card.

**Facet integration**
- New optional `renderFacetValue(columnName, value)` hook on the query-visualization config (threaded through `QueryVisualizationWrapper` and both `QueryWrapperPlotNav` / `SearchQueryWrapperPlotNav`). `EnumFacetFilter` uses it to render custom value nodes, falling back to the plain text label. Generic and backward-compatible (facet options now accept a `ReactNode` label; counts stay column-aligned when labels truncate). Wired so the **Data Use Modifiers** facet renders DUO tags.

## Notes
- The DUO term→code/category/icon mapping is hand-maintained for now; if adopted, it should be sourced from the ontology / governanceDUO programmatically.
- The annotation is the ontology code; the displayed tag is a separate lookup, so display can evolve independently of the annotation (aligns with the Croissant data-model work).
- Shared query components are touched but all additions are optional — no behavior change for portals/facets that don't opt in.


apps/portals/nf/src/config/synapseConfigs/datasets.ts
```diff
@@ -5,6 +5,14 @@ import { datasetsSearchIndexId, datasetsSql } from '../resources'
 import { columnAliases as sharedColumnAliases } from './commonProps'
 import { studyColumnIconConfigs } from './studies'
 import { SearchQueryWrapperPlotNavProps } from 'synapse-react-client/components/SearchQueryWrapperPlotNav/SearchQueryWrapperPlotNav'
+import { createElement } from 'react'
+import DuoTermTags from 'synapse-react-client/components/GenericCard/DuoTermTags/DuoTermTags'
+
+// Render the "Data Use Modifiers" facet values as DUO tags (PORTALS-4282).
+const renderDuoFacetValue = (columnName: string, value: string) =>
+  columnName === 'dataUseModifiers'
+    ? createElement(DuoTermTags, { terms: [value], truncate: true })
```
jay-hodgson: If we can stadardize duo term tag value rendering, then maybe this logic should not live in the portal config.  Instead, maybe the configuration would just set the dataUseModifiersColumnName and the facet values would be rendered appropriately

apps/portals/nf/src/pages/DuoMockup/DuoMockup.tsx
```diff
@@ -0,0 +1,156 @@
+import { useMemo } from 'react'
+import { Box, Container, Stack, Typography } from '@mui/material'
```
jay-hodgson: I see why this would be a useful demo.  But this could be a much more realistic test demo if we instead created a new Storybook story to demonstrate the different Duo values (via mock data responses)

packages/synapse-react-client/src/components/GenericCard/DuoTermTags/DuoTermTags.tsx
```diff
@@ -0,0 +1,189 @@
...
+
+const termIcon = (term: DuoTerm) => {
+  const Icon = DUO_ICONS[term.code] ?? HelpOutline
+  return <Icon />
+}
+
+// Color carries meaning: green = a use you may make, red = commercial use
+// prohibited, blue = an action the requester must take, amber = a limit on
+// who/where/when/what. Every term is colored.
+const chipSx = (category: DuoCategory) => {
```
jay-hodgson: I'm certain the Sage Design Team will have recommendations here

apps/portals/nf/src/config/synapseConfigs/datasets.ts
jay-hodgson: I do appreciate how generic this is, but it's a problem if all portal configurations need to define the same handler (if they want to render DUO facet values)

allaway:
**Open question — facet/pill chip sizing**

In the "results filtered by" (active-filter / selected-criteria) row, the DUO chips render with the MUI `small` chip size, which is very slightly shorter than the other (non-DUO) facet pills next to them (e.g. the plain `Aaron Kujawa` pill). It's subtle but noticeable when they sit side by side.

Screenshot 2026-06-19 at 2 57 08 PM

A few ways we could resolve it:

1. **Adopt this chip design for all facet/active-filter pills** — switch every pill to the `small` chip with the baked-in remove (×) button, so DUO and non-DUO pills are visually consistent. Most uniform, but the largest blast radius.
2. **Match the new DUO chips to the existing pill height** — keep the current pill styling for non-DUO values and just nudge the DUO chip's vertical dimension to line up. Smallest change, but leaves two slightly different pill implementations.
3. **Drop the custom DUO styling in these chips entirely** — keep only the value re-mapping so the codes render as human-readable names, and ignore the color/icon/delete-button treatment for DUO pills. Simplest, but loses the visual language we get on the cards/facet.


(cc @adamhindman)

apps/portals/nf/src/pages/DuoMockup/DuoMockup.tsx
allaway: @jay-hodgson thank you! yeah, this was just for my own experimentation/visual testing. I was intending to delete it. Instead, I can add it to the Storybook!

apps/portals/nf/src/config/synapseConfigs/datasets.ts
allaway: Thanks Jay, that makes sense, I wasn't sure how flexible to make it.  I am always pushing for more opinionated approaches in these scenarios, so I should walk the walk lol.

allaway:
Re: my previous comment, this is what i was thinking about for option 1:

Screenshot 2026-06-19 at 3 06 42 PM

packages/synapse-react-client/src/components/GenericCard/DuoTermTags/DuoTermTags.tsx
allaway: tagging @adamhindman

allaway:
### Review feedback incorporated

Pushed a round of changes addressing the feedback on this PR:

**1. Truncation scoped to the facet only** — DUO chips now show the **full term name on cards** (and in the active-filter pills); the fixed-width ellipsis truncation applies **only in the facet sidebar**, where space is tight. Full name remains available via the tooltip everywhere.

**2. DUO rendering extended to the active-filter pills** — the "results filtered by" (selected-criteria) row now renders DUO values with the same tag treatment as the cards and facet, via the generic `renderFacetValue` hook (non-DUO columns are unaffected).

**3. Active-filter pill reworked into a single deletable chip** — the first pass left the remove "✕" detached from the chip and the pill a different height than its neighbors. It's now one MUI deletable chip (✕ baked inside), tinted to the term's category.

**4. Resolved the chip-sizing open question — went with option 1** — every active-filter pill (DUO and plain-value alike) is now a uniform MUI `small` chip with a built-in delete icon, so they all share one shape/height. Note this restyles the active-filter row across **all** portals/apps using `TotalQueryResults`, not just NF — that uniformity is the point, but flagging the broader surface.

**5. DUO Mockup page removed; demo moved to Storybook** — deleted the NF `DUO Mockup` route/page and added a `DuoTermTags` Storybook story (`Explore/GenericCard/DuoTermTags`) covering the three real contexts: cards, the facet sidebar (with truncation), and the active-filter pills. This lives next to the component with no network/data dependency, so **there is no longer a mockup artifact to remove before merge** — the earlier callout about that no longer applies.

Tests/type-check/lint are green, and all stories verified rendering in Storybook.

packages/synapse-react-client/src/components/GenericCard/DuoTermTags/DuoTermTags.tsx
adamhindman: I'm skeptical that the meaning behind the colors will be obvious to users without a key of some kind. If it's not clear, it would just be visually distracting. My vote would be to make them all the same, neutral color.

allaway:
### Standardized DUO facet/pill rendering (re: @jay-hodgson's comment on the per-portal handler)

Good call — this is now addressed in `9dbb5622b9f`. Portals no longer define a `renderFacetValue` handler for DUO.

The DUO column is declared **once**, via the card schema field that already drives the card metadata row:

```ts
genericCardSchema: {
  // ...
  dataUseModifiersColumnName: 'dataUseModifiers',
}
```

From there, SRC does the rest:
- `QueryWrapperPlotNav` / `SearchQueryWrapperPlotNav` derive the DUO column from `cardConfiguration.genericCardSchema.dataUseModifiersColumnName` and pass it down.
- `QueryVisualizationWrapper` builds the renderer internally via a new `createDuoFacetValueRenderer(columnName)` factory (truncated tags in the facet sidebar; full-name deletable chips in the active-filter pills).
- The generic `renderFacetValue` hook is kept as an **escape hatch** — if a portal passes one explicitly, it still wins — but it's no longer required for the DUO case.

Net effect on the NF config: the `renderDuoFacetValue` handler (and its `DuoTermTags`/`createElement` imports) is **deleted**; only the one-line column declaration remains. Any other portal gets DUO facet + pill rendering for free by setting that same field.

Verified end-to-end in the NF dev server with no portal handler: the Data Use Modifiers facet and the active-filter pills both render DUO chips. Type-check, build, lint, and tests are green.

packages/synapse-react-client/src/components/GenericCard/DuoTermTags/DuoTermTags.tsx
allaway: That seems like a totally reasonable way to go. Would you be more inclined towards the standard grey color of the other pills, or something else? 

One other alternative is to try to extend this coloring: https://github.com/EBISPOT/DUO#duo-pictograms, which is blue for data use requirements, and then green/yellow/red for hierarchical restrictions.

allaway: But I don't know that is necessarily more intuitive than what is already there.

adamhindman:
@allaway @jay-hodgson 

Re: standardizing on the MUI chips:

I completely agree that we should standardize on one chip style, and not mix the two heights. I do not have a strong opinion about whether we standardize on using the MUI style or our existing style. Our chips are a little larger, providing a slightly more generous click target (in theory at least), but I have no doubt that the MUI component is well-tested and is fine to use. 

Putting an icon into the chips is an interesting innovation. It makes perfect sense in the context of DUO, where there are icons associated with each filter option, and those icons may even be known to the user already. It doesn't make sense for most other filters, so obviously we'd want that icon to be optional. I doubt you were suggesting otherwise, I just want to be clear!

I've already commented a bit about how I'm skeptical of using colors to differentiate chips on cards. When chips are used for filters, I think it may be even more important not to complicated their meaning. I think it's enough that a user can recognize a chip as representing a filter, without attaching a second level of meaning to it (i.e. the category of the filter). Just my thought, happy to hear a rebuttal!

adamhindman:
@allaway Oh, also noting that the icons used in the video do not match the [EBISPOT glyphs](https://github.com/EBISPOT/DUO/tree/master/doc/figs/pictograms/duo_svg) linked in the ticket. I assume Claude just did his thing, but for production we would use these 'real' glyphs.

allaway:
Thanks for the review @adamhindman !

>I completely agree that we should standardize on one chip style, and not mix the two heights. I do not have a strong opinion about whether we standardize on using the MUI style or our existing style. Our chips are a little larger, providing a slightly more generous click target (in theory at least), but I have no doubt that the MUI component is well-tested and is fine to use.

My very minor preference for the MUI chips is just that the X is vertically centered and i think they are a little more visually appealing, but this is something that could be fixed in our chips. I would leave it up to @jay-hodgson to decide on what direction makes sense here. 

>I've already commented a bit about how I'm skeptical of using colors to differentiate chips on cards.

Agree - I'll get rid of the varying colors. 

>@allaway Oh, also noting that the icons used in the video do not match the [EBISPOT glyphs](https://github.com/EBISPOT/DUO/tree/master/doc/figs/pictograms/duo_svg) linked in the ticket. I assume Claude just did his thing, but for production we would use these 'real' glyphs.

This was actually an intentional choice on my part. the glyphs only exist for 8 out of 24 DUO codes, and the readme describes these as proposed, so I'm not sure they are officially adopted. I also didn't find any other repository that used these.

allaway:
chip coloring is now based on portal color scheme values: 

Screenshot 2026-06-22 at 11 17 57 AM

allaway:
I filed an issue about the missing pictograms, if they provide an update we can revise these icons later. 

Otherwise, @jay-hodgson / @nickgros  i think this is ready for review. Thank you!

packages/synapse-react-client/src/components/GenericCard/DuoTermTags/DuoTermTags.tsx
```diff
@@ -0,0 +1,214 @@
...
+
+// `@mui/icons-material` (the classic Material Icons set) has no DNA/helix glyph,
+// so the genetic-studies term uses Google's Material Symbols "genetics" icon,
+// inlined here as an MUI SvgIcon so it inherits chip sizing and color like the
+// rest. (Material Symbols, Apache-2.0.)
+const GeneticsIcon = (props: SvgIconProps) => (
+  <SvgIcon viewBox="0 0 24 24" {...props}>
+    <path d="M5 23v-1q0-3.475 1.45-5.637t4-4.363q-2.55-2.2-4-4.362T5 2V1h2v1q0 .275.013.513T7.05 3h9.9q.025-.25.038-.488T17 2V1h2v1q0 3.475-1.45 5.638t-4 4.362q2.55 2.2 4 4.363T19 22v1h-2v-1q0-.275-.012-.513T16.95 21h-9.9q-.025.25-.037.488T7 22v1zM8.45 7h7.1q.325-.475.563-.95T16.55 5h-9.1q.2.55.437 1.038T8.45 7M12 10.7q.5-.425.975-.85t.9-.85h-3.75q.425.425.9.85t.975.85M10.125 15h3.75q-.425-.425-.9-.85T12 13.3q-.5.425-.975.85t-.9.85M7.45 19h9.1q-.2-.55-.437-1.037T15.55 17h-7.1q-.325.475-.562.95T7.45 19" />
+  </SvgIcon>
+)
```
nickgros: This should go in `synapse-react-client/src/assets/icons`

packages/synapse-react-client/src/components/GenericCard/DuoTermTags/DuoTermTags.tsx
```diff
@@ -0,0 +1,214 @@
...
+        flexWrap: 'wrap',
+        gap: 0.75,
+        minWidth: 0,
+        maxWidth: '100%',
+      }}
+    >
+      {sorted.map((t, i) => {
+        const url = duoRegistryUrl(t)
+        return (
+          <Tooltip
```
nickgros: This is kind of a nit but when these are rendered in the EnumFacetFilter component, the tooltip placement blocks the other options:

Image

You wouldn't want to hard-code the placement to 'right' because then the problem gets moved to the facet selections:

Image

Since the tooltip is just arbitrarily rendered, it might be more complex to support configuring this than it's worth.

packages/synapse-react-client/src/components/GenericCard/DuoTermTags/duoTerms.ts
```diff
@@ -0,0 +1,288 @@
...
+
+const BY_NAME = new Map(TERMS.map(t => [normalize(t.name), t]))
+const BY_CODE = new Map(TERMS.map(t => [t.code.toUpperCase(), t]))
+// A few alternate phrasings seen in the data that don't match the names above.
+const ALIASES: Record<string, string> = {
+  'not for profit non commercial use only': 'DUO:0000018',
+  'not for profit use only': 'DUO:0000018',
+  'non commercial use only': 'DUO:0000046',
+  'population origins or ancestry research': 'DUO:0000011',
+}
```
nickgros: Ok if this is the full extent of the normalization/aliasing, but I would strongly prefer this to be fixed upstream instead if possible. Portals should not be responsible for transforming/cleaning data other than what is minimally required to present the UI

packages/synapse-react-client/src/components/GenericCard/DuoTermTags/duoTerms.ts
```diff
@@ -0,0 +1,288 @@
...
+  const match =
+    BY_NAME.get(key) ?? BY_CODE.get(ALIASES[key]?.toUpperCase() ?? '')
+  if (match) {
+    return match
+  }
+  const abbr = value
+    .split(/\s+/)
+    .map(w => w[0]?.toUpperCase() ?? '')
+    .join('')
+    .slice(0, 4)
```
nickgros: is this abbreviation always correct? if not, may need to hard-code in the TERMS

packages/synapse-react-client/src/components/GenericCard/DuoTermTags/DuoTermTags.tsx
```diff
@@ -0,0 +1,214 @@
...
+                  </Link>
+                ) : (
+                  t.code
+                )}
+                <br />
+                {t.definition}
+              </>
+            }
+          >
+            <Chip
```
nickgros: Could this component and SelectionCriteriaPill use a shared base component, which would guarantee chip/icon/hover styles are identical? For example right now the hover 'close' icon color is different.

packages/synapse-react-client/src/components/widgets/facet-nav/SelectionCriteriaPill.tsx
```diff
@@ -1,41 +1,67 @@
...
-          <button
-            onClick={onRemoveFilter}
-            className="SelectionCriteriaPill__btnRemove"
-            title="deselect"
-          >
-            <Close />
-          </button>
-        )}
-      </div>
+      <Chip
```
nickgros: Using the MUI base styles and dropping most of our customizations seems to be a great simplification, thanks

packages/synapse-react-client/src/components/widgets/facet-nav/SelectionCriteriaPill.tsx
```diff
@@ -1,41 +1,67 @@
...
   readonly key: string
   readonly innerText: string
   readonly tooltipText: string
+  /**
+   * Optional custom node to render in place of the plain-text label (e.g. DUO
+   * tags). The node is expected to be self-contained — carrying its own tooltip
+   * and its own remove affordance — so the pill's tooltip, chrome, and remove
+   * button are all suppressed when this is provided.
+   */
+  readonly renderedInnerText?: ReactNode
```
nickgros: Maybe instead, SelectionCriteriaPill could accept a custom icon and a tooltip? Then it just happens to render the `Chip` (or custom component that uses `Chip`) in the same way as the DUO facet renderer? WDYT?

It would be more plumbing, but it would be more structured and you wouldn't have to worry about handling a 'remove facet' button in the custom renderer--it would just get added here.

packages/synapse-react-client/src/components/widgets/facet-nav/SelectionCriteriaPill.test.tsx
```diff
@@ -0,0 +1,81 @@
+import { render, screen } from '@testing-library/react'
+import userEvent from '@testing-library/user-event'
+import { describe, test, expect, vi } from 'vitest'
+import SelectionCriteriaPill from './SelectionCriteriaPill'
+
+describe('SelectionCriteriaPill', () => {
```
nickgros, line 7: thanks for the test backfill

packages/synapse-react-client/src/components/widgets/facet-nav/SelectionCriteriaPill.test.tsx
```diff
@@ -0,0 +1,81 @@
...
+    expect(screen.getByTestId('duo-tag')).toBeInTheDocument()
+    // ...the plain text is not...
+    expect(screen.queryByText('Non-Commercial Use Only')).toBeNull()
+    // ...and the pill drops its chrome so the node reads as the chip.
+    expect(
+      document.querySelector('.SelectionCriteriaPill--custom'),
+    ).toBeInTheDocument()
+  })
+
+  test('does not render its own delete icon for a custom node (the node owns removal)', () => {
```
nickgros: As mentioned in another comment, I wonder if we can remove this path and not have the node own removal.

packages/synapse-react-client/src/components/widgets/facet-nav/SelectionCriteriaPills.tsx
```diff
@@ -221,6 +222,18 @@ function getPillPropsFromFacetFilters(
         return {
           key: `facet-${selectedFacet.concreteType}-${selectedFacet.columnName}-${facetValue}`,
           innerText: innerText,
+          renderedInnerText: renderFacetValue?.(
```
nickgros: Following a thread of other comments I left...maybe renderFacetValue should return a structure like 

```ts
{
    value: ReactNode
    tooltipTitle: ReactNode
    icon: ReactNode // (or text that we transform using IconSvg)
}
```

Then those get passed in to SelectionCriteriaPill instead of `renderedInnerText`

nickgros (commented):
Had a few thoughts and suggestions, let me know what you think

packages/synapse-react-client/src/components/GenericCard/DuoTermTags/DuoTermTags.tsx
nickgros: Actually, in another set of comments I proposed how you could pass structured data to `SelectionCriteriaPill`. IF you do that, then you could hard-code the placement in a `Tooltip` component rendered by `SelectionCriteriaPill`, fixing this problem

allaway:
@nickgros took a pass at the refactor you suggested (pushed) - wanted to check I understood you correctly before going further.

- Pulled the chip into a shared `FacetValueChip` (icon + tooltip + optional remove), and `renderFacetValue` now returns structured content `{ value, tooltipTitle, icon }` instead of a self-contained node.
- `SelectionCriteriaPill` renders that through `FacetValueChip` and owns the remove button - dropped the `renderedInnerText` escape hatch entirely.
- The facet sidebar (`EnumFacetFilter`), the active-filter pills, and the DUO tags on cards all go through the same chip now, so styling/hover/close-icon are identical.
- Since the chip is rendered in one place, I hard-coded the tooltip placement there, which should fix the sidebar overlap you flagged.

Does this match what you were picturing? Still have your `duoTerms.ts` comments (upstream normalization + the abbreviation question) and the icon-location one to address.

packages/synapse-react-client/src/components/GenericCard/DuoTermTags/duoTerms.ts
allaway: Agreed - removed the ALIASES map entirely. The portal still does case/punctuation-insensitive name matching, but no longer aliases alternate phrasings. Filed NFOSI-1977 to canonicalize the `dataUseModifiers` annotations upstream (https://sagebionetworks.jira.com/browse/NFOSI-1977); until then those few values render as neutral "unknown" tags.

packages/synapse-react-client/src/components/GenericCard/DuoTermTags/duoTerms.ts
allaway: Good catch - the auto-abbreviation only applied to unknown values, and a guessed initialism could be wrong. Known terms already carry a hard-coded `abbr` in TERMS; for unknown values we now don't show a fabricated abbreviation at all (the tooltip just shows the raw name).

packages/synapse-react-client/src/components/widgets/facet-nav/SelectionCriteriaPill.tsx
```diff
@@ -1,68 +1,34 @@
...
-  readonly tooltipText: string
-  /**
-   * Optional custom node to render in place of the plain-text label (e.g. DUO
-   * tags). The node is expected to be self-contained — carrying its own tooltip
-   * and its own remove affordance — so the pill's tooltip, chrome, and remove
-   * button are all suppressed when this is provided.
-   */
-  readonly renderedInnerText?: ReactNode
+  /** The chip label (plain text, or structured content from `renderFacetValue`). */
+  readonly innerText: ReactNode
```
nickgros: May just rename this `label` for consistency/easier to pass-thru

packages/synapse-react-client/src/components/GenericCard/DuoTermTags/DuoTermTags.tsx
```diff
@@ -84,25 +74,11 @@ const DUO_ICONS: Record<string, ComponentType> = {
   'DUO:0000012': RuleOutlined, // Research Specific Restrictions
 }
 
-const termIcon = (term: DuoTerm) => {
+const termIcon = (term: DuoTerm): ReactElement => {
```
nickgros: Prefer `ReactNode` over `ReactElement` return type

packages/synapse-react-client/src/components/QueryVisualizationWrapper/QueryVisualizationContext.tsx
```diff
@@ -4,12 +4,28 @@ import {
...
 
+/**
+ * Structured rendering for a single facet value, returned by `renderFacetValue`.
+ * Consumers (the facet sidebar, active-filter pills) render this through the
+ * shared `FacetValueChip`, so styling, tooltip, and the remove affordance stay
+ * consistent and live in one place rather than in each renderer.
+ */
+export type RenderedFacetValueChip = {
+  /** The chip label. */
+  value: ReactNode
```
nickgros: May want to rename this to `label` for consistency/easier pass-thru

packages/synapse-react-client/src/components/GenericCard/DuoTermTags/DuoTermTags.stories.tsx
```diff
@@ -0,0 +1,179 @@
...
+import { EnumFacetFilterOption } from '@/components/widgets/query-filter/EnumFacetFilter/EnumFacetFilterOption'
+import SelectionCriteriaPill from '@/components/widgets/facet-nav/SelectionCriteriaPill'
+import FacetValueChip from '@/components/widgets/facet-nav/FacetValueChip'
+import DuoTermTags, { createDuoFacetValueRenderer } from './DuoTermTags'
+import { ALL_DUO_TERMS } from './duoTerms'
+
+// Build a DUO value's chip content (icon + label + tooltip) the way the facet
+// sidebar and active-filter pills do, then render it through the shared chip.
+const renderDuoFacetValue = createDuoFacetValueRenderer('duo')
+function duoChip(code: string, truncate = false) {
```
nickgros: A function that returns a ReactNode is basically a React component, so should be treated as such (capitalize the first letter, has a single `props` arg, invoke with JSX/createElement instead of function call)


```suggestion
function DuoChip(props: {code: string, truncate?: boolean}) {
  const { code, truncate = false } = props
```

packages/synapse-react-client/src/components/GenericCard/DuoTermTags/DuoTermTags.stories.tsx
```diff
@@ -0,0 +1,179 @@
...
+      </Typography>
+      {ALL_DUO_TERMS.map((t, i) => (
+        <EnumFacetFilterOption
+          key={t.code}
+          id={t.code}
+          inputType="checkbox"
+          isDropdown={false}
+          checked={false}
+          count={(i * 7) % 130}
+          label={duoChip(t.code, true)}
```
nickgros: Treat as React component
```suggestion
          label={<DuoChip code={t.code} truncate={true} />}
```

nickgros (commented):
@allaway refactor is exactly what I had in mind, code looks much cleaner to me. Had a few minor comments on those changes

packages/synapse-react-client/src/components/widgets/facet-nav/FacetValueChip.tsx
```diff
@@ -0,0 +1,60 @@
...
+      label={label}
+      onDelete={onDelete}
+      sx={{
+        ...chipSx,
+        ...(truncate ? { maxWidth: `min(${MAX_CHIP_WIDTH}px, 100%)` } : {}),
+      }}
+    />
+  )
+  return tooltipTitle ? (
+    <Tooltip title={tooltipTitle} placement="top">
```
nickgros: For the Tooltip `placement`, I was thinking for the EnumFacetFilter component, it should be `"right"` and for the SelectionCriteriaPills it should be `"top"`, so it might be a prop (if you agree, maybe your choice is better, it's subjective)

packages/synapse-react-client/src/components/GenericCard/DuoTermTags/DuoTermTags.tsx
allaway: Done - moved it to `src/assets/icons/GeneticsIcon.tsx` alongside the other icon assets; `DuoTermTags` imports it now.

allaway:
Thanks for the reviews and feedback @nickgros  - I think I got everything now!

packages/synapse-react-client/src/components/GenericCard/DuoTermTags/DuoTermTags.tsx
```diff
@@ -74,7 +64,7 @@ const DUO_ICONS: Record<string, ComponentType> = {
   'DUO:0000012': RuleOutlined, // Research Specific Restrictions
 }
 
-const termIcon = (term: DuoTerm): ReactElement => {
+const termIcon = (term: DuoTerm): ReactNode => {
```
nickgros: Missed this one, this is technically a React component too:

```suggestion
const TermIcon = (props: {term: DuoTerm}): ReactNode => {
  const  { term } = props
```

packages/synapse-react-client/src/components/GenericCard/DuoTermTags/DuoTermTags.tsx
```diff
@@ -170,7 +160,7 @@ export function createDuoFacetValueRenderer(duoColumnName: string) {
     }
     const t = resolveDuoTerm(value)
     return {
-      value: t.name,
+      label: t.name,
       icon: termIcon(t),
```
nickgros:
```suggestion
      icon: <TermIcon term={t} />,

nickgros (commented):
@allaway One tiny case I missed before (sorry), and can you ensure the tests pass?

Then I think this is good to merge!

allaway:
🤞 everything passes now!

nickgros (approved):

PR #2933, opened 2026-06-22
PORTALS-4305

jay-hodgson:
Screenshot 2026-06-29 at 5 11 59 PM
Screenshot 2026-06-29 at 5 08 37 PM

Normal bar chart rendering improvement:
Screenshot 2026-06-29 at 5 06 57 PM

Previously, code would force any bar chart to be full width.
Screenshot 2026-06-29 at 5 14 07 PM

ELITE portal custom plot:
Screenshot 2026-06-30 at 9 34 38 AM


ankimno:
These look great. I appreciate updating the code base so that the charts are not forced into full width. I suggest we deploy and then revisit/iterate as needed based on user feedback.

packages/synapse-react-client/src/components/widgets/facet-nav/PlotsContainer.tsx
```diff
@@ -212,6 +212,16 @@ function PlotsContainer(props: PlotsContainerProps) {
     }
   }, [plotUiStateArray])
 
+  // When all charts become visible (transition to 'LESS'), trigger a resize event so
+  // Plotly recalculates dimensions for charts that were previously hidden (display: none).
+  useEffect(() => {
```
jay-hodgson, line 217: bug discovered during implementation

packages/synapse-react-client/src/components/widgets/facet-nav/FacetNavPanel.tsx
```diff
@@ -54,19 +54,12 @@ const maxLabelLength: number = 19
...
-    visible: false,
-    showgrid: false,
-  },
-}
+// Note: do NOT define layout as a module-level constant.
+// Plotly.js mutates the layout object it receives, and sharing one object across
+// multiple chart instances causes cross-instance corruption (e.g. when the expand
+// modal mounts a second FacetNavPanel with the same layout object, Plotly's
+// mutations to the inner chart's layout will silently break the outer chart).
+// Instead, each component instance computes its own layout via useMemo.
```
nickgros: OK to keep this comment content but I suggest merging it with the comment above the `useMemo` below. Since the module-level `layout` is gone, this information is out-of-place

nickgros (approved):

nickgros:
IMO the plots seem a bit bare without axis labels (but I don't think that's a new problem)

jay-hodgson:
Screenshot 2026-07-02 at 9 20 21 AM

nickgros:
@jay-hodgson sorry I was not totally clear! I was thinking we would also benefit from

- Titles for the custom plot X axes ("Number of files", "Size (Bytes)", "Number of Participants")
- Axis tick marks for the non-custom barplot (to give a sense of absolute values). Not sure if it matters to add a label for non-custom plots since it would pretty much always just be "Count". Also provides visual uniformity since we have ticks on the custom plots.

I think that the custom Y axis title that you added is helpful too though, since at a glance it might not be clear that each bar represents a single dataset.

jay-hodgson:
With axis labels on all custom plots, and showing y axis tick marks if showing a bar chart:
Screenshot 2026-07-02 at 9 48 49 AM

PR #2934, opened 2026-06-22
PORTALS-4308: ADKP Homepage Redesign Cleanup

kianamcc:
Jira: https://sagebionetworks.jira.com/browse/PORTALS-4308

### DIFF
--- a/apps/portals/adknowledgeportal/src/config/footerConfig.ts
+++ b/apps/portals/adknowledgeportal/src/config/footerConfig.ts
@@ -1,6 +1,8 @@
 import { FooterConfig } from '@sage-bionetworks/synapse-portal-framework/types/portal-config'
 
 const footer: FooterConfig = {
+  contactUs: 'https://sagebionetworks.jira.com/servicedesk/customer/portal/12',
+  help: 'https://help.adknowledgeportal.org/apd/',
   about: 'https://help.adknowledgeportal.org/apd/About.1994096680.html',
   variant: 'adkp',
 }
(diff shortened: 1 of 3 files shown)


apps/portals/adknowledgeportal/src/config/footerConfig.ts
```diff
@@ -1,6 +1,8 @@
 import { FooterConfig } from '@sage-bionetworks/synapse-portal-framework/types/portal-config'
 
 const footer: FooterConfig = {
+  contactUs: 'https://sagebionetworks.jira.com/servicedesk/customer/portal/12',
```
kianamcc, line 4: Should these be gated by the feature flag or is it ok to have these show for homepage v1 as well? If gated, we can add a new prop to the footerConfig for feature flag enabled links?

apps/portals/adknowledgeportal/src/pages/HomePageV2.tsx
```diff
@@ -58,28 +58,20 @@ function HomePageInternal() {
     title: 'Our portal has more than a petabyte of data..',
     buttonText: 'Explore Alzheimer’s Data',
     buttonLink: '/Explore/Data',
-    subtitle: (
-      <>
-        <div style={{ marginBottom: '10px' }}>
-          Including over 265,000 files from 60+ assays and in 50+ different file
```
kianamcc, line 64: Asked to remove if these numbers aren't updated in real-time.

apps/portals/adknowledgeportal/src/config/footerConfig.ts
jay-hodgson: I'd confirm with Jo, but I suspect this is ok to roll out independent of the new home page.

jay-hodgson (approved):

apps/portals/adknowledgeportal/src/config/footerConfig.ts
kianamcc: @jay-hodgson got the OK from Jo that these can show in the current homepage as well!

PR #2935, opened 2026-06-23
PORTALS-4277: Explore tab focus style does not appear on Firefox

kianamcc:
Jira: https://sagebionetworks.jira.com/browse/PORTALS-4277

Screenshot 2026-06-22 at 5 23 20 PM
Screenshot 2026-06-22 at 5 23 33 PM
Collections
Programs

### DIFF
--- a/apps/portals/arkportal/src/config/style/_style_overrides.scss
+++ b/apps/portals/arkportal/src/config/style/_style_overrides.scss
@@ -1,3 +1,8 @@
+// Accessible keyboard focus-ring color. See PORTALS-4277.
+:root {
+  --synapse-focus-color: #ac7058;
+}
+
 #header {
   height: fit-content;
   background-image: none;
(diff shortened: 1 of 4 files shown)


apps/portals/arkportal/src/config/style/_style_overrides.scss
```diff
@@ -1,3 +1,8 @@
+// Accessible keyboard focus-ring color. See PORTALS-4277.
+:root {
+  --synapse-focus-color: #ac7058;
```
kianamcc, line 3: Had claude compute the WCAG contrast ratio of each portal’s primary action color against the white tab background. Ark and Elite failed this test. Suggested darkened hues to meet this criteria. Could follow-up with Adam to see if he is ok with these colors?

jay-hodgson: Nice find!  Please send a note to Adam in a jira ticket comment (on PORTALS-4277) so we have additional visibility of this change, but I think we should make the change!

jay-hodgson (approved):

PR #2937, opened 2026-06-24
SWC-7853 - Display and edit task status on CurationTaskCard

afwillia:
- Add curation task status pill to CurationTaskCard
- Move taskId into description div
- Create settings button that opens useCreateOrUpdateCurationTaskDialog if user has edit permission to the project
- Add dropdown to modify task status in useCreateOrUpdateCurationTaskDialog

image

expanded content
image

### DIFF
--- a/packages/synapse-react-client/src/features/curator/dashboard/components/CurationTaskCard.tsx
+++ b/packages/synapse-react-client/src/features/curator/dashboard/components/CurationTaskCard.tsx
@@ -167,7 +167,9 @@ export default function CurationTaskCard(props: CurationTaskCardProps) {
               <Typography variant="body1">Task ID: {taskId}</Typography>
             )}
           </Box>
-          <Typography variant="body1">{description}</Typography>
+          {description && (
+            <Typography variant="body1">{description}</Typography>
+          )}
           <div className={styles.userChipContainer}>
             {principalIds.map(principalId => (
               <UserOrTeamChip key={principalId} principalId={principalId} />
--- a/packages/synapse-react-client/src/features/curator/dashboard/components/CurationTaskCard.tsx
+++ b/packages/synapse-react-client/src/features/curator/dashboard/components/CurationTaskCard.tsx
@@ -3,16 +3,7 @@ import CreateOrUpdateCurationTaskDialog from '@/features/entity/metadata-task/co
 import useOpenCuratorFromTaskButton from '@/features/entity/metadata-task/hooks/useOpenCuratorButton'
 import { OPEN_CURATOR_NO_PERMISSION_ON_SOURCE_ERROR_MESSAGE } from '@/features/entity/metadata-task/utils/constants'
 import useGetEntityBundle from '@/synapse-queries/entity/useEntityBundle'
-import {
-  Box,
-  Button,
-  Card,
-  Chip,
-  Divider,
-  IconButton,
-  Link,
-  Typography,
-} from '@mui/material'
+import { Box, Card, Chip, Divider, IconButton, Typography } from '@mui/material'
 import {
   TaskBundle,
   TaskStatusStateEnum,
@@ -22,7 +13,6 @@ import styles from './CurationTaskCard.module.scss'
 import NextStepButton from './NextStepButton'
 import sharedStyles from './shared.module.scss'
 import UserOrTeamChip from './UserOrTeamChip'
-import TableChartOutlinedIcon from '@mui/icons-material/TableChartOutlined'
 import SettingsOutlinedIcon from '@mui/icons-material/SettingsOutlined'
 import { useState } from 'react'
 
@@ -136,7 +126,6 @@ export default function CurationTaskCard(props: CurationTaskCardProps) {
   } = useUiForTask(taskBundle)
 
   const [isSettingsOpen, setIsSettingsOpen] = useState(false)
-  const [isExpanded, setIsExpanded] = useState(false)
 
   const projectId = taskBundle.task?.projectId
   const { data: bundle } = useGetEntityBundle(projectId, undefined, {
@@ -159,14 +148,7 @@ export default function CurationTaskCard(props: CurationTaskCardProps) {
       <div className={styles.cardContent}>
         <div className={styles.mainContent}>
           <div className={styles.titleChipContainer}>
-            <Link
-              component="button"
-              variant="headline3"
-              onClick={() => setIsExpanded(v => !v)}
-              sx={{ textAlign: 'left' }}
-            >
-              {title}
-            </Link>
+            <Typography variant="headline3">{title}</Typography>
             {taskType && <TaskTypeChip label={taskType} />}
             {canEdit && (
               <IconButton
@@ -183,45 +165,28 @@ export default function CurationTaskCard(props: CurationTaskCardProps) {
               <Typography variant="body1">Task ID: {taskId}</Typography>
             )}
           </Box>
+          <Typography variant="body1">{description}</Typography>
           <div className={styles.userChipContainer}>
             {principalIds.map(principalId => (
               <UserOrTeamChip key={principalId} principalId={principalId} />
             ))}
           </div>
-          {isExpanded && (
-            <div className={styles.expandedContent}>
-              <Typography variant="body1">{description}</Typography>
-              <Button
-                variant="contained"
-                fullWidth
-                onClick={onClickAction}
-                disabled={isLoading || isPending}
-                startIcon={<TableChartOutlinedIcon />}
-              >
-                {buttonText}
-              </Button>
-            </div>
-          )}
         </div>
         <Box sx={{ display: 'flex', alignItems: 'center' }}>
           <TaskStatusChip state={statusState} />
         </Box>
-        {!isExpanded && (
-          <>
-            <Divider
-              orientation="vertical"
-              flexItem
-              sx={{ display: { xs: 'none', md: 'block' } }}
-            />
-            <NextStepButton
-              className={styles.cardButton}
-              buttonText={buttonText}
-              onClick={onClickAction}
-              disabled={isLoading}
-              loading={isPending}
-            />
-          </>
-        )}
+        <Divider
+          orientation="vertical"
+          flexItem
+          sx={{ display: { xs: 'none', md: 'block' } }}
+        />
+        <NextStepButton
+          className={styles.cardButton}
+          buttonText={buttonText}
+          onClick={onClickAction}
+          disabled={isLoading}
+          loading={isPending}
+        />
       </div>
       <CreateOrUpdateCurationTaskDialog
         key={String(isSettingsOpen)}
--- a/packages/synapse-react-client/src/features/curator/dashboard/components/CurationTaskCard.tsx
+++ b/packages/synapse-react-client/src/features/curator/dashboard/components/CurationTaskCard.tsx
@@ -3,7 +3,15 @@ import CreateOrUpdateCurationTaskDialog from '@/features/entity/metadata-task/co
 import useOpenCuratorFromTaskButton from '@/features/entity/metadata-task/hooks/useOpenCuratorButton'
 import { OPEN_CURATOR_NO_PERMISSION_ON_SOURCE_ERROR_MESSAGE } from '@/features/entity/metadata-task/utils/constants'
 import useGetEntityBundle from '@/synapse-queries/entity/useEntityBundle'
-import { Box, Card, Chip, Divider, IconButton, Typography } from '@mui/material'
+import {
+  Box,
+  Card,
+  Chip,
+  Divider,
+  IconButton,
+  Skeleton,
+  Typography,
+} from '@mui/material'
 import {
   TaskBundle,
   TaskStatusStateEnum,
@@ -136,12 +144,9 @@ export default function CurationTaskCard(props: CurationTaskCardProps) {
 
   function onClickAction() {
     if (hasPermissionToOpenGrid) {
-        onClickNextStep()
+      onClickNextStep()
     } else {
-        displayToast(
-          OPEN_CURATOR_NO_PERMISSION_ON_SOURCE_ERROR_MESSAGE,
-          'danger',
-        )
+      displayToast(OPEN_CURATOR_NO_PERMISSION_ON_SOURCE_ERROR_MESSAGE, 'danger')
     }
   }
 
@@ -162,7 +167,11 @@ export default function CurationTaskCard(props: CurationTaskCardProps) {
             )}
           </div>
           <Box sx={{ display: 'flex', gap: 4 }}>
-            <Typography variant="body1">{bundle?.entity?.name}</Typography>
+            {isLoading ? (
+              <Skeleton width={100} />
+            ) : (
+              <Typography variant="body1">{bundle?.entity?.name}</Typography>
+            )}
             {taskId && (
               <Typography variant="body1">Task ID: {taskId}</Typography>
             )}
--- a/packages/synapse-react-client/src/features/curator/dashboard/components/CurationTaskCard.tsx
+++ b/packages/synapse-react-client/src/features/curator/dashboard/components/CurationTaskCard.tsx
@@ -3,7 +3,10 @@ import CreateOrUpdateCurationTaskDialog from '@/features/entity/metadata-task/co
 import useOpenCuratorFromTaskButton from '@/features/entity/metadata-task/hooks/useOpenCuratorButton'
 import { OPEN_CURATOR_NO_PERMISSION_ON_SOURCE_ERROR_MESSAGE } from '@/features/entity/metadata-task/utils/constants'
 import { Box, Button, Card, Chip, Divider, Typography } from '@mui/material'
-import { TaskBundle } from '@sage-bionetworks/synapse-client'
+import {
+  TaskBundle,
+  TaskStatusStateEnum,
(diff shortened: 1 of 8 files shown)


packages/synapse-react-client/src/synapse-queries/curation/task/useCurationTask.ts
```diff
@@ -116,10 +116,12 @@ export function useUpdateCurationTaskStatus(
         { taskId: taskStatus.taskId!, taskStatus },
       ),
     onSuccess: (data, variables, context) => {
-      // Invalidate both the status query and the task query using the ID key
       queryClient.invalidateQueries({
         queryKey: keyFactory.getCurationTaskIdKey(variables.taskId!),
       })
+      queryClient.invalidateQueries({
+        queryKey: keyFactory.getAllCurationTaskListKey(),
+      })
```
nickgros, line 124: Nice catch, makes sense

packages/synapse-react-client/src/features/curator/dashboard/components/CurationTaskCard.tsx
```diff
@@ -86,28 +118,52 @@ export default function CurationTaskCard(props: CurationTaskCardProps) {
...
     <Card className={classNames(sharedStyles.card, styles.card)}>
       <div className={styles.cardContent}>
         <div className={styles.mainContent}>
           <div className={styles.titleChipContainer}>
             <Typography variant="headline3">{title}</Typography>
             {taskType && <TaskTypeChip label={taskType} />}
-            {taskId && <TaskTypeChip label={`TaskId ${taskId}`} />}
+            {canEdit && (
+              <Button
+                sx={{ display: 'flex' }}
```
nickgros: Use MUI `IconButton` and can drop `flex` to get a nice rounded interactive shape

packages/synapse-react-client/src/features/entity/metadata-task/components/CreateOrUpdateCurationTaskDialog.tsx
```diff
@@ -565,6 +606,35 @@ export default function CreateOrUpdateCurationTaskDialog(
         </Grid>
         <Grid size={{ xs: 12, sm: 6 }}>
           <Stack gap={3}>
+            {isEditMode && (
+              <FormControl fullWidth>
```
nickgros: Use our `StyledFormControl`

packages/synapse-react-client/src/features/curator/dashboard/components/CurationTaskCard.tsx
```diff
@@ -130,6 +186,14 @@ export default function CurationTaskCard(props: CurationTaskCardProps) {
           loading={isPending}
         />
       </div>
+      <CreateOrUpdateCurationTaskDialog
+        open={isSettingsOpen}
```
nickgros, line 197: Should add a prop `key={String(isSettingsOpen)}` to ensure the state of the dialog resets if it is closed

packages/synapse-react-client/src/features/curator/dashboard/components/CurationTaskCard.tsx
```diff
@@ -86,28 +118,52 @@ export default function CurationTaskCard(props: CurationTaskCardProps) {
...
+                sx={{ display: 'flex' }}
+                onClick={() => setIsSettingsOpen(true)}
+                aria-label="Task settings"
+              >
+                <SettingsOutlinedIcon />
+              </Button>
+            )}
           </div>
-          <Typography variant="body1">{description}</Typography>
+          <Box sx={{ display: 'flex', gap: 6 }}>
```
nickgros: nit: `gap: 4` would be closer to design

packages/synapse-react-client/src/features/curator/dashboard/components/CurationTaskCard.tsx
```diff
@@ -86,28 +118,52 @@ export default function CurationTaskCard(props: CurationTaskCardProps) {
...
+
   return (
     <Card className={classNames(sharedStyles.card, styles.card)}>
       <div className={styles.cardContent}>
         <div className={styles.mainContent}>
           <div className={styles.titleChipContainer}>
             <Typography variant="headline3">{title}</Typography>
             {taskType && <TaskTypeChip label={taskType} />}
-            {taskId && <TaskTypeChip label={`TaskId ${taskId}`} />}
+            {canEdit && (
```
nickgros, line 151: `canEdit` is part of the equation here. If you are the assignee (without project edit permission), you can still edit the properties in the `TaskStatus` (including the `state`).

In this PR, should we enable assignees without edit access to edit this property? You could check if the assignee is the current user or belongs to that team with the existing `useGetIsPrincipalIdUserOrMemberOfTeam`. Or for this case, should we show a different editor that only allows you to update the `TaskStatus`?

If we stick with a single editor dialog, we would need to also update `CreateOrUpdateCurationTaskDialog` to only allow changes to the `TaskStatus` fields if that user is an assignee but does not have project edit permission.

packages/synapse-react-client/src/features/entity/metadata-task/components/CreateOrUpdateCurationTaskDialog.tsx
```diff
@@ -289,7 +316,21 @@ export default function CreateOrUpdateCurationTaskDialog(
...
-      updateTask(payload)
+      const statusChanged =
+        pendingStatusState !== undefined &&
+        pendingStatusState !== currentTaskStatus?.state
+      updateTask(payload, {
+        onSuccess: updatedTask => {
+          if (statusChanged && currentTaskStatus != null) {
+            updateTaskStatus(
+              { ...currentTaskStatus, state: pendingStatusState },
+              { onSuccess: () => onSuccess(updatedTask) },
```
nickgros: Both the task and task status share the same etag, so the second call will always fail unless you grab the new etag from the task.

Also, if we decide to allow non-project-editors to update the task status, the first call will always fail for those users.

More of a style nit, but I don't really like chaining `onSuccess` in this way--prefer we use `mutateAsync` and use `async`/`await` if we are going to chain calls.

packages/synapse-react-client/src/features/curator/dashboard/components/CurationTaskCard.tsx
afwillia: The [ticket AC](https://sagebionetworks.jira.com/browse/SWC-7853) states `Users without project edit access can view, but not edit, Task Status and Due Date.`. So we can leave as-is and see if we need more granular controls in the future. Maybe adding a status toggle to the pill so we don't overload assignees with extra information they cannot edit.

packages/synapse-react-client/src/features/entity/metadata-task/components/CreateOrUpdateCurationTaskDialog.tsx
afwillia: I updated this to check if the task properties change in addition to task status. If not, skip the task update. If updating both, use the etag from the first response in the second request.

afwillia:
@nickgros I also made some style changes requested in in the ticket. The task title is now a link that expands the card to show instructions and a larger next step button.

packages/synapse-react-client/src/features/curator/dashboard/components/CurationTaskCard.tsx
```diff
@@ -86,50 +131,108 @@ export default function CurationTaskCard(props: CurationTaskCardProps) {
...
     isPending,
   } = useUiForTask(taskBundle)
 
+  const [isSettingsOpen, setIsSettingsOpen] = useState(false)
+  const [isExpanded, setIsExpanded] = useState(false)
+
+  const projectId = taskBundle.task?.projectId
+  const { data: permissions } = useGetEntityPermissions(projectId ?? '')
+  const { data: project } = useGetEntity(projectId)
+  const canEdit = permissions?.canEdit ?? false
```
nickgros: Instead of getting both the entity and the permissions separately, you can `useGetEntityBundle` which will include both.

packages/synapse-react-client/src/features/curator/dashboard/components/CurationTaskCard.tsx
```diff
@@ -86,50 +131,108 @@ export default function CurationTaskCard(props: CurationTaskCardProps) {
     taskType,
     principalIds,
     buttonText,
+    statusState,
     onClickNextStep,
     hasPermission,
```
nickgros: Not your code, but could you rename this to `hasPermissionToOpenGrid` or `hasViewPermissionOnSource`?

packages/synapse-react-client/src/features/curator/dashboard/components/CurationTaskCard.tsx
```diff
@@ -86,50 +131,108 @@ export default function CurationTaskCard(props: CurationTaskCardProps) {
...
+  const canEdit = permissions?.canEdit ?? false
+
+  const onClickAction = hasPermission
+    ? onClickNextStep
+    : () => {
+        displayToast(
+          OPEN_CURATOR_NO_PERMISSION_ON_SOURCE_ERROR_MESSAGE,
+          'danger',
+        )
+      }
```
nickgros: Style nit: instead of changing the identity of the function on a ternary, just invoke the logic directly. I think this is more readable
```suggestion
  function onClickAction() {
    if (hasPermissionToOpenGrid) {
        onClickNextStep()
    } else {
        displayToast(
          OPEN_CURATOR_NO_PERMISSION_ON_SOURCE_ERROR_MESSAGE,
          'danger',
        )
    }
  }
```

packages/synapse-react-client/src/features/curator/dashboard/components/CurationTaskCard.tsx
```diff
@@ -86,50 +131,108 @@ export default function CurationTaskCard(props: CurationTaskCardProps) {
...
           <div className={styles.titleChipContainer}>
-            <Typography variant="headline3">{title}</Typography>
+            <Link
+              component="button"
+              variant="headline3"
+              onClick={() => setIsExpanded(v => !v)}
+              sx={{ textAlign: 'left' }}
+            >
+              {title}
+            </Link>
```
nickgros: A `Link` to expand content is unconventional and I think awkward. As a user, I would expect this to take me somewhere, and I would not expect it to expand to show the instructions.

Would it make more sense to use a pattern like our "Show More" component provides? Or some other clearly 'expandable' UI pattern?

For real world tasks, are these description strings large? Maybe it's always OK to show them without the expand/collapse?

packages/synapse-react-client/src/features/curator/dashboard/components/CurationTaskCard.tsx
```diff
@@ -86,50 +131,108 @@ export default function CurationTaskCard(props: CurationTaskCardProps) {
...
+                onClick={() => setIsSettingsOpen(true)}
+                aria-label="Task settings"
+              >
+                <SettingsOutlinedIcon />
+              </IconButton>
+            )}
           </div>
-          <Typography variant="body1">{description}</Typography>
+          <Box sx={{ display: 'flex', gap: 4 }}>
+            <Typography variant="body1">{project?.name}</Typography>
```
nickgros: In your story, the project never loads, which makes this look strange. So that probably needs to be fixed. Additionally, consider showing a MUI `Skeleton` while `project` is loading or missing.


```suggestion
            {projectIsLoading ? <Skeleton width={100} /> : <Typography variant="body1">{project?.name}</Typography>}
```

packages/synapse-react-client/src/features/curator/dashboard/components/CurationTaskCard.tsx
```diff
@@ -86,50 +131,108 @@ export default function CurationTaskCard(props: CurationTaskCardProps) {
...
+              <Typography variant="body1">Task ID: {taskId}</Typography>
+            )}
+          </Box>
           <div className={styles.userChipContainer}>
             {principalIds.map(principalId => (
               <UserOrTeamChip key={principalId} principalId={principalId} />
             ))}
           </div>
+          {isExpanded && (
+            <div className={styles.expandedContent}>
```
nickgros: Not sure if we need the expandable, but if we keep it, use a MUI `Collapse` which will animate the expand/collapse, which makes it more gentle and easy for the user to see what changed

packages/synapse-react-client/src/features/curator/dashboard/components/CurationTaskCard.tsx
```diff
@@ -86,50 +131,108 @@ export default function CurationTaskCard(props: CurationTaskCardProps) {
...
+              <Typography variant="body1">{description}</Typography>
+              <Button
+                variant="contained"
+                fullWidth
+                onClick={onClickAction}
+                disabled={isLoading || isPending}
+                startIcon={<TableChartOutlinedIcon />}
+              >
+                {buttonText}
+              </Button>
```
nickgros: This button looks kinda weird, I wonder if we can just keep the existing button if/when we expand?

packages/synapse-react-client/src/features/entity/metadata-task/components/CreateOrUpdateCurationTaskDialog.tsx
```diff
@@ -286,12 +310,42 @@ export default function CreateOrUpdateCurationTaskDialog(
...
+
+      // Only call updateTask when mutable task fields actually changed.
+      // This lets users who can only update status (not the task itself)
+      // save without hitting a 403 on the task PUT.
+      const taskFieldsChanged =
+        payload.dataType !== task.dataType ||
+        payload.instructions !== task.instructions ||
+        payload.assigneePrincipalId !== task.assigneePrincipalId ||
+        payload.taskProperties?.suggestedAuthorizationMode !==
+          task.taskProperties?.suggestedAuthorizationMode
```
nickgros: If we are only letting users with full edit permissions edit the status, then I don't think you need this. We'll just end up updating twice, which is OK. 

In fact, I prefer we get rid of it if we can since it's a place we could forget to extend the logic if we add a new field.

If we need to keep it, can we do a full object equality check (`isEqual(payload, task)`)?

packages/synapse-react-client/src/features/entity/metadata-task/components/CreateOrUpdateCurationTaskDialog.tsx
```diff
@@ -565,6 +619,35 @@ export default function CreateOrUpdateCurationTaskDialog(
...
+                  </MenuItem>
+                  <MenuItem value={TaskStatusStateEnum.IN_PROGRESS}>
+                    {TASK_STATUS_IN_PROGRESS_LABEL}
+                  </MenuItem>
+                  <MenuItem value={TaskStatusStateEnum.COMPLETED}>
+                    {TASK_STATUS_COMPLETED_LABEL}
+                  </MenuItem>
+                  <MenuItem value={TaskStatusStateEnum.CANCELED}>
+                    {TASK_STATUS_CANCELED_LABEL}
+                  </MenuItem>
```
nickgros: Total nit but it could be cool to reuse the colorful chips here as shown on the card (kinda like how Jira does it)

packages/synapse-react-client/src/features/curator/dashboard/components/CurationTaskCard.tsx
afwillia: I agree the link is awkward. I think it would be worth merging this showing the description, then getting design feedback.

packages/synapse-react-client/src/features/entity/metadata-task/components/CreateOrUpdateCurationTaskDialog.tsx
```diff
@@ -725,7 +700,7 @@ export default function CreateOrUpdateCurationTaskDialog(
           )}
           <Button
             variant="contained"
-            onClick={() => void handleSave()}
+            onClick={() => void handleSave().catch(noop)}
```
nickgros, line 704: Is there a reason we need `.catch(noop)` (a lint rule perhaps)? If not, remove it. If we do need it, add a comment like `// Error handling is already covered in the mutation options`

```suggestion
            onClick={() => void handleSave()}
```

packages/synapse-react-client/src/features/curator/dashboard/components/CurationTaskCard.tsx
```diff
@@ -179,50 +158,37 @@ export default function CurationTaskCard(props: CurationTaskCardProps) {
...
+            {bundle?.entity?.name ? (
+              <Typography variant="body1">{bundle.entity.name}</Typography>
+            ) : (
+              <Skeleton width={100} />
+            )}
             {taskId && (
               <Typography variant="body1">Task ID: {taskId}</Typography>
             )}
           </Box>
+          <Typography variant="body1">{description}</Typography>
```
nickgros: description might be optional/empty string, might want to
```suggestion
          {description && <Typography variant="body1">{description}</Typography>}
````

nickgros (approved):

packages/synapse-react-client/src/features/entity/metadata-task/components/CreateOrUpdateCurationTaskDialog.tsx
afwillia: I'll add that comment, this is to prevent a rejected promise from throwing an additional error

PR #2940, opened 2026-06-24
PORTALS-4315: follow-up (text changes)

jay-hodgson:


jay-hodgson:
Ann Novakowski reviewed this copy/text change and approved.

PR #2941, opened 2026-06-24
SWC-7826 - style upsert key column header part 2

afwillia:
pass isUpsertKey to columnFactory and update ColumnHeaderWithTooltip to display primary key icon and help text

image

### DIFF
--- a/packages/synapse-react-client/src/components/DataGrid/components/ColumnHeaderWithTooltip.tsx
+++ b/packages/synapse-react-client/src/components/DataGrid/components/ColumnHeaderWithTooltip.tsx
@@ -13,7 +13,7 @@ type ColumnHeaderWithTooltipProps = {
 }
 
 const UPSERT_KEY_TOOLTIP_TEXT =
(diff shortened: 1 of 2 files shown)


packages/synapse-react-client/src/components/DataGrid/components/ColumnHeaderWithTooltip.tsx
```diff
@@ -1,16 +1,20 @@
...
   description?: string
   isRequired?: boolean
+  isUpsertKey?: boolean
   showPinIcon?: boolean
   isPinned?: boolean
   onTogglePin?: () => void
 }
 
+const UPSERT_KEY_TOOLTIP_TEXT =
+  'This property field is the primary key. It is used to match the existing rows when applying updates. If a value already exists the row is updated; if not, a new row is added.  If multiple primary keys are defined, all primary key values must match for an existing row to be updated.'
```
kianamcc: Nit: I think there's a double space between "...a new row is added." and "If multiple..." but looks like the tooltip is collapsing this!

kianamcc (approved):

PR #2943, opened 2026-06-25
PORTALS-4321

jay-hodgson:


jay-hodgson:
Still renders without error:
Screenshot 2026-06-25 at 2 24 14 PM

afwillia (approved):

PR #2947, opened 2026-06-28
PORTALS-4330, PORTALS-4331: Fix NF dataset "On Synapse" link and remove survey banner

allaway:
## Summary

Resolves two NF Data Portal tickets:

### PORTALS-4330 — Dataset card "On Synapse" link misconfigured
Dataset cards render the `studyId` label via a `labelLinkConfig` entry with `resolveEntityName: true`. That branch in `SynapseCardLabel` ignored `urlParamStyle` and always built an **in-portal** URL (`/Explore/Studies?studyId=synXXX`), which has no matching route and falls back to the Explore page.

`EntityLink` already resolves the entity name for the link text and defaults to linking to `https://www.synapse.org/Synapse:{id}`. The fix stops overriding that default with the broken in-portal href.

`resolveEntityName` is used in exactly 3 places, all NF-only and all previously broken the same way, so this also fixes the study card's "Next/Previous Phase Project" links.

### PORTALS-4331 — Remove the data access survey banner
Removed the `SurveyToast` banner from the NF portal app shell (`RootApp.tsx`), mirroring the prior GENIE (PORTALS-3901) and AD Knowledge Portal (PORTALS-3899) removals. Also tidied a stale comment in `routes.ts`.

> Note on survey storage (ticket follow-up): the banner linked to a Google Form; responses live in that form's Responses tab (and any linked Google Sheet) under the owning Google account. The codebase only held the public form link.

## Test plan
- `SynapseCardLabel` unit tests pass (15/15), including a new test asserting the `resolveEntityName` branch links to Synapse.org and does not pass an in-portal `link` override.
- `synapse-react-client` type-check passes.

### DIFF
--- a/apps/portals-e2e/src/configs/bannersConfig.ts
+++ b/apps/portals-e2e/src/configs/bannersConfig.ts
@@ -1,6 +1,6 @@
 import { Portal } from './routesConfig'
 
-export type BannerType = 'cookies' | 'survey' | 'surveyDialog'
+export type BannerType = 'cookies' | 'surveyDialog'
 
 type BannerConfig = Record<Portal, BannerType[]>
 
@@ -15,7 +15,7 @@ const bannerConfig: BannerConfig = {
   digitalhealth: ['cookies'],
   eliteportal: ['cookies'],
   namhub: ['cookies'],
-  nf: ['cookies', 'survey'],
+  nf: ['cookies'],
   stopadportal: ['cookies'],
   genie: ['cookies'],
 }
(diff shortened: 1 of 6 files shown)


allaway:
I think this is ready for review, but the test feels like it might be a bit overkill to me...happy to remove it so.

jay-hodgson:
all NF end2end tests are failing after this change because it is looking to dismiss the survey banner.  Could you please update the configuration for NF in the e2e tests?
For example:
`src/explore.spec.ts:481:7 › Explore › Tools @nf › go to Explore/Tools › dismiss survey banner`

jay-hodgson (changes requested):
(need to update NF e2e tests)

allaway:
Good catch, thanks Jay. Fixed in c4d04924316 - the NF entry in the e2e `bannersConfig` still listed `survey`, so `dismissBanners` was trying to close a banner that no longer exists. Dropped it from nf's list (and removed the now-unused `survey` banner type, since nf was the only portal using it). `surveyDialog` for AD Knowledge is untouched.

allaway:
Checks pass (i just synced with main so retriggered them....)

jay-hodgson (approved):

PR #2949, opened 2026-06-29
PORTALS-4332: Allow portal configuration to define how to sort the se…

jay-hodgson:
…arch hits

### DIFF
--- a/apps/portals/classic/src/config/synapseConfigs/studies.ts
+++ b/apps/portals/classic/src/config/synapseConfigs/studies.ts
@@ -71,6 +71,11 @@ export const studiesSearch: SearchQueryWrapperPlotNavProps = {
   shouldDeepLink: false,
   searchIndexId: studiesSearchIndexId,
   autocompleteFieldName: 'studyName',
+  initQueryRequest: {
+    query: {
+      sort: [{ column: 'metadataStatus', direction: 'ASC' }],
+    },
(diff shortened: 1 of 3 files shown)


jay-hodgson:
Screenshot 2026-06-29 at 1 16 14 PM

apps/portals/classic/src/config/synapseConfigs/studies.ts
```diff
@@ -71,6 +71,11 @@ export const studiesSearch: SearchQueryWrapperPlotNavProps = {
   shouldDeepLink: false,
   searchIndexId: studiesSearchIndexId,
   autocompleteFieldName: 'studyName',
+  initQueryRequest: {
+    query: {
+      sort: [{ column: 'metadataStatus', direction: 'ASC' }],
```
BryanFauble, line 76: Does this translate over to `columnName` that the API would be expecting?

https://rest-docs.synapse.org/rest/org/sagebionetworks/repo/model/search/SortField.html

BryanFauble (approved):
LGTM! Just a question on the translation layer

apps/portals/classic/src/config/synapseConfigs/studies.ts
jay-hodgson: Yes

PR #2951, opened 2026-06-30
PORTALS-4334: resize & restyle v3 navbar search box

allaway:
## Summary

Adjusts the sticky-search navbar search box (`HeaderSearchBox variant="v3"`) to approximately match the NF Research Tools Database design in Figma, with one deliberate deviation for legibility.

**Figma reference:** https://www.figma.com/design/aSr8SUVMX0EETgF3Qxvyrw/NF-Research-Tools-Database?node-id=2577-3277

## Changes

In the shared `HeaderSearchBoxV3.module.scss` and `_Navbar.scss`:

- **Height** — the box used the default small MUI input (~24–40px), which looked disproportionately short against the navbar strip. Set to a fixed `44px` on all breakpoints.
- **Mobile MENU spacing** — the mobile sticky-search navbar is a fixed-height `flex-wrap` container, so its default `align-content: stretch` injected free vertical space *between* the search box and the "MENU" link. When the box grew taller, that stretch (not padding) is what pushed MENU down. Fixed by packing the wrapped lines to the top (`align-content: flex-start`) and reducing the MENU top padding, so the taller box and MENU stay close and MENU is no longer displaced.
- **Resting border** — changed from `#eee` to `--synapse-gray-500`. This is a slight, intentional departure from the Figma: at `#eee` the unhighlighted (unfocused) search bar was very low-contrast against the near-white navbar and hard to see. The focused state stays primary blue.

## Scope

This is a shared v3-navbar change, so it applies to the portals that use the `with-sticky-search` navbar layout:

- **NF** (always)
- **ELITE** (always)
- **AD Knowledge Portal** (V2 layout only)

All other portals use the `default` navbar layout and are unaffected.

## Test plan

- [x] Ran NF locally and verified the desktop navbar search box height and resting/focused border against the design.
- [x] Verified on mobile widths that the taller box is preserved and the "MENU" link sits directly below it without being pushed down (compared against a pre-PR build side by side).
- [x] Spot-checked ELITE (shares the v3 box) at desktop and mobile widths.
- [ ] ADKP (V2) uses the same shared navbar component; its V2 layout is behind the `ADKP_HOMEPAGE_V2` feature flag.

### DIFF
--- a/apps/synapse-portal-framework/src/components/HeaderSearchBoxV3.module.scss
+++ b/apps/synapse-portal-framework/src/components/HeaderSearchBoxV3.module.scss
@@ -34,6 +34,7 @@
   :global(.MuiInputBase-root) {
     border-radius: 30px;
     background: #fcfcfc;
+    height: 44px;
   }
 
   :global(.MuiInputBase-input)::placeholder {
--- a/apps/synapse-portal-framework/src/components/HeaderSearchBoxV3.module.scss
+++ b/apps/synapse-portal-framework/src/components/HeaderSearchBoxV3.module.scss
@@ -34,10 +34,6 @@
   :global(.MuiInputBase-root) {
     border-radius: 30px;
     background: #fcfcfc;
-    // The default MUI small input (~40px) looks short against the navbar strip.
-    // Fixed rather than derived from --synapse-nav-top-height, which doubles at
-    // the mobile breakpoint and would overflow the mobile navbar grid.
-    height: 44px;
   }
 
   :global(.MuiInputBase-input)::placeholder {
--- a/apps/synapse-portal-framework/src/components/HeaderSearchBoxV3.module.scss
+++ b/apps/synapse-portal-framework/src/components/HeaderSearchBoxV3.module.scss
@@ -27,11 +27,16 @@
 
 .searchField {
   border-radius: 30px;
-  border: 1px solid #eee;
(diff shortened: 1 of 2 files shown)


jay-hodgson (approved):

jay-hodgson (changes requested):
Just spot checked ELITE, and NF, and this height (based on this calculation) causes a bug on mobile.

Image

allaway:
@jay-hodgson thanks for the look! Just to be clear, is the bug that the search bar is now below the logo instead of next to it? or something else?

jay-hodgson:
> @jay-hodgson thanks for the look! Just to be clear, is the bug that the search bar is now below the logo instead of next to it? or something else?

Apologies for not being more clear.  The mobile 'Menu' drop-down link has been pushed down out of the area, and is now barely visible.

allaway:
Fixed in d242ecd. Root cause was the mobile sticky-search navbar being a fixed-height `flex-wrap` container: its default `align-content: stretch` injected free space between the search box and MENU, so the taller box pushed MENU down. Switched to `align-content: flex-start` (+ trimmed the MENU top padding) so the box and MENU stay packed together. The taller 44px box is kept on all breakpoints; verified on mobile against a pre-PR build that MENU is no longer displaced, and spot-checked ELITE at desktop and mobile. Ready for another look when you have a sec.

jay-hodgson (approved):

PR #2952, opened 2026-06-30
PORTALS-4313: Entity Page Redesign:  Update styles for EntityActionMenu

kianamcc:
Jira: https://sagebionetworks.jira.com/browse/PORTALS-4313?atlOrigin=eyJpIjoiZjUyNTBiOWRiNzdlNDAzNGJlYWQ1NzUxNTBlZTJiY2UiLCJwIjoiaiJ9

SWC implementation (has pictures): https://github.com/Sage-Bionetworks/SynapseWebClient/pull/5978

### DIFF
--- a/packages/synapse-react-client/src/components/DropdownSelect/DropdownSelect.tsx
+++ b/packages/synapse-react-client/src/components/DropdownSelect/DropdownSelect.tsx
@@ -44,6 +44,7 @@ export default function DropdownSelect(props: DropdownSelectProps) {
     buttonGroupAriaLabel,
     onButtonClick,
     buttonText,
+    sx,
     anchorRef: externalAnchorRef,
     ...rest
   } = props
@@ -88,6 +89,9 @@ export default function DropdownSelect(props: DropdownSelectProps) {
         variant={variant}
         ref={anchorRef}
         aria-label="split button"
+        sx={{
+          ...sx,
+        }}
         {...rest}
       >
         <Button
--- a/packages/synapse-react-client/src/components/DropdownSelect/DropdownSelect.tsx
+++ b/packages/synapse-react-client/src/components/DropdownSelect/DropdownSelect.tsx
@@ -44,7 +44,6 @@ export default function DropdownSelect(props: DropdownSelectProps) {
     buttonGroupAriaLabel,
     onButtonClick,
     buttonText,
-    sx,
     anchorRef: externalAnchorRef,
     ...rest
   } = props
@@ -89,9 +88,6 @@ export default function DropdownSelect(props: DropdownSelectProps) {
         variant={variant}
         ref={anchorRef}
         aria-label="split button"
-        sx={{
-          ...sx,
-        }}
         {...rest}
       >
         <Button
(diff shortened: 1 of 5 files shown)


packages/synapse-react-client/src/components/DropdownSelect/DropdownSelect.tsx
```diff
@@ -88,6 +89,9 @@ export default function DropdownSelect(props: DropdownSelectProps) {
         variant={variant}
         ref={anchorRef}
         aria-label="split button"
+        sx={{
+          ...sx,
```
jay-hodgson: Why destructure + spread?

packages/synapse-react-client/src/components/entity/page/action_menu/EntityActionMenu.tsx
```diff
@@ -211,19 +218,45 @@ export default function EntityActionMenu(props: EntityActionMenuProps) {
...
     },
     items: mapAndFilterMenuGroups(
       layout.primaryMenuActions,
       actionConfiguration,
     ),
   }
 
+  // Optional per-layout style overrides applied to every action button (e.g. a
+  // different border color for entity pages vs the project header).
+  const menuButtonOverrides = (layout.menuButtonSx ?? {}) as unknown as Record<
```
jay-hodgson: Is double cast necessary?  Possible to use the MUI sx array form in the new Box below?

packages/synapse-react-client/src/components/DropdownSelect/DropdownSelect.tsx
kianamcc: This can actually be removed now, it was leftover when I originally set the border radius on ButtonGroup and had to destructure sx so {...rest} wouldn't override it. I switched to styling firstButton and lastButton for the corners so the destructure is no longer needed.

packages/synapse-react-client/src/components/entity/page/action_menu/EntityActionMenu.tsx
kianamcc: There was a type issue with nesting SxProps under the selector & .MuiButton-root. Looking into it to see if we can get rid of the double cast

kianamcc: we can drop double cast, nested selector is expecting SystemStyleObject<Theme> though

kianamcc: The reason why I put the styles here was so that it doesn't bleed into other components. However, if we are ok with it, we could apply the styles directly on the buttons in dropdownmenu/complexmenu so we don't have to do the whole type cast thing? I think search results cards buttons would adopt this new style as well though

jay-hodgson: Ok to leave the double cast if necessary, I just wanted to flag it!

jay-hodgson (approved):

PR #2966, opened 2026-07-09
PORTALS-4317: Entity Page Redesign: Add Sidebar, and move Entity TitleBarProperties into it (initially)

kianamcc:
Jira: https://sagebionetworks.jira.com/browse/PORTALS-4317?atlOrigin=eyJpIjoiNGEzMDgzNDY4ODQ1NDliY2FmZGFhMmFkNTljMzJjOTgiLCJwIjoiaiJ9

SWC (has images): https://github.com/Sage-Bionetworks/SynapseWebClient/pull/5985


packages/synapse-react-client/src/components/EntitySidebar/EntitySidebar.module.scss
```diff
@@ -0,0 +1,89 @@
+$sidebar-header-color: #616972;
```
kianamcc: Will try to think of a better name for this because it's used as button text color as well, not just header.

kianamcc:
I think we can keep the TitleBarProperties component as is for now until we figure out what to do with the conditions for use button which is called in that component. But let me know if you want just the properties removed or whatever else you prefer!

jay-hodgson (approved):

PR #2980, opened 2026-07-14
PORTALS-4360: NAM hub datasets

jay-hodgson:
See https://sagebionetworks.jira.com/browse/PORTALS-4360 for more information

### DIFF
--- a/apps/portals/namhub/src/pages/DatasetDetailsPage.tsx
+++ b/apps/portals/namhub/src/pages/DatasetDetailsPage.tsx
@@ -4,8 +4,11 @@ import { useGetPortalComponentSearchParams } from '@sage-bionetworks/synapse-por
 import { ColumnSingleValueFilterOperator } from '@sage-bionetworks/synapse-types'
 import { CardContainerLogic } from 'synapse-react-client/components/CardContainerLogic/CardContainerLogic'
 import QueryWrapperPlotNav from 'synapse-react-client/components/QueryWrapperPlotNav/QueryWrapperPlotNav'
-import { datasetsSql } from '@/config/resources'
-import { datasetCardConfiguration } from '@/config/synapseConfigs/datasets'
+import { datasetsSql, rgbIndex } from '@/config/resources'
+import {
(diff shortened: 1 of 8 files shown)


jay-hodgson:
Jineta confirmed this is the expectation, marking this draft PR as Ready for Review.

apps/portals/namhub/src/pages/DatasetDetailsPage.tsx
```diff
@@ -0,0 +1,56 @@
...
+import QueryWrapperPlotNav from 'synapse-react-client/components/QueryWrapperPlotNav/QueryWrapperPlotNav'
+import { datasetsSql } from '@/config/resources'
+import { datasetCardConfiguration } from '@/config/synapseConfigs/datasets'
+
+function DatasetDetailsPage() {
+  const { id } = useGetPortalComponentSearchParams()
+  return (
+    <DetailsPage
+      header={
+        <CardContainerLogic
```
kianamcc, line 18: It looks like the detail page is showing the raw 'projectId' label? If that is the case, I think we pass column alias to CardContainerLogic?

apps/portals/namhub/src/pages/DatasetDetailsPage.tsx
```diff
@@ -0,0 +1,56 @@
...
+      resourcePrimaryKey={['id']}
+    >
+      <DetailsPageContent
+        content={[
+          {
+            title: 'Files',
+            id: 'Files',
+            element: id ? (
+              <QueryWrapperPlotNav
+                rgbIndex={5}
```
kianamcc: minor,, maybe we can export and import here rgbIndex from datasets.ts

kianamcc (approved):

apps/portals/namhub/src/pages/DatasetDetailsPage.tsx
jay-hodgson: Nice catch!

kianamcc (approved):

PR #2996, opened 2026-07-17
PORTALS-4367

jay-hodgson:


kianamcc (approved):
Got it.. they want to keep that search term pill removable.

PR #3005, opened 2026-07-20
SWC-7884 - Upgrade OpenAPI schema

nickgros:


packages/synapse-react-client/src/features/curator/dashboard/components/CurationTaskCard.tsx
```diff
@@ -55,6 +55,24 @@ function useUiForTask(taskBundle: TaskBundle) {
         isPending,
         hasPermission,
       }
+    case 'org.sagebionetworks.repo.model.curation.execution.SampleSheetGenerationExecutionProperties':
```
nickgros, line 58: The rest of the changes in the file accommodate the spec change. For example, this switch needs to be exhaustive. Just using some placeholders for now

packages/synapse-react-client/src/features/entity/metadata-task/components/CreateOrUpdateCurationTaskDialog.tsx
```diff
@@ -171,9 +172,14 @@ export default function CreateOrUpdateCurationTaskDialog(
     string | undefined
   >(() => task?.assigneePrincipalId)
 
-  const initialAuthMode = toAuthorizationModeOption(
-    task?.taskProperties?.suggestedAuthorizationMode,
-  )
+  const initialAuthMode =
+    task?.taskProperties &&
+    instanceOfGridSupportedTaskProperties(task?.taskProperties)
```
nickgros, line 177: Many of these grid-task-specific hooks now need specific type guards

packages/synapse-react-client/src/features/entity/metadata-task/utils/types.ts
```diff
@@ -0,0 +1,18 @@
+import {
+  GridSupportedTaskProperties,
+  instanceOfFileBasedMetadataTaskProperties,
+  instanceOfRecordBasedMetadataTaskProperties,
+} from '@sage-bionetworks/synapse-client'
+
+export function instanceOfGridSupportedTaskProperties(
```
nickgros, line 7: Our generator does not create a function for this model, maybe because it's an interface only? Not sure. It's easy enough to make one but risks drift.

Might be fixed if we upgrade the generator, but that's always a risky, brittle change

afwillia (approved):

PR #3007, opened 2026-07-21
Improve Vitest VSCode config

nickgros:


synapse-web-monorepo.code-workspace
```diff
@@ -11,8 +11,8 @@
   ],
   "settings": {
     // General settings
-    "typescript.tsdk": "node_modules/typescript/lib",
+    "js/ts.tsdk.path": "node_modules/typescript/lib",
```
nickgros, line 14: This setting was renamed

synapse-web-monorepo.code-workspace
```diff
@@ -11,8 +11,8 @@
...
     // Vitest extension settings
-    "vitest.maximumConfigs": 99,
+    "vitest.workspaceConfig": "vitest.config.ts",
```
nickgros, line 16: `maximumConfigs` was removed, now Vitest instructs to use [projects](https://vitest.dev/guide/projects) (formerly known as workspaces, the extension still calls them that)

synapse-web-monorepo.code-workspace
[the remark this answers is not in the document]
nickgros: VSCode literally says this is deprecated and to use the new value

jay-hodgson (approved):

PR #3031, opened 2026-07-23
SWC-7849 - add curator help links

afwillia:
card
grid


packages/synapse-react-client/src/components/DataGrid/SynapseGrid.tsx
```diff
@@ -458,13 +458,26 @@ function SynapseGridInner({
...
+                    <GridMenuButton
+                      variant="outlined"
+                      startIcon={<HelpOutline />}
+                      onClick={() =>
+                        window.open(
+                          'https://docs.synapse.org/synapse-docs/managing-metadata-with-curator',
+                          '_blank',
+                          'noopener,noreferrer',
+                        )
+                      }
```
nickgros: Prefer using `href` (MUI `Button` supports it) which will turn this into an `a` tag which is more accessible & better user experience

```suggestion
                      href="https://docs.synapse.org/synapse-docs/managing-metadata-with-curator"
                      target="_blank"
                      rel="noopener noreferrer"
```

nickgros (approved):

PR #3032, opened 2026-07-23
PORTALS-4397

jay-hodgson:
### Problem
The portal search page previously auto-navigated to the tab with the **highest result count**.
This doesn't reflect relevance — a tab with many low-quality matches could win over a tab with
fewer but highly relevant results.

### Solution
Auto-navigate to the tab whose **top search hit has the highest relevance score** instead.
Fall back to the first tab when no tab has a score.

#### `SearchQueryUseQueryOptions.ts`
- **`searchQueryResultsToQueryResultBundle`**: includes `maxScore` (the relevance score of the
  first/top-ranked hit) as a non-standard field in the returned `QueryResultBundle` object.
- **`metadataQuery`** (inside `getSearchQueryUseQueryOptions`): now requests `HITS` with
  `size: 1` so the top hit's score is available in the metadata response. Previously `HITS` was
  omitted, making the score unreachable. Fetching only 1 hit keeps the overhead minimal; facet
  aggregation counts and `totalHits` are unaffected by `size`.

#### `PortalSearchTabs.tsx`
- Adds `score?: number` to `PortalSearchTabConfig` — tracked alongside the existing `count`.

#### `PortalSearchPage.tsx`
- Adds a `getMaxScore` helper that reads the `maxScore` field from the serialized bundle JSON.
- `onQueryResultBundleChange` now mutates and tracks `score` per tab in addition to `count`,
  using the same pattern already used for counts.
- The auto-navigation `reduce` now finds the tab with the **highest `score`** rather than the
  highest `count`. If no tab has a score, `maxScoreTab` is `undefined` and `searchPageTabs[0]`
  is used as the fallback. The existing `roleMapping` override is preserved.

### DIFF
--- a/apps/synapse-portal-framework/src/components/PortalSearch/PortalSearchPage.tsx
+++ b/apps/synapse-portal-framework/src/components/PortalSearch/PortalSearchPage.tsx
@@ -28,6 +28,13 @@ function getQueryCount(queryResultBundleJSON: string) {
   return queryCount
 }
 
+function getMaxScore(queryResultBundleJSON: string): number | undefined {
+  const bundle = JSON.parse(queryResultBundleJSON) as QueryResultBundle & {
+    maxScore?: number
+  }
+  return bundle.maxScore
+}
+
 export function PortalSearchPage(props: PortalSearchPageProps) {
   const {
     selectedTabIndex,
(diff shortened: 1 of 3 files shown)


nickgros (approved):
If one index has highly boosted fields and others do not, then that index will tend to win, since these scores aren't normalized. Still worth trying IMO

PR #3036, opened 2026-07-24
SWC-7903: Fix routing issue

nickgros:
Caused by bad relative `navigate`


packages/synapse-react-client/src/features/entity/metadata-task/components/MetadataTasksPageRouter.tsx
```diff
@@ -51,8 +40,8 @@ export default function MetadataTasksPageRouter(
             element: <MetadataTasksPageInternal projectId={projectId} />,
           },
           {
-            path: 'create/*',
-            element: <CreateRouteRenderer projectId={projectId} />,
+            path: 'create',
+            children: getCurationTaskFlowRoutes({ projectId, exitPath: '/' }),
```
nickgros, line 44: Refactor to use static routes (preferred over the dynamic component routes that this previously used)

packages/synapse-react-client/src/features/entity/metadata-task/components/MetadataTasksPageRouter.tsx
```diff
@@ -51,8 +40,8 @@ export default function MetadataTasksPageRouter(
             element: <MetadataTasksPageInternal projectId={projectId} />,
           },
           {
-            path: 'create/*',
```
nickgros, line 54: This `/*` caused the bug as it changes how relative navigation works in React router.

The flow is something like:

1. User is on `/Synapse:syn123/metadata/create` to create a task
2. User finishes creating the task, the component calls `navigate(`../confirmation/${task.taskId}`)`

Before: Browser navigates to `/Synapse:syn123/metadata/confirmation/${task.taskId}` (incorrect)
After: Browser navigates to `/Synapse:syn123/metadata/create/confirmation/${task.taskId}` (correct)

afwillia (approved):

PR #3045, opened 2026-07-27
PORTALS-4373: update openapi spec

jay-hodgson:
(includes the eDUC preview service)


nickgros (approved):
can you also update the @sage-bionetworks/synapse-client package version in this PR?

PR #3050, opened 2026-07-28
PORTALS-4386

jay-hodgson:
Adds the ability for ACT to select a DocuSign template.
Screenshot 2026-08-12 at 11 51 46 AM


packages/synapse-react-client/src/components/SetManagedAccessRequirementFields/SetManagedAccessRequirementFields.tsx
```diff
@@ -112,6 +115,20 @@ export const SetManagedAccessRequirementFields = forwardRef(
       return undefined
     }, [updatedAr?.ducTemplateFileHandleId, updatedAr?.id])
 
+    const { data: eDucTemplatePage, isLoading: isLoadingEDucTemplates } =
+      useListEDucTemplates()
```
nickgros: Need to handle pagination? `useInfiniteQuery` or `getAllOfPaginatedService`

nickgros (approved):

packages/synapse-react-client/src/components/SetManagedAccessRequirementFields/SetManagedAccessRequirementFields.tsx
jay-hodgson: You're absolutely right.  Let's use useInfiniteQuery with an effect (since getAllOfPaginatedService is in the ol SynapseClient.ts layer)

PR #3053, opened 2026-07-28
SWC-7890

nickgros:
image

image

Ready for review, but do not merge (backend dependency has not yet been merged)


packages/synapse-react-client/src/components/SynapseChat/components/AddFilesDialog/AddFilesDialog.tsx
```diff
@@ -0,0 +1,114 @@
...
+      onCancel={onClose}
+      title="Add files"
+      maxWidth="sm"
+      content={
+        <>
+          <BasicFileHandleUpload
+            ref={uploadRef}
+            allowMultipleUpload={true}
+            disableFolderUpload={true}
+            disableDragAndDrop={true}
```
jay-hodgson, line 69: Does DnD not work in this case?

packages/synapse-react-client/src/components/SynapseChat/GridAgentChat.tsx
```diff
@@ -35,13 +40,15 @@ export function GridAgentChat({
...
   const chatState = useChatState(agentSession)
+  const useGridAgentV2 = useGetFeatureFlag(FeatureFlagEnum.GRID_AGENT_V2) // TODO: remove this once the feature flag is fully rolled out
 
   // Create session context for grid sessions
   const sessionContext: GridAgentSessionContext = {
     concreteType:
       'org.sagebionetworks.repo.model.agent.GridAgentSessionContext',
     gridSessionId,
     usersReplicaId,
+    experimental: useGridAgentV2,
```
jay-hodgson, line 53: Does setting `experimental` in the `GridAgentSessionContext` have any unintended side effects?

packages/synapse-react-client/src/components/SynapseChat/GridAgentChat.tsx
```diff
@@ -61,6 +68,7 @@ export function GridAgentChat({
         setExternalSession={setAgentSession}
         externalChatState={chatState}
         suggestedPrompts={suggestedPrompts}
+        allowAttachments={useGridAgentV2}
```
jay-hodgson, line 73: Setting `allowAttachments` to `useGridAgentV2` is more clear, but I'm not sure what other features useGridAgentV2 is guarding

jay-hodgson (changes requested):
Check test failure(s), and verify that this works properly with the actual backend API.  Otherwise it looks great (just had a few questions)!

packages/synapse-react-client/src/components/SynapseChat/components/AddFilesDialog/AddFilesDialog.tsx
nickgros: We don't actually have any DnD logic in synapse-react-client right now. This boolean just controls the displayed copy. The DnD logic lives in SWC's [GlobalApplicationStateImpl](https://github.com/Sage-Bionetworks/SynapseWebClient/blob/2ef9f64faa656aa12644cb66672afe0811b1f442/src/main/java/org/sagebionetworks/web/client/GlobalApplicationStateImpl.java#L504). To support DnD here, we would likely want to replace that altogether with DnD attached to the specific React components that should support DnD file upload (the Uploader, EntityTreeTable, others?)

afwillia (approved):
Approved conditional on fixing the broken CI!

packages/synapse-react-client/src/components/SynapseChat/GridAgentChat.tsx
nickgros: `experimental` opts-in to using the new multi-agent architecture. Using this is required to support attachments.

Added a clarifying comment.

packages/synapse-react-client/src/components/SynapseChat/GridAgentChat.tsx
nickgros: On the backend it causes you to use the new multi-agent architecture, where the user's messages are picked up by a specific "Supervisor" agent (Curie) that has access to sub-agents.

The new supervisor has different prompts and more tools than the existing agent, but they are API compatible with two exceptions:

- Attachment support (experimental only)
- The trace information will be in a different format. Our code should already tolerate different trace formats, since it can change between models (e.g. Sonnet 3 -> 4 went from XML to JSON)

jay-hodgson (approved):

PR #3056, opened 2026-07-29
PORTALS-4401

jay-hodgson:
Show the "completion" status indicating if the NIH_RESEARCHER_AUTH_SERVICE oauth provider is linked to the current account.

Screenshot 2026-07-29 at 10 32 54 AM

Screenshot 2026-07-29 at 10 19 59 AM


packages/synapse-types/src/IdentityProvider.ts
```diff
@@ -0,0 +1,19 @@
+// https://rest-docs.synapse.org/rest/org/sagebionetworks/repo/model/auth/OAuthIdentityProvider.html
+
+export type OAuthProvider =
```
nickgros, line 3: Since these are all pure `type`s it is safe to make this change without bumping the synapse-types version

nickgros (approved):

PR #3059, opened 2026-07-29
PORTALS-4327: Entity Page Redesign:  Entity thread list table

kianamcc:
Jira: https://sagebionetworks.jira.com/browse/PORTALS-4327?atlOrigin=eyJpIjoiZTc3MzFkYTc3YzhjNGY3MmI0ZjAwMjlmNjA1NzRmYWYiLCJwIjoiaiJ9

SWC: https://github.com/Sage-Bionetworks/SynapseWebClient/pull/6017


packages/synapse-react-client/src/components/Forum/DiscussionEmpty.tsx
```diff
@@ -0,0 +1,30 @@
...
+}
+
+function DiscussionEmpty({ onViewForumClicked }: DiscussionEmptyProps) {
+  return (
+    <Stack className={styles.container}>
+      <Illustrations.Discussions2 />
+      <Box>
+        <Typography className={styles.title}>No discussions yet</Typography>
+        <Typography className={styles.subtitle}>
+          Mentions of this file will be shown here.{' '}
```
jay-hodgson, line 16: Do we know that the entity is always a file?

jay-hodgson (approved):

packages/synapse-react-client/src/components/Forum/DiscussionEmpty.tsx
kianamcc: The only entity that renders a discussion section today are file entities and so this component was made to only render if isFile is true, but if we want to render this UI on different pages as well we can update it!

PR #3062, opened 2026-07-30
SWC-7890: chat look & feel tweaks

nickgros:
- Track all messages with client-side ID. Enables optimistically tracking pending messages to avoid rendering a frame with an empty user message.
- Add an animation for new interactions
- Block send until the turn is complete

---


packages/synapse-react-client/src/components/SynapseChat/SynapseChat.test.tsx
```diff
@@ -163,6 +167,16 @@ describe('SynapseChat - suggestedPrompts', () => {
     })
   })
 
+  it('disables the send button while awaiting a response', async () => {
```
nickgros, line 170: Currently, there is a bug where you can send consecutive messages before you get a response back, but it's buggy and typically all messages except the first are lost. 

We must either disable the send button or queue messages up. I opted to disable the send button for now.

packages/synapse-react-client/src/components/SynapseChat/SynapseChat.tsx
```diff
@@ -125,7 +109,17 @@ export function SynapseChat({
 
   const internalChatState = useChatState(agentSession, onChatResponse)
   const chatState = externalChatState ?? internalChatState
-  const { pendingMessage, chatJobIds, sendChat } = chatState
+  const { interactions, isAwaitingResponse, sendChat } = chatState
```
nickgros, line 112: Instead of separating out `chatJobIds` and a `pendingMessage`, we combine the two into a single `interactions` list, which is optimistically updated when we send a message. This eliminates a few frames of a weird empty message/skeleton thing and IMO simplifies the logic in this component

packages/synapse-react-client/src/components/SynapseChat/SynapseChat.tsx
```diff
@@ -125,7 +109,17 @@ export function SynapseChat({
...
+
+  // Only interactions added after this component instance mounted should play the entry
+  // animation. Callers may lift state so history survives a `Dialog` closing, but the
+  // `Dialog` unmounts its children, so every reopen would otherwise re-animate the whole
+  // conversation.
+  const presentAtMountRef = useRef<Set<string> | undefined>(undefined)
```
nickgros, line 118: Added an animation when the chat DOM elements appear, but it should not always run (e.g. if you close and open the chat dialog in a grid session)

packages/synapse-react-client/src/components/SynapseChat/SynapseChatInteraction.module.scss
```diff
@@ -0,0 +1,25 @@
+// Entry animation for a chat bubble/response block mounting.
+$entry-duration: 500ms;
+$entry-easing: cubic-bezier(0.2, 0, 0, 1);
+$entry-translate-y: 10px;
```
nickgros, line 4: 500ms and 10px may seem dramatic but with this easing it's a pretty subtle animation. We can tweak it if needed but IMO it is better than nothing and makes chat a lot less jarring.

packages/synapse-react-client/src/components/SynapseChat/useChatState.ts
```diff
@@ -1,43 +1,82 @@
...
+  const { mutate: sendChatMessageToAgent, isPending: isAwaitingResponse } =
+    useSendChatMessageToAgent(
+      {
+        onMutate: (newChatMessage: AgentChatRequest) => {
+          const id = String(nextInteractionId.current++)
+          setInteractions(prev => [
+            ...prev,
+            { id, userMessage: newChatMessage.chatText },
+          ])
+        },
```
nickgros, line 46: This `onMutate` is essential new logic. Every time a user sends a message, assign it a client-generated ID and add it to the list of interactions.

The client-generated ID allows us to drive an optimistic update (so for the user, it feels like the message goes out instantly) and eliminates a lot of UI bugs (like not being able to show your message until the server accepts it).

It also gives us a single list of messages, so the view does not need to disambiguate between a complete message and a pending one.

packages/synapse-react-client/src/components/SynapseChat/useChatState.ts
```diff
@@ -1,43 +1,82 @@
...
+          if (pendingIndex === -1) {
+            return prev
+          }
+          const next = [...prev]
+          next[pendingIndex] = {
+            ...next[pendingIndex],
+            jobId: status.jobId,
+          }
+          return next
+        })
```
nickgros, line 77: Once we get a response from the async job polling, populate the interaction with the asyncJobId. The component that actually renders the message will take it from there.

packages/synapse-react-client/src/components/SynapseChat/SynapseChatInteraction.tsx
```diff
@@ -58,21 +63,31 @@ export function SynapseChatInteraction({
...
+  }, [scrollIntoView])
 
-  const isLoading = !chatResponseText && !chatErrorReason
+  useEffect(() => {
+    // once the response (or an error) arrives, scroll all the way to the bottom so the full
+    // answer - which may be taller than the viewport - is brought into view.
+    if (scrollIntoView && !isLoading) {
+      responseRef.current?.scrollIntoView({ behavior: 'smooth', block: 'end' })
+    }
+  }, [scrollIntoView, isLoading])
```
nickgros, line 90: Mitigates an issue where you can send a message that starts to trigger scrolling, but then the agent's response appears offscreen. This still isn't perfect, but this is closer.

packages/synapse-react-client/src/components/SynapseChat/SynapseChatInteraction.test.tsx
```diff
@@ -11,11 +11,19 @@ const defaultProps: SynapseChatInteractionProps = {
...
     wrapper: createWrapper(),
   })
 }
+
 describe('SynapseChatInteraction tests', () => {
+  const scrollIntoViewMock = vi.fn()
+
+  beforeEach(() => {
+    scrollIntoViewMock.mockClear()
+    window.HTMLElement.prototype.scrollIntoView = scrollIntoViewMock
```
jay-hodgson, line 24: Thanks for adding tests for the scrollIntoView interaction

jay-hodgson (approved):

afwillia (approved):

PR #3067, opened 2026-07-31
SWC-7911 bug fix pasting long strings clears cell

afwillia:
Fix is in https://github.com/sage-bionetworks/react-datasheet-grid so bump the SRC version


afwillia:
Closing now that the library has been incorporated into the monorepo

PR #3074, opened 2026-08-10
PORTALS-4312: Entity Page Redesign:  Update style of EntityTreeTable in SRC

kianamcc:
Jira: https://sagebionetworks.jira.com/browse/SWC-7918

Screenshot 2026-08-10 at 11 27 40 AM

Screenshot 2026-08-10 at 11 28 14 AM


packages/synapse-react-client/src/components/EntityTreeTable/components/EntityNameCell.tsx
```diff
@@ -67,6 +67,8 @@ export const NameCell: React.FC<CellContext<EntityBundleRow, unknown>> = ({
           entity={entityHeader}
           link={!disableEntityLinks}
           onEntityIdClicked={onEntityIdClicked}
+          iconColor="#4D535A"
```
jay-hodgson: Shame this is hard-coded.  Is there a way to get this from the theme?

jay-hodgson (approved):

packages/synapse-react-client/src/style/abstracts/_cssVariables.scss
```diff
@@ -71,6 +71,19 @@
   --synapse-gray-200: #f1f3f5;
   --synapse-gray-100: #fbfbfc;
 
+  // Redesign greyscale colors
+  --synapse-redesign-gray-50: #f9f9fa;
```
jay-hodgson: Since we may have a new set of gray colors in the future, what do you think about using "v2" instead of "redesign" in all of these variables?

packages/synapse-react-client/src/theme/palette/Palettes.ts
```diff
@@ -76,6 +76,19 @@ export const palette: PaletteOptions = {
     200: '#f1f3f5',
     100: '#fbfbfc',
   },
+  redesignGrey: {
```
jay-hodgson: This could be greyV2, for example

jay-hodgson (approved):

PR #3077, opened 2026-08-10
Fix broken lockfile and `pnpm dedupe`

nickgros:
Broke the lockfile merging #3069, this should fix it


pnpm-workspace.yaml
```diff
@@ -1,10 +1,10 @@
 packages:
```
nickgros, line 1: lines were re-ordered automatically by pnpm,  no changes

PR #3079, opened 2026-08-11
SWC-7901 - Add JSON Schema client methods and query hooks

nickgros:
- Add SynapseClient methods and KeyFactory keys for listing JSON
  Schema organizations, schemas, and schema versions
- Add useListOrganizations, useListJsonSchemasInfinite, and
  useListJsonSchemaVersions query hooks with tests
- Add MSW mocks and handlers for the new listing endpoints

---

### DIFF
--- a/packages/synapse-react-client/src/synapse-client/SynapseClient.ts
+++ b/packages/synapse-react-client/src/synapse-client/SynapseClient.ts
@@ -3716,15 +3716,11 @@ export const getAllOfPaginatedService = async <T>(
   const results: T[] = []
 
   while (existsMoreData) {
-    try {
-      const data = await fn(limit, offset)
-      results.push(...data.results)
-      offset += data.results.length
-      if (data.results.length < limit) {
-        existsMoreData = false
-      }
-    } catch (e) {
-      throw Error(`Error on getting paginated results ${e}`)
+    const data = await fn(limit, offset)
+    results.push(...data.results)
+    offset += data.results.length
+    if (data.results.length < limit) {
+      existsMoreData = false
     }
   }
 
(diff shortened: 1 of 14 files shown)


packages/synapse-react-client/src/synapse-client/SynapseClient.ts
```diff
@@ -3716,15 +3716,11 @@ export const getAllOfPaginatedService = async <T>(
...
   while (existsMoreData) {
-    try {
-      const data = await fn(limit, offset)
-      results.push(...data.results)
-      offset += data.results.length
-      if (data.results.length < limit) {
-        existsMoreData = false
-      }
-    } catch (e) {
-      throw Error(`Error on getting paginated results ${e}`)
```
nickgros, line 3727: Update these utilities to avoid unnecessarily wrapping the error.

For example, we may cancel the fetch and trigger an `AbortError`, and we want to detect it as such.

packages/synapse-react-client/src/synapse-queries/jsonschema/useListOrganizations.ts
```diff
@@ -0,0 +1,39 @@
...
+ */
+export function useListOrganizations(
+  options?: Partial<UseQueryOptions<Organization[], SynapseClientError>>,
+) {
+  const { synapseClient, keyFactory } = useSynapseContext()
+  return useQuery<Organization[], SynapseClientError>({
+    ...options,
+    queryKey: keyFactory.getListOrganizationsQueryKey(),
+    queryFn: ({ signal }) =>
+      getAllOfNextPageTokenPaginatedService<Organization>(
```
nickgros, line 21: Note: We do not have an organization search service today. This will fetch all organizations (<250 in production today) to support a client-side search.

Since this is a niche tool, I think this will work for now and add the service(s) we need if/when we see meaningful performance issues.

jay-hodgson: Just out of curiosity, what is the page size (which results in <250/x requests)?

jay-hodgson (approved):

packages/synapse-react-client/src/synapse-queries/jsonschema/useListOrganizations.ts
nickgros: Page size is 50

PR #3080, opened 2026-08-11
SWC-7901 - Add JSON Schema selection state hooks

nickgros:
- Add useJsonSchemaSelection to manage schema/version selection state
- Add useJsonSchemaOrganizations to load organizations for selection
- Add VersionSelectionType to distinguish required vs. optional
  version selection

---

### DIFF
--- a/packages/synapse-react-client/src/components/JsonSchema/JsonSchemaPicker/VersionSelectionType.ts
+++ b/packages/synapse-react-client/src/components/JsonSchema/JsonSchemaPicker/VersionSelectionType.ts
@@ -1,0 +1,16 @@
+/**
+ * Documents the possible JsonSchemaPicker version selection configurations.
+ */
+export enum VersionSelectionType {
+  /**
+   * A specific, numbered semantic version must be selected. This is the default. Use this when
+   * the consumer needs a reference to a specific version of the schema, rather than "whichever
+   * version is newest."
+   */
+  REQUIRED = 'REQUIRED',
+  /**
+   * A specific numbered version may be selected, or the user may choose "Latest" to always
+   * reference whichever version of the schema is newest at the time of use.
+   */
+  LATEST_ALLOWED = 'LATEST_ALLOWED',
+}
--- a/packages/synapse-react-client/src/components/JsonSchema/JsonSchemaPicker/useJsonSchemaOrganizations.ts
+++ b/packages/synapse-react-client/src/components/JsonSchema/JsonSchemaPicker/useJsonSchemaOrganizations.ts
@@ -1,0 +1,54 @@
+import { useListOrganizations } from '@/synapse-queries'
+import { Organization } from '@sage-bionetworks/synapse-client'
+import { useEffect, useMemo, useRef, useState } from 'react'
+
+/**
+ * Option displayed as a disabled placeholder row while Organizations are still being fetched,
+ * so the loading state is reflected within the option list itself. This avoids relying on the
+ * Autocomplete's `loading` prop, which blocks the list from use while loading.
+ */
+export const LOADING_ORGANIZATIONS_OPTION: Organization = {}
+
+/**
+ * Lists every JSON Schema Organization, and optionally preselects one matching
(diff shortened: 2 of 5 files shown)


packages/synapse-react-client/src/components/JsonSchema/JsonSchemaPicker/VersionSelectionType.ts
```diff
@@ -0,0 +1,16 @@
+/**
+ * Documents the possible JsonSchemaPicker version selection configurations.
+ */
+export enum VersionSelectionType {
```
nickgros, line 4: Similar to entity finder, we can support different 'cases' where a version should or should not be picked.

packages/synapse-react-client/src/components/JsonSchema/JsonSchemaPicker/useJsonSchemaSelection.ts
```diff
@@ -0,0 +1,179 @@
...
+    setSelectedVersionInfo(undefined)
+    onSelectionChange(
+      resolveSelection(newSelectedSchema, undefined, versionSelectionType),
+    )
+
+    const requestId = ++requestIdRef.current
+
+    if (
+      newSelectedSchema &&
+      versionSelectionType === VersionSelectionType.REQUIRED
```
nickgros, line 146: Similar to the EntityFinder, the versions of the selected schema are lazy-loaded. The load is triggered at the time of selection. If selecting a version is required, then we fetch the versions in the click handler and pick the latest version after the fetch completes.

We track a `requestIdRef` to prevent a race where e.g. 

1. Selection 1 is started
2. Selection 2 is started
3. Selection 2's version fetch completes and `applyVersionSelection` is triggered
3. Selection 1's version fetch completes and `applyVersionSelection` is triggered, unexpectedly switching the user's selection back to (1)

packages/synapse-react-client/src/components/JsonSchema/JsonSchemaPicker/useJsonSchemaOrganizations.ts
```diff
@@ -0,0 +1,54 @@
...
+
+/**
+ * Lists every JSON Schema Organization, and optionally preselects one matching
+ * `defaultOrganizationName` once it becomes available.
+ */
+export function useJsonSchemaOrganizations(defaultOrganizationName?: string) {
+  const [selectedOrganization, setSelectedOrganization] =
+    useState<Organization | null>(null)
+
+  const { data, isLoading, isError, error } = useListOrganizations()
```
nickgros, line 20: As mentioned in #3079 , this fetches all organizations to support a client-side filter.

The UI flow is 

1. Pick an organization from the list
2. Pick a schema published by that organization

packages/synapse-react-client/src/components/JsonSchema/JsonSchemaPicker/useJsonSchemaOrganizations.ts
```diff
@@ -0,0 +1,54 @@
...
+    if (!hasAppliedDefaultOrganizationRef.current && defaultOrganizationName) {
+      const match = organizations.find(
+        org => org.name === defaultOrganizationName,
+      )
+      if (match) {
+        setSelectedOrganization(match)
+        hasAppliedDefaultOrganizationRef.current = true
+      }
+    }
+  }, [organizations, defaultOrganizationName])
```
nickgros, line 44: The picker will support a prop for a 'default' selected organization (which I think will be useful for ACT in the future JSON Schema AR editor).

We need this effect to conditionally sync the default value with the selected value _after_ the full list of organizations is fetched.

jay-hodgson: And in the unlikely event that the provided `defaultOrganizationName` does not match any `org.name`, it looks like no issue would be encountered (`setSelectedOrganization` would not be called, which is fine).

jay-hodgson (approved):

PR #3084, opened 2026-08-11
PORTALS-4325: Entity Page Redesign:  new Access and Permissions component

kianamcc:
Jira: https://sagebionetworks.jira.com/jira/software/c/projects/PLFM/boards/238?assignee=712020%3Aa7fbbeb0-d7a6-443a-9dd4-246363ff09dd&selectedIssue=PORTALS-4325

Screenshot 2026-08-11 at 2 07 51 PM
Screenshot 2026-08-11 at 2 08 11 PM
Screenshot 2026-08-11 at 2 08 53 PM
Not signed in:
Screenshot 2026-08-11 at 2 10 18 PM

### DIFF
--- a/packages/synapse-react-client/src/components/EntitySidebar/AccessAndPermissions.tsx
+++ b/packages/synapse-react-client/src/components/EntitySidebar/AccessAndPermissions.tsx
@@ -65,8 +65,10 @@ function AccessAndPermissions({
     ? 'contained'
     : 'outlined'
 
-  const checkIcon = <CheckCircleRoundedIcon sx={{ color: '#3E8379' }} />
-  const blockIcon = <BlockFlippedIcon sx={{ color: '#D42322' }} />
+  const checkIcon = (
+    <CheckCircleRoundedIcon className={styles.checkCircleIcon} />
+  )
+  const blockIcon = <BlockFlippedIcon className={styles.blockIcon} />
 
   const downloadIcon =
     canDownload && !hasUnmetAccessRequirement ? checkIcon : blockIcon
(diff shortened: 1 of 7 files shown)


kianamcc:
@jay-hodgson We have "request access" in both sidebar sections that open the same modal... is that ok?

jay-hodgson:
> @jay-hodgson We have "request access" in both sidebar sections that open the same modal... is that ok?

Does this new section mean that we can remove the existing Access entry?  (are there cases where the new component is not shown, where we would still want to show the other Access entry?)

packages/synapse-react-client/src/components/EntitySidebar/AccessAndPermissions.tsx
```diff
@@ -0,0 +1,148 @@
...
+    restrictionLevel !== undefined &&
+    restrictionLevel !== RestrictionLevel.OPEN
+  const accessButtonText = hasUnmetAccessRequirement
+    ? 'Request to download'
+    : 'View Terms'
+  const accessButtonVariant = hasUnmetAccessRequirement
+    ? 'contained'
+    : 'outlined'
+
+  const checkIcon = <CheckCircleRoundedIcon sx={{ color: '#3E8379' }} />
```
jay-hodgson: Move these color values to AccessAndPermissions.module.scss ?  Or perhaps even add to the theme?

jay-hodgson (approved):

kianamcc:
> (are there cases where the new component is not shown, where we would still want to show the other Access entry?)

This new section is rendered unconditionally. I think it covers the same access modal entry points so the old one could be removed but we would lose the access icon displays (though this is probably redundant anyways with the new sections status rows)

Screenshot 2026-08-12 at 12 53 33 PM

Screenshot 2026-08-12 at 12 55 02 PM

We could go ahead and remove them, leave it as is for now with duplication, or we could update hasAccess so that we only show the icon and keep only one entry point for opening the modal?

I'm leaning towards just removing haveAccess display in the About section.

packages/synapse-react-client/src/components/EntitySidebar/AccessAndPermissions.tsx
kianamcc: Moving to module scss but if there is reuse in the future I will move to theme.

PR #3085, opened 2026-08-12
SWC-7924

nickgros:
- fix style issue where remove button could be clipped. give the chip a margin that matches the button offset to ensure it is visible

|Before|After|
|---|---|
| image | image|

### DIFF
--- a/packages/synapse-react-client/src/components/SynapseChat/components/AttachmentChip/AttachmentChip.module.scss
+++ b/packages/synapse-react-client/src/components/SynapseChat/components/AttachmentChip/AttachmentChip.module.scss
@@ -1,11 +1,13 @@
 .chip {
   position: relative;
+  margin-top: var(--remove-button-offset-y, 5px);
+  margin-left: var(--remove-button-offset-x, 8px);
 }
 
 .removeBadge:global(.MuiButtonBase-root) {
   position: absolute;
-  top: -6px;
-  left: -6px;
+  top: calc(-1 * var(--remove-button-offset-y, 5px));
+  left: calc(-1 * var(--remove-button-offset-x, 8px));
   opacity: 0;
   transition: opacity 100ms;
 }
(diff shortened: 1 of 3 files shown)


packages/synapse-react-client/src/components/SynapseChat/components/AttachmentChip/AttachmentChip.module.scss
```diff
@@ -1,11 +1,13 @@
 .chip {
   position: relative;
+  margin-top: var(--remove-button-offset-y, 5px);
+  margin-left: var(--remove-button-offset-x, 8px);
 }
 
 .removeBadge:global(.MuiButtonBase-root) {
   position: absolute;
-  top: -6px;
-  left: -6px;
```
nickgros, line 8: The button is positioned such that the top and left edges of the button are outside of the chip container. To ensure it isn't clipped off by the containers' boundaries, add a bit of margin to the `.chip`. The margin is computed using the button offset to ensure consistency and also make it clear that the margin is related to the button location.

Another way to solve this would be with `overflow: visible`, but we would have to ensure we apply that style to every un-padded container that holds this element. `margin` keeps the change isolated, and we just need to tweak gap values in the `ChatAttachmentStrip` container.

packages/synapse-react-client/src/components/SynapseChat/components/ChatAttachmentStrip/ChatAttachmentStrip.module.scss
```diff
@@ -3,8 +3,12 @@
 $chip-min-width: 140px;
 
 .strip {
+  --remove-button-offset-x: 8px;
+  --remove-button-offset-y: 5px;
```
nickgros, line 7: Changed the offset from (-6, -6) to (-8, -5) just because I thought it looked slightly better.

packages/synapse-react-client/src/components/SynapseChat/components/ChatAttachmentStrip/ChatAttachmentStrip.module.scss
```diff
@@ -3,8 +3,12 @@
...
+
   display: flex;
-  gap: 10px;
+  column-gap: calc(10px - var(--remove-button-offset-x));
+  row-gap: calc(10px - var(--remove-button-offset-y));
```
nickgros, line 11: Since the chips now have an inner margin, use those margins to tweak the container's `gap` properties. This ensures the layout does not visually change.

afwillia (approved):

PR #3086, opened 2026-08-12
SWC-7925

nickgros:
Fix chat attachment filename disappearing while message is processing

pendingAttachments was withdrawn as soon as an interaction's jobId was assigned (on the first PROCESSING poll), not once the job completed, so there was a real window with neither the draft-time filename nor the server-resolved attachmentStatuses available. Gate on isLast instead, so the fallback filename survives sent -> processing -> completed for the last interaction.

### DIFF
--- a/packages/synapse-react-client/src/components/SynapseChat/SynapseChat.test.tsx
+++ b/packages/synapse-react-client/src/components/SynapseChat/SynapseChat.test.tsx
@@ -1,11 +1,19 @@
 import { useChatState } from '@/components/SynapseChat/useChatState'
-import { mockAgentSession, mockChatAttachment } from '@/mocks/chat/mockChat'
+import {
+  mockAgentChatRequest,
+  mockAgentSession,
+  mockChatAttachment,
+  mockChatJobStatus,
+} from '@/mocks/chat/mockChat'
+import usePollAsynchronousJob from '@/synapse-queries/asynchronous/usePollAsynchronousJob'
 import {
   useCreateAgentSession,
   useGetChatAgentTraceEvents,
   useUpdateAgentSession,
 } from '@/synapse-queries/chat/useChat'
+import { getUseQuerySuccessMock } from '@/testutils/ReactQueryMockUtils'
 import { createWrapper } from '@/testutils/TestingLibraryUtils'
+import { FileHandleAssociateType } from '@sage-bionetworks/synapse-client'
 import { render, screen } from '@testing-library/react'
 import userEvent from '@testing-library/user-event'
 import {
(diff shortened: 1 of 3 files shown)


packages/synapse-react-client/src/components/SynapseChat/SynapseChat.tsx
```diff
@@ -299,7 +298,10 @@ export function SynapseChat({
                   animateEntry={!presentAtMount.has(interaction.id)}
                   isAwaitingResponse={isAwaitingResponse}
                   pendingAttachments={
-                    isPending ? lastSentAttachments : undefined
+                    // Only one interaction can be in flight at a time (the composer is disabled
+                    // while isAwaitingResponse), so lastSentAttachments always corresponds to
+                    // whichever interaction is last.
+                    isLast ? lastSentAttachments : undefined
                   }
```
nickgros, line 305: `pendingAttachments` should reference the 'last' interaction. A 'pending' interaction that is separate state from the `interactions` list was removed in the last refactor--that is now simply just the last interaction in the list. So we use `isLast` and drop `isPending` altogether.

packages/synapse-react-client/src/components/SynapseChat/SynapseChat.test.tsx
```diff
@@ -303,7 +325,7 @@ describe('SynapseChat - allowAttachments', () => {
     expect(mockSendChat).toHaveBeenCalledExactlyOnceWith('hello', undefined)
   })
 
-  it('shows a rich attachment chip on the pending message once sent, since the polled async job (which only reports the bare fileHandleId) is not yet registered', async () => {
+  it('shows a rich attachment chip on the pending message once sent, before the interaction has a jobId', async () => {
```
nickgros, line 328: This existing test handles the first 'intermediate state' - the message was sent, but Synapse hasn't assigned it a job ID.

packages/synapse-react-client/src/components/SynapseChat/SynapseChat.test.tsx
```diff
@@ -325,8 +347,76 @@ describe('SynapseChat - allowAttachments', () => {
 
     // The composer's own inline chip is cleared on send (see the "clears both" test above), but
     // the pending turn in the message list should still show a rich chip, via SynapseChat's
-    // lastSentAttachments state, until the interaction gains a jobId and SynapseChatMessage
-    // takes over from the polled async job's requestBody.
+    // lastSentAttachments state.
+    expect(screen.getByText('report.pdf')).toBeInTheDocument()
+  })
+
+  it('keeps showing the filename once the interaction gains a jobId, while the job is still PROCESSING', async () => {
```
nickgros, line 354: And a regression test for the broken case. The message was sent and has a job ID, but is still PROCESSING.

jay-hodgson (approved):

PR #3088, opened 2026-08-13
SWC-7928 reorder grid columns

afwillia:
Add a button to the grid that opens a dialog to reorder columns
button
dialog

### DIFF
--- a/packages/synapse-react-client/src/components/DataGrid/SynapseGrid.tsx
+++ b/packages/synapse-react-client/src/components/DataGrid/SynapseGrid.tsx
@@ -472,16 +472,7 @@ function SynapseGridInner({
                     spacing={1}
                     sx={{ justifyContent: 'flex-end' }}
                   >
-                    <GridMenuButton
-                      variant="outlined"
-                      startIcon={<HelpOutline />}
-                      href="https://docs.synapse.org/synapse-docs/managing-metadata-with-curator"
-                      target="_blank"
-                      rel="noopener noreferrer"
-                      sx={{ mr: 'auto' }}
-                    >
-                      Help and Instructions
-                    </GridMenuButton>
+                    {/* Keep this sync indicator first in the stack, otherwise it will cause other buttons to shift */}
                     {(!hasCompletedInitialSync || isSyncing) && (
                       <Box
                         sx={{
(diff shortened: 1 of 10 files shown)


packages/synapse-react-client/src/components/DataGrid/utils/computeDefaultColumnOrder.ts
```diff
@@ -0,0 +1,32 @@
+import { JSONSchema7 } from 'json-schema'
+
+/**
+ * Computes the "default" column display order: the order properties are defined in the JSON schema,
+ * with any columns not present in the schema appended afterward in their identity-index order.
+ */
+export function computeDefaultColumnOrder(
+  columnNames: string[],
+  jsonSchema: JSONSchema7 | undefined,
+): number[] {
```
nickgros, line 12: This is a good idea! You might also consider including an optional `upsertKeyColumnNames` parameter and make sure that those are always first, when provided.

packages/synapse-react-client/src/components/DataGrid/SynapseGrid.tsx
```diff
@@ -487,6 +498,12 @@ function SynapseGridInner({
                     )}
```
nickgros, line 489: This is kind of unrelated but while you're here can you move this status so it's first in this `Stack` (and add a comment that it should stay there)?

The reason is because this appears and disappears, so the position of the "Help and Instructions" button doesn't stay consistent. A user could try to click it right as it moves.

packages/synapse-react-client/src/components/DataGrid/components/ReorderColumnsButton.tsx
```diff
@@ -0,0 +1,45 @@
...
+          onSave={newColumnOrder => {
+            onReorder(newColumnOrder)
+            setShowDialog(false)
+          }}
+          onCancel={() => setShowDialog(false)}
+        />
+      )}
+      <GridMenuButton
+        variant="outlined"
+        startIcon={<SwapVert />}
```
nickgros: May want to use SwapHoriz (or get design feedback) on this. SwapVert I think is closer to how the dialog UX works but SwapHoriz is probably closer to how a user would find the button they want (because they want to move the actual columns left or right)

nickgros (approved):
Looks great, had a few suggested enhancements but that doesn't block merging if you want to go with this

nickgros (approved):

PR #3092, opened 2026-08-17
SWC-7862 curator grid remove columns

afwillia:
For record sets only, support column removal and restoration in the column reordering dialog.


packages/synapse-react-client/src/components/DataGrid/SynapseGrid.tsx
```diff
@@ -504,6 +507,7 @@ function SynapseGridInner({
                       columnOrder={modelSnapshot?.columnOrder ?? []}
                       jsonSchema={jsonSchema}
                       upsertKey={upsertKey}
+                      canRemoveColumns={isRecordSet}
```
nickgros, line 510: I was originally thinking that we would only let users remove columns that are not enumerated in the JSON Schema -- that way a user could change the grid after a sync to remove 'outdated' columns (after making sure data is migrated to the new columns), but they cannot actually make it so the grid does not match the data model (schema).

Can you check with Christina to identify the true requirements if we add this feature? Should we allow any grid editor to remove any column?

packages/synapse-react-client/src/components/DataGrid/components/ReorderColumnsDialog.tsx
```diff
@@ -38,46 +47,104 @@ export default function ReorderColumnsDialog(props: ReorderColumnsDialogProps) {
...
       open={open}
       title="Reorder Columns"
       content={
         <Box>
+          {canRemoveColumns && (
+            <Typography variant="body2" color="text.secondary" sx={{ mb: 1 }}>
+              Removed columns are hidden from the grid, but their data is
+              preserved.
+            </Typography>
+          )}
```
nickgros, line 102: I'm ok with allowing columns to be recovered but I am wary of adding language like this because I don't want to make it a guarantee that the data is preserved. We may decide to drop it at a later date or as a result of a future grid interaction.

packages/synapse-react-client/src/components/DataGrid/components/ReorderColumnsButton.test.tsx
```diff
@@ -6,9 +6,9 @@ import ReorderColumnsButton from './ReorderColumnsButton'
 describe('ReorderColumnsButton', () => {
   const columnNames = ['a', 'b']
   const columnOrder = [0, 1]
+  // 'a' is not part of the schema, so it's the only column removable by default
```
nickgros, line 9: I was expecting to see a test that verifies that `a` can be removed but `b` cannot, because `b` is in the schema (matching the use case where the RecordSet schema is updated, and now there are outdated property names).


You'll have to collect the top-level properties of the schema to figure out what we should allow removing. We have an existing util for this: `getPossibleTopLevelPropertiesInObjectSchema`

You might also look at the backend utility function [`collectTopLevelProperties`](https://github.com/Sage-Bionetworks/Synapse-Repository-Services/blob/23403ad59b6898875ebe55ee281c9dad9dd427f8/lib/models/src/main/java/org/sagebionetworks/repo/model/schema/JsonSchemaProperties.java#L34) which was implemented more recently and I think handles more scenarios (e.g. `$ref`s). You might decide to update our utility to match this implementation.

afwillia:
@nickgros thanks for the pointer to the backend schema tool. I updated the SRC schema property parser to mirror that logic and ended up doing away with `getPossibleTopLevelPropertiesInObjectSchema`. Let me know what you think.

nickgros (approved):
Looks good!

PR #3093, opened 2026-08-17
PORTALS-4377

jay-hodgson:
Small update to the AR editor.  isDUCRequired must be set to true to use the eDUC template.
Screenshot 2026-08-17 at 2 30 19 PM
Screenshot 2026-08-17 at 1 42 13 PM

Screenshot 2026-08-17 at 1 42 49 PM
Screenshot 2026-08-17 at 1 42 30 PM
Note, fails to fetch due to CORS.  Verified that directly loading the synapse.org FHA portal servlet URL successfully downloads the file.  The preview should work properly in the staging/prod environment.
Screenshot 2026-08-17 at 2 34 09 PM

### DIFF
--- a/packages/synapse-react-client/src/utils/hooks/useFetchBlobUrl.ts
+++ b/packages/synapse-react-client/src/utils/hooks/useFetchBlobUrl.ts
@@ -1,0 +1,38 @@
+import { useEffect, useState } from 'react'
+
+/**
+ * Fetches a URL with credentials and returns a blob object URL.
+ * The object URL is revoked when the hook unmounts or the URL changes.
+ */
+export function useFetchBlobUrl(url: string | undefined): {
+  blobUrl: string | undefined
+  error: Error | undefined
+} {
+  const [blobUrl, setBlobUrl] = useState<string | undefined>()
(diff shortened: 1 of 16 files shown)


packages/synapse-react-client/src/utils/hooks/useFetchBlobUrl.ts
```diff
@@ -0,0 +1,38 @@
+import { useEffect, useState } from 'react'
+
+/**
+ * Fetches a URL with credentials and returns a blob object URL.
+ * The object URL is revoked when the hook unmounts or the URL changes.
+ */
+export function useFetchBlobUrl(url: string | undefined): {
```
nickgros, line 8: Please take a look at `usePreFetchResource` to see if it makes sense to extend an existing hook in that file, or to use some of the logic that's already there.

If we need a new hook like this one, we should use react-query instead of a bare `useEffect`

jay-hodgson: Updated to continue to be two hooks with different intents that now share an implementation utility (`useCreateUrlForData`)

nickgros (approved):

PR #3094, opened 2026-08-18
PORTALS-4378

jay-hodgson:

PR #3095, opened 2026-08-18
PORTALS-4403

jay-hodgson:
Builds off of https://github.com/Sage-Bionetworks/synapse-web-monorepo/pull/3094

Adds templates table, and eDUC templates tab to AR dashboard (visible to ACT only)

Screenshot 2026-08-18 at 3 39 52 PM
After clicking on the "Validate" button on each row.
Screenshot 2026-08-18 at 3 40 03 PM

Also, a dedicated set of stories for the templates table itself (all valid, all invalid, and empty).
Screenshot 2026-08-18 at 3 33 55 PM
Screenshot 2026-08-18 at 3 34 04 PM
Screenshot 2026-08-18 at 3 34 11 PM

### DIFF
--- a/packages/synapse-react-client/src/components/dataaccess/EDucTemplateTable.tsx
+++ b/packages/synapse-react-client/src/components/dataaccess/EDucTemplateTable.tsx
@@ -3,8 +3,16 @@ import {
   useListEDucTemplates,
 } from '@/synapse-queries'
 import { formatDate } from '@/utils/functions/DateFormatter'
-import { CheckCircleTwoTone, ErrorTwoTone } from '@mui/icons-material'
-import { Alert, Box, Button, Skeleton, Typography } from '@mui/material'
+import { CheckCircleTwoTone, ErrorTwoTone, Replay } from '@mui/icons-material'
+import {
+  Alert,
+  Box,
+  Button,
+  IconButton,
+  Skeleton,
+  Tooltip,
+  Typography,
+} from '@mui/material'
 import { EDucTemplate } from '@sage-bionetworks/synapse-client'
 import {
   createColumnHelper,
@@ -61,7 +69,7 @@ function ValidationCell(props: { templateId: string | undefined }) {
   const { templateId } = props
   const [enabled, setEnabled] = useState(false)
 
-  const { data, error, isFetching } = useGetEDucTemplateValidation(
+  const { data, error, isFetching, refetch } = useGetEDucTemplateValidation(
     templateId ?? '',
     { enabled: enabled && Boolean(templateId) },
   )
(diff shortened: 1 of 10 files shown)


packages/synapse-react-client/src/components/dataaccess/EDucTemplateTable.tsx
```diff
@@ -0,0 +1,194 @@
...
+  columnHelper.display({
+    id: 'validation',
+    header: props => <ColumnHeader {...props} title={'Validation status'} />,
+    cell: ({ row }) => <ValidationCell templateId={row.original.templateId} />,
+  }),
+]
+
+function ValidationCell(props: { templateId: string | undefined }) {
+  const { templateId } = props
+  const [enabled, setEnabled] = useState(false)
```
kianamcc: Can the naming for this be more explicit? hasRequestedValidation for example.

packages/synapse-react-client/src/components/dataaccess/EDucTemplateTable.tsx
```diff
@@ -0,0 +1,194 @@
...
+  )
+
+  if (!templateId) return null
+
+  if (!enabled) {
+    return (
+      <Button
+        variant={'outlined'}
+        size={'small'}
+        onClick={() => setEnabled(true)}
```
kianamcc: Once set to true does it stay true when switching between tabs?

packages/synapse-react-client/src/components/dataaccess/EDucTemplateTable.tsx
```diff
@@ -0,0 +1,194 @@
...
+      >
+        Validate
+      </Button>
+    )
+  }
+
+  if (isFetching) {
```
kianamcc, line 91: If a template is invalid, user fixes it in DocuSign and comes back, we don't validate it again right? So we continue to show a stale reason? Should we have some sort of re-validate button?

jay-hodgson: You're absolutely right.  I'll add this functionality...

packages/synapse-react-client/src/components/dataaccess/EDucTemplateTable.tsx
jay-hodgson: Renaming

packages/synapse-react-client/src/components/dataaccess/EDucTemplateTable.tsx
jay-hodgson: Validation states are reset when tab is changed (but result is probably cached).  I think this is ok for an ACT-only feature

jay-hodgson:
Implemented @kianamcc 's suggestion to provide the ability to revalidate:
Screenshot 2026-08-19 at 12 54 10 PM

kianamcc (approved):

PR #3103, opened 2026-08-24
PORTALS-4426: NF tools database re-architecture

BelindaBGarana:
cc @anngvu for awareness
Addresses nf-osi/nf-research-tools-schema#217 ("Review synapse-web-monorepo to see how we can get more info on cards for new NF tool types"), found while adding new facet columns to the NF tools data source (`syn51730943`).

**`tools.tsx` `secondaryLabels`:**
- Removed `cellLineDisease`/`animalModelDisease`/`diseaseType` and `modelofManifestation`/`animalModelOfManifestation` — none of these columns exist anymore. The LinkML migration on the NF research-tools-schema side (nf-osi/nf-research-tools-schema docs/MIGRATION.md) unified them into shared `geneticDisorder`/`manifestation` columns across resource types.
- Added `geneticDisorder`, `manifestation`, `organ`, `tumorType` (now shared across multiple resource types post-migration).
- Added facets for the 4 tool types that had **none at all**: Computational Tool, Organoid Protocol, Patient-Derived Model, Clinical Assessment Tool.
- Added Genetic Reagent's `insertName`/`insertSpecies`/`vectorType` (previously missing entirely) plus `selectableMarker`.
- Added a few newly-available facets for existing types: `tissue`/`resistance` (Cell Line), `animalState` (Animal Model), `conjugate` (Antibody).

Verified all of these columns exist on the live `syn51730943` view with real populated data across every resource type before opening this PR (ranging from partial to full population depending on the field, consistent with real-world data completeness).

**`resources.ts`:**
- `usageRequirementsSql` was still querying the retired `Resource` table (`syn26450069`, stopped being written to when Phase 7 of the migration retired it) — switched to `syn51730943`, which carries `usageRequirements` directly like every other tool field now.
- `toolsSearchQueryConfig`'s `fieldBoosts` had a `diseaseType` boost — that column doesn't exist post-migration (folded into `geneticDisorder`), so the boost was a silent no-op. Fixed to `geneticDisorder`.

**Not build-tested locally** — pre-commit lint tooling (oxlint) wasn't available in this checkout (dependencies not installed). `secondaryLabels` is typed `string[]` (no strict key union), so this is a low-risk content-only change, but please have CI/a reviewer confirm before merge.

https://claude.ai/code/session_012cwbpfDLVQqbNrio62qn1q

### DIFF
--- a/apps/portals/nf/src/config/resources.ts
+++ b/apps/portals/nf/src/config/resources.ts
@@ -18,7 +18,11 @@ export const observationsSql = 'SELECT * FROM syn51735464'
 export const investigatorSql = `SELECT investigatorName as "firstName", ' ' as "lastName", institution, investigatorSynapseId as "USERID" FROM syn51734029 WHERE (investigatorName IS NOT NULL OR investigatorSynapseId IS NOT NULL)`
 export const developmentPublicationSql = `SELECT * FROM syn51735467`
 export const fundingAgencySql = `SELECT funderName as "Funding Agency" FROM syn51734076`
-export const usageRequirementsSql = `SELECT usageRequirements as "Usage Restrictions" FROM syn26450069 WHERE usageRequirements IS NOT NULL`
+// syn26450069 (the legacy Resource table) was retired in Phase 7 of the LinkML
+// migration (nf-osi/nf-research-tools-schema docs/MIGRATION.md) and is no longer
+// written to -- usageRequirements now lives on syn51730943 (the live, unified
+// Tool view) directly, same as every other tool.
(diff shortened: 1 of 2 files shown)


BelindaBGarana:
I think the failed test should actually pass now (I was changing the MV a bit at the time, trying to add more facets. I think there may be a bug with MVs getting facets from the tables in their defining SQL? Documented attempts here https://github.com/nf-osi/nf-research-tools-schema/issues/247, filed ticket https://sagebionetworks.jira.com/browse/PLFM-9906)

BelindaBGarana:
Created PORTALS ticket for this https://sagebionetworks.jira.com/browse/PORTALS-4426, would appreciate a quick review @jay-hodgson since this will be highly visible next week at a workshop

jay-hodgson (approved):

PR #3104, opened 2026-08-24
PORTALS-4383: create in-flight eDUC table and add to the user access …

jay-hodgson:
…request history page

Note, this asks for all DARs and fliters client-side to eDUC requests (opened https://sagebionetworks.jira.com/browse/PLFM-9907 to move this to the server).  Will show nothing if no eDUC DARs are found

Screenshot 2026-08-24 at 9 40 11 AM

### DIFF
--- a/packages/synapse-react-client/src/components/dataaccess/UserAccessRequestHistory/InFlightEDucSignaturesTable.tsx
+++ b/packages/synapse-react-client/src/components/dataaccess/UserAccessRequestHistory/InFlightEDucSignaturesTable.tsx
@@ -1,0 +1,117 @@
+import { useListUserDataAccessRequests } from '@/synapse-queries'
+import ColumnHeader from '@/components/TanStackTable/ColumnHeader'
+import StyledTanStackTable from '@/components/TanStackTable/StyledTanStackTable'
+import { SkeletonTable } from '@/components/Skeleton'
+import { Alert, Box, Typography } from '@mui/material'
+import {
+  AccessRequestSummary,
+  AccessRequestSummaryStatusEnum,
+} from '@sage-bionetworks/synapse-client'
+import {
+  createColumnHelper,
+  getCoreRowModel,
+  useReactTable,
+} from '@tanstack/react-table'
+import { useMemo } from 'react'
+
+const IN_FLIGHT_STATUSES: readonly AccessRequestSummaryStatusEnum[] = [
+  'sent',
+  'delivered',
+  'completed',
+] as const
+
+const STATUS_DISPLAY: Record<string, string> = {
+  sent: 'Signatures pending',
+  delivered: 'Signatures pending',
+  completed: 'Ready to submit',
+}
+
+const columnHelper = createColumnHelper<AccessRequestSummary>()
+const columns = [
+  columnHelper.accessor('accessRequirementName', {
+    header: props => <ColumnHeader {...props} title={'Request type'} />,
+    enableSorting: false,
+    enableColumnFilter: false,
+  }),
+  columnHelper.display({
+    id: 'signatureProgress',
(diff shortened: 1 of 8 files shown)


packages/synapse-react-client/src/components/dataaccess/UserAccessRequestHistory/InFlightEDucSignaturesTable.tsx
```diff
@@ -0,0 +1,117 @@
...
+  } = useListUserDataAccessRequests()
+
+  const rows = useMemo(
+    () =>
+      (accessRequestList?.results ?? []).filter(
+        summary =>
+          summary.isEDuc === true &&
+          summary.status !== undefined &&
+          (IN_FLIGHT_STATUSES as readonly string[]).includes(summary.status),
+      ),
```
nickgros: Ideally server could allow us to provide a filter in the API, may be worth creating a ticket, or add a comment here with the ticket ID if it already exists

packages/synapse-react-client/src/components/dataaccess/UserAccessRequestHistory/InFlightEDucSignaturesTable.tsx
```diff
@@ -0,0 +1,117 @@
...
+/**
+ * Lists the requester's active eDUC signatures (routed / signed but not yet submitted).
+ * Renders nothing when the filtered list is empty.
+ */
+export function InFlightEDucSignaturesTable() {
+  const {
+    data: accessRequestList,
+    isLoading,
+    error,
+  } = useListUserDataAccessRequests()
```
nickgros: need to handle pagination (use infinite + load more) ?

nickgros (approved):

packages/synapse-react-client/src/components/dataaccess/UserAccessRequestHistory/InFlightEDucSignaturesTable.tsx
jay-hodgson: Yes, mentioned in the PR comment but not the code.  I'll update the code to include while I shift to using an infinite page (and load all pages for client-side filtering)

packages/synapse-react-client/src/components/dataaccess/UserAccessRequestHistory/InFlightEDucSignaturesTable.tsx
```diff
@@ -0,0 +1,135 @@
...
+    isFetchingNextPage,
+    fetchNextPage,
+  } = useListUserDataAccessRequestsInfinite()
+
+  // Auto-fetch remaining pages so client-side filtering sees the complete list.
+  useEffect(() => {
+    if (hasNextPage && !isFetchingNextPage) {
+      void fetchNextPage()
+    }
+  }, [hasNextPage, isFetchingNextPage, fetchNextPage])
```
nickgros: Not a fan of waiting for a re-render to trigger each next fetch, would prefer we instead do one of:

- A "Load More" button (might not be good with a client-side filter)
- Load more when we hit a ref at the bottom of the list
- Change the queryFn to fetch all pages inside the queryFn (e.g. the helper `getAllOfPaginatedService`)  and use `useQuery` instead of `useInfiniteQuery`

If we want to load all pages before we show the list, then the last option is the same functionally but more straightforward and performant.

jay-hodgson: I'll remove the auto-fetch effect.  I don't think the "Load More" button is a viable option for this case, but I'll implement one of the other 2 approaches.

nickgros (approved):

PR #3105, opened 2026-08-24
PORTALS-4425

jay-hodgson:


hallieswan (approved):
Looks good to me! Confirmed that searching for "anything" with the "Researcher" role directs me to the "Publications" tab on this branch, instead of the "Datasets" tab as on staging.

PR #3113, opened 2026-08-26
PORTALS-4430: NF age facet (years label)

BelindaBGarana:
## Changes
- NF portal `commonProps.ts`: `age` → `Donor Age (years)` alias.

## Removed from scope of this PR
- `RangeFacetFilterUI.tsx`: `DOUBLE` columns now use `RangeSlider` when `columnMin`/`columnMax` are available (mirrors the existing `INTEGER` branch), falling back to the plain `Range` inputs otherwise — same fallback behavior as before.
- `Range.tsx`: `step: 'any'` on both number inputs.
No behavior change for `INTEGER`/`DATE` facets or for `DOUBLE` facets that lack `columnMin`/`columnMax`.

## Context
While cleaning up the NF Data Portal's `Age` facet data (nf-osi/nf-research-tools-schema#268 has background — donor ages like `10M`/`1Y1M` were converted to fractional years, changing the Synapse column type from STRING to DOUBLE), found three small, related rough edges in this codebase:

1. **No sliding-scale UI for DOUBLE range facets.** `RangeFacetFilterUI` only uses the nice `RangeSlider` widget for `INTEGER`-typed columns; `DOUBLE` columns fell through to plain `From`/`To` text inputs. `RangeSlider` itself has no integer-only logic (just uses `Number()` throughout) — the restriction was purely this component's conditional, so this extends it to `DOUBLE` with a finer step (0.01) appropriate for fractional values.
2. **Native browser validation tooltip on decimal entry.** The plain-input `Range` component's number `TextField`s had no `step` set, defaulting to `step=1` — so typing a decimal value (like a fractional-year age) into the `To`/`From` fields triggered the browser's native step-mismatch constraint-validation tooltip. Added `step: 'any'`.
3. **No unit on the Age facet label.** Added `age: 'Age (years)'` to the NF portal's `columnAliases`.

### DIFF
--- a/packages/synapse-react-client/src/components/widgets/query-filter/RangeFacetFilterUI.tsx
+++ b/packages/synapse-react-client/src/components/widgets/query-filter/RangeFacetFilterUI.tsx
@@ -145,39 +145,6 @@ export function RangeFacetFilterUI(props: RangeFacetFilterProps) {
                   />
                 ))}
 
-              {columnType === 'DOUBLE' &&
-                (columnMin && columnMax ? (
-                  <RangeSlider
-                    key="RangeSlider"
-                    domain={[columnMin, columnMax]}
-                    initialValues={{
-                      min: parseFloat(currentMin ?? columnMin),
-                      max: parseFloat(currentMax ?? columnMax),
-                    }}
-                    // Fractional values are common for DOUBLE columns (e.g. an age in years
-                    // derived from a donor's age in months), so use a fine-grained step
-                    // rather than the integer step used for INTEGER columns.
-                    step={0.01}
-                    onApplyClicked={onRangeValueSelected}
-                  >
-                    {'>'}
-                  </RangeSlider>
-                ) : (
-                  <Range
-                    key="Range"
-                    type="number"
-                    initialValues={
-                      currentMin || currentMax
-                        ? {
-                            min: currentMin ? parseFloat(currentMin) : undefined,
-                            max: currentMax ? parseFloat(currentMax) : undefined,
-                          }
-                        : undefined
-                    }
-                    onApplyClicked={onRangeValueSelected}
-                  />
-                ))}
-
               {columnType === 'DATE' && (
                 <Range
                   key="Range"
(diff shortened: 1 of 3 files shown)


jay-hodgson:
Is a range slider with step of 0.01 usable?

jay-hodgson:
Is a slider shown for this case?  I think the column min and max values are unavailable for search indexes (which are required for a slider)

jay-hodgson:
Pulled down this branch and tested.  Documented the results here: https://sagebionetworks.jira.com/browse/PORTALS-4430?focusedCommentId=340319

packages/synapse-react-client/src/components/widgets/query-filter/RangeFacetFilterUI.tsx
```diff
@@ -145,6 +145,39 @@ export function RangeFacetFilterUI(props: RangeFacetFilterProps) {
...
+                    key="RangeSlider"
+                    domain={[columnMin, columnMax]}
+                    initialValues={{
+                      min: parseFloat(currentMin ?? columnMin),
+                      max: parseFloat(currentMax ?? columnMax),
+                    }}
+                    // Fractional values are common for DOUBLE columns (e.g. an age in years
+                    // derived from a donor's age in months), so use a fine-grained step
+                    // rather than the integer step used for INTEGER columns.
+                    step={0.01}
```
jay-hodgson: With a minimal range of 0 to 100 and a step size of 0.01, it would potentially require 10,000 arrow clicks to traverse the range.  This could be a usability nightmare, unless other steps are supported by the UI (like page-up/down).
I'm going to reject this PR for now (until this is manually tested)

jay-hodgson (changes requested):
See comments.  These changes do not fix https://sagebionetworks.jira.com/browse/PORTALS-4430, and may in fact introduce additional usability issues.

packages/synapse-react-client/src/components/widgets/query-filter/RangeFacetFilterUI.tsx
BelindaBGarana: agree, downscoped to label rename and looks good to me now in local build

jay-hodgson:
I added a comment, but it was irrelevant (the diff that github showed me included changes that are not currently part of the PR).  This column alias is actually included in 
https://github.com/Sage-Bionetworks/synapse-web-monorepo/pull/3117/changes#r3868088053
(which also fixes PORTALS-4433 and PORTALS-4434), so I think this PR can be closed.  
This PR title and description are inaccurate and include a lot of extra information that only conflates what change is actually being made.

BelindaBGarana:
replaced with https://github.com/Sage-Bionetworks/synapse-web-monorepo/pull/3117

PR #3116, opened 2026-08-26
PORTALS-4419

jay-hodgson:
### Motivation

The default `multi_match` clause applies `fuzziness: AUTO`, which allows up to 2 edits on tokens ≥ 5 characters. For a query like `syn12345`, this means the search will also match neighbouring IDs (`syn12344`, `syn12346`, `syn12340`, etc.) — almost never what a user wants when they paste an entity ID.

This mirrors the existing handling for double-quoted phrases: identifiers, like phrases, need to be treated as exact tokens rather than fuzzy free-text.

### Change

In `SearchQueryUseQueryOptions.ts`:

- Added a `containsSynapseId()` helper using `/\bsyn\d+(?:\.\d+)?\b/iu` to detect Synapse entity ID tokens (with optional `.version` suffix, case-insensitive, on word boundaries).
- Extended the existing quoted-phrase fast-path in `buildQueryClause` so queries containing a Synapse ID also route to `simple_query_string` regardless of the configured `queryStrategy`. Field boosts are preserved.

`simple_query_string` doesn't apply fuzziness by default, so IDs match exactly.

### Behavior

| Query                  | Before (multi_match)                 | After                                            |
|------------------------|--------------------------------------|--------------------------------------------------|
| `syn12345`             | fuzzy match — hits `syn12344` etc.   | `simple_query_string` — exact                    |
| `SYN12345`             | fuzzy                                | `simple_query_string` — exact (case-insensitive) |
| `syn12345.7`           | fuzzy                                | `simple_query_string` — exact, versioned         |
| `syn12345 cancer`      | fuzzy on both                        | `simple_query_string` — ID stays exact           |
| `synapse` / `synonyms` | unchanged (`multi_match`)            | unchanged (`multi_match`)                        |
| `syn` (no digits)      | unchanged                            | unchanged                                        |
| `asyn12345`            | unchanged (no word boundary)         | unchanged                                        |

### Tests

Added 9 tests to `SearchQueryUseQueryOptions.test.ts` covering:

- Bare IDs route to `simple_query_string` across every configured strategy.
- Case-insensitive detection (`SYN12345`).
- Versioned IDs (`syn12345.7`).
- Mixed ID + free-text (`syn12345 cancer`).
- Field-boost preservation on the ID path.
- Negative cases: `synapse`, `synonyms`, bare `syn`, and embedded `asyn12345` all keep the default `multi_match` behavior.

### DIFF
--- a/packages/synapse-react-client/src/components/SearchQueryWrapper/SearchQueryUseQueryOptions.ts
+++ b/packages/synapse-react-client/src/components/SearchQueryWrapper/SearchQueryUseQueryOptions.ts
@@ -128,6 +128,19 @@ function containsQuotedPhrase(text: string): boolean {
   return /"[^"]+"/u.test(text)
 }
 
+/**
+ * Returns true if the query text contains a Synapse entity ID token
+ * (e.g. `syn12345` or `syn12345.4`). `multi_match` with fuzziness AUTO can match
+ * neighbouring IDs (e.g. `syn12345` → `syn12344`), which is almost never desired
+ * for identifier lookups. Callers use this to route ID-containing queries to
+ * `simple_query_string`, which does not apply fuzziness.
+ */
+function containsSynapseId(text: string): boolean {
+  // `syn` followed by digits (optionally `.version`), on word boundaries so
+  // ordinary words like `synapse` or `synonyms` don't match.
+  return /\bsyn\d+(?:\.\d+)?\b/iu.test(text)
+}
+
 function buildQueryClause(
   queryText: string,
   config: SearchQueryConfig = {},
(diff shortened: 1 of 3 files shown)


packages/synapse-react-client/src/components/SearchQueryWrapper/SearchQueryUseQueryOptions.ts
```diff
@@ -128,18 +128,31 @@ function containsQuotedPhrase(text: string): boolean {
...
 }
 
+/**
+ * Returns true if the query text contains a Synapse entity ID token
+ * (e.g. `syn12345` or `syn12345.4`). `multi_match` with fuzziness AUTO can match
+ * neighbouring IDs (e.g. `syn12345` → `syn12344`), which is almost never desired
+ * for identifier lookups. Callers use this to route ID-containing queries to
+ * `simple_query_string`, which does not apply fuzziness.
+ */
+function containsSynapseId(text: string): boolean {
```
kianamcc, line 139: Would this be useful to have in our RegularExpressions.ts?

kianamcc (approved):

packages/synapse-react-client/src/components/SearchQueryWrapper/SearchQueryUseQueryOptions.ts
jay-hodgson: I'm so happy you pointed this out, as it was bothering me as well (just not to the point of updating).  I'll refactor it!

PR #3117, opened 2026-08-27
PORTALS-4433, PORTALS-4434

jay-hodgson:
### Problem
Selecting **"Not Assigned"** on a range facet (e.g. NF → Tools → Age) caused OpenSearch to fail with:

> `number_format_exception: For input string: "org.sagebionetworks.UNDEFINED_NULL_NOTSET"`

### Root cause
`RangeFacetFilter` represents the "Not Assigned" selection by setting both `min` and `max` to the sentinel string `VALUE_NOT_SET` (`org.sagebionetworks.UNDEFINED_NULL_NOTSET`). In [SearchQueryUseQueryOptions.ts](packages/synapse-react-client/src/components/SearchQueryWrapper/SearchQueryUseQueryOptions.ts), `toSearchIndexQuery` translated any range selection into an OpenSearch `range: { field: { gte, lte } }` clause — which fails to parse the sentinel string as a number/date.

### Fix
In [SearchQueryUseQueryOptions.ts](packages/synapse-react-client/src/components/SearchQueryWrapper/SearchQueryUseQueryOptions.ts):

- When `min === max === VALUE_NOT_SET`, emit a "field missing" clause instead of a range clause:
  ```json
  { "bool": { "must_not": [{ "exists": { "field": "<columnName>" } }] } }
  ```
- Extended the internal `FilterClause` union to include the new missing-field variant.
- Updated the per-facet aggregation "drop own column" filter so the new clause is correctly excluded from its own column's aggregation (keeping the column's counts wide).

### Tests
Added 3 tests in [SearchQueryUseQueryOptions.test.ts](packages/synapse-react-client/src/components/SearchQueryWrapper/SearchQueryUseQueryOptions.test.ts):

- Translates a range facet "Not Assigned" selection into a `must_not exists` `post_filter`.
- Translates "Not Assigned" for DATE range facets without applying date conversion.
- Excludes a "Not Assigned" range clause from its own column's aggregation filter.

Also updated a column alias for NF age (Belinda asserts in https://github.com/Sage-Bionetworks/synapse-web-monorepo/pull/3113 that this should be rendered as `Donor Age (years)` across the NF portal)

### DIFF
--- a/apps/portals/nf/src/config/synapseConfigs/commonProps.ts
+++ b/apps/portals/nf/src/config/synapseConfigs/commonProps.ts
@@ -8,6 +8,7 @@ export const columnAliases = {
   nf1Genotype: 'NF1 Genotype',
   nf2Genotype: 'NF2 Genotype',
   studyId: 'On Synapse',
+  age: 'Donor Age (years)',
 }
 
 export const searchConfiguration = {
--- a/packages/synapse-react-client/src/components/SearchQueryWrapper/SearchQueryUseQueryOptions.ts
+++ b/packages/synapse-react-client/src/components/SearchQueryWrapper/SearchQueryUseQueryOptions.ts
@@ -38,6 +38,7 @@ import dayjs from 'dayjs'
 import { useCallback, useMemo } from 'react'
 import { KeyFactory } from '@/synapse-queries/KeyFactory'
 import { SYNAPSE_ENTITY_ID_TOKEN_REGEX } from '@/utils/functions/RegularExpressions'
+import { VALUE_NOT_SET } from '@/utils/SynapseConstants'
 
 // ─── Local OpenSearch DSL types ─────────────────────────────────────────────
 // Narrow typings for the OpenSearch structures we build (query DSL, sort) and
@@ -76,6 +77,8 @@ export type SearchQueryConfig = {
 type FilterClause =
   | { terms: Record<string, string[]> }
   | { range: Record<string, { gte?: string; lte?: string }> }
+  // "Field is not set" — emitted when a range facet selects "Not Assigned".
+  | { bool: { must_not: [{ exists: { field: string } }] } }
 
 type MultiMatchClause = {
   multi_match: {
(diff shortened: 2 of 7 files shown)


jay-hodgson:
Verified that this fixes the reported issue when running the NF portal locally off this branch

apps/portals/nf/src/config/synapseConfigs/commonProps.ts
```diff
@@ -8,6 +8,7 @@ export const columnAliases = {
   nf1Genotype: 'NF1 Genotype',
   nf2Genotype: 'NF2 Genotype',
   studyId: 'On Synapse',
+  age: 'Donor Age (years)',
```
jay-hodgson, line 11: Unrelated column alias update

packages/synapse-react-client/src/components/widgets/Range.tsx
```diff
@@ -167,7 +167,7 @@ export function Range(props: RangeProps): React.ReactNode {
                 })
               }
               slotProps={{
-                htmlInput: { 'aria-label': 'min' },
+                htmlInput: { 'aria-label': 'min', autoComplete: 'off' },
```
jay-hodgson, line 170: An attempt to fix PORTALS-4434 (could not repro locally, so I could not manually verify that this will fix the browser autofill behavior)

kianamcc:
temp

packages/synapse-react-client/src/components/SearchQueryWrapper/SearchQueryUseQueryOptions.ts
```diff
@@ -324,6 +327,16 @@ export function toSearchIndexQuery(
         f.concreteType === FACET_COLUMN_RANGE_REQUEST_CONCRETE_TYPE_VALUE,
     )
     .forEach(f => {
+      // "Not Assigned" selection: the RangeFacetFilter UI sets both bounds to
+      // VALUE_NOT_SET. Translate to an OpenSearch "field missing" clause rather
+      // than emitting the sentinel string as a range bound (which would fail
+      // with NumberFormatException on numeric/date fields).
+      if (f.min === VALUE_NOT_SET && f.max === VALUE_NOT_SET) {
```
kianamcc, line 334: What happens if the user picks "Not Assigned" and then switches back to "Range" and clicks Apply without typing anything? Will it still hit this guard?

jay-hodgson: Nice catch!  This exposes an existing bug, where we seed inputs with NaN when switching from "Not Assigned" → "Range" and cause an unexpected filter (like "NaN to NaN", or "NaN to 300").  I'll fix this case as well, since I'm here

kianamcc (approved):

PR #3119, opened 2026-08-27
PORTALS-4413-follow-up

kianamcc:
Jira: https://sagebionetworks.jira.com/browse/PORTALS-4413

Screenshot 2026-08-27 at 11 29 20 AM

### DIFF
--- a/packages/synapse-react-client/src/components/SynapseChat/SynapseChat.tsx
+++ b/packages/synapse-react-client/src/components/SynapseChat/SynapseChat.tsx
@@ -34,8 +34,7 @@ import { ReactComponent as CurieAvatarHead } from '@/assets/illustrations/curie_
 const CURIE_GREETING = 'Hi! How can I help you today?'
 
 const CURIE_ACCESS_LEVEL_SUFFIX: Record<AgentAccessLevel, string> = {
-  [AgentAccessLevel.PUBLICLY_ACCESSIBLE]:
-    ' and can only read public data in Synapse.',
+  [AgentAccessLevel.PUBLICLY_ACCESSIBLE]: ' and can only read public data.',
   [AgentAccessLevel.READ_YOUR_PRIVATE_DATA]:
     ' and can read your public and private data in Synapse on your behalf.',
   [AgentAccessLevel.WRITE_YOUR_PRIVATE_DATA]:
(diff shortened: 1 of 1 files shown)


packages/synapse-react-client/src/components/SynapseChat/SynapseChat.tsx
```diff
@@ -33,6 +33,12 @@ import { ReactComponent as CurieAvatarHead } from '@/assets/illustrations/curie_
 
 const CURIE_GREETING = 'Hi! How can I help you today?'
 
+const CURIE_ACCESS_LEVEL_SUFFIX: Record<AgentAccessLevel, string> = {
+  [AgentAccessLevel.PUBLICLY_ACCESSIBLE]: ' and only reads public data.',
+  [AgentAccessLevel.READ_YOUR_PRIVATE_DATA]: '',
```
kianamcc: Curie chat defaults to public but just added other cases as placeholders for now.

jay-hodgson: Sounds good, but I would put something reasonable for each value.  Perhaps something like 'and can read data in Synapse on your behalf', or 'and can read and write data in Synapse on your behalf'?

jay-hodgson (approved):

packages/synapse-react-client/src/components/SynapseChat/SynapseChat.tsx
```diff
@@ -34,9 +34,12 @@ import { ReactComponent as CurieAvatarHead } from '@/assets/illustrations/curie_
 const CURIE_GREETING = 'Hi! How can I help you today?'
 
 const CURIE_ACCESS_LEVEL_SUFFIX: Record<AgentAccessLevel, string> = {
-  [AgentAccessLevel.PUBLICLY_ACCESSIBLE]: ' and only reads public data.',
-  [AgentAccessLevel.READ_YOUR_PRIVATE_DATA]: '',
-  [AgentAccessLevel.WRITE_YOUR_PRIVATE_DATA]: '',
+  [AgentAccessLevel.PUBLICLY_ACCESSIBLE]:
+    ' and can only read public data in Synapse.',
```
jay-hodgson: Let's maintain the  ' and can only read public data.' for this case (without the Synapse reference)

jay-hodgson (approved):

PR #3120, opened 2026-08-27
SWC-7934 Preserve schema types on CSV upload

afwillia:
When uploading a CSV to a grid, preserve the column types defined in the schema. In some cases the backend will guess which column type to use. For example, `entityId` will be coerced to `entityId` even if the schema defined it as a `string`.


packages/synapse-react-client/src/components/table/CsvPreview/reconcileCsvImportSchema.ts
```diff
@@ -0,0 +1,73 @@
...
+ *  - Otherwise, if the column is declared in a custom schema as a string, that declaration wins.
+ *  - Otherwise, if the column already exists (e.g. a RecordSet system/metadata column like
+ *    `entityId`, `id`, `path`, or `color`, which are not part of the custom schema), it's presumed
+ *    to be a string, since it would otherwise render as plain text (columns absent from a schema
+ *    default to a text cell).
+ * A column that is not already known by any of the above is a brand-new column, so the CSV-inferred
+ * type is used as-is. The string presumptions also leave wide text suggestions (MEDIUMTEXT,
+ * LARGETEXT) as-is, since those are already string types that hold more than a STRING can.
+ */
+export function reconcileCsvImportSchema(
```
nickgros, line 41: To me it seems like using a JSON Schema to come up with a sensible set of ColumnModel objects is a generalizable feature that might make more sense in the backend. If a user imports a CSV using the Python client, the Python client could also use these recommended column models.

One possible approach would be to extend [POST /table/upload/csv/preview/async/start](https://rest-docs.synapse.org/rest/POST/table/upload/csv/preview/async/start.html) to give [UploadToTablePreviewRequest](https://rest-docs.synapse.org/rest/org/sagebionetworks/repo/model/table/UploadToTablePreviewRequest.html) an optional parameter `schema$id`, where the client may reference the relevant JSON Schema. Relevant backend function would be [`RecordSetSchemaResolver.getReconciledSchema`](https://github.com/Sage-Bionetworks/Synapse-Repository-Services/blob/76ddb873baeaf3a2334ef672f0656f55341c976a/services/repository-managers/src/main/java/org/sagebionetworks/repo/manager/table/RecordSetSchemaResolver.java#L99) (this method is really similar and if we stick with a client-side approach, we may want to ensure these two utilities are aligned).

A different approach would be to convert the backend CSV import service to work use the engine the synchronize services use (which use the RecordSetSchemaResolver utility), which might be able to derive the correct column types on-the-fly. That might be a large project, though.

Maybe the first set of questions to answer is: 

1. Does the Python client support CSV import into a grid? If yes...
2. How does the Python client determine the column models today?
3. Would the Python client benefit from this kind of logic or behavior in the backend?

PR #3128, opened 2026-08-31
PORTALS-4432

nickgros:
- Update aridhia-client to expand spec. Spec is now derived from three components:
  - Hand-written gateway spec for token exchange
  - Full spec retrieved from the internet
  - Script that filters and fixes the full spec and merges the two to generate the final OpenAPI spec.
- ALS portal can use C-PATH dev environment with local environment vars
- Utils for general react-query/Aridhia integration


packages/aridhia-client/src/codegen/filterSpec.ts
```diff
@@ -1,20 +1,29 @@
-import { readFileSync, writeFileSync } from 'fs'
+import { mkdirSync, readFileSync, writeFileSync } from 'fs'
 import { join } from 'path'
 
-// Define the endpoints we want to keep
+// Endpoints kept from the vendored upstream FAIR spec (fair-api-full.json)
```
nickgros, line 4: @jay-hodgson : Was the previously committed `fair-api.json` the post-filtered spec?

I decided to commit the pre-filtered spec retrieved from the internet, and derive + .gitignore the filtered & fixed spec. IMO this makes it clear if upstream changes occur that would require changes to this script.

packages/synapse-react-client/src/aridhia-queries/AridhiaKeyFactory.ts
```diff
@@ -0,0 +1,70 @@
...
+import { QueryKey } from '@tanstack/react-query'
+
+/**
+ * Co-location of react-query keys for Aridhia (C-Path FAIR / RDCA-DAP) queries.
+ *
+ * Mirrors the session-isolation convention in `synapse-queries/KeyFactory.ts`: every key is
+ * prefixed with a hash of the current Synapse access token (or the literal `'anonymous'`), so
+ * one user's cached Aridhia data can never leak to the next user of the same browser tab.
+ */
+export class AridhiaKeyFactory {
```
nickgros, line 11: Not all methods are used yet but they will be in future PR(s)

CLAUDE.md
```diff
@@ -105,10 +105,19 @@ These are the standard tools and technologies that are used by multiple packages
 - react-router
 - MUI
 
-## Comments
-
-Default to no comment. Code should be self-explanatory through clear naming and structure; a comment is a last resort for genuinely non-obvious intent — a subtle invariant, a workaround for external behavior, or a "why this and not the obvious thing" that the code can't express on its own. Never comment to restate what the code does, and don't narrate routine logic. It is always useful to document regex patterns - i.e. what they are intended to match. When in doubt, leave it out.
-Never write "historical context" comments in a refactor — e.g. `// previously this used X`, `// changed from the old approach`, `// no longer does Y`. Once a PR merges, the old behavior is irrelevant and lives in git history. Comment what the code does now, not what it used to do.
+## Code Comments
```
nickgros, line 108: Expand on code comment rules (highly duplicated from similar rules I added to Synapse-Repository-Services)

packages/aridhia-client/src/codegen/filterSpec.ts
jay-hodgson: Yes, it was post-filtered, and not having the full json and filtering script checked in was keeping me up at night, so thank you!

apps/portals/ampals/.env
```diff
@@ -1,3 +1,13 @@
 VITE_PORTAL_NAME="ALS Knowledge Portal"
 VITE_PORTAL_DESCRIPTION="Welcome to the ALS Knowledge Portal, a comprehensive resource for ALS researchers. This platform is a core component of the AMP ALS initiative, a public private partnership among the National Institutes of Health (NIH), U.S. Food and Drug Administration (FDA), Foundation for the National Institutes of Health (FNIH), Critical Path Institute (C-Path), industry, academia, foundations, and advocacy groups. Explore rich datasets, build custom cohorts, and access the latest publications to advance your research. Join a collaborative community committed to rapidly advancing the development of effective new treatments for ALS!"
 VITE_PORTAL_KEY="ampals"
+
+# Aridhia / C-Path DAP. To point local dev at the C-Path dev hub, create
```
jay-hodgson, line 5: Excellent, a way to easily set the endpoint

jay-hodgson (approved):

PR #3130, opened 2026-08-31
PORTALS-4436: ARK Portal: Collections icons too large

kianamcc:
Jira: https://sagebionetworks.jira.com/browse/PORTALS-4436

Screenshot 2026-08-31 at 1 51 48 PM
Screenshot 2026-08-31 at 11 28 16 AM
Screenshot 2026-08-31 at 11 29 46 AM


packages/synapse-react-client/src/style/components/_cards.scss
```diff
@@ -200,15 +200,17 @@
   .SRC-imageThumbnail {
     vertical-align: top;
     padding-right: 15px;
-    max-width: 215px;
+    max-width: 150px;
```
kianamcc: @nickgros lmk if you want smaller!

apps/portals/adknowledgeportal/src/config/style/_style_overrides.scss
```diff
@@ -57,15 +57,6 @@ footer {
   }
 }
 
-.ProgramCardList {
```
kianamcc, line 60: removing dead code

nickgros:
I think your change is reasonable default behavior and looks a lot better in these default cases!

Some regressions I caught on visual inspection, and a suggestion on how to address each:

|Component|Before|After|Notes|
|---|---|---|---|
| AMP-ALS Collections |image |image | The images seem to be cropped in an offset to attempt to center the logo. After your changes, a centered logo would appear 'centered'. Consider reaching out to AMP-ALS owners about uploading re-cropped images (preferred) or adding an override. |
|NF Inititatives| image | image | Consider adding specific overrides so there is no gap between the image and the top and left of the card (in the _before_, there is a gap on the left of the card, but I think it would be good to remove it!). There is already an `.InitiativeCardList` selector but it might not be properly applied. |
| NF Our Partners (Home Page) | image | image | I think it looks fine after your changes, but it is different. If you choose to add an override to match the before, I would give it a bit more top margin than it originally had |

kianamcc:
@nickgros 
Screenshot 2026-09-01 at 4 50 03 PM
Screenshot 2026-09-01 at 4 50 37 PM
Screenshot 2026-09-01 at 4 53 05 PM
Screenshot 2026-09-01 at 5 01 14 PM

nickgros (approved):

PR #3132, opened 2026-08-31
move react-datasheet-grid package into monorepo

afwillia:


packages/react-datasheet-grid/src/utils/domParser.ts
```diff
@@ -1,8 +1,23 @@
...
+ * otherwise be picked up by `textContent` and land in a pasted cell value.
+ */
 export const parseDom = (html: string): Document => {
-  return parser.parseFromString(html, 'text/html')
+  return parser.parseFromString(
+    // <style> is allowed by default, but like <script> its text is not visible
+    // content, so it must not be able to reach a cell value via textContent.
+    DOMPurify.sanitize(html, { FORBID_TAGS: ['style'] }),
+    'text/html',
+  )
```
nickgros, line 22: This is never actually used to render HTML, so sanitization is not really necessary, but it's OK to do in case this is (mis)used in that way in the future. All this does is parse to look for `<table>` data.

packages/react-datasheet-grid/e2e/xssPaste.spec.ts
```diff
@@ -0,0 +1,131 @@
+import { test, expect, Page } from '@playwright/test'
+
+// Clipboard HTML is attacker-controlled: a user can be tricked into copying
+// crafted markup from a malicious page. It is parsed with DOMParser, which
+// produces an inert document, and only textContent is ever read from it — these
+// tests hold that boundary in a real browser, which is something JSDOM cannot
+// demonstrate (it does not execute scripts at all).
+test.describe('pasting hostile HTML', () => {
```
nickgros: Did these fail before the sanitization change? I am under the impression that the HTML document was never rendered, so these scripts would have never been executed in the first place. I'm 50/50 on whether we really need speculative tests like this.

packages/synapse-react-client/.storybook/main.mts
```diff
@@ -69,9 +69,30 @@ const config: StorybookConfig = {
...
+      plugins: (config.plugins ?? []).filter(
+        plugin =>
+          !(
+            plugin &&
+            typeof plugin === 'object' &&
+            'name' in plugin &&
+            plugin.name === 'storybook:react-docgen-plugin'
+          ),
+      ),
+    }
```
nickgros, line 91: not ideal to have to rip it out like this, but OK

nickgros (approved):

packages/react-datasheet-grid/e2e/xssPaste.spec.ts
afwillia: These didn't exist before the sanitization change. They were written to character a possible paste bug related to the domParser false positive. I agree we can do away with this.

PR #3134, opened 2026-09-01
SWC-7964 Show error message when 2FA reset fails

nickgros:
## Problem
In `useChangePasswordFormState`, the `resetTwoFactorAuth` mutation (triggered when a user requests a 2FA reset while changing their password) had no `onError` handler and did not expose its `error` state. If the reset request failed (e.g. 400 due to incorrect current password), the failure was silently swallowed — the user saw no feedback at all.

## Fix
- Destructure `error` from `useResetTwoFactorAuth()` and render an error `Alert` with the failure reason when the reset request fails during the 2FA-reset-confirmation step.
- Added `getResetTwoFactorAuthBadRequestHandler` MSW mock handler (mirrors the existing `getBadRequestChangePasswordHandler` pattern).
- Added a regression test (`shows an error message when requesting 2FA reset fails`) that drives the reset flow against the new failure handler and asserts the error message renders. Verified this test fails against the pre-fix code.

## Testing
`pnpm --filter synapse-react-client test ChangePassword.test.tsx` — 10/10 passing.

JIRA: SWC-7964

### DIFF
--- a/packages/synapse-react-client/src/components/ChangePassword/ChangePassword.test.tsx
+++ b/packages/synapse-react-client/src/components/ChangePassword/ChangePassword.test.tsx
@@ -590,7 +590,7 @@ describe('ChangePassword tests', () => {
     expect(otpInputs).toHaveLength(6)
   })
 
-  it('supports requesting 2FA reset using a password', async () => {
+  it('supports requesting 2FA reset using the current password', async () => {
     const userId = MOCK_USER_ID
     const twoFaToken = 'mock-2fa-token'
     server.use(
@@ -655,7 +655,8 @@ describe('ChangePassword tests', () => {
       expect(reset2faSpy).toHaveBeenCalledWith({
         userId: MOCK_USER_ID,
         twoFaResetEndpoint: expect.any(String),
-        // The current password must be used. A twoFaToken returned by the changePassword service cannot be used to reset 2FA.
+        // The current password must be used. A twoFaToken returned by the changePassword service cannot be used to
+        // reset 2FA: the 2fa/reset endpoint only accepts a twoFaToken minted for an AUTHENTICATION (login) challenge.
         password: currentPassword,
       })
 
--- a/packages/synapse-react-client/src/components/ChangePassword/ChangePassword.test.tsx
+++ b/packages/synapse-react-client/src/components/ChangePassword/ChangePassword.test.tsx
@@ -4,7 +4,10 @@ import {
   getSuccessfulChangePasswordHandler,
 } from '@/mocks/msw/handlers/changePasswordHandlers'
 import { getFeatureFlagsOverride } from '@/mocks/msw/handlers/featureFlagHandlers'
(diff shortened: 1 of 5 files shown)


nickgros:
New flow:

image

After click:

image

packages/synapse-react-client/src/components/ChangePassword/ChangePassword.test.tsx
```diff
@@ -587,7 +590,7 @@ describe('ChangePassword tests', () => {
     expect(otpInputs).toHaveLength(6)
   })
 
-  it('supports requesting 2FA reset using a password', async () => {
+  it('supports requesting 2FA reset using the current password', async () => {
```
nickgros, line 593: When investigating, the agent decided to add some extra comments and clarify some pronouns. I think it's reasonable clarification.

packages/synapse-react-client/src/components/ChangePassword/ChangePassword.test.tsx
```diff
@@ -661,4 +665,58 @@ describe('ChangePassword tests', () => {
       expect(confirmPasswordField).not.toBeInTheDocument()
     })
   })
+
+  it('shows an error message when requesting 2FA reset fails', async () => {
```
nickgros, line 669: Ensure we show the swallowed error message.

In the specific Service Desk ticket that prompted this PR, the user should have never been able to get this far. There are other cases where this shared code path should still show this error message.

packages/synapse-react-client/src/components/ChangePassword/ChangePasswordWithToken.test.tsx
```diff
@@ -401,4 +403,51 @@ describe('ChangePasswordWithToken tests', () => {
     const otpInputs = await getTOTPInputs()
     expect(otpInputs).toHaveLength(6)
   })
+
+  it('shows Service Desk recovery guidance instead of an interactive 2FA reset, since this flow has no credential the reset endpoint accepts', async () => {
```
nickgros, line 407: This is the test for the new case: `ChangePasswordWithToken` handles a user attempting to reset password with just an emailed token. We don't have enough info to let you reset 2FA, so we guide you to the service desk.

packages/synapse-react-client/src/components/ChangePassword/ChangePasswordWithToken.tsx
```diff
@@ -47,7 +59,40 @@ export default function ChangePasswordWithToken(
...
+                >
+                  Synapse Service Desk
+                </Link>{' '}
+                for help recovering your account.
+              </Typography>
+              <Typography variant={'body1'} sx={{ my: 2 }} align={'center'}>
+                If you have a Google or ORCID account linked to your Synapse
+                account, you may also be able to regain access by signing in
+                with that account instead.
+              </Typography>
```
nickgros, line 92: Please review copy

jay-hodgson: Seems accurate to me

jay-hodgson (approved):

xschildw (approved):

PR #3135, opened 2026-09-01
PORTALS-4432 - Aridhia hooks

nickgros:
Add types and hooks for each Aridhia request


packages/synapse-react-client/src/aridhia-queries/useGetAridhiaWorkflow.ts
```diff
@@ -0,0 +1,42 @@
...
+import { AridhiaKeyFactory } from './AridhiaKeyFactory'
+import { AridhiaError, useAridhiaQuery } from './useAridhiaQuery'
+
+/**
+ * Fetches information about a workflow's form fields (`about`/`project` sections), keyed by the
+ * workflow's own `key`. Callers derive `workflowKey` from `DatasetSettings.workflow_key`.
+ *
+ * Deliberately returns the raw response body instead of the generated model. openapi-generator's
+ * `FromJSON` mapper only copies the properties the OpenAPI spec declares for a given model,
+ * silently dropping any additional properties.
```
jay-hodgson, line 14: This suggests the open API spec for the fair services is inaccurate (returns fields not described in the spec), is that right?

jay-hodgson (approved):

packages/synapse-react-client/src/aridhia-queries/useGetAridhiaWorkflow.ts
nickgros: Yes. It required inspecting the network requests made by RDCA-DAP UI to achieve parity and figure out which calls were missing info. The spec also is lacking descriptions for properties, so it's not the most reliable source, but it's a decent starting point.

PR #3137, opened 2026-09-01
PORTALS-4441

jay-hodgson:
### `@sage-bionetworks/synapse-types`

**`packages/synapse-types/src/Table/QueryFilter.ts`**

- Extended `ColumnSingleValueFilterOperator` enum with the 8 new operators shipped by the backend:
  `GREATER_THAN`, `LESS_THAN`, `GREATER_THAN_OR_EQUAL`, `LESS_THAN_OR_EQUAL`,
  `NOT_EQUAL`, `IS_NULL`, `IS_NOT_NULL`, `BETWEEN`.
- Added `BooleanOperator = 'AND' | 'OR'`.
- Added `FILTER_GROUP_CONCRETE_TYPE_VALUE` constant + `FILTER_GROUP_CONCRETE_TYPE`
  type alias (`'org.sagebionetworks.repo.model.table.FilterGroup'`).
- Added `FilterGroup` interface (`operator`, `not`, `children`, `isDefiningCondition`),
  mirroring the shape from `@sage-bionetworks/synapse-client`.
- Extended the `QueryFilter` union to include `FilterGroup`.

### `synapse-react-client`

**`src/utils/types/IsType.ts`**

- New `isFilterGroup` type guard, following the existing hand-written
  `isTypeViaConcreteTypeFactory` pattern in that file.

**`src/utils/functions/SqlFunctions.ts`**

- Extended the `getAdditionalFilters` switch so URL-search-params → additional-filters
  handles every new operator:
  - `EQUAL` / `NOT_EQUAL` / `GREATER_THAN` / `LESS_THAN` / `GREATER_THAN_OR_EQUAL`
    / `LESS_THAN_OR_EQUAL` — single-value pattern with `[searchParams[key]]`.
  - `IN` / `BETWEEN` — multi-value via `splitAndTrim`.
  - `IS_NULL` / `IS_NOT_NULL` — `values: []`.
  - `LIKE` (unchanged), `HAS` / `HAS_LIKE` (unchanged).
- Added `operator satisfies never` exhaustiveness guard so future enum additions
  break the type-check at this exact site.
- Preserved the existing test contract — `getAdditionalFilters` now returns a
  valid filter for every operator (verified by the pre-existing
  "handles every filter type" test).

**`src/components/QueryBuilder/QueryBuilderTypes.ts`** (new)

- Internal QB UI types: `QBNode`, `QBGroup`, `QBCondition`, `QBConditionOp`,
  `QBCombinator`.
- `QBConditionOp` covers 15 ops: `is_any_of`, `is_all_of`, `between`, `gt`,
  `gte`, `lt`, `lte`, `equal`, `not_equal`, `has_value`, `no_value`, `contains`,
  `starts_with`, `ends_with`, `is_exactly`. Negation is expressed via a
  `QBGroup { not: true }` — there is no `is_none_of` leaf op.
- Type narrows: `isQBGroup`, `isQBCondition`.

**`src/components/QueryBuilder/queryBuilderTranslation.ts`** (new)

- `qbNodeToApiFilter(node)` — QB tree → `QueryFilter | null`.
  - Empty groups and incomplete conditions return `null` (partial input never
    reaches the outgoing query).
  - `is_any_of` on `*_LIST` columns emits `ColumnMultiValueFunctionQueryFilter { HAS }`;
    on single-value columns emits `ColumnSingleValueQueryFilter { IN }`.
  - `is_all_of` emits `HAS` (multi-value list columns only).
  - Range ops emit their matching `ColumnSingleValueFilterOperator`; `BETWEEN`
    packs `[rangeMin, rangeMax]`.
  - Text ops escape `%` / `_` / `\` in user text so a literal `50%` isn't
    treated as a wildcard.
  - Groups emit `FilterGroup { concreteType, operator, not, children }`,
    omitting `not` when false.
- `apiFilterToQBNode(filter)` — `QueryFilter` → QB tree for initial hydration
  (used when the initial query already carries a `FilterGroup`).
  - Documented lossy inversions: `HAS` → `is_any_of` (can't distinguish from
    `is_all_of` at the API layer); `EQUAL` on text → `equal` (not `is_exactly`);
    `TextMatchesQueryFilter` and other non-QB filters are left in the baseline.
- `selectedFacetsToQBGroup(facets)` — mode-switch translation from
  `selectedFacets` into a QB tree.
- Pure helpers: `isListColumn`, `escapeLike`, `unescapeLike`,
  `classifyLikePattern`.

**`src/components/QueryBuilder/queryBuilderTranslation.test.ts`** (new)

- 68 unit tests covering:
  - Column-kind classification (`isListColumn`).
  - LIKE escape / unescape round-trip on `%`, `_`, `\`.
  - `classifyLikePattern` mapping (`%foo%`, `foo%`, `%foo`, `foo`, plus
    escaped-wildcard cases).
  - Every `QBConditionOp` → API translation, including incomplete-input null
    guards, boolean values in `values[0]`, and multi-value `HAS` vs
    single-value `IN`.
  - Group serialization: empty group / all-incomplete children → `null`;
    partial children pruned; `not: true` preserved; nested groups recurse.
  - `apiFilterToQBNode` for every `ColumnSingleValueFilterOperator` plus the
    LIKE-pattern classification path and the HAS → `is_any_of` lossy path.
  - Full round-trip through a nested `FilterGroup`.
  - `selectedFacetsToQBGroup` for both value and range facet shapes.

### Verification

- `pnpm nx run @sage-bionetworks/synapse-types:type-check` ✓
- `pnpm nx run synapse-react-client:type-check` ✓
- 68 translation tests ✓
- All 20 pre-existing `sqlFunctions.test.ts` tests ✓ (including the
  "handles every filter type" case)

### Out of scope for this ticket

- No component / UI changes (tracked in PORTALS-4442).
- `ColumnSubQueryFilter` / `SubQueryFilter` (tracked with the Cohort Builder
  page work).
- No context wiring, `hasResettableFilters` update, or mode-transition logic
  in `QueryVisualizationWrapper` (part of PORTALS-4442).

### DIFF
--- a/packages/synapse-react-client/src/components/QueryBuilder/queryBuilderTranslation.ts
+++ b/packages/synapse-react-client/src/components/QueryBuilder/queryBuilderTranslation.ts
@@ -446,7 +446,7 @@ function newId(): string {
   ) {
     return crypto.randomUUID()
   }
-  // Fallback for environments without crypto.randomUUID (older jsdom, etc.).
+  // Monotonic counter is sufficient — QB ids are client-only React keys, not secrets.
   idCounter += 1
-  return `qb-${idCounter}-${Math.random().toString(36).slice(2, 10)}`
+  return `qb-${idCounter}`
 }
--- a/packages/synapse-react-client/src/components/QueryBuilder/queryBuilderTranslation.ts
+++ b/packages/synapse-react-client/src/components/QueryBuilder/queryBuilderTranslation.ts
@@ -176,7 +176,13 @@ function conditionToApiFilter(
             [text!],
           )
     default: {
+      // Compile-time exhaustiveness guard; the log fires only if a new op is
+      // added to QBConditionOp without a case here.
       op satisfies never
+      console.warn(
+        `conditionToApiFilter: unrecognized QB op, skipping condition for column "${columnName}"`,
+        { op },
+      )
       return null
     }
   }
--- a/packages/synapse-react-client/src/components/QueryBuilder/queryBuilderTranslation.ts
+++ b/packages/synapse-react-client/src/components/QueryBuilder/queryBuilderTranslation.ts
@@ -105,32 +105,14 @@ function conditionToApiFilter(
   if (columnName === null) return null
 
   switch (op) {
-    case 'is_any_of': {
-      if (values.length === 0) return null
-      // Multi-value list columns use HAS on ColumnMultiValueFunctionQueryFilter;
-      // single-value columns use IN on ColumnSingleValueQueryFilter.
-      if (isListColumn(columnType)) {
-        return makeMultiValueFilter(columnName, values)
-      }
-      return makeSingleValueFilter(
-        columnName,
-        ColumnSingleValueFilterOperator.IN,
-        values,
-      )
-    }
-    case 'is_all_of': {
-      if (values.length === 0) return null
-      return makeMultiValueFilter(columnName, values)
-    }
-    case 'between': {
-      if (rangeMin == null || rangeMin === '') return null
-      if (rangeMax == null || rangeMax === '') return null
-      return makeSingleValueFilter(
-        columnName,
-        ColumnSingleValueFilterOperator.BETWEEN,
-        [rangeMin, rangeMax],
-      )
-    }
+    case 'is_any_of':
+      return anyOfFilter(columnName, columnType, values)
+    case 'is_all_of':
+      return values.length === 0
+        ? null
+        : makeMultiValueFilter(columnName, values)
+    case 'between':
+      return betweenFilter(columnName, rangeMin, rangeMax)
     case 'gt':
       return rangeBoundFilter(
         columnName,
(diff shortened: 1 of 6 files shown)


packages/synapse-react-client/src/utils/functions/SqlFunctions.ts
```diff
@@ -150,8 +167,20 @@ export const getAdditionalFilters = (
               }
               return filter
             }
+            default:
+              // exhaustive
+              operator satisfies never
+              return undefined
```
nickgros: Maybe add a log?

packages/synapse-react-client/src/components/QueryBuilder/queryBuilderTranslation.ts
```diff
@@ -0,0 +1,452 @@
...
+      return isBlank(text)
+        ? null
+        : makeSingleValueFilter(
+            columnName,
+            ColumnSingleValueFilterOperator.EQUAL,
+            [text!],
+          )
+    default: {
+      op satisfies never
+      return null
```
nickgros, line 186: Consider logging

nickgros (approved):

PR #3144, opened 2026-09-03
PORTALS-4385

jay-hodgson:
## PORTALS-4384 — Actions column on the In-Flight eDUC Signatures table

Added a fourth **Actions** column to `InFlightEDucSignaturesTable` with three per-row link-style actions.

### Actions
| Action | Behavior |
| --- | --- |
| **Review Signatures and Submit** | `<Link>` to the new deep link `/request/:requestId/signature` (see PORTALS-4385). |
| **Modify Request** | Opens a modal `AccessRequirementList` at `RequestDataStep.UPDATE_RESEARCH_PROJECT`, seeded from the row's `accessRequirementId` (new field from PLFM-9903). Disabled when the summary has no `accessRequirementId`. |
| **Cancel Request** | Opens a confirmation `Dialog` warning that the signature envelope will be voided. Confirm calls `useVoidDataAccessRequestSignature(requestId)`; on success a toast fires, the dialog closes, and the query invalidation removes the row from the client-side in-flight filter. |

### Implementation notes
- Column set built by a `buildColumns(actions)` factory so cell renderers close over per-row `onModify` / `onCancel` callbacks; memoized inside the component.
- A small `ModifyRequestModal` sub-component lazily fetches the AR via `useGetAccessRequirements<ManagedACTAccessRequirement>(arId, { staleTime: Infinity })` before mounting the wizard.
- Cancel dialog uses MUI `Dialog` with `Keep Request` (outlined) and `Cancel Request` (contained error, loading state).

### Tests
- `InFlightEDucSignaturesTable.test.tsx` grew from 4 to 8 tests: empty, 4 headers including Actions, deep-link href, Modify opens wizard (asserts `renderAsModal` + `initialWizardEntry.step`), Cancel opens dialog + confirms mutation, Keep Request backs out without mutating, loading skeleton, error alert.
- `AccessRequirementList` is stubbed in the test file so the suite focuses on table behavior.

---

## PORTALS-4385 — Signature-status deep-link route

Added a shareable deep-link route that opens the wizard on the signature-status step for a specific request.

### Route
- **Path:** `/request/:requestId/signature`
- Constants live in `RouteConstants.ts`: `USER_ACCESS_HISTORY_REQUEST_SUBPATH`, `REQUEST_ID_PARAM`, `USER_ACCESS_HISTORY_REQUEST_SIGNATURE_SUBPATH`.
- Registered in `UserAccessRequestHistoryPlace.tsx` as a sibling of the index and submission routes (outside the breadcrumb layout).

### Route renderer (`RequestSignatureRouteRenderer`)
1. If the user is not authenticated, redirects to OneSage login via `storeRedirectURLForOneSageLoginAndGotoURL(oneSageURL.toString())`.
2. Looks up the summary in `useListAllUserDataAccessRequests` (implicit "not-a-participant" check — a request id absent from the user's own list is treated as not found).
3. Fetches the AR with `useGetAccessRequirements<ManagedACTAccessRequirement>`.
4. If summary or AR is missing/failed → `<Navigate to="/" replace />`.
5. Otherwise renders:
   ```tsx
   <UserAccessRequestHistoryPage />
   <AccessRequirementList
     renderAsModal
     onHide={() => { void navigate('/') }}
     initialWizardEntry={{
       step: RequestDataStep.SIGNATURE_STATUS,
       managedACTAccessRequirement: accessRequirement,
     }}
   />

Integration test (.mov):
https://github.com/user-attachments/assets/cd376980-3dab-4d0b-8a25-be77ace7f7cf


packages/synapse-react-client/src/components/dataaccess/UserAccessRequestHistory/UserAccessRequestHistoryPlace.tsx
```diff
@@ -37,6 +53,68 @@ function UserSubmissionPageRouteRenderer() {
...
+  const { requestId } = useParams<{ [REQUEST_ID_PARAM]: string }>()
+  const { isAuthenticated } = useSynapseContext()
+  const oneSageURL = useOneSageURL()
+  const navigate = useNavigate()
+
+  useEffect(() => {
+    if (isAuthenticated === false) {
+      storeRedirectURLForOneSageLoginAndGotoURL(oneSageURL.toString())
+    }
+  }, [isAuthenticated, oneSageURL])
```
nickgros: Maybe this part should be extracted to a reusable hook `useRedirectToLoginIfNotAuthenticated` (or maybe you can think of a more terse name)?

```tsx
  const { isAuthenticated } = useSynapseContext()
  const oneSageURL = useOneSageURL()

  useEffect(() => {
    if (isAuthenticated === false) {
      storeRedirectURLForOneSageLoginAndGotoURL(oneSageURL.toString())
    }
  }, [isAuthenticated, oneSageURL])
```

packages/synapse-react-client/src/components/dataaccess/UserAccessRequestHistory/UserAccessRequestHistoryPlace.tsx
```diff
@@ -37,6 +53,68 @@ function UserSubmissionPageRouteRenderer() {
...
+
+  const { data: summaries, isLoading: isLoadingSummaries } =
+    useListAllUserDataAccessRequests(undefined, {
+      enabled: isAuthenticated === true,
+    })
+  const summary = summaries?.find(s => s.requestId === requestId)
+  const accessRequirementId = summary?.accessRequirementId
```
nickgros, line 65: We will definitely need a service to get one by ID. If/when we go to paginated, there is no guarantee the summary will be on the first page. If it's not, then L98 would immediately redirect.

packages/synapse-react-client/src/components/dataaccess/UserAccessRequestHistory/UserAccessRequestHistoryPlace.tsx
```diff
@@ -111,6 +189,11 @@ const routes: RouteObject[] = [
         path: '/',
         children: [
           { index: true, element: <UserAccessRequestHistoryPage /> },
+          {
+            // Signature status deep link — no breadcrumbs, dialog overlays the history page.
+            path: `${USER_ACCESS_HISTORY_REQUEST_SUBPATH}/:${REQUEST_ID_PARAM}/${USER_ACCESS_HISTORY_REQUEST_SIGNATURE_SUBPATH}`,
+            element: <RequestSignatureRouteRenderer />,
+          },
```
nickgros, line 185: If the user navigates from the base `UserAccessRequestHistoryPage` to this page, the original UserAccessRequestHistoryPage will be unmounted and any internal state would be lost--not 100% sure if there is internal state there, but if there is, losing it would result in unexpected behavior.

I think the more conventional way to do this is to do something like
```tsx
{
  path: '/',
  element: <UserAccessRequestHistoryPage />,
  children: [
    {
      index: true,
      element: null,
    },
    {
      path: `${USER_ACCESS_HISTORY_REQUEST_SUBPATH}/:${REQUEST_ID_PARAM}/${USER_ACCESS_HISTORY_REQUEST_SIGNATURE_SUBPATH}`,
      element: <RequestSignatureRouteRenderer />,
    },
  ],
},
```

Where `UserAccessRequestHistoryPage` would be modified to render an `<Outlet />` and `RequestSignatureRouteRenderer` only renders the dialog.

jay-hodgson: That makes sense, I'll plan to update

packages/synapse-react-client/src/components/dataaccess/UserAccessRequestHistory/UserAccessRequestHistoryPlace.tsx
jay-hodgson: I should probably find a way to reuse `AuthenticationGuard`...

nickgros (approved):

PR #3147, opened 2026-09-04
SWC-7966 Fix clipped pledge and terms-of-use layout on mobile

nickgros:
Below MUI's `md` breakpoint the pledge card was clipped rather than scrollable: the desktop right panel stacked under the form and covered later pledge items, the fixed 150px icon and checkbox columns pushed the "I agree" control off-screen, and the explanatory copy painted over the Continue button.

Layout
- `LeftRightPanel.StyledInnerContainer` (SRC) now owns the canonical 900px two-column card and its mobile single-column stacking. SageAccountWeb's `StyledInnerContainer` extends it instead of redefining the layout, contributing only panel padding and letting the column grow past the viewport so tall content is never clipped.
- Add `getWideDesktopTwoColumnSx` and `getMobileBodyScrollSx`, replacing the 1200px `sx` block that was duplicated inline in `ProfileValidation` and `TermsOfUsePage`. Scoping the wide layout to `up('md')` keeps it from fighting the mobile single-column rules. `RegisterAccount1` opts into document-body scrolling via the latter.
- Below 900px, `_terms-and-conditions.scss` drops the decorative icon column and stacks/centers the checkbox column so the row fits the viewport.
- Show `TermsOfUseMobileAccordion` in place of the desktop right panel on narrow viewports, and hide the panel itself below `md`.
- Split `TermsOfUseRightPanelText` into `TermsOfUseExplanation`: `...Content` for the accordion and `...Section`, which adds the heading, for the right panel.
- Reword the terms button to "View Complete Terms and Conditions for Use".

Tests
- Split SageAccountWeb's Vitest config into `jsdom` and `browser` projects. The browser project runs Playwright/Chromium, since jsdom cannot evaluate the CSS cascade, media queries, or hit testing that these regressions require. Browser files run serially because the MSW service worker is shared across the per-file iframes.
- Serve MSW through a service worker for browser tests and reject unhandled requests, which had been passed through to the production API.
- Register a `termsOfService` group in SRC's shared MSW handler set, using the neutral UP_TO_DATE status. The handlers previously existed but were reachable only from one story, leaving those endpoints unmocked for every  other consumer.
- Add `renderInBrowser` (app theme, style engine, stylesheet, and a fresh QueryClient per render) plus `viewports.ts`, and assert the pledge layout at both mobile and desktop widths so the mobile fix cannot silently regress the desktop layout.

Screenshots:

image

image

image

image

### DIFF
--- a/apps/SageAccountWeb/public/mockServiceWorker.js
+++ b/apps/SageAccountWeb/public/mockServiceWorker.js
@@ -1,0 +1,347 @@
+/* eslint-disable */
+/* tslint:disable */
+
+/**
+ * Mock Service Worker.
+ * @see https://github.com/mswjs/msw
+ * - Please do NOT modify this file.
+ */
+
+const PACKAGE_VERSION = '2.14.6'
+const INTEGRITY_CHECKSUM = '4db4a41e972cec1b64cc569c66952d82'
+const IS_MOCKED_RESPONSE = Symbol('isMockedResponse')
+const activeClientIds = new Set()
+
+addEventListener('install', function () {
+  self.skipWaiting()
+})
+
+addEventListener('activate', function (event) {
+  event.waitUntil(self.clients.claim())
+})
+
+addEventListener('message', async function (event) {
+  const clientId = Reflect.get(event.source || {}, 'id')
+
+  if (!clientId || !self.clients) {
+    return
+  }
+
+  const client = await self.clients.get(clientId)
+
+  if (!client) {
+    return
+  }
+
+  const allClients = await self.clients.matchAll({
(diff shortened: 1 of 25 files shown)


apps/SageAccountWeb/public/mockServiceWorker.js
```diff
@@ -0,0 +1,347 @@
+/* eslint-disable */
```
nickgros, line 1: Auto generated MSW file

jay-hodgson (approved):
Thank you for adding the browser tests, and MSW setup and mock data

PR #3149, opened 2026-09-08
chore: enable oxlint typescript/prefer-optional-chain rule

nickgros:
## Summary
- Enable `typescript/prefer-optional-chain` in `.oxlintrc.json` as `error`, matching the existing `prefer-*` rules. Motivation is that Sonarcloud flags these, so if we decide to continue using it, the ruleset should match.
- Fix all 91 pre-existing violations across 65 files via `oxlint --fix` / `--fix-suggestions`, manually correcting two spots where the automated fix changed behavior/types:
  - `queryBuilderTranslation.ts`: `columnType?.endsWith('_LIST')` returned `boolean | undefined` instead of `boolean` — added `?? false`.
  - `PortalSearchPage.tsx`: `roleTab`'s inferred type included `""` (from the original `&&` chain), which `?.` can't guard against like `&&` could — restructured as a ternary so `roleTab` is cleanly `PortalSearchTabConfig | undefined`.
- Add `mise.toml` pinning `node`/`pnpm`/`java` to the versions this repo's `engines` field and OpenAPI codegen step expect, so `pnpm type-check` works without a separately-installed JDK.

## Test plan
- [x] `oxlint --config .oxlintrc.json .` — zero `prefer-optional-chain` violations
- [x] `pnpm type-check` — all 44 tasks pass
- [x] `pnpm lint` — all 26 projects pass

### DIFF
--- a/.oxlintrc.json
+++ b/.oxlintrc.json
@@ -142,6 +142,7 @@
     "typescript/no-wrapper-object-types": "error",
     "typescript/only-throw-error": "error",
     "typescript/prefer-as-const": "error",
+    "typescript/prefer-optional-chain": "error",
     "typescript/prefer-namespace-keyword": "error",
     "typescript/prefer-promise-reject-errors": "error",
     "typescript/require-await": "error",
--- a/mise.toml
+++ b/mise.toml
@@ -1,0 +1,4 @@
+[tools]
+java = "corretto-21"
+node = "22.13.1"
+pnpm = "10.33.0"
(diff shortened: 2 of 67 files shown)


mise.toml
```diff
@@ -0,0 +1,4 @@
+[tools]
```
nickgros, line 1: I personally use `mise` to manage my environment, so committing this will simplify some of my workflows (e.g. starting up this repo in a container that doesn't have Java or Node)

.oxlintrc.json
```diff
@@ -142,6 +142,7 @@
     "typescript/no-wrapper-object-types": "error",
     "typescript/only-throw-error": "error",
     "typescript/prefer-as-const": "error",
+    "typescript/prefer-optional-chain": "error",
```
nickgros, line 145: Sonarcloud flags these as warnings, so let's add the rule so it gets auto-fixed.

afwillia (approved):

PR #3157, opened 2026-09-08
PORTALS-4442

jay-hodgson:
Demo of Cohort Builder UI (.mov):
https://github.com/user-attachments/assets/fe975276-8a6a-4131-b998-82ca6664a208

Adds a new Query Builder (QB) UI that composes complex `FilterGroup`s of AND/OR/NOT-nested
conditions, mutually exclusive with the existing facet nav in `QueryWrapperPlotNav`.

## Behavior

- **Toggle in `TopLevelControls`.** New `showQueryBuilderControl` and `defaultShowQueryBuilder`
  props on `QueryWrapperPlotNav` control a "Show / Hide Query Builder" button and its initial
  state. When the QB is showing, the facet nav and its toggle are hidden and the plot-nav
  grid drops to full-width layout via a new `isShowingQueryBuilder` class.
- **Live-debounced apply.** Tree edits translate to a `FilterGroup` and are pushed into the
  outgoing `additionalFilters` via `executeQueryRequest(..., { debounce: true })` (750ms),
  reusing the existing `ImmutableTableQuery` debounce plumbing. Initial mount is skipped so
  URL-hydrated filters aren't clobbered.
- **Group semantics.** `+ Condition` and `+ Condition Group` prepend to the target group so
  newly-added rows appear on top. Empty groups render a "No conditions yet, add one above"
  placeholder. Root group cannot be deleted.
- **Faceted-column defaults.** A column with an enumeration facet always classifies as
  `enum` (regardless of raw type), so faceted STRING / INTEGER / DATE / etc. all default the
  op to `is_any_of` and render facet values as pills.
- **Facet-source query strips QB filters.** `useQBFacetSourceMetadata` fires a companion
  metadata query with any `FilterGroup` entries removed from `additionalFilters`, so the QB
  property picker + pill values remain stable while the user builds (mitigates the Synapse
  facet API's lack of self-exclusion for `additionalFilters`).
- **Live SQL preview.** `useQBCombinedSql` requests `BUNDLE_MASK_COMBINED_SQL` for the QB's
  proposed filter, fired only while the SQL tab of the summary bar is active.
- **Plain-English summary.** `qbTreeToReadable` renders a live, styled Reads-As summary
  (AND/OR/NOT connectives colored; NOT groups parenthesized; blank rows skipped).
- **Friendly column names.** Property picker labels and the summary route through
  `QueryVisualizationContext.getColumnDisplayName` (aliases + `unCamelCase`), falling back
  to `unCamelCase` when the QV wrapper isn't present.

## New files (`packages/synapse-react-client/src/components/QueryBuilder/`)

- `QueryBuilderTypes.ts` — `QBNode` / `QBGroup` / `QBCondition` union + guards. *(from earlier
  PR merged into main; extended here.)*
- `queryBuilderTranslation.ts` — updated to use the shared `newQBNodeId` counter. *(existing
  file, minor edits only.)*
- `queryBuilderId.ts` — shared `newQBNodeId()` (crypto.randomUUID w/ monotonic fallback).
- `queryBuilderMetadata.ts` — `classifyColumn`, `availableOpsForKind`, `labelForOp`.
- `QueryBuilderStore.ts` + `.test.ts` — pure immutable tree ops (`addConditionToGroup`,
  `addChildGroup`, `insertNodeAt`, `clearGroup`, `updateCondition`, `updateGroup`,
  `removeNode`, `moveNode` with cycle-rejection), factories, and `hasCompleteCondition`.
  63 tests.
- `QueryBuilderInternalContext.ts` — React context bridging root controls to nested rows
  (column models, facet results, dispatch handlers, `getColumnDisplayName`).
- `FilterConditionRow.tsx` + `.module.scss` — property Autocomplete, operator Select, value
  input per kind (pills / range / text / boolean), drag handle + trash.
- `FilterGroupNode.tsx` + `.module.scss` — AND/OR toggle, NOT toggle, action row, empty
  state, colored left-bracket per combinator/NOT.
- `qbTreeToReadable.tsx` + `.test.tsx` — plain-English summary. 7 tests.
- `QueryBuilderControls.tsx` + `.module.scss` + `.stories.tsx` — top-level panel; controlled
  `{ tree, onTreeChange, onlyFacetedColumns? }`. 4 stories against `syn9630847`.
- `QueryBuilderComponents.test.tsx` — 16 component-level tests.
- `useQBCombinedSql.ts` — SQL preview hook.
- `useQBFacetSourceMetadata.ts` — QB-filter-stripped facet source hook.

## Modified files outside the QB folder

- `components/QueryWrapperPlotNav/QueryWrapperPlotNav.tsx` — new `showQueryBuilderControl` /
  `defaultShowQueryBuilder` props; local `qbTree` state; debounced `executeQueryRequest`
  sync; renders `<QueryBuilderControls />` in place of `<FacetFilterControls />` when the
  toggle is on; forces the collapsed grid class; stamps `isShowingQueryBuilder`.
- `components/QueryWrapperPlotNav/QueryWrapperPlotNav.stories.tsx` — adds
  `FileViewWithQueryBuilder` story pointing at `syn9630847`.
- `components/SynapseTable/TopLevelControls/TopLevelControls.tsx` — new
  `showQueryBuilderControl` / `showQueryBuilder` / `onToggleQueryBuilder` props and toggle
  button.
- `style/components/_query-wrapper-plot-nav.scss` — adds `.QueryBuilderControls` grid-row
  and `.isShowingQueryBuilder .FacetFilterControls { display: none }`.

## Tests

154 QB-related tests pass (63 store + 68 translation + 16 component + 7 summary).
Full package `type-check` clean.

## V1 scope caveats (deferred)

- DnD library integration (drag handle is currently a placeholder IconButton).
- Full context wiring in `QueryVisualizationWrapper` (`filterMode` / `queryBuilderTree` in
  context; QB tree currently lives as local `useState` in `QueryWrapperPlotNavContents`).
- `hasResettableFilters` extension to recognize `FilterGroup` completeness.
- QB → Facets exit confirmation dialog.
- `SelectionCriteriaPills` skipping `FilterGroup` entries.
- Facet-value counts on QB pills.
- Per-condition self-exclusion for facet dropdowns (currently strips the whole `FilterGroup`).

Screenshot 2026-09-15 at 3 55 42 PM


packages/synapse-react-client/src/components/SynapseTable/TopLevelControls/TopLevelControls.tsx
```diff
@@ -202,6 +211,26 @@ const TopLevelControls = (props: TopLevelControlsProps): React.ReactNode => {
...
+                  icon={showQueryBuilder ? 'close' : 'filter'}
+                  wrap={false}
+                />
+              }
+              sx={{
+                ml: 2,
+                fontWeight: 400,
+                fontSize: '14px',
+                textDecoration: 'none !important',
+              }}
```
nickgros: Nit: share styles with facet filter button

packages/synapse-react-client/src/components/QueryBuilder/useQBFacetSourceMetadata.ts
```diff
@@ -0,0 +1,52 @@
...
+ *
+ * Rationale: Synapse's facet-values computation applies `additionalFilters`
+ * globally (unlike `selectedFacets`, which are self-excluded per facet). Once
+ * the QB commits a `FilterGroup`, the current query's facet results collapse
+ * to only the values that pass the QB filter — leaving the user unable to
+ * add or edit values that were filtered out. Firing a separate metadata
+ * query with the QB's own `FilterGroup` removed gives the QB stable
+ * dropdowns while still honoring any non-QB filters and `selectedFacets`.
+ */
+export function useQBFacetSourceMetadata() {
```
nickgros, line 23: Since this isn't affected by the query builder's filters changing, we'll probably want to take a look at if we can fix the suspense boundaries to prevent the full component from flashing when the QB changes are applied (in a future pass).

Or maybe even easier, has design considered having an "Apply" button? I think debounce works for the facet case since the UI is unlikely to jump and future facet options depend on the current facet selection, but the query builder feels like a more slow, intentional process where unexpected changes & flashing should be avoided.

packages/synapse-react-client/src/components/QueryWrapperPlotNav/QueryWrapperPlotNav.tsx
```diff
@@ -171,15 +190,46 @@ export function QueryWrapperPlotNavContents(
...
+  // ImmutableTableQuery machinery (750ms). Skip the initial mount so we don't
+  // clobber any additionalFilters hydrated from the URL before the user has
+  // touched the tree.
+  const skipInitialQbSync = useRef(true)
+  useEffect(() => {
+    if (!showQueryBuilder) return
+    if (skipInitialQbSync.current) {
+      skipInitialQbSync.current = false
+      return
+    }
```
nickgros: Ideally, the app rendering this component would handle pulling the query out of the URL before it gets to this component. We've made incremental changes to get closer to this, but that work still is not done.

If QueryWrapperPlotNav still handles pulling the query out of the URL, would it be possible for the effect that does that to emit a boolean to indicate that the work is done, and use that here instead of a ref that assumes the work is done in the first render pass?

packages/synapse-react-client/src/components/QueryBuilder/QueryBuilderControls.tsx
```diff
@@ -0,0 +1,198 @@
...
+                if (next != null) setSummaryMode(next)
+              }}
+              aria-label="Query summary display mode"
+            >
+              <ToggleButton
+                value="plain-english"
+                aria-label="Show plain English"
+              >
+                Plain English
+              </ToggleButton>
```
nickgros: More of a design/copy nit but I don't really consider this 'plain English'...maybe we call this the 'Summary' and the other is 'SQL'?

Image

packages/synapse-react-client/src/components/QueryBuilder/QueryBuilderControls.tsx
```diff
@@ -0,0 +1,198 @@
...
+export function QueryBuilderControls(props: QueryBuilderControlsProps) {
+  const { tree, onTreeChange, onlyFacetedColumns = false } = props
+  const { columnModels, facetResults } = useQBFacetSourceMetadata()
+
+  const activeTree = tree ?? defaultQBGroup()
+
+  // Use the outer QueryVisualizationContext's display-name resolver when
+  // available; fall back to the shared unCamelCase util when the QB is
+  // rendered outside a QueryVisualizationWrapper (e.g. standalone stories).
+  const queryVizContext = useContext(QueryVisualizationContext)
```
nickgros: Is this ever not available? It should always be present. Should use the wrapper `useQueryVisualizationContext`?

packages/synapse-react-client/src/components/QueryBuilder/FilterConditionRow.module.scss
```diff
@@ -0,0 +1,104 @@
...
+  --qb-pill-color: var(--synapse-primary-action-color);
+}
+
+.dragHandle {
+  cursor: grab;
+  color: var(--synapse-gray-deemphasized);
+}
+
+.remove {
+  color: var(--synapse-gray-600) !important;
```
nickgros, line 21: Lots of `!important`s that I don't know if are really necessary. Not super worried about it in a style module since scope is limited but might be worth a 2nd look

packages/synapse-react-client/src/components/QueryBuilder/useQBFacetSourceMetadata.ts
jay-hodgson: Like the idea of an "Apply" button (Adam suggested "Update Results").  So I'll change to this paradigm.

packages/synapse-react-client/src/components/QueryBuilder/QueryBuilderControls.tsx
jay-hodgson: Will change to "Summary" after discussing with Adam

packages/synapse-react-client/src/components/QueryBuilder/QueryBuilderInternalContext.tsx
```diff
@@ -0,0 +1,57 @@
...
+  updateConditionAt: (
+    conditionId: string,
+    patch: Partial<Omit<QBCondition, 'kind' | 'id'>>,
+  ) => void
+  removeConditionAt: (conditionId: string) => void
+}
+
+export const QueryBuilderInternalContext =
+  createContext<QueryBuilderInternalContextType | null>(null)
+
```
nickgros, line 48: nit: expose a provider wrapper, not the context itself (easier to refactor/add logic to the provider in the future)

```suggestion
const QueryBuilderInternalContext =
  createContext<QueryBuilderInternalContextType | null>(null)
  
export function QueryBuilderInternalContextProvider() {
  // ...
}
```

packages/synapse-react-client/src/components/QueryBuilder/queryBuilderOperations.ts
```diff
@@ -0,0 +1,387 @@
+import { newQBNodeId } from './queryBuilderId'
```
nickgros, line 1: nit: This module is not really a "Store" since it's not storing mutable state. Recommend renaming the module to something like `queryBuilderOperations.ts`

nickgros (approved):

packages/synapse-react-client/src/components/QueryBuilder/FilterConditionRow.module.scss
jay-hodgson: I'll update what I can (and verify in the UI), and add a note for why other `!important` directives are needed

nickgros (approved):

PR #3158, opened 2026-09-09
PORTALS-4454: ADKP autocomplete suggestions in search

jay-hodgson:
### New search index config (ADKP autocomplete)
- Added `autocompleteSuggestionsSearchIndexId in `apps/portals/adknowledgeportal/src/config/resources.ts`
- Added `autocompleteSuggestionsSearchIndexConfig` (`autocompleteFieldName: 'suggestions'`) in `apps/portals/adknowledgeportal/src/config/searchConfig.tsx`
- Wired it into:
  - `pages/Search.tsx` → `ResourceSearchPage`'s `searchIndexConfig` prop
  - `pages/Home.tsx` → `AdknowledgeHeader`'s new `searchIndexConfig` prop
  - `config/navbarConfig.tsx` → new `NavbarConfig.searchIndexConfig` field (for the sticky global nav search bar)

### Framework changes to support the above
- `AdknowledgeHeader.tsx` (synapse-portal-framework): now accepts an optional `searchIndexConfig` prop instead of being hardcoded, since ADKP-specific config can't live in the shared package
- `NavbarConfig` / `Navbar.tsx`: added optional `searchIndexConfig` field, threaded into the sticky-search `HeaderSearchBox`

### Bug fix: rounded border/shadow clash on autocomplete search fields
- `PortalFullTextSearchField.tsx`: moved `className` from the outer `<Autocomplete>` to the inner `<TextField>` (alongside `sx`), so the header variant's rounded border/shadow (from SCSS module) and the default border/shadow (from `sx`) apply to the same element instead of two mismatched nested boxes


apps/portals/adknowledgeportal/src/config/resources.ts
```diff
@@ -36,6 +36,7 @@ export const targetEnablingResourcesDetailsPageSql =
 export const cavaticaConnectAccountURL =
   'https://help.adknowledgeportal.org/apd/CAVATICA.4271145000.html'
 
+export const autocompleteSuggestionsSearchIndexId = 'syn77320025'
```
jay-hodgson: Before review, this needs to be replaced with the SearchIndex that pulls from syn77320025, and tested locally

jay-hodgson: Updated and tested.  Resolving this comment

jay-hodgson:
Screenshot 2026-09-09 at 9 17 46 AM

kianamcc (approved):

PR #3167, opened 2026-09-10
PORTALS-4475

jay-hodgson:
## Fix: don't refetch `/preview` after voiding signature envelope

**File:** `ManualUploadDucStep.tsx`

**Problem**
Uploading a signed DUC on a DAR with an active signature envelope failed the
save step. Sequence:
1. User uploads signed PDF.
2. Client calls DELETE /signature to void the envelope.
3. Void's onSuccess broadly invalidates the DAR query tree.
4. React Query auto-refetches `GET /dataAccessRequest/{id}/preview`, which is
   a stateful endpoint that creates a new envelope and resets the DAR to draft.
5. Client `updateDar(..., ducFileHandleId)` is rejected by the backend because
   the DAR now has both `eDucSignatureEnvelopeId` (re-set by /preview) and a
   pending `ducFileHandleId` update — a state the backend does not allow.

**Change**
Added local state `isSubmittingSignedDuc` that flips to `true` before
`voidSignatureAsync(...)`. It's OR'd into the `enabled` predicate of
`useGetDataAccessRequestPreview` so the observer becomes inactive before the
invalidation fires. Inactive observers are not auto-refetched.

**Test**
Added a regression test in `ManualUploadDucStep.test.tsx` that:
- Mounts with `eDucSignatureEnvelopeId` set
- Counts `/preview`, `DELETE /signature`, and `POST /dataAccessRequest` calls
- Triggers the upload
- Asserts `/preview` was called exactly once (mount only) after the full
  upload → void → refetch → update sequence

All 14 tests in the file pass; adjacent test files (`AccessRequirementList`,
`InFlightEDucSignaturesTable`) also pass unchanged.

Manually verified that with these changes I can now eject out of the eDUC workflow (void signatures) and submit a data access request for review using the manually uploaded document.

### DIFF
--- a/packages/synapse-react-client/src/components/AccessRequirementList/ManagedACTAccessRequirementRequestFlow/ManualUploadDucStep/ManualUploadDucStep.tsx
+++ b/packages/synapse-react-client/src/components/AccessRequirementList/ManagedACTAccessRequirementRequestFlow/ManualUploadDucStep/ManualUploadDucStep.tsx
@@ -71,12 +71,24 @@ export default function ManualUploadDucStep(props: ManualUploadDucStepProps) {
     throwOnError: true,
   })
 
+  // Once we start the void → refetch → updateDar sequence, the preview query must not run again:
+  // GET /dataAccessRequest/{id}/preview has server-side side effects (creates a new signature
+  // envelope, resets DAR to draft) that repopulate `eDucSignatureEnvelopeId` on the DAR. The
+  // backend then rejects the follow-up updateDar because we're trying to set a manually-uploaded
+  // `ducFileHandleId` on a DAR that still has an active signature envelope. Because
+  // `useVoidDataAccessRequestSignature` invalidates the root DAR query key on success, an active
+  // preview observer would otherwise be auto-refetched.
+  const [isSubmittingSignedDuc, setIsSubmittingSignedDuc] = useState(false)
+
   const {
     data: previewFileHandle,
     isLoading: isLoadingPreview,
     error: previewError,
   } = useGetDataAccessRequestPreview(dataAccessRequest?.id ?? '', {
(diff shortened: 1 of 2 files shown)


packages/synapse-react-client/src/components/AccessRequirementList/ManagedACTAccessRequirementRequestFlow/ManualUploadDucStep/ManualUploadDucStep.tsx
```diff
@@ -71,12 +71,24 @@ export default function ManualUploadDucStep(props: ManualUploadDucStepProps) {
...
     data: previewFileHandle,
     isLoading: isLoadingPreview,
     error: previewError,
   } = useGetDataAccessRequestPreview(dataAccessRequest?.id ?? '', {
-    enabled: Boolean(dataAccessRequest?.id) && !downloadHrefOverride,
+    enabled:
+      Boolean(dataAccessRequest?.id) &&
+      !downloadHrefOverride &&
+      !isSubmittingSignedDuc,
   })
```
nickgros, line 92: Does Bruce plan to change the backend to make this 'stateless' (IOW, is this workaround temporary)?

If not, could you instead change `useGetDataAccessRequestPreview` to use `useMutation`? That would provide a more correct API to handle a request that results in a backend side-effect.

jay-hodgson: Bruce is planning to change the backend so the preview call is stateless (create an envelope, return the preview, and then discard the envelope), tracked here:
https://sagebionetworks.jira.com/browse/PLFM-9938

nickgros (approved):
LGTM, fine workaround for a temporary problem

PR #3173, opened 2026-09-10
PORTALS-4470: update custom text for externally hosted datasets

jay-hodgson:
Screenshot 2026-09-10 at 1 08 10 PM


kianamcc:
Maybe it's just me, but I would've expected the link to be more easily accessible, without clicking 'show more', right under the 'how to download'!

kianamcc (approved):

PR #3176, opened 2026-09-11
PORTALS-4476: Update NF portal search configs from (re)tuned results

anngvu:
Updates existing tools config and adds a `SearchQueryConfig` to the NF datasets, publications and studies searches. Also have to make one small synapse-react-client API change this config update depends on.  Added `includeUnlistedFields` for more control, defaulting to `true` **so existing configs are unaffected**, because it turns out that we do want the ability to exclude fields in some cases where `*` pulls back in columns that add too much noise.

All new weights/shapes are transcribed from recently updated benchmarks, scored against each of their golden set on the live indexes. Each index wanted a different shape, which is the main reason these are per-index configs rather than one shared default, e.g.

- **Cross-fields** wins where queries spread terms across several columns, e.g. a real query seen in GA4 for "human+nerve+plexiform+neurofibroma+bulk+RNA-seq" that essentially needs info from 4 different columns.
- **Fuzziness** wins on some others where search may be more complicated so there needs to be more typo tolerance.
- Datasets is deliberately unboosted, just how the scores turned out.

### DIFF
--- a/apps/portals/nf/src/config/resources.ts
+++ b/apps/portals/nf/src/config/resources.ts
@@ -110,8 +110,9 @@ export const publicationsSearchQueryConfig: SearchQueryConfig = {
 }
 export const studiesSearchIndexId = 'syn75081633'
 // Tuned in nf-osi/opensearch-ops (benchmark/studies/fields.yaml).
-// includeUnlistedFields:true is load-bearing, not just the default -- the `*` it appends is
-// what keeps the keyword-mapped ENTITYID columns searchable.
+// includeUnlistedFields:true because the `*` enables search for keyword-mapped ENTITYID 
+// cols not explicitly here; accessRequirements or others at 1 despite `*` to convey
+// they were measured-and-unboosted, and weights may still change as content evolves.
 export const studiesSearchQueryConfig: SearchQueryConfig = {
   queryStrategy: 'MULTI_MATCH_CROSS_FIELDS',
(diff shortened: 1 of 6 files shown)


apps/portals/nf/src/config/resources.ts
```diff
@@ -42,21 +42,121 @@ export const enabledAnalysisPlatforms: ExternalAnalysisPlatform[] = [
...
+    summary: 3,
+    dataType: 3,
+    diseaseFocus: 2,
+    fundingAgency: 2,
+    grantDOI: 2,
+    studyStatus: 2,
+    dataStatus: 2,
+    clinicalTrialID: 2,
+    alternateDataRepository: 2,
+    accessRequirements: 1,
```
jay-hodgson, line 135: You could add a comment that notes these `1` boosted fields are being explicitly called out, even though the `*` will implicitly set these to 1 (because `includeUnlistedFields` is set to `true` in this instance).
Maybe I should ask, did you mean to set `includeUnlistedFields` to `true` in the studiesSearchQueryConfig?

jay-hodgson (approved):

apps/portals/nf/src/config/resources.ts
anngvu: Updated comment, yes studiesSearchQueryConfig was only one where `includeUnlistedFields` should be `true`

PR #3179, opened 2026-09-11
PORTALS-4447 Gate RDCA-DAP DAR wizard on account eligibility

nickgros:
- Detect isEligibilityFailure from wizard queries/mutation and surface a shared RdcaDapEligibilityExplainer before and during the wizard
 - Block AridhiaAccessStatus from offering the wizard when ineligible; add AccessBlockedByRDCADAPAccountNotLinked icon state
 - Add not_authorized error code for FAIR 403 "not authorised" responses
 - Fix datasets to be an array (matches live API) in the OpenAPI patch, request-matching util, and mocks
 - Add DAR status popover and FAIR portal URL helpers for pending/approved requests
 - Remove unused AMPALS_RDCA_DAP_FORM_ENABLED feature flag
 - Dedupe RDCA_DAP_URL into a shared constant

Screenshots:

image

image

image

image

Account is not linked with RDCA-DAP:

image

After clicking the 'unlinked' icon':

image

### DIFF
--- a/apps/portals/ampals/src/components/AmpAlsAccessColumn.tsx
+++ b/apps/portals/ampals/src/components/AmpAlsAccessColumn.tsx
@@ -11,22 +11,25 @@ type SourceValue = 'Synapse' | 'GEO' | 'Critical Path Institute' | null
 
 const ID_COLUMN_NAME = 'id'
 const SOURCE_COLUMN_NAME = 'source'
-const URL_COLUMN_NAME = 'url'
 const DATASET_CODE_COLUMN_NAME = 'dataset_code'
 
 function AccessBySource(props: {
   source: SourceValue
   id: string
   datasetCode: string
-  url?: string
 }) {
-  const { source, id, url, datasetCode } = props
+  const { source, id, datasetCode } = props
 
   switch (source) {
     case 'GEO':
       return <AccessIcon restrictionUiType={RestrictionUiType.Accessible} />
     case 'Critical Path Institute':
-      return <AridhiaAccessStatus url={url} datasetCode={datasetCode} />
+      return (
+        <AridhiaAccessStatus
+          fairPortalUrl={import.meta.env.VITE_ARIDHIA_FAIR_PORTAL_URL}
+          datasetCode={datasetCode}
+        />
+      )
     case 'Synapse':
     default:
       return <HasAccessV2 entityId={id} showButtonText={false} />
@@ -45,7 +48,6 @@ function AmpAlsAccessColumnCell<TValue = unknown>(
 
   const idColumnIndex = getColumnIndex(ID_COLUMN_NAME, selectColumns)
   const sourceColumnIndex = getColumnIndex(SOURCE_COLUMN_NAME, selectColumns)
-  const urlColumnIndex = getColumnIndex(URL_COLUMN_NAME, selectColumns)
   const datasetCodeColumnIndex = getColumnIndex(
     DATASET_CODE_COLUMN_NAME,
     selectColumns,
@@ -53,27 +55,20 @@ function AmpAlsAccessColumnCell<TValue = unknown>(
   if (
     idColumnIndex == null ||
     sourceColumnIndex == null ||
-    urlColumnIndex == null ||
     datasetCodeColumnIndex == null
   ) {
     console.warn(
-      "AmpAlsAccessColumn: 'id', 'source', 'url' or 'dataset_code' column not found",
+      "AmpAlsAccessColumn: 'id', 'source' or 'dataset_code' column not found",
     )
     return null
   }
 
   const id = row.original.values[idColumnIndex]!
   const sourceValue = row.original.values[sourceColumnIndex] as SourceValue
-  const url = row.original.values[urlColumnIndex] ?? undefined
   const datasetCode = row.original.values[datasetCodeColumnIndex]!
 
   return (
-    <AccessBySource
-      source={sourceValue}
-      id={id}
-      url={url}
-      datasetCode={datasetCode}
-    />
+    <AccessBySource source={sourceValue} id={id} datasetCode={datasetCode} />
   )
 }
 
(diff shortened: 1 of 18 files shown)


packages/aridhia-client/src/codegen/filterSpec.ts
```diff
@@ -349,6 +349,31 @@ function applyUpstreamDefectPatches(spec: OpenAPISpec): OpenAPISpec {
     fieldItemSchema.properties.default_options = { type: 'string' }
   }
 
+  // 7. `RequestListItem` and `RequestFull` both declare `datasets` as a single `RequestDataset`
+  //    object, but the live `GET /fair/requests` and `GET /fair/requests/{code}` responses
+  //    return `datasets` as an array of `RequestDataset` (a request can span multiple
+  //    datasets). Confirmed against live gateway responses.
+  {
```
nickgros, line 356: Fix a regression where current request status could not be seen because the OpenAPI Schema does not match the actual behavior. I introduced this when I rebuilt the schemas, the earlier version of `aridhia-client` handled this correctly.

apps/portals/ampals/src/components/AmpAlsAccessColumn.tsx
```diff
@@ -11,22 +11,25 @@ type SourceValue = 'Synapse' | 'GEO' | 'Critical Path Institute' | null
...
-const URL_COLUMN_NAME = 'url'
 const DATASET_CODE_COLUMN_NAME = 'dataset_code'
 
 function AccessBySource(props: {
   source: SourceValue
   id: string
   datasetCode: string
-  url?: string
 }) {
-  const { source, id, url, datasetCode } = props
```
nickgros: We will compute the URL using the dataset code only, (caught because I expected to go to C-Path Dev and actually ended up on prod)

packages/synapse-react-client/src/aridhia-queries/useAridhiaQuery.test.tsx
```diff
@@ -255,6 +255,33 @@ describe('useAridhiaMutation', () => {
     },
   )
 
+  it('normalizes a FAIR-side 403 "not authorised" response into an eligibility-failure AridhiaError', async () => {
```
nickgros: Regression test where we don't find out a user has not linked their account until we make a request to the FAIR API. The token exchange gateway always returns a token, even with a brand new account that RDCA-DAP has never seen.

packages/synapse-react-client/src/components/Aridhia/AridhiaAccessStatus.test.tsx
```diff
@@ -95,13 +114,9 @@ describe('AridhiaAccessStatus', () => {
   afterEach(() => server.resetHandlers())
   afterAll(() => server.close())
 
-  it('opens the request wizard in a dialog when no request exists for this dataset and the feature flag is enabled', async () => {
```
nickgros: Removed the feature flag

packages/synapse-react-client/src/components/Aridhia/AridhiaAccessStatus.test.tsx
```diff
@@ -227,4 +267,52 @@ describe('AridhiaAccessStatus', () => {
       expect(screen.queryByRole('dialog')).not.toBeInTheDocument(),
     )
   })
+
+  it('keeps the submitted confirmation screen open when the post-submission requests refetch changes the background access icon', async () => {
```
nickgros: Mid-development regression where the confirmation dialog would disappear because it was unintentionally unmounted.

packages/aridhia-client/src/codegen/filterSpec.ts
jay-hodgson: This filterSpec approach allows the next engineer (perhaps you or I) to not encounter the same headaches that I passed to you, so THANK YOU!

packages/synapse-react-client/src/aridhia-queries/useAridhiaQuery.ts
```diff
@@ -115,6 +120,18 @@ export async function toAridhiaError(error: unknown): Promise<AridhiaError> {
         isEligibilityFailure: bodyErrorText !== 'Invalid parameter',
       })
     }
+    // The gateway's `/authenticate` always mints a token regardless of RDCA-DAP linkage; a FAIR
+    // resource call (e.g. `/fair/requests`) is what actually rejects an unlinked account, as a
+    // 403 with this message in its own `{ error: { message } }` envelope.
+    if (
+      httpStatus === 403 &&
+      bodyErrorText === 'Not authorised for this operation'
```
jay-hodgson: This is so precise, I hope the message doesn't change (or even change case)!
Would changing this condition to instead contain the case insensitive string "not authorised" or "not authorized" be too broad (and catch other cases)?

packages/synapse-react-client/src/components/Aridhia/RdcaDapEligibilityDialog.tsx
```diff
@@ -0,0 +1,59 @@
...
+export function RdcaDapEligibilityExplainer() {
+  return (
+    <Stack spacing={2}>
+      <Typography>
+        You need a linked RDCA-DAP account before you can request access to this
+        dataset.
+      </Typography>
+      <Typography>
+        To create a new account or link an existing account, open RDCA-DAP and
+        sign in with your ALS Knowledge Portal (Synapse) account.
```
jay-hodgson, line 25: I know this is ALSKP only, and will be for the foreseeable future, but I still wonder if this "ALS Knowledge Portal (Synapse)" should be hard-coded.

jay-hodgson (approved):

packages/synapse-react-client/src/components/Aridhia/RdcaDapEligibilityDialog.tsx
nickgros: Agree. I think the right tool to generalize this is the `useSourceApp` utility, but that could be a decent refactoring effort since it's all in SageAccountWeb and I'm not sure if it would need to be modified to function correctly on other sites.

nickgros: Agree. I think the right tool to generalize this is the `useSourceApp` utility, but that could be a decent refactoring effort since it's all in SageAccountWeb and I'm not sure if it would need to be modified to function correctly on other sites. So IMO let's defer this work for now

PR #3184, opened 2026-09-14
Optimize CI workflows

nickgros:
Optimize CI workflows: fewer redundant clones/rebuilds, correctness fixes in npmpublish, dead-weight removal

- **Checkouts**: add `filter: tree:0` to every `fetch-depth: 0` checkout. This repo's `.git` is 5.2GB; a full-history clone was costing 125-155s per job, often more than the job's actual work (e.g. Storybook's build). Partial-clone filtering keeps the commit graph `nx affected` needs but defers blob/tree downloads.

- **`end-to-end-test-portals.yml`**: split into a `detect-affected-portals` job + an affected-driven dynamic matrix. Previously all 11 portal legs always spun up (checkout + install + affected-check) even when zero portals were affected — measured ~14 min of pure overhead on ~35% of PRs. The matrix now only contains portals nx actually marked affected, and skips entirely when that list is empty. Also: raised `max-parallel` 5→6 (cuts the "many portals affected" case from 3 waves to 2); dropped `fetch-depth`/`nx-set-shas` from the `e2e-tests`/`merge-e2e-reports` jobs, which call `nx run` directly and never needed full history; moved the portal list out of an inline shell string into a checked-in `portals_e2e.json`; removed now-unreachable dead code in `merge-e2e-reports` (the directory-existence guards became no-ops once `e2e-tests` only runs when a portal is affected and always uploads a blob).

- **`npmpublish.yml`**: gate the 7-package PR matrix on `nx affected` instead of rebuilding and dry-run-publishing every package on every PR. Fixed two real bugs introduced while wiring this up (caught in review, not by CI): the affected-check had no resolvable `nx affected` base ref under the default depth-1 clone (added `fetch-depth: 0` + `filter: tree:0` + `nrwl/nx-set-shas`, PR-only); and the check piped `nx` directly into `jq -e` inside an `if`, which — under GitHub's default `bash -e` with no `pipefail` — silently treated any real `nx` crash as "not affected" rather than failing the job. Restructured to capture `nx`'s output into a variable first and added `set -euo pipefail`, so a genuine failure is now loud. The resulting `should_run` output is applied uniformly across all four gated steps (previously the guard was copy-pasted on three and silently omitted on a fourth).

- **`build-test.yml`**: `test`, `lint`, and `type-check` no longer `needs: build`. None of them actually consume `build`'s output — each computes its own `^build` dependency chain via `nx affected` regardless — so gating them behind `build` only added `build`'s own runtime to the critical path for no benefit. `build` still runs (in parallel, not as a gate): it's independently useful signal, since it verifies every affected project can be bundled by Vite, including static-prerender-heavy portals like `nf`, which `test`/`lint`/`type-check` don't otherwise exercise.

- **Nx cache**: attempted a fix (caching `.nx/workspace-data` alongside `.nx/cache`, since Nx's database cache was rejecting every restored artifact as "unrecognized"). That did stop the rejection warning, but verifying against live CI runs on two real test PRs showed Nx still doesn't reuse cached task outputs across machines — `lint`/`type-check` jobs on separate runners were still fully re-executing the codegen steps `build` had already completed. This tracks with Nx's own docs: cross-machine cache trust without Nx Cloud isn't really supported (the old `NX_REJECT_UNKNOWN_LOCAL_CACHE` escape hatch was removed in the DB-cache rewrite, and the self-hosted remote-cache packages were deprecated for the CREEP vulnerability). Since it doesn't work, removed the cache-sharing attempt entirely rather than leaving dead weight: deleted the `STORE_NX_CACHE` input and cache step from `pnpm-setup-action`, and the 7 call sites that set it.

- **`deploy-sage-account-web-docker.yml`**: quoted shell variables, `tr '[:upper:]'/'[:lower:]'` instead of literal bracket-class chars, `"${GIT_REF##*/}"` instead of `sed`, moved context values into step `env:` — closes 11 shellcheck findings surfaced by `actionlint`. No behavior change.

Verified with `actionlint` (0 findings across the repo) and by comparing real CI runs on two throwaway test PRs (a portal-only change and a shared-package change) against a pre-change baseline run on the same fork.


.github/actions/pnpm-setup-action/action.yml
```diff
@@ -1,10 +1,5 @@
 name: 'Set up pnpm and configure environment'
 description: 'Installs Node, pnpm, fetches cache (if exists), and runs pnpm install'
-inputs:
-  STORE_NX_CACHE:
```
nickgros, line 4: Non-cloud solutions are no longer supported by Nx, this has not worked for a while. Only options, if we want to use Nx's cache, are to use Nx Cloud or host our own service using their API spec: https://nx.dev/docs/reference/deprecated/self-hosted-cache-packages


Since it's broken, let's remove it

.github/workflows/build-test.yml
```diff
@@ -20,12 +20,10 @@ jobs:
       - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
         with:
           fetch-depth: 0
+          filter: tree:0
```
nickgros, line 23: `filter: tree:0` downloads the entire git tree, but does NOT download the actual patch content (which can be downloaded on-demand as needed).

These jobs no longer have to download the 5+ GB repo contents.

.github/workflows/deploy-sage-account-web-docker.yml
```diff
@@ -39,18 +39,22 @@ jobs:
         run: echo "${{ secrets.GITHUB_TOKEN }}" | docker login ghcr.io -u ${{ github.actor }} --password-stdin
       - name: Push image
         working-directory: apps/SageAccountWeb
+        env:
+          REPO_OWNER: ${{ github.repository_owner }}
+          GIT_REF: ${{ github.ref }}
+          MODE: ${{ matrix.mode }}
```
nickgros, line 45: Fixing lint issues caught by actionlint/shellcheck

.github/workflows/end-to-end-test-portals.yml
```diff
@@ -12,74 +12,52 @@ concurrency:
...
-            challenges,
-            digitalhealth,
-            eliteportal,
-            genie,
-            namhub,
-            nf,
-            stopadportal,
-          ]
+  detect-affected-portals:
+    if: ${{ (github.event_name == 'push' && github.ref_name == 'main') || github.event_name == 'pull_request' }}
```
nickgros, line 16: Determine if portals changed before running their e2e.

.github/workflows/portals_e2e.json
```diff
@@ -0,0 +1,13 @@
+[
```
nickgros, line 1: Portals which have e2e tests extracted to a single point (this list is now referenced twice)

.github/workflows/end-to-end-test-portals.yml
```diff
@@ -12,74 +12,52 @@ concurrency:
...
+          all_json=$(cat .github/workflows/portals_e2e.json)
+          portals=$(jq -cn --argjson affected "$affected_json" --argjson all "$all_json" '[$all[] | select(IN($affected[]))]')
+          echo "Affected portals: $portals"
+          echo "portals=$portals" >> "$GITHUB_OUTPUT"
+  e2e-tests:
+    needs: detect-affected-portals
+    if: ${{ needs.detect-affected-portals.result == 'success' && needs.detect-affected-portals.outputs.portals != '[]' }}
+    strategy:
+      fail-fast: false
+      max-parallel: ${{ github.repository_owner == 'Sage-Bionetworks' && 6 || 10 }}
```
jay-hodgson, line 42: Reason why max-parallel is set to 6 for `Sage-Bionetworks` repo owner but 10 otherwise?

.github/workflows/portals_e2e.json
jay-hodgson: Will this list need to be updated over time?  Do we have the necessary prompts to guide llms in the future to update this list when adding new portals (for example)?

jay-hodgson (approved):

.github/workflows/end-to-end-test-portals.yml
nickgros: There's a limit for concurrent machines for the entire organization. We limit it to avoid interfering with other Sage-Bionetworks repos. Bumped in this PR from 5 to 6 because now we have 11 portals with e2e tests, so (assuming all individual jobs take roughly the same amount of time, $$n$$), they should complete in $$2n$$ instead of $$3n$$.

.github/workflows/portals_e2e.json
nickgros: Updated portals-e2e/README.md, which I assume is what an LLM would follow when adding a new one

PR #3186, opened 2026-09-14
PORTALS-4473 - Anonymous chat session support

jay-hodgson:
Enables **anonymous chat** with agents the ACT has opened via `allowAnonymousChatSession`, using an attempt-based flow (try to create the session; on 401/403 fall back to login). No dependency on the ACT-only `actSettings` GET endpoint.

### useChat.ts
- useCreateAgentSession now forwards onError to the caller (previously only onSuccess was forwarded), matching useUpdateAgentSession.

### SynapseChat.tsx
- Reads isAuthenticated from useSynapseContext.
- New optional prop `onSessionCreationUnauthenticated?: () => void`.
- On session-creation failure with a **401/403 for an anonymous user**, invokes onSessionCreationUnauthenticated (caller prompts login) and suppresses the inline error Alert. Authenticated failures still show the error inline.
- Hides the access-level (data levels) dropdown when `!isAuthenticated`, so anonymous users aren't offered private-data levels that would likely fail server-side.

### SynapsePortalChatDialog.tsx
- Wires onSessionCreationUnauthenticated to the existing OneSage login redirect (useOneSageURL + storeRedirectURLForOneSageLoginAndGotoURL).
- Anonymous users can now open the dialog; they're only redirected if the agent rejects the anonymous session.

### CurieChatWidget.tsx
- Removed the hard-coded "must be logged in → redirect on click" gate; the widget now opens the chat for everyone and lets the session attempt decide (redirect happens only on failure).

### Tests
- SynapseChat.test.tsx — added cases: callback fires on anonymous 401/403, not for authenticated users, not for non-auth errors, error suppressed when deferring to login, error shown when no login fallback.
- SynapsePortalChatDialog.test.tsx — new file verifying the OneSage redirect wiring.
- CurieChatWidget.test.tsx — reworked to drop the auth-redirect mocks and assert anonymous users open the chat.

Additional fix to UX, specific to the case when an anonymous user attempts to use a chat that does not allow anonymous usage:
In this case, an anonymous user opens the chat → session creation fails → they're redirected to OneSage login. The full-page navigation discards the dialog's open state, so on return the chat is gone.
Now we attempt to persist the chat intent (variant and initial message), and restore the state after login.  Use helpers for this (see ChatReopenAfterLogin and portalChatSessionStorage)


packages/synapse-react-client/src/synapse-queries/chat/useChat.ts
```diff
@@ -51,6 +51,11 @@ export function useCreateAgentSession(
         await options.onSuccess(newAgentSession, variables, ctx)
       }
     },
+    onError: async (err, variables, ctx) => {
+      if (options?.onError) {
+        await options.onError(err, variables, ctx)
+      }
+    },
```
nickgros: Just spread `options`

```suggestion
    ...options,
    mutationFn: (request: CreateAgentSessionRequest) =>
      synapseClient.agentChatServicesClient.postRepoV1AgentSession({
        createAgentSessionRequest: request,
      }),
```

apps/synapse-portal-framework/src/components/ChatReopenAfterLogin.tsx
```diff
@@ -0,0 +1,35 @@
...
+import { useChatDialogContext } from './ChatDialogContext'
+
+/**
+ * Reopens the portal chat dialog after an anonymous user completes the login round-trip.
+ *
+ * The reopen intent is stored just before the login redirect (see SynapsePortalChatDialog); here it
+ * is consumed once the session has initialized, and the chat is reopened only if the user returned
+ * authenticated (otherwise reopening would immediately redirect again).
+ */
+export function ChatReopenAfterLogin() {
```
nickgros: Make this a hook since it doesn't return any JSX. Maybe just call it inside of `ChatDialogContextProvider`? You might need to add a wrapper over the basic `ChatDialogContext.Provider`, but that's a pattern we use for most of our custom contexts
```suggestion
export function useReopenChatAfterLogin() {
```

nickgros (changes requested):
Two minor issues to fix before merging, approach LGTM

apps/synapse-portal-framework/src/components/ChatDialogContext.tsx
nickgros, line 17: Likely no need to export this anymore (to avoid the case where a new caller uses this instead of the new provider wrapper)
```suggestion
const ChatDialogContext = createContext<
```

nickgros (approved):

apps/synapse-portal-framework/src/components/ChatDialogContext.tsx
jay-hodgson: Added a comment above this export.
```
// Do not consume this raw context directly; use ChatDialogContextProvider (which also wires up
// side effects like post-login chat reopen). Exported only so tests can provide a value in isolation.
```

PR #3192, opened 2026-09-14
updated b2ai config table versions

Sigfried:
now includes Harry's new data from
https://github.com/bridge2ai/b2ai-standards-registry/pull/529

but the explore page doesn't get updated because it now uses a search index rather than the DST_denormalized table and it doesn't update automatically. i don't have access to fix this. needs attention from someone at Sage: @BryanFauble, @jay-hodgson?


jay-hodgson (approved):

jay-hodgson:
I also updated https://www.synapse.org/Synapse:syn74909093/tables/ definingSql to point to syn65676531.99

PR #3200, opened 2026-09-16
SWC-7922 Hide grid until initial data load completes

afwillia:
Show a loading skeleton instead of grid data, buttons, and validation messages while the client loads an existing grid session.


packages/synapse-react-client/src/components/DataGrid/useDataGridWebsocket.ts
```diff
@@ -356,10 +358,29 @@ export function useDataGridWebSocket(options?: UseDataGridWebSocketOptions) {
...
+  // until the replay drains. Latching it means later exchanges — mid-session
+  // patches, or a reconnect to the same session — never take it back to false.
+  if (
+    state.hasCompletedInitialSync &&
+    !state.isSyncing &&
+    isModelRenderable(state.model)
+  ) {
+    hasCompletedInitialLoadRef.current = true
+  }
+
```
nickgros, line 373: Storing the state in the ref isn't the best option, since it's impure and may not trigger a re-render.

-----

## Recommended design

The most robust option is to make this an explicit reducer state transition:

```ts name=useDataGridWebsocket.ts
interface WebSocketState {
  // ...
  hasCompletedInitialLoad: boolean
}

type WebSocketAction =
  // ...
  | { type: 'INITIAL_LOAD_COMPLETE' }
```

Then dispatch `INITIAL_LOAD_COMPLETE` only after all required conditions are satisfied, and reset it only when changing to a different session/replica.

For example, an effect could observe the relevant state:

```ts name=useDataGridWebsocket.ts
useEffect(() => {
  if (
    !state.hasCompletedInitialLoad &&
    state.hasCompletedInitialSync &&
    !state.isSyncing &&
    isModelRenderable(state.model)
  ) {
    dispatch({ type: 'INITIAL_LOAD_COMPLETE' })
  }
}, [
  state.hasCompletedInitialLoad,
  state.hasCompletedInitialSync,
  state.isSyncing,
  state.model,
  modelSnapshot,
])
```

The reducer would make the transition one-way:

```ts name=useDataGridWebsocket.ts
case 'INITIAL_LOAD_COMPLETE':
  return {
    ...state,
    hasCompletedInitialLoad: true,
  }
```

This gives you:

- an explicit one-time transition;
- a guaranteed re-render when the flag changes;
- no ref mutation during render;
- clear reset behavior for a different session;
- protection against later sync activity turning the flag off.

packages/synapse-react-client/src/components/DataGrid/useDataGridWebsocket.ts
```diff
@@ -368,6 +419,6 @@ export function useDataGridWebSocket(options?: UseDataGridWebSocketOptions) {
     errorEstablishingWebsocketConnection:
       state.connectionError ?? errorEstablishingWebsocketConnection,
     websocketError: state.websocketError,
-    hasSufficientData: isModelRenderable(state.model),
+    hasSufficientData: isModelRenderable(state.model, modelSnapshot),
```
nickgros: **Dead code**. `hasSufficientData` (`useDataGridWebsocket.ts:422`) is still returned by the hook but SynapseGrid — its only production reader — was switched to `hasCompletedInitialLoad`; only tests (lines 146, 211, 304, 330, 661, 692) still read it. Confirmed via grep. Fix: delete the field, retarget those test assertions.

packages/synapse-react-client/src/components/DataGrid/useDataGridWebsocket.ts
```diff
@@ -343,23 +381,36 @@ export function useDataGridWebSocket(options?: UseDataGridWebSocketOptions) {
...
-    const orderReady = columnOrder.length >= 1
-    const rowsReady = rows.length >= 0
-    return columnsReady && orderReady && rowsReady
-  }
+  }, [
+    state.hasCompletedInitialLoad,
+    state.hasCompletedInitialSync,
+    state.isSyncing,
+    state.model,
+    modelSnapshot,
```
nickgros: Correctness issue (from LLM):

> **Defect:** uses `!state.isSyncing` as a proxy for "replay finished," but `CONNECTION_CLOSED` also sets `isSyncing: false` without touching `hasCompletedInitialSync`/`model`.
>
> **Repro (suspected, confidence 0.55):** mid-replay, if the socket drops (network blip/sleep) after `GRID_READY`+`SYNC_ENDED` but before the next `ResponseComplete`, `CONNECTION_CLOSED` flips `isSyncing` false while `hasCompletedInitialSync` is already true and the model is renderable — all four gate conditions hold on a disconnected, partially-loaded model. The one-way flag latches true; grid/buttons/validation show with stale/incomplete data, then the reconnect replay streams in live — the exact confusion this PR exists to prevent.
>
> **Fix suggested:** also require `state.isConnected`, or clear `hasCompletedInitialSync` on `CONNECTION_CLOSED` while the load flag is still false.


IMO it's an edge case, but probably worth fixing, since the fix seems simple

packages/synapse-react-client/src/components/DataGrid/components/GridLoadingState.tsx
```diff
@@ -0,0 +1,56 @@
...
+
+/**
+ * Placeholder shown in place of the grid until its first connection has finished
+ * loading data. Rendering this instead of the grid keeps users from editing cells
+ * or watching rows stream in while the initial data is still being replayed.
+ */
+export default function GridLoadingState(props: GridLoadingStateProps) {
+  return (
+    <Stack spacing={1} sx={{ mb: 1.25 }}>
+      <Typography variant="sectionTitle">Setting up grid</Typography>
```
nickgros: Move `"Setting up grid"` into GRID_LOADING_MESSAGES also

packages/synapse-react-client/src/components/DataGrid/components/GridLoadingState.test.tsx
```diff
@@ -0,0 +1,51 @@
...
+    'reports progress when $name',
+    ({ props, expectedMessage }) => {
+      render(<GridLoadingState {...props} />)
+      expect(screen.getByRole('status')).toHaveTextContent(expectedMessage)
+    },
+  )
+
+  it('renders a skeleton in place of the grid', () => {
+    render(<GridLoadingState hasReplicaId hasPresignedUrl isConnected />)
+    expect(screen.getByText('Setting up grid')).toBeInTheDocument()
```
nickgros: Move to GRID_LOADING_MESSAGES and reference in test

packages/synapse-react-client/src/components/DataGrid/useDataGridWebsocket.ts
```diff
@@ -343,23 +381,36 @@ export function useDataGridWebSocket(options?: UseDataGridWebSocketOptions) {
...
-  function isModelRenderable(model: GridModel | null) {
-    if (!model?.api.getSnapshot() || !modelSnapshot) {
-      return false
+  useEffect(() => {
+    if (
+      !state.hasCompletedInitialLoad &&
+      state.hasCompletedInitialSync &&
+      !state.isSyncing &&
+      isModelRenderable(state.model, modelSnapshot)
+    ) {
```
nickgros: Testability nit (from LLM)

**Tangled effect** — the initial-load gate (`!hasCompletedInitialLoad && hasCompletedInitialSync && !isSyncing && isModelRenderable(...)`) lives inside a dispatching `useEffect` plus a module-private `isModelRenderable`. Tests can only reach it by mocking 4 modules and driving a full connect→waitFor→act sequence; the fixed `useCRDTModelView` stub means empty-columns/empty-columnOrder branches are never exercised. **Prescription:** extract `export function isInitialLoadReady(state, snapshot): boolean`. **Payoff:** `it.each` over states/snapshots with no module mocks, no `renderHook`, no `waitFor`.

packages/synapse-react-client/src/components/DataGrid/useDataGridWebsocket.test.ts
```diff
@@ -612,4 +612,216 @@ describe('useDataGridWebSocket', () => {
...
+      expect(result.current.hasCompletedInitialLoad).toBe(true)
+
+      act(() => {
+        config.onSyncStart!()
+      })
+
+      expect(result.current.isSyncing).toBe(true)
+      expect(result.current.hasCompletedInitialLoad).toBe(true)
+
+      // The transition is one-way: a completed later exchange can't reset it
```
nickgros: Testability nit (from LLM):

> **Wide interface** — the one-way reset logic is fully decided by the private `websocketReducer`, but each reducer-level behavior ("stays true on reconnect," "resets on new session") is only reachable through the full hook harness (4 mocks + async orchestration), a ~10:1 setup-to-assert ratio.
>
> **Prescription:** `export function websocketReducer(...)` (already exports `initialWebSocketState`). 
>
> **Payoff:** `expect(websocketReducer(state, action).hasCompletedInitialLoad).toBe(false)` — synchronous, no mocks.

packages/synapse-react-client/src/components/DataGrid/useDataGridWebsocket.test.ts
```diff
@@ -612,4 +612,216 @@ describe('useDataGridWebSocket', () => {
...
+      act(() => {
+        config.onGridReady!()
+        config.onSyncEnd!()
+      })
+      expect(result.current.hasCompletedInitialLoad).toBe(true)
+    })
+
+    it('stays true when reconnecting to the same session', async () => {
```
nickgros: > **Vacuous / can't-fail test**: `useDataGridWebsocket.test.ts` "stays true when reconnecting to the same session" (~784-791) passes even if `CONNECT_REQUESTED` transiently reset the flag to false, because the `INITIAL_LOAD_COMPLETE` effect re-fires and re-latches it true within the same `act()` before the assertion reads it — exactly the flicker that would re-show the skeleton. Fix: record every rendered value and assert no `false` appears post-reconnect.

packages/synapse-react-client/src/components/DataGrid/components/GridLoadingState.test.tsx
```diff
@@ -0,0 +1,51 @@
...
+    expect(
+      document.querySelectorAll('.MuiSkeleton-root').length,
+    ).toBeGreaterThan(0)
```
nickgros: Should test for an accessible role (you might need to add an ARIA role)

nickgros (changes requested):
New approach looks good! Had a handful of minor issues caught in LLM review that I agree with, just need to work those out and should be good

PR #3202, opened 2026-09-16
PORTALS-4439: Mobile view - fix programs grid for screen sizes ~<375px

kianamcc:
Jira: https://sagebionetworks.jira.com/browse/PORTALS-4493
Ann comment: https://sagebionetworks.jira.com/browse/PORTALS-4477?focusedCommentId=342660

Screenshot 2026-09-16 at 1 48 02 PM
Tested with these mobile devices:
Screenshot 2026-09-16 at 1 52 13 PM


apps/synapse-portal-framework/src/components/adknowledge/AdknowledgePrograms/AdknowledgePrograms.module.scss
```diff
@@ -23,6 +23,10 @@
   align-items: flex-start;
   padding: 0 2rem;
 
+  @media (max-width: 450px) {
+    padding: 0;
```
kianamcc, line 27: Technically we could run into the same issue for realllyyy small screen sizes like 320px and under but that is for older gen phones and we would probably need a new design instead of hexes for screen sizes that small so this will hopefully be good enough.

jay-hodgson (approved):

PR #3210, opened 2026-09-17
SWC-7984 - Hooks and validation

nickgros:


packages/synapse-react-client/package.json
```diff
@@ -32,6 +32,8 @@
     "@apidevtools/json-schema-ref-parser": "^14.2.1",
     "@aws-sdk/client-s3": "^3.1064.0",
     "@aws-sdk/lib-storage": "^3.1064.0",
+    "@dnd-kit/helpers": "^0.5.0",
+    "@dnd-kit/react": "^0.5.0",
```
nickgros, line 36: DnD support (used in a later PR)

packages/synapse-react-client/src/mocks/msw/handlers/asyncJobHandlers.ts
```diff
@@ -1,67 +1,121 @@
...
+ * (`POST /repo/v1/asynchronous/job` + `GET /repo/v1/asynchronous/job/{jobId}`). Callers that need
+ * to support more than one `concreteType` in the same test/story MUST merge their entries into a
+ * single `dispatch` map and call this once.
+ *
+ * @param dispatch maps each supported request `concreteType` to a function producing that job's response body
+ * @param options.asyncTypeServicePaths registers a dedicated start/get path pair instead of the generic endpoint; see {@link AsyncTypeServicePaths}
+ * @param options.backendOrigin the backend origin to use for the handlers
+ * @param options.serviceSpecificEndpointResponseStatus the status code for the dedicated response path (`asyncTypeServicePaths` only); also determines whether both GET endpoints report `COMPLETE` or `FAILED`
+ */
+export function generateAsyncJobHandlers(
```
nickgros, line 81: Modified to enable mocking multiple async services at the same time.

packages/synapse-react-client/src/utils/jsonschema/generateDataAccessSchema.ts
```diff
@@ -0,0 +1,133 @@
+/**
+ * Temporary code before PLFM-9949 is complete.
```
nickgros: Temporary client-side code to enable 'draft' mode for ACT where they can iterate on a form template + JSON Schema and get a live preview. This also powers our mock services (so maybe it's not 100% temporary). Ideally we can get this removed and use the backend service before releasing, but seems to work fine, so it's not super risky IMO.

packages/synapse-react-client/src/components/DataGrid/components/ReorderColumnsDialog.tsx
```diff
@@ -34,13 +35,6 @@ export type ReorderColumnsDialogProps = {
   onCancel: () => void
 }
 
-function moveItem(order: number[], from: number, to: number): number[] {
```
nickgros, line 37: Moved to shared util where it was slightly refined

packages/synapse-react-client/src/mocks/accessRequirement/mockFormTemplates.ts
```diff
@@ -27,7 +24,6 @@ export const mockGenomicsTemplate: FormTemplate = {
             'ui:widget': 'textarea',
             'ui:options': { rows: 5 },
           },
-          submissionContext: FormTemplateFieldSubmissionContextEnum.ALWAYS,
```
nickgros, line 30: Modified the design to move this `submissionContext` property info to the JSON Schema instead of in the FormTemplate. Lots of utility code in this PR to support this change.

This field will be removed from the backend soon, at which point we can remove these properties from the auto-generated models.

packages/synapse-react-client/src/mocks/accessRequirement/mockJsonSchemas.ts
```diff
@@ -51,21 +62,44 @@ export const mockClinicalSchema: RJSFSchema = {
...
+          irbApprovalNumber: {
+            type: 'string',
+            title: 'IRB Approval Number',
+            description:
+              'If your institution requires IRB approval, provide the approval number.',
+          },
+        },
+      },
+    },
+  ],
```
nickgros, line 100: Example where 'REQUEST' | 'RENEWAL' submissionContext is respected in the JSON Schema.

packages/synapse-react-client/src/mocks/msw/handlers/asyncJobHandlers.ts
```diff
@@ -1,67 +1,140 @@
...
+ * Build a single-entry `AsynchronousJobDispatch` for one `concreteType`, narrowing `handler`'s
+ * request parameter to that type's variant of `AsynchronousRequestBody` so callers never have to
+ * re-cast the generic request themselves.
  */
-export function generateAsyncJobHandlers<
-  TRequestBody = unknown,
-  TResponseBody extends DefaultBodyType | ErrorResponse =
-    | DefaultBodyType
-    | ErrorResponse,
+export function dispatchEntry<
```
nickgros, line 45: Refactored this async job MSW code so it can now mock multiple async job types on the generic `POST /asynchronous/job` and `GET /asynchronous/job/{jobId}` at the same time.

packages/synapse-react-client/src/utils/jsonschema/submissionContext.ts
```diff
@@ -0,0 +1,412 @@
+/**
+ * Convention for gating a JSON Schema property to a data access REQUEST or RENEWAL submission.
```
nickgros, line 2: In a recent change to the technical design, we moved `submissionContext` (whether a question should be asked in a `REQUEST` or `RENEWAL` context only) to be represented by conditional logic in the JSON Schema, instead of as a property on the `FormTemplate`.

This change eliminates a complicated schema-preprocessing step that was necessary before validation. This step was also the driving reason for the server to own logic related to how the client (likely web client only) displays the form.

The flipside of this is that now synapse-react-client has to maintain logic to help ACT users modify moderately complex schemas (with conditional logic) via the UI. This new utility is the bulk of that support.

packages/synapse-react-client/src/utils/jsonschema/generateDataAccessSchema.ts
```diff
@@ -0,0 +1,88 @@
...
+ * by that step's fields, with the step's title/description encoded as `jsonSchema.title` /
+ * `jsonSchema.description`. A field whose property only applies to the other `requestType` (per
+ * the `x-synapse-submissionContext` convention -- see `./submissionContext`) is omitted.
+ *
+ * @param template The form template defining the steps and fields.
+ * @param schema The full JSON Schema referenced by the form template.
+ * @param requestType The type of data access request, used to filter fields based on context.
+ * @returns A bundle of per-step (jsonSchema, uiSchema) objects suitable for RJSF.
+ */
+export function generateDataAccessSchema(
```
nickgros, line 34: In a recent change to the technical design, we changed this functionality from being a backend-driven service to now be the client's responsibility. Part of the motivation was that we removed a complicated step that originally justified putting this logic in the server (explained in another comment on this PR).

In all likelihood, only the web client would have ever used this service, which has no scaling concerns that justify keeping it on the backend, so it makes more sense to maintain it here as a utility.

packages/synapse-react-client/src/components/FormTemplateEditor/useSaveFormTemplate.ts
```diff
@@ -0,0 +1,104 @@
...
+  | { ok: true; template: FormTemplate }
+  | { ok: false; validationErrors: FormTemplateFieldValidationError[] }
+
+/**
+ * Owns the create-schema, create-template, and update-template mutations, and sequences them
+ * into a single save mutation. A validation failure resolves as `{ ok: false, validationErrors }`
+ * rather than rejecting; a network failure (including a 412 conflict from a concurrent edit)
+ * rejects normally, surfaced via the returned mutation's `error`.
+ */
+export function useSaveFormTemplate(
```
nickgros, line 41: Composite mutation to combine saving a JSON Schema and FormTemplate via UI coming in a later PR.

We could consider asking for a backend service to combine these two changes + validation into a single transaction, but I think it's low risk for the client to run them in sequence--changes to the JSON Schema or FormTemplate are _not_ automatically applied to existing ARs

packages/synapse-react-client/package.json
jay-hodgson: Yay - after this is merged I'll go back to the Query Builder and make the DnD handles functional

packages/synapse-react-client/src/components/FormTemplateEditor/formTemplateValidation.ts
```diff
@@ -0,0 +1,142 @@
...
+   * schema-required property that isn't bound to any field -- there is no slot to anchor it to. */
+  stepIndex?: number
+  fieldIndex?: number
+  schemaPath: string
+  message: string
+}
+
+/**
+ * Replicates the three rules the server enforces on `FormTemplateField` at create/update time
+ * (see the `FormTemplateField` doc comment), so the editor can block save and point the error at
```
jay-hodgson, line 65: It's unfortunately that it sounds like this validation logic is duplicated on the client.  Perhaps later the server api could provide a pre-check validation service, that would return a structured response with per-field errors.  
This does not block approval of the PR

packages/synapse-react-client/src/components/FormTemplateEditor/useSaveFormTemplate.ts
jay-hodgson: Per Nick, this should say "not automatically applied to existing ARs" (because the AR is locked to a FormTemplate version, which is locked to a JSON Schema version).

packages/synapse-react-client/src/components/FormTemplateEditor/utils.ts
```diff
@@ -0,0 +1,44 @@
...
+ * (unlike fields, whose identity is their `schemaPath`), so the editor
+ * tracks one here. `uiKey` never leaves this component tree — strip it with
+ * `toFormTemplateSteps` before handing steps back to a caller.
+ */
+export type EditableFormTemplateStep = FormTemplateStep & { uiKey: string }
+
+/** Create a new empty step with a fresh identity. */
+export function createEditableStep(): EditableFormTemplateStep {
+  return {
+    title: 'New Step',
```
jay-hodgson, line 19: Should we autofill the title on a new step?

packages/synapse-react-client/src/mocks/msw/handlers/asyncJobHandlers.ts
```diff
@@ -1,67 +1,140 @@
...
+/**
+ * A dedicated async service-specific start/get path pair.
+ */
+type AsyncTypeServicePaths = {
+  requestPath: string
+  responsePath: (tokenParam: string) => string
+}
+
+/**
+ * Generates MSW handlers for asynchronous jobs, dispatching by request `concreteType`.
```
jay-hodgson, line 68: Very useful!

packages/synapse-react-client/src/utils/jsonschema/generateDataAccessSchema.ts
jay-hodgson: This makes sense. 
How many mapper coding pattern solutions are we going to introduce to this codebase? :D

jay-hodgson (approved):

packages/synapse-react-client/src/components/FormTemplateEditor/utils.ts
nickgros: Can you say more about what you think we should do? Are you suggesting we leave it blank to force the user to enter a more meaningful title?

packages/synapse-react-client/src/components/FormTemplateEditor/formTemplateValidation.ts
nickgros: Tracking in PLFM-9984

PR #3217, opened 2026-09-17
PORTALS-4495

jay-hodgson:
# Facet Filtering (FF)  ↔ Query Builder (QB) transition support

Wires the Query Builder toggle in `TopLevelControls` to a smart handler in
`QueryWrapperPlotNav` that manages state and the outgoing query on both
sides of the transition.

## FF → QB (silent, lossless)

When the user turns the QB on:

- Reads `selectedFacets` from `currentQueryRequest`.
- If non-empty, translates them into a QB tree via
  `selectedFacetsToQBGroup(selectedFacets, columnModels)` and dispatches a
  single `executeQueryRequest` that:
  - clears `selectedFacets`,
  - strips any prior `FilterGroup` from `additionalFilters`,
  - and appends the newly-translated `FilterGroup` produced by
    `qbNodeToApiFilter`.
- If empty, leaves the tree at `defaultQBGroup()` so the user starts with
  the empty-state helper text.
- Flips `showQueryBuilder` to `true`.

Table results stay identical across the swap — the filter is just
represented differently.

## QB → FF (QB conditions removed, confirmation required)

When the user turns the QB off:

- If the current tree has at least one complete condition
  (`hasCompleteCondition(qbTree)`) OR the query already carries a
  `FilterGroup` in `additionalFilters`, opens a `ConfirmationDialog`
  (`"Hide Query Builder"` in red).
- Otherwise, tears down immediately.
- On confirm: resets the tree to `defaultQBGroup()`, strips any
  `FilterGroup` entries from `additionalFilters`, and flips
  `showQueryBuilder` to `false`.
- On cancel: closes the dialog; QB stays open with its state intact.

## Supporting changes

- **`queryBuilderMetadata.ts`**: reordered `classifyColumn` so `_LIST` →
  `list` and `hasFacet` → `enum` run before the `columnType == null` →
  `unknown` short-circuit. Fixes a rendering bug where FF-translated
  conditions (which had `columnType: null`) fell back to an
  operator-less, plain-text-input row.
- **`queryBuilderTranslation.ts`**: `selectedFacetsToQBGroup` accepts an
  optional `columnModels` argument (deep-readonly-friendly) and threads
  the real `columnType` into every translated condition. Ensures LIST
  facets round-trip as `HAS` (`ColumnMultiValueFunctionQueryFilter`),
  not `IN` (`ColumnSingleValueQueryFilter`).
- **`QueryWrapperPlotNav.tsx`**:
  - Added `isConfirmingHideQb` state and a `ConfirmationDialog`.
  - Added `enterQueryBuilder()`, `exitQueryBuilder()`, and
    `handleToggleQueryBuilder()` helpers wired into `onToggleQueryBuilder`
    on `TopLevelControls`.
  - Destructures `data: queryMetadata` from the existing `useQuery` call
    so `enterQueryBuilder` can pass `columnModels` to the translation.

## Reused primitives

- `selectedFacetsToQBGroup` (translation)
- `qbNodeToApiFilter` (translation)
- `hasCompleteCondition` (operations)
- `isFilterGroup` (type guard, `utils/types/IsType`)
- `ConfirmationDialog` (existing component)

## Tests added

- `selectedFacetsToQBGroup`: LIST columnType resolution + fallback to
  `null` when the column isn't in the models (2 cases).
- `FilterConditionRow`: renders the operator picker + pill group when
  `columnType` is `null` but a facet result exists (1 case).


packages/synapse-react-client/src/components/QueryWrapperPlotNav/QueryWrapperPlotNav.tsx
```diff
@@ -198,12 +207,77 @@ export function QueryWrapperPlotNavContents(
...
+  }
+
+  const exitQueryBuilder = () => {
+    setQbTree(defaultQBGroup())
+    executeQueryRequest(prev => ({
+      ...prev,
+      query: {
+        ...prev.query,
+        additionalFilters: prev.query.additionalFilters?.filter(
+          f => !isFilterGroup(f),
```
nickgros, line 258: Might this unexpectedly remove a locked filter?

nickgros (approved):

packages/synapse-react-client/src/components/QueryWrapperPlotNav/QueryWrapperPlotNav.tsx
jay-hodgson: It shouldn't pose an issue on exit.  
However, if a QueryBuilder condition is added that targets the same column as a "locked" column filter, I think the locked filter would get replaced. I'll investigate.
Thanks for asking!

jay-hodgson: The bug was worse.  I think the locked column filter was lost as soon as any condition is applied.

packages/synapse-react-client/src/components/QueryBuilder/QueryBuilderControls.tsx
```diff
@@ -50,13 +51,22 @@ export function QueryBuilderControls(props: QueryBuilderControlsProps) {
...
-      query: {
-        ...prev.query,
-        additionalFilters: filterGroup ? [filterGroup] : undefined,
-      },
-    }))
+    executeQueryRequest(prev => {
+      // Preserve any non-QB additionalFilters (e.g. lockedColumn constraints
+      // introduced by portal detail pages) — Update Results only owns the
+      // FilterGroup slot.
+      const preserved = (prev.query.additionalFilters ?? []).filter(
```
nickgros, line 58: is `prev` the right one to look at for locked column preservation? I would assume `initialQueryRequest` is the one to look at, but maybe `prev` correctly handles a case I am not thinking of

jay-hodgson: I think this will work, because we only want to protect the QB FilterGroup, and everything else from prev.additionalFilters should be preserved (including the lock column filter, or any other dynamically-added filters other components have injected since then).  On "Update Results", we only want to clear the QB filters

PR #3223, opened 2026-09-18
PORTALS-4494: Add example search terms for CLASSIC

jay-hodgson:
Screenshot 2026-09-18 at 1 54 35 PM


kianamcc (approved):
I think the background color of the pills could have better contrast against the background. 🤔

kianamcc (approved):

